PR #17, opened 2017-10-26
Test fix

faboweb:


faboweb:
/go is not on PATH i think

README.md
```diff
@@ -68,3 +70,12 @@ Then build and run the app:
 npm run pack && npm run build:darwin
 open builds/cosmos-ui-darwin-x64/cosmos-ui.app
 ```
+
+## Testing
```
mappum: :+1:

app/src/main/index.js
```diff
@@ -180,19 +192,37 @@ async function startTendermint (root) {
     'node',
     '--home', root
   ]
-  if (DEV) args.push('--log_level', 'info')
+  // if (DEV) args.push('--log_level', 'info')
```
mappum: Why is this commented out? Won't this mean development mode now uses `error` log level? We might miss some useful messages.

app/src/main/index.js
```diff
@@ -180,19 +192,37 @@ async function startTendermint (root) {
...
+    }
+  })
 
   let rpc = RpcClient('localhost:46657')
   let status = () => new Promise((resolve, reject) => {
-    // ignore errors, since we'll just poll until we get a response
-    rpc.status((err, res) => resolve(res))
+    rpc.status((err, res) => {
+      // ignore connection errors, since we'll just poll until we get a response
+      if (err && err.code !== 'ECONNREFUSED') {
```
mappum: 👍 Nice

app/src/main/index.js
faboweb: Was recommended by Frey

mappum: OK

app/src/renderer/vuex/modules/user.js
```diff
@@ -64,7 +64,13 @@ export default ({ commit, node }) => {
 
   const actions = {
     async signIn ({ commit, state }, seedWords) {
-      let privkey = dg.mnemonicToPrivKey(seedWords)
+      let privkey
+      try {
+        privkey = dg.mnemonicToPrivKey(seedWords)
+      } catch (e) {
+        commit('notifyError', {title: 'Generating private key failed', body: e.message})
```
mappum: Maybe this should say "Deriving private key failed" instead of "generating", since this is for deriving a key generated during the fundraiser.

PR #50, opened 2017-11-09
Switch to Gaia

### LINKED ISSUE #33: Switch to Gaia
I as user
Want to be on the production system
So that I can participate in the actual cosmos network.

AC:
- [ ] The wallet is working using gaia-server

mappum:
(This PR depends on #48 being merged)

Resolves #33

This will let us actually use gaia-based networks (baseserver is not compatible, so balance will always show 0). We can currently only use older networks like `tak`

AC:
- [x] Switch to use `gaia-1` network by default
- [x] Switch to `gaiaserver`

### DIFF
--- a/test/unit/specs/main.spec.js
+++ b/test/unit/specs/main.spec.js
@@ -1,6 +1,5 @@
 const fs = require('fs-extra')
 const {join} = require('path')
-const rmdir = require('../../../app/src/helpers/rmdir.js')
 
 function sleep (ms) {
   return new Promise((resolve) => setTimeout(resolve, ms))


mappum:
We now have a working `gaia server` command in `cosmos/gaia`, but it's on a WIP branch. So if we merge this, things will work in development, but it would temporarily break production builds since our scripts pull that code using `go build` which only takes what is on master (it's not possible to specify another branch). So to finish this PR for merge we could either:
- Modify the build scripts to checkout the WIP branch for `gaia`, which would make production builds work now, although they would be targeting an unstable WIP branch

OR
- Merge the PR as-is, letting us work on top of this in development but leaving production builds non-functional until the `gaia server` command makes it into a `cosmos/gaia` master release

What do you guys think? @faboweb @keppel @nylira @jolesbi

faboweb:
I vote for merging this and communicate, that we are blocked on production build while gaiaserver is not in master.

mappum:
This PR is almost done, but I'm having some trouble fixing the tests for the switching the light-client command from `baseserver serve` to `gaia server serve`.

mappum:
This is ready for merge 🎉 just needs another review

mappum:
Hm, merging `develop` into this is making a test fail (not hanging this time)

faboweb:
On it right now

PR #53, opened 2017-11-11
changed testnet genesis path

faboweb:


mappum:
We won't be able to make this change until we switch from `baseserver` to `gaiaserver` (the light-client specific to gaia-based networks). BTW, I also made this change in #50

faboweb:
For me

Issue #34, opened 2017-11-07
Remove Webpack


faboweb:
I as a developer
Want to remove webpack from the project
So that I have less steps building the project and have better stacktraces leading to faster development.

AC:
- [ ] Webpack is removed
- [ ] All imports work
- [ ] Vue templates rendering at runtime (or precompile)
- [ ] NPM scripts work with the new setup

faboweb:
My thoughts on this:
* we will in any case have a build step
  * copy dependencies and assets to the dist folder
    * we need to figure out which dependencies those are
  * we need to transpile .vue files
* we need to load the dependencies at runtime
  * therefor we need a module loader
* we already have sourcemapped stacktraces
  * those are long and could be cropped to only show files in src/

Currently I would vote against removing webpack and splitting all these tasks into single ones.
We can crop the stacktrace.

faboweb:
We aligned to go with only cropped stacktraces

PR #56, opened 2017-11-12
shrinking stacktrace to just the files in the project

faboweb:
before:
image

after:
image

implementing #34


mappum:
Very cool, all that other stuff is just noise

PR #61, opened 2017-11-15
Use version from root package.json

### LINKED ISSUE #39: Use main package.json for version
To test compatibility we check the version in the app/package.json against a saved one. This will probably never change so we should check against the top-level package.json.

keppel:
The compatibility check still uses semver, so a version change is only gonna be "incompatible" if the major version increments. Fixes #39.


mappum:
>The compatibility check still uses semver, so a version change is only gonna be "incompatible" if the major version increments.

That was intentional. As this is a UI and there wouldn't otherwise be dependents which we could make breaking changes for, we consider changing the format of the locally stored data a breaking change.

keppel:
Yep, makes sense! 👍

PR #82, opened 2017-11-21
fix refresh button location + design

### LINKED ISSUE #72: Refresh Balance Design
The refresh button on the balance page was implemented, but there was no design applied.

nylira:
Fixes #72


mappum:
It looks good, but I'm worried that the button being in the footer will make it not obvious enough. It's already not standard that we have a refresh button, so I have a feeling some people will be sitting there wondering why their incoming balance isn't showing up.

Maybe the solution is something like text under the balances that says "Last updated at <time>"? What are your thoughts @nylira @jolesbi ?

nylira:
Yeah visibility is something I have to do. The plan is to move those footer buttons to the top of the screen for desktop

mappum:
Ah perfect

mappum:
Hm, not sure why the Circle tests froze.

PR #84, opened 2017-11-21
Fixes the cursor to a hand on vote components (also disables selecting vote text)

### LINKED ISSUE #75: Wrong Cursor in Proposal Voting View
The cursor doesn't show a pointer when hovering the voting options.

nylira:
Fixes #75


mappum:
The Circle tests froze, just like #82

PR #98, opened 2017-11-22
Fix validators page

### LINKED ISSUE #78: Mock Validators
Child of: #24 

As long as we don't have data from the block-chain, all possible view elements should be viewable so we can start testing early with users.

For the validators it shows an empty list for. We should mock this.

mappum:
Resolves #78


app/src/renderer/vuex/modules/validators.js
```diff
@@ -10,18 +11,29 @@ export default ({ commit, node }) => {
     },
     setValidators (state, validators) {
       state.validators = validators
+      console.log('vallies', validators)
```
faboweb: should this stay in?

mappum: Oops, nope.

PR #101, opened 2017-11-22
Dark theme colors

nylira:
Colors for dark-theme are here. Don't merge it yet though, I'm going through the app to make sure it works with these colors.

I will work on improving the layout in a different PR. Layout improvements will include more padding and better large display layouts.


mappum:
The new colors are super good, I'm excited for this to get merged 💯

nylira:
There is endless polish to be done with UI design, but this PR (for updating the color scheme to match cosmos.network) is ready to go!

screen shot 2017-11-22 at 3 30 48 pm
screen shot 2017-11-22 at 3 29 26 pm

webpack.renderer.config.js
```diff
@@ -96,11 +100,15 @@ let rendererConfig = {
...
     alias: {
-      'components': path.join(__dirname, 'app/src/renderer/components'),
-      'renderer': path.join(__dirname, 'app/src/renderer'),
-      '@': path.join(__dirname, 'app/src/renderer')
+      'components': resolve('app/src/renderer/components'),
+      'renderer': resolve('app/src/renderer'),
+      '@': resolve('app/src/renderer'),
+      assets: resolve('app/src/renderer/assets'),
+      comp: resolve('app/src/renderer/components'),
+      common: resolve('app/src/renderer/components/common'),
```
nylira: @faboweb can you help figure out why CircleCI/Jest isn't finding `common/AnchorCopy.vue` when it runs the app unit test? I'm thinking it's related to the Webpack `resolve.alias` module I used:

nylira: This seems related https://stackoverflow.com/questions/42629925/testing-with-jest-and-webpack-aliases

faboweb: U already found the solution. Jest uses their own mapper because it doesn't touch webpack. Therefor we need to have module mappers in webpack and jest.

faboweb:
Feedback on the design:
- Yellow and Green are very close in contrast.
- Yellow actually means dismissal of the proposal right? It doesn't look like that.
- The number symbolizes most voted option in percent of total possible votes, right? This could mean that we show a green 20 or a red 15. Which feels weird. How about showing the amount of votes relative to the voted?
- Personal: The colors do not match well with the rest of the colors.

nylira:
Yes, the colors are placeholders. The saturated green/red/blue are pulled from standard app notification colors (success, warning, failure). They don't fit in the app design because they are meant to stand out. Let me try using the app colors.

nylira:
screen shot 2017-11-23 at 11 00 09 am

faboweb:
Looks fancy :) I like it. Also the 4 numbers actually work pretty well for me. Everything is spotable pretty fast.

jolesbi:
re: color conversation ... we can sort this out later ... but i do think once we establish our own "standard app notification colors" (i thought we did already) we should always use those when appropriate. if the feeling is that they are too much or they don't feel right in a particular use case (like colouring this circle-graph) than we need to discuss a different way of using the colors to show the information. the colors are standard — but the circle graph is not.

✌️ 🎨 📈

app/src/renderer/components/common/AppHeader.vue
```diff
@@ -8,7 +8,7 @@ nav#app-header: .container
   router-link.header-item.header-item-logo(to="/")
     img(src="~@/assets/images/cosmos.png")
   app-menu(v-if="config.activeMenu === 'app' || config.desktop")
-  app-menu-user(v-if="config.activeMenu === 'user' || config.desktop")
+  // app-menu-user(v-if="config.activeMenu === 'user' || config.desktop")
```
jolesbi: comment?

nylira: We don't have an app-menu-user currently because this code was taken from the cosmos-forum. We may have in the future?

app/src/renderer/components/common/NiListItemLoading.vue
```diff
@@ -0,0 +1,64 @@
+<template lang="pug">
+.li-loading
+  .ni-votes
+  .text
+    .title ████████████████████
```
jolesbi: anybody seeing this black box?

nylira: This is also a component from cosmos-forum. The black box is shown using the `dim` color and this component is used to show that the app is in the middle of loading list items.

app/src/renderer/components/govern/ChartVotes.vue
```diff
@@ -57,10 +61,16 @@ export default {
...
             backgroundColor: [
-              'hsl(210,41%,49%)',
-              'hsl(60,18%,70%)',
-              'hsl(345,75%,49%)',
-              'hsl(210,32%,12%)'
+              'hsl(0,0%,100%)',
+              'hsl(233,96%,60%)',
+              'hsl(326,96%,59%)',
+              'hsl(233,13%,50%)'
+              /*
```
jolesbi: what's with these?

nylira: `ChartVotes` is used in the `LiProposal` component -- it's the CircleGraph. The graph colors don't seem to be editable through CSS so I hardcoded them in here.

jolesbi:
a lot of great improvements here!

nylira:
@jolesbi Thanks! Agreed that we need to sort out the graph colors. The notification colors are set --- red for danger, yellow for warning, green for success. We have to talk about whether they make sense for the circle graph though (in another issue).

faboweb:
@jolesbi where those colors the standard notification colors?

PR #104, opened 2017-11-23
Jordan/25 wallet tests

jolesbi:
yikes. so this took me longer than it should have — but i learned quite a bit in the process. thanks for your help @faboweb !


jolesbi:
screen shot 2017-11-23 at 1 55 39 pm

jolesbi:
@faboweb we need some help here. can't figure out how to fix CircleCI 🌮

faboweb:
I want to do a small test tomorrow if we need the added dependencies on root level.

PR #105, opened 2017-11-23
Improve desktop page-based navigation.

### LINKED ISSUE #88: Move footer buttons to the app header for larger devices.
They're too hard to find on tablet portrait and above devices. Affects #82

nylira:
I've implemented a new NiPageHeader/NiToolBar design that:

1. Works for mobile and desktop
2. Has hover labels for desktop
3. Has much better positioning on desktop

screen shot 2017-11-23 at 2 03 01 pm

I've also added a help button to the top right of the screen, because we need something like that. It opens to a modal for users to get help on our Matrix, Forum, or GitHub (for now).

screen shot 2017-11-23 at 2 02 54 pm

Fixes #88

### DIFF
--- a/app/src/renderer/components/staking/PageNominate.vue
+++ b/app/src/renderer/components/staking/PageNominate.vue
@@ -1,5 +1,7 @@
 <template lang='pug'>
 page(:title='pageTitle')
+  div(slot="menu"): tool-bar
+
   form-struct(:submit='onSubmit')
     form-group(:error='$v.fields.id.$error'
       field-id='form-nominate-id' field-label='Keybase ID')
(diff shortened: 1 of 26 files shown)


jolesbi:
amazing! well done!

jolesbi:
i find the placement of the 'copy' button a bit confusing. what exactly is being copied if you click it? would it be better directly next to the copyable value?

nylira:
Yeah that is a good point. We should put the copy button next to the address.

app/src/renderer/components/staking/PageNominate.vue
```diff
@@ -122,6 +124,7 @@ import FormGroup from '../common/NiFormGroup'
 import FormMsg from '../common/NiFormMsg'
 import FormStruct from '../common/NiFormStruct'
```
jolesbi: don't need the .. anymore

nylira: Yep, will work on a PR to remove all of those relative links (soon).

PR #110, opened 2017-11-24
Delegration (Delegation integration)

### LINKED ISSUE #24: Delegation
As a user
I want to be able to delegate my atoms to a validator
so that I can support the network in validation and earn commission.

AC:
- [x] User is able to view all the validators
- [x] User is able to view the profile of each validator
  - [ ] Validator Name
  - [ ] Validator Website
  - [ ] Validator Description (this can hold links to external ressources like an audit)
  - [ ] atoms self-bonded
  - [ ] atoms delegated to them
  - [ ] commission rate
    - https://github.com/cosmos/gaia/blob/proposals/spec.md#fee-pool-commission
  - [ ] how long they have been validator
- [x] User is able to bond a certain amount of atoms to a selected validator
- [ ] User is able to unbond atoms off a selected validator
- [ ] User is able to see all the validators he has bonded atoms with (and the amount of atoms per validator)
- [ ] User is able to see the amount of commission they have earned

Note:
- Does a user need to unbond all atoms at once or can he unbound partial amounts?

mappum:
Resolves #24 

Sets up a mock API server, with a mocked delegation API which can be un-mocked as soon as it becomes stable. I'm still in the process of integrating the vuex delegation logic with this mock API.


mappum:
I implemented the delegation vuex logic (fetching candidates from API and building txs when delegating) on top of the mock API, so this PR should be ready. Most of the delegation behavior will be inaccessible until we have a way to sign in, which should happen soon with @jolesbi's welcome screen.

Other than some minor UI polish and a way to persist the delegation data in a JSON file, I think we've developed the delegation stuff as far as we can until we wait for the backend to solidify.

app/src/renderer/vuex/modules/candidates.js
```diff
@@ -1,67 +1,38 @@
-import { PubKey } from 'tendermint-crypto'
-
-const CANDIDATE_INTERVAL = 1000 // update every second
+function pubkeyToString (pubkey) {
+  let type = pubkey.type === 'ed25519' ? '01' : '02'
```
faboweb: This would benefit from a comment I think.

app/src/renderer/vuex/modules/candidates.js
```diff
@@ -1,67 +1,38 @@
...
-      state.list.push(candidate)
-      state.map[pubkey] = candidate
 
-      console.log('new candidate', candidate)
+      state.push(candidate)
     }
   }
 
   const actions = {
     async getCandidates ({ commit }) {
```
faboweb: here you want "dispatch" instead of "commit" as of line 26

app/src/main/index.js
```diff
@@ -404,9 +405,14 @@ async function main () {
     await initBaseserver(chainId, baseserverHome, nodeIP)
   }
 
-  log('starting baseserver')
+  log('starting gaia server')
   baseserverProcess = await startBaseserver(baseserverHome)
-  log('baseserver ready')
+  log('gaia server ready')
+
+  if (!TEST) {
```
faboweb: Should we maybe introduce a flag "MOCK"?

app/src/renderer/node.js
```diff
@@ -6,7 +6,8 @@ const RpcClient = require('tendermint')
 let sleep = (ms) => new Promise((resolve) => setTimeout(resolve, ms))
 
 module.exports = async function (nodeIP) {
-  let rest = RestClient()
+  // TODO: once we're done with the mock API server, switch this from port 8999 to 8998
+  let rest = RestClient('http://localhost:8999')
```
faboweb: And maybe use the "MOCK" flag here to so we can't forget about it.

faboweb:
sorting doesn't work for the candidates list

jolesbi:
the name of this PR is 💯

mappum:
@nylira I'm not sure how to get sorting working with this setup, could you help? It looks like the `PanelSort` component changes the appearance of the column headers, but doesn't update the rows in `PageCandidates`.

nylira:
Looking great Matt! @mappum I'll get sorting working in a different PR.

PR #112, opened 2017-11-26
attempt to improve cosmos-ui <select> for windows

### LINKED ISSUE #107: Dropdown/Select Style
The dropdown/select looks fishy on windows:

image

nylira:
I wasn't able to test this because I'm on my Mac laptop and I don't have easy access to Windows currently, but I believe I've styled the options for select to better match the app theme. @faboweb please try it out and let me know.

If it still looks sketchy, we can build an entirely custom `<select>` and `<option>` instead of using HTML5 native controls. This is probably a low priority feature though.

Fixes #107


faboweb:
Approving the update.
I am not sure how the select should look like. Could you maybe post a screenshot into #107?

nylira:
As I said before, I don't have access to Windows so I can't take the screenshot right now. It should fit in with the app better, using `txt` for text color and `input-bg` for background color.

faboweb:
The question was for a screenshot how it looks like on your system, not on windows. So that I can compare the should state with the is state.

PR #111, opened 2017-11-25
spring cleaning (in autumn)

nylira:
* remove the relative Vue and Stylus links in favor of Webpack Resolve.Alias links
* fixed issues where css colors were undefined after switch to dark-theme
* fixed page navigation that wasn't converted to the new desktop-friendly system
* attempted to replace all 1px borders with variable (px = 0.0625rem) borders (for easier adjustment later)
* attempted to replace all fonts defined in pixels with variable font sizes (xs, sm, lg, xl)
* attempted to replace all magic color variables with standard colors
* remove old, commented out code
* removes PageReceive #103
* resolves loading bg not same as app bg #106 

This PR hangs on the the last `npm run test`, not sure what the issue is:

screen shot 2017-11-25 at 11 10 25 am

Also, I can't get gaia to run when I run `npm run testnet`. Is this an issue on develop too?

screen shot 2017-11-25 at 11 12 23 am

PR #113, opened 2017-11-27
components for DataLoading, DataEmpty, and DataError states

### LINKED ISSUE #97: no data states need to be informative and communicative
every route / feature should have messaging when data is not present.

for example, if netmon is down — it should say... 

"Sorry, even though you're running / connected to a full node we can't get display this data for you right now — try again later or contact us to let us know" 

... or some such message. A nice little graphic too would go a long way.

nylira:
@jolesbi pointed out the fact that we need some user-friendly components to let the user know why data isn't loading. Here's what I'm proposing to solve that issue:

## NiDataLoading
An animated spinner and some placeholder text to let the user know that data is loading.

screen shot 2017-11-27 at 9 47 06 am

## NiDataEmpty
For when there's no data yet, even after the loading process (e.g. transactions)

screen shot 2017-11-27 at 10 07 40 am

## NiDataError
Used when there's some error causing the data to not load at all. We're also using this for pages that are not done yet (e.g. Network Monitor index page)

screen shot 2017-11-27 at 9 25 57 am

Resolves #97 

This PR requires #111 to be merged first.


faboweb:
I like the smiley :)

app/src/renderer/components/common/NiDataEmpty.vue
```diff
@@ -0,0 +1,15 @@
+<template lang="pug">
+data-msg(icon="info_outline")
+  div(slot="title") N/A
```
jolesbi: if i understand this correctly, this component will be used when there are no errors but also no data to display. this is a great first step but in this situation we could provide some context + a call to action. 

for example, on the transaction history page it could say ...

```
(slot="title") No Transaction History 
(slot="subtitle") Looks like you haven't sent or received any transactions yet. Head over to the [send page]() to make your first transaction!
```

nylira: This is a great idea. I'll push an update to this effect:

screen shot 2017-11-28 at 12 00 17 pm

PR #139, opened 2017-11-29
Fix navigation on PageProposal

nylira:
screen shot 2017-11-29 at 9 37 52 am

to

screen shot 2017-11-29 at 9 37 38 am


faboweb:
Thx for going the extra mile with multiple screenshot. This makes it really easy to understand the change!

Issue #132, opened 2017-11-28
Validator Profile


faboweb:
As a user
I want to be able to see details for validators
so that I can make sophisticated decisions on delegating my atoms.

AC (profile shows):
- [x] Validator Name
- [x] Validator Website
- [x] Validator Description (this can hold links to external ressources like an audit)
- [x] atoms self-bonded
- [x] atoms delegated to them
- [x] commission rate
  -https://github.com/cosmos/gaia/blob/proposals/spec.md#fee-pool-commission
- [x] how long they have been validator
- [x] max commission rate (the rate that it will never go higher than)
- [x] max commission increase rate (the max amount it can increase over time)

faboweb:
Current state:
![image](https://user-images.githubusercontent.com/5869273/33314727-d38a9b42-d42e-11e7-9732-c2cf8a5b192c.png)

faboweb:
I assume shares equal "atoms delegated to them"? This isn't very clear to me.

jolesbi:
@faboweb that's a good point. i'd like to see more tooltips available with small definitions / descriptions. there is a lot of vocabulary that can be overwhelming / difficult to understand.

jolesbi:
some things missing from the screenshot include ...

- max commission rate (the rate that it will never go higher than)
- max commission increase rate (the max amount it can increase over time)

@rigelrozanski is it possible to display the denomination of the fee shares?

jolesbi:
next release will include `moniker` for validators.

let's make sure we include this field in the UI :)

nylira:
@jolesbi who sets the max commission rate and max commission increase rate? And can the validator make changes to their profile? If they can make changes--for example, fix a typo on their description or website url--what prevents them from changing their max commission rate and increase rate?

nylira:
We have another Validator Profile page on the Network Monitor section of the app.

<img width="778" alt="screen shot 2017-11-30 at 11 50 48 am" src="https://user-images.githubusercontent.com/172531/33410700-cf34eca4-d5c4-11e7-8ab3-2a3963543646.png">

How can we reconcile the differences between the two pages?

nylira:
[...]

PR #142, opened 2017-11-30
Add CommissionMax, CommissionMaxRate, Url to Validator Profile

nylira:
Also improves the fake data by using a fake data generator (`casual`).

screen shot 2017-11-30 at 12 53 58 pm

Part of #132


faboweb:
Very cool idea with the generator! :)

Issue #128, opened 2017-11-28
Simplify NiAppMenu


nylira:
With the recent removal of viewing lists of delegators, the receive page, and the nomination page, the sidebar seems overloaded with section titles. The page that do exist are a little empty and the content seems too thinly spread out when they should be shown together.

## Current Navigation

* **WALLET**
  * Balances
  * Send
  * Transactions
* **GOVERN**
  * Proposals
* **STAKE**
  * Delegate
* **MONITOR**
  * Blockchain
  * Validators

## Proposed Navigation
I want to propose a primary app navigation with just four buttons.

### 1. Wallet

The wallet page shows your primary address and balances on the top of the page. Below that is a list of most recent 20-30 transactions.

*Page Nav*: [Index], Transaction History (this page is a longer version of the recent transactions list)

*Page Actions*: Send Tokens (pressing "Send Tokens" will create a modal popup allowing you to fill in details about the tx).

### 2. Govern

The govern page shows a list of current proposals.

*Page Nav*: [Index], Finished Proposals (finished proposals are proposals that have already either been accepted or rejected)

### 3. Stake
The delegation game is not a thing anymore. The most important thing users will want to see on the Stake page is their current basket of validators (it's sort of like an index fund).

*Page Nav*: [Index], Edit Basket of Validators (this page shows a detailed, filterable list of all current validators)

### 4. Monitor
The monitor page shows important blockchain details at the top, and then a list of validators below that. The list of validators can be sorted and filtered.

*Page Nav*: [Index], Search Blockchain (this page will let you search for transactions, blocks, etc. in the cosmos network)

Thoughts on this?

jolesbi:
@nylira do you still want to discuss this / change the app architecture?

nylira:
Yeah, let's change the navigation, and make everything nouns. My updated proposal:

* Balances - actions: view/filter balance, view/filter transactions, create new sendtx
* Proposals - actions: view proposals, filter proposals, create proposal
* Delegates - actions: view delegates, filter delegates, bond to delegates
* Monitor - actions: view status of the network, search for tx
* User - actions: view profile, edit account settings, change account

There is merit to making a Transactions page as a main navigation item. I think it can cause confusion though. Let's say you're a user and you want to send some tokens (aka create a transaction). Would you go to the Balances page or the Transactions page? It makes the most sense to merge these pages.

nylira:
It would be great to have 4 custom icons, one for each page of the app.

[...]

PR #143, opened 2017-11-30
Improve Delegate Index View

nylira:
This is not the final design we need for the Delegate page, but it's an improvement. What is better:

* Shows useful data like voting power, country, commission
* Sorting works
* Removed some `v-if="signedIn"` qualifiers because the user will already be signed in if they can access this page.

Old:
screen shot 2017-11-30 at 1 56 20 pm

New:
screen shot 2017-11-30 at 2 06 39 pm

Part of #132


jolesbi:
awesome.

app/src/renderer/components/staking/CardCandidate.vue
```diff
@@ -1,25 +1,25 @@
 <template lang='pug'>
-transition(name='ts-card-candidate'): div(:class='cssClass')
+transition(name='ts-card-candidate'): div(:class='styles')
   .card-candidate-container
     .values
-      .value.keybaseID
```
jolesbi: are we not using keybase? or just not important to display here?

jolesbi:
in the screenshot the "keybase ID" column is displaying names not ID's...?

app/src/renderer/components/staking/PageCandidates.vue
```diff
@@ -44,8 +44,8 @@ export default {
   computed: {
     ...mapGetters(['candidates', 'filters', 'shoppingCart', 'user']),
     pageTitle () {
-      if (this.user.signedIn) return `Candidates (${this.candidatesNum} Selected)`
-      else return 'Candidates'
+      if (this.user.signedIn) return `Delegate (${this.candidatesNum} Selected)`
```
jolesbi: i'm not sure about this. we need to confirm — but i think candidate is still the correct term.

nylira:
@jolesbi keybaseID can be a name. E.g. mine is [https://keybase.io/peng](https://keybase.io/peng). I don't think the system is case sensitive but it's okay because these keybaseIDs are fake.

jolesbi:
the fake data generator is cool.

app/src/renderer/components/staking/PageCandidates.vue
mappum: This is right, delegate is being used as a verb here. If that's not clear then maybe it could say `Delegate (X Candidates Selected)`.

nylira: Should it say "Candidates Selected" or "Validators Selected"? I feel like having the "Candidates" label doesn't make sense. It will just cause users to be confused. If a user stakes their atoms to a Candidate instead of a Validator, what happens to their staking income? Only the percentage of atoms they stake to Validators will earn them staking atoms? If that is the case, who would ever stake atoms to a Candidate?

nylira: Separately: the page name is Delegate, so the page title should also be Delegate. If the user clicks on "Delegate" and they go to the "Candidates" page, it's confusing. That's why I made this change. This relates to issue #128, where I made the case that we should settle on either "Stake" or "Delegate" as the page name instead of using both terms.

mappum: Definitely "Candidates". You would delegate to a non-validator candidate if you want to help push them up into the top 100. There is probably a bias to delegating to validators, but users should be encouraged to delegate to the candidates they think would make best validators to ensure the network is secure.

Related to that, we will need to ensure the default sorting order doesn't make any candidates unfairly get more visibility (but it's not obvious what the best order is). Will probably also want to warn when delegating to a validator with too much voting power so they don't get an unsafe amount of power.

nylira: Sounds good. Let's use "Delegate (X Candidates Selected)".

Yeah we'll have to experiment with different sorting methods.

jolesbi: yep. good conversation here! the challenge is that validators and validator candidates are dynamic. so sometimes you're going to delegate to a validator while other times you could be delegating to a candidate. this could even change in the time it takes to type in the amount and submit...

i'm not sure the right way to think about this. and the language around it is challenging. an analogy...

- a politician is someone running for mayor (a candidate)
- a politician is someone who is the mayor (a validator)

they are both politicians. we don't have a word that covers both which i think is the source of the trouble.

Issue #35, opened 2017-11-07
SDK Requirements


faboweb:
* historical transactions
  * send/receive in one transactions
  * filter by
    * block-height (start and end)
    * limit (default 50)
    * cap by some maximum
  * output in JSON
  * stream new transactions
    * MVP: refresh button
    * desired: via Websockets
* delegation
  * all candidates

jolesbi:
i think ideally this would link to ticket #'s in the SDK repo... 🌮

rigelrozanski:
Awesome I will keep this is mind

faboweb:
Ethan and me aligned on following requirements for the LCD:

The LCD provides following endpoints:
// for wallet transactions
`GET /tx/coins/{address of sender or receiver}`

// for staking transactions
`GET /tx/bondings/delegator/{address of delegator bonding / unbonding}`
`GET /tx/bondings/validator/{address of validator delegators bond to or unbond from}`

// for fee withdrawal transactions
`GET /tx/cashout/delegator/{address  of delegator}`
`GET /tx/cashout/validator/{address  of validator}`

// for proposal transactions
`GET /tx/votes/{proposal}`

All of those queries take following optional arguments:
`
minheight,
maxheight
`

Coin-transactions look like this:
```
[{
  tx: {
    inputs: [{
      sender: String,
      coins: [{
        amount: Number,
        denom: String
      }]
    }],
    outputs [{
      receiver: String,
      coins: [{
        amount: Number,
        denom: String
      }]
    }]
  },
  time: DateTime,
  height: Number
}]
```

Tags to index those are:
`
coin.sender,
coin.receiver,
height
`

Bonding-transactions look like this:
```
[{
  tx: {
    type: "bond" | "unbond",
    validator: String,
    address: String, // of delegator
    shares: Number
  },
  time: DateTime,
  height: Number
}]
```

Tags to index those are:
`
bond.delegator,
bond.validator,
height
`

Cashout-transactions look like this:
```
[{
  tx: {
    address: String, // of validator
    receiver: String,
    outputs [{
      address: String,
      coins: [{
        amount: Number,
        denom: String
      }]
    }]
  },
  time: DateTime,
  height: Number
}]
```

Tags to index those are:
`
cashout.delegator,
cashout.validator,
height
`

We probably need as well:

// for proposals
`GET /proposals?status='active'|'closed'`

// TO BE DEFINED

// for candidates
`GET /candidates`

[...]

PR #146, opened 2017-12-01
Mocked Transaction History

mappum:
This PR hooks up the transaction history to the newly proposed transaction history API (#35), and adds some of those proposed APIs to the mock server.

### DIFF
--- a/app/src/renderer/components/wallet/PageTransactions.vue
+++ b/app/src/renderer/components/wallet/PageTransactions.vue
@@ -43,13 +43,13 @@ export default {
     ...mapGetters(['filters', 'transactions']),
     filteredTransactions () {
       let query = this.filters.transactions.search.query
-      let list = orderBy(this.transactions, ['id', 'desc'])
+      let list = this.transactions
 
-      if (this.filters.transactions.search.visible) {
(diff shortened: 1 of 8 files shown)


nylira:
Are the mocked transactions supposed to be showing up in the UI? I can't see any, and I also have this error in the console:

screen shot 2017-12-01 at 1 07 13 pm

mappum:
Make sure you reinstall the dependencies, I updated the `cosmos-sdk` module dependency.

app/src/renderer/components/wallet/PageTransactions.vue
```diff
@@ -40,16 +41,10 @@ export default {
...
-      let query = this.filters.transactions.search.query
-      let list = orderBy(this.transactions, ['id', 'desc'])
-
-      if (this.filters.transactions.search.visible) {
-        return list.filter(i => includes(i.id.toLowerCase(), query))
-      } else {
-        return list
-      }
+      return this.transactions
+      // TODO: restore searchability? (what part of the tx are we searching?)
```
faboweb: full text? so amount, sender, receiver. It is a local search anyway.

PR #154, opened 2017-12-02
sidebar and li improvements

jolesbi:
aiming for more consistency in styles, clearer information architecture, better readability

i also changed the words in the sidebar a little bit for the reasons above. i still think there is room for improvement.

@nylira i adjusted the notification colours slightly — i found them just a tiny bit too electric. happy to discuss. also, added a secondary border color which i think really helps.

screen shot 2017-12-01 at 4 59 36 pm


app/src/renderer/styles/variables.styl
```diff
@@ -43,15 +44,16 @@ sans = -apple-system, BlinkMacSystemFont, "Segoe UI", "Roboto", "Oxygen", "Ubunt
...
-warning = hsl(60,100%,50%)
-danger = hsl(0,100%,50%)
+warning = hsl(52,100%,50%)
+danger = hsl(0,100%,40%)
 
 // sizes
 aw = 64rem
 width-main = 100%
 width-side = 16rem
+max-width = 600px
```
nylira: Are we using this anywhere? And if we are, can it be a multiple of 1rem? (16px)

app/src/renderer/styles/variables.styl
```diff
@@ -25,6 +25,7 @@ dim = hsl(app-hue, 13%, 50%)
 
 // borders
 bc = hsl(app-hue, 22%, 23%)
+bc-dim = app-fg
```
nylira: Really like the look of this border color for reducing visual noise.

app/src/renderer/styles/variables.styl
```diff
@@ -15,7 +15,7 @@ accent = hsl((mhue + app-hue)/2, 96%, 62%)
 app-bg = hsl(app-hue, 36%, 13%)
 
 app-bg-alpha = alpha(app-bg, 92%)
-app-bg-hover = hsl(app-hue, 25%, 11%)
+app-bg-hover = hsl(app-hue, 33%, 16%)
```
nylira: Yes, it can use more contrast here.

app/src/renderer/components/common/NiPart.vue
```diff
@@ -36,16 +36,16 @@ export default {
 .ni-part-container
   margin 0 auto
 
+  &:last-child
+    padding-bottom: 1.5rem
```
nylira: Can we get this to affect 768px and higher? The padding is too much on small screens. Looks fantastic on larger screens though.

nylira:
I really like these improvements for desktop -- the extra space gives the design room to breathe, and the faint border color is less harsh. The `active` style for the sidebar looks good too. 

One thing - can we can get the `li` padding to only affect only 768px and above though? It's too roomy for smaller displays:

screen shot 2017-12-04 at 10 39 52 am

app/src/renderer/components/common/NiPart.vue
jolesbi: yep!

app/src/renderer/styles/variables.styl
jolesbi: awesome.

app/src/renderer/styles/variables.styl
jolesbi: using it for forms. yep. i can switch it rem's

PR #168, opened 2017-12-05
Improve transaction pages

### LINKED ISSUE #148: Transaction History Design
Current UI:
image

Wasn't there already a more sophisticated design by Peng?

nylira:
* redesigned CardTransaction and renamed it to LiTransaction. Reasoning: Cards can be laid out in a 2D grid, while ListItems are laid out in one dimension only
* the new LiTransaction layout works for small and large screens
* the new layout supports one to infinite coins per transaction
* there is a hover effect that links to the TODO: transaction view page
* updated the transaction colors to match the site (@jolesbi should we use these colors or `danger` and `success`? I was thinking to relegate`danger` and `success` only to error/achievment states, but it is also common to show incoming transactions in green and outgoing in red)
* updated the associated test

## Old Design (with multiple coins)

screen shot 2017-12-05 at 2 22 20 pm

## New Design (with multiple coins)

screen shot 2017-12-05 at 1 51 21 pm

Fixes #148


faboweb:
I like the structure I like the added Icons!
I feel the separation between transactions is not so visible. Also the contrast for the positive values in comparison to the background feels to less.
One idea: What about putting the icons on the first row always? Having them centered feels chaotic.

I feel not strong about any of this and we could also first merge and adjust later.

app/src/renderer/components/wallet/LiTransaction.vue
```diff
@@ -0,0 +1,179 @@
...
+    date () {
+      return dateUnixAgo(this.transactionValue.time)
+    }
+  },
+  data: () => ({
+    num: num
+  }),
+  methods: {
+    viewTransaction () {
+      this.$store.commit('notify', {
```
faboweb: :+1:

nylira:
@faboweb I like your suggestions, and have thought about those issues too. Let's do it in the future. Merging this.

PR #147, opened 2017-12-01
trying to spruce up our native experience

jolesbi:
- changed some config language
- added some icons

closes #138

Issue #138, opened 2017-11-28
electron professional stuff


jolesbi:
- [ ] correct icons
- [ ] name the app

PR #170, opened 2017-12-05
Maximize Window on Start

faboweb:
Relates to #138 
Pulled out from #147


faboweb:
I am opposed to this. This feels unnatural for me. Apps I know don't maximize on start.

nylira:
During development I have the dev tools open nearly all the time -- so I like having the window maximized. However, I agree that it shouldn't be a production default. Is there a way to specify develop only?

faboweb:
@jolesbi did you close this by accident? I added https://github.com/cosmos/cosmos-ui/pull/170/commits/fe62a1293e3565604a759ab6f57e675ba5315f7f

PR #184, opened 2017-12-06
Jordan/send fixes

nylira:
* default coin is set on the Send page if there is only one coin available
* removes the currently useless "Max" button
* improves DataEmpty message


faboweb:
New snapshot shows 'None Available'. Should not be the case.

app/src/renderer/components/wallet/PageSend.vue
```diff
@@ -123,6 +121,11 @@ export default {
     },
     ...mapActions(['walletSend'])
   },
+  mounted () {
+    if (this.denominations.length === 1) {
+      this.fields.denom = this.denominations[0].value
```
mappum: Nice 👍

jolesbi:
@faboweb can you take a look at the  `PageSend` tests? i can't figure out why the onSubmit is not working as expected. also, should we even be calling `onSubmit`?

PR #195, opened 2017-12-07
Fabo/144 sign in hookup

faboweb:
We need to decide what happens if you login to a new account. Currently we assume just one account 'default'. I created a temporary version where the user is able to remove this account.

If the user has an account 'default'
image
image

If the user does not have an account 'default'
image


faboweb:
Tests are still missing.

faboweb:
Currently the login screen will appear every time the user opens the app.

mappum:
Good job on this @faboweb 👏

I'll look through this PR in more detail, but I think the concept of the initial sign in screen makes sense

jolesbi:
@faboweb what do you mean 'default' account? i think every user should see the same welcome modal no matter what.

jolesbi:
also, let's hide the hardware option (and any others) since it doesn't actually work for now.

nylira:
@jolesbi the hardware sign in is still being hidden with the `devMode` option, so that's fine.

nylira:
@jolesbi by default account, Fabo means the account the user created in a previous session. 

Let me know if I'm wrong, but this is how it should be working:

When the user downloads the UI and runs it for the first time, they will see "Create new account"/"Restore account from seed" options. "Sign in with hardware" will be hidden with the `devMode` flag. Why this happens: because the folder `~/.cosmos-ui` has not been created yet.

When the user exits the app and starts it again, they will see the "Sign in with password"/"Remove account" options. This is the "default account" @faboweb means.

mappum:
@jolesbi and I just talked for a bit, and we thought of a change that could simplify this flow a bit. We can increment the app version in the `package.json`, which will make the app automatically "Remove account" the first time they run this release, removing the need for the first screenshot above.

Also, when running the app after that, we can detect an existing account and go right into the "Sign in with password" prompt, with an option to back out to the initial welcome screen (last screenshot) if the user wants to switch accounts.

I'll make a PR for this change on top of this PR, @faboweb @nylira let me know what you think.

nylira:
@mappum that sounds like a fine change. However, are you able to successfully create a new account through this flow currently? I can't do it. Am I on the wrong version of gaia or something?

screen shot 2017-12-08 at 11 01 58 am

mappum:
>Am I on the wrong version of gaia or something?

Yes, the `keys/recover` endpoint which is giving the error was added in a change @faboweb made to cosmos-sdk, which is not currently merged into gaia. So we'll have to make a PR to gaia before we can really test this.

PR #198, opened 2017-12-07
moved mobile hamburger menu to right side, got rid of fixed toolbar

### LINKED ISSUE #181: toolbar can stay in the topbar on mobile
@nylira i think it would be better to leave the toolbar in the top bar on mobile. this will provide a more consistent experience.

jolesbi:
closes #181 

screen shot 2017-12-07 at 1 57 00 pm


jolesbi:
let's wait for @nylira on this one!

nylira:
I still don't agree with putting the action buttons up in the title bar for smaller screens.

A lot of the title gets cut off here, including how many candidates are currently selected. You are no longer able to click the "Delegate" button while halfway down the page:
screen shot 2017-12-08 at 11 04 59 am

You can no longer search through proposals or create a new proposal while scrolling down this page:
screen shot 2017-12-08 at 11 06 26 am

You can no longer go back or create a comment on a proposal while scrolling:
screen shot 2017-12-08 at 11 08 33 am

app/src/renderer/components/common/AppHeader.vue
```diff
@@ -66,7 +66,7 @@ export default {
     position fixed
     top 0
     left 0
-    z-index 100
+    z-index 2000
```
nylira: 1000 is the z-index of modals - I'm going to revert this one because I don't think the app header should show on top of modals

nylira:
I'm strongly for keeping the fixed toolbar for small widths. However, let's make it only apply to displays < 568px. Displays wider than that can handle the inline toolbar.

jolesbi:
ok i see your point. thanks for the merge!

PR #199, opened 2017-12-07
 improved contrast ratio of colours based on accessibility test

### LINKED ISSUE #167: audit colours for accessibility and contrast ratio
recommend we use lea verou's tool to establish a WCAG 2.0 compliant colour scheme (DARK-THEME FO LYFE) 🎨 

i'll make a PR with the closest colours from `variables.styl` that pass the accessibility contrast check at any font-size on our background colour. 

http://leaverou.github.io/contrast-ratio/

jolesbi:
i modified the colours to make sure they all rank well in terms of contrast ratio and accessibility. here are the "new" colours with very good rankings. 

ideally, all our colours for text would be above 7.0 (AAA). i lightened the background from 13% to 16% so the scores are a bit off. but this is more or less what i'm proposing.

screen shot 2017-12-07 at 2 55 18 pm
screen shot 2017-12-07 at 2 49 10 pm
screen shot 2017-12-07 at 3 02 52 pm
screen shot 2017-12-07 at 3 02 45 pm
screen shot 2017-12-07 at 2 41 01 pm
screen shot 2017-12-07 at 2 40 15 pm
screen shot 2017-12-07 at 2 39 32 pm

closes #167


nylira:
Really liking these changes in general! It's looking good and easier to read. I'm going to revert the background lightening and bright text darkening because I feel like the design loses a lot of 'pop' and contrast with those changes. Also going to revert the button change because they should be done in the `@nylira/vue-button` lib, and I have some ideas for improving that (in a separate pr)

PR #200, opened 2017-12-07
use status colors for transactions

jolesbi:
screen shot 2017-12-07 at 3 50 25 pm 1


jolesbi:
you're right. let's address this in the next release when transaction history will be in. sorry for this distraction.

Issue #144, opened 2017-11-30
Test SignIn Methods / Hook them up


faboweb:
Follow up of #54 

Requirements were:
* create new account / generate new key (w optional mnemonic)
* log in with hardware / ledger wallet
* log in with mnemonic (from fundraiser)
* log in with local keystore / JSON file

jolesbi:
addressed by #202 and #208

PR #202, opened 2017-12-08
Sign-in Changes

mappum:
Implemented the changes mentioned in #144:
>@jolesbi and I just talked for a bit, and we thought of a change that could simplify this flow a bit. We can increment the app version in the package.json, which will make the app automatically "Remove account" the first time they run this release, removing the need for the first screenshot above.
>
>Also, when running the app after that, we can detect an existing account and go right into the "Sign in with password" prompt, with an option to back out to the initial welcome screen (last screenshot) if the user wants to switch accounts.

Also, updated the `cosmos-sdk` client dependency so we can use the `recoverKey` method (we will still need to merge the key recovery into gaia and update to that version).


nylira:
On initial load, I'm supposed to be seeing either the sign-in-with-pw page or the create/restore/hardware welcome page right? I'm seeing this instead:

screen shot 2017-12-08 at 1 22 20 pm

PR #206, opened 2017-12-08
Make sends more reliable

mappum:
@ebuchman was testing out the send functionality by spamming sends, and found that the account nonce was not being properly incremented with each successive transaction, so only the first one was actually going through (even though the UI showed success). This PR ensures we see error results on tx send, ensures the account nonce properly increases as we send, and puts sends in a queue to ensure only one happens at a time.


ebuchman:
@mappum what happened to adding tests here? Too difficult to test ? If so that's something we should probably work on since this is relatively critical functionality

mappum:
Will add tests before releasing

PR #208, opened 2017-12-08
Fix sign in flow

faboweb:
Following up on:

> Proposed flow by https://github.com/cosmos/cosmos-ui/pull/202 doesn't work. If you have an account, you can not recover another account or create a new one. All account creations write the  account 'default' to disc. If there is already such an account, this will fail. Therefor the user actively needs to remove the old 'default' account before signing in to  a new account.
If you want to show the signin-password view directly we need to also provide a way to "change" (remove) the account from there.

Welcome:
image

New Account:
image

Default if account 'default' exists:
image

Delete account (leads back to sign in with password):
image


jolesbi:
@faboweb thanks for taking care of this.

what is the problem with what @nylira said in [this PR](https://github.com/cosmos/cosmos-ui/pull/202)?
>On initial load, I'm supposed to be seeing either the sign-in-with-pw page or the create/restore/hardware welcome page right?

having "remove account" buttons is really a very strange user experience at this stage of the welcome flow. 

>If you have an account, you can not recover another account or create a new one.
...
- if user has default account welcome flow starts on "enter password"
- if user doesn't have default account then welcome flow starts on initial screen without option to sign in

what's the problem with the above points? is it the scenario where someone has a default account but cannot find / remember their password? in this situation — we should let them create a new account as if everything is 😎 and then change the default account behind the scenes.

🕺 🍪 💰 🌵

faboweb:
I don't feel comfortable with letting users delete their account without confirming this explicitly

jolesbi:
ok that makes sense.

why do we have to delete accounts? can't we just change the default?

ebuchman:
Why do we need to remove accounts ?

faboweb:
We have one active account currently. The account with the name 'default'
shows us, that there is already an account created. To switch accounts, we
delete this account 'default' and then create a new one. To prevent the
user from accidentally overwrite an existing account, the user has to
actively remove this account 'default'.
When we have the possibility to choose the account you want to use (like in
the CLI), this will probably not be needed anymore! But we want to lunch.
Multiple accounts can probably be part of the next release if desired.

Ethan Buchman <notifications@github.com> schrieb am Fr., 8. Dez. 2017,
20:10:

mappum:
>I don't feel comfortable with letting users delete their account without confirming this explicitly

I agree, which is why the version check doesn't actually delete anything, it just backs up the old data by copying it to a new directory. That gives us the best UX plus no risk of deleting data that might be important

faboweb:
But the version check would only back it up once right?

Matt Bell <notifications@github.com> schrieb am Fr., 8. Dez. 2017, 20:35:

mappum:
>But the version check would only back it up once right?

Oh right, so it only handles the case of people who might have an old key (with the old hard coded password).

For switching accounts, what do you think of automatically backing up the existing `default` key when the user creates/recovers a new key? We could use the key recovery to rename the key to 'default1', 'default2', etc. Then as a benefit, once we have a UI for key selection, the user could select from those keys.

ebuchman:
So do we even need thi "delete" function for this release? why not just force one key for now and we can introduce key names for next release ?

mappum:
I'm OK with how this PR works for now, I think it will be rare that a user wants to switch accounts on this release (but if they do, the option is there).

>why not just force one key for now and we can introduce key names for next release ?

We do force one key right now, which is the reason for the key deletion (for switching to another key)

ebuchman:
ok sure!

PR #218, opened 2017-12-11
Fix bonding/unbonding ui

nylira:
* fixed the bonding/unbonding mockup in the UI
* users have a hardcoded atom count they can allocate to validator candidates
* form validations and description text for too few and too many atoms allocated
* validator candidates can be deleted from the shopping cart, with confirmation
* todo: testing @faboweb 
  * can you help me figure out why `PageDelegate.spec.js` is failing? I think it's related to vuex and shoppingCart.js 
* todo: @mappum 
  *  create mockServer.js api calls for bonding/unbonding
  * can you help me figure out why the form breaks after a successful allocation form submission? It seems related to `async submitDelegation ()`

## Start State
screen shot 2017-12-11 at 2 10 54 pm

## User starts delegating
screen shot 2017-12-11 at 2 11 08 pm

## Delegating more than exists
screen shot 2017-12-11 at 2 11 03 pm

## Success state
screen shot 2017-12-11 at 2 20 57 pm


mappum:
todo: @mappum
  - [x] create mockServer.js api calls for bonding/unbonding
  - [ ] can you help me figure out why the form breaks after a successful allocation form submission? It seems related to `async submitDelegation ()`

>create mockServer.js api calls for bonding/unbonding
We already have it one step better, the real rest API is in gaia :)

app/src/renderer/components/common/AppMenu.vue
```diff
@@ -8,7 +8,7 @@ menu.app-menu
     part(title='Governance' v-if="config.devMode")
       list-item(to="/proposals" exact @click.native="close" title="Proposals")
     part(title='Stake' v-if="config.devMode")
-      list-item(to="/staking" exact @click.native="close" title="Delegates")
+      list-item(to="/staking" exact @click.native="close" title="Delegate")
```
jolesbi: hahaha. we keep switching this back and forth. the reason i changed it to `Delegates` with an 's' is for a couple reasons.

1. Transactions, Proposals, Validators, `Delegates` (nouns) ... but i guess i see how `Delegate` fits with send (verbs)

2. when you click the link `Delegates` what you see is a list of delegates (noun), not a screen for delegating (verb).

this highlights a bigger issue that you brought up in #128 and that i've been thinking about — ideally, our sidebar would use consistent language. actions (verbs) should take place within the context of important nouns (Balances, Transactions, Proposals, Delegates, Validators). your idea to include send as a modal on balances makes a ton of sense to me for this reason. also, why `Delegate`, the action, makes sense as a feature modal of the `Delegates` page. bonding ATOMs is a type of transaction and should have a similar UX to sending.

see also #223

app/src/renderer/components/staking/PageCandidates.vue
```diff
@@ -44,7 +44,7 @@ export default {
   computed: {
     ...mapGetters(['candidates', 'filters', 'shoppingCart']),
     pageTitle () {
-      return `Delegates (${this.candidatesNum} Candidates Selected)`
+      return `Delegate (${this.candidatesNum} Candidates Selected)`
```
jolesbi: I don't think "Candidates Selected" belongs in the page title.

app/src/renderer/components/wallet/PageSend.vue
```diff
@@ -91,15 +91,15 @@ export default {
       this.sending = false
       this.$v.$reset()
     },
-    onSubmit () {
+    async onSubmit () {
```
jolesbi: what does this have to do with bonding/unbonding?

app/src/renderer/vuex/modules/user.js
```diff
@@ -13,7 +13,7 @@ export default ({ commit, node }) => {
   }
 
   const emptyUser = {
-    atoms: 0,
+    atoms: 2097152,
```
jolesbi: can't we use real data here yet?

app/src/renderer/components/staking/PageDelegate.vue
```diff
@@ -1,37 +1,33 @@
 <template lang="pug">
-page(:title="pageTitle")
+page.page-delegate(title="Delegate Atoms")
   div(slot="menu"): tool-bar
     router-link(to='/staking')
       i.material-icons arrow_back
       .label Change Candidates
```
jolesbi: i think back arrows should always say "back"

app/src/renderer/components/wallet/PageSend.vue
faboweb: This was me improving some tests while fixing the tests. Should have maybe gone into another PR.

app/src/renderer/components/staking/PageDelegate.vue
nylira: Agreed.

app/src/renderer/vuex/modules/user.js
nylira: Not yet. We can put it in mockServer.js though, @mappum ?

app/src/renderer/components/common/AppMenu.vue
nylira: Read your reasoning -- let's change it to Delegates.

app/src/renderer/components/staking/PageCandidates.vue
nylira: Let's put it in the label for the Bond button instead.

nylira: Hmm, let's keep this here until we design a new component to put this in.

nylira:
I did some renaming as suggested on #223 for improved UX & clarity. There are improvements that can be made. Primarily with real atoms. But PR is ready to be merged.

* Validator Candidates, Candidates -> Delegates
* Delegating, Allocating, Staking -> Bonding

nylira:
@jolesbi does this PR still need changes?

Issue #194, opened 2017-12-07
Dev-Mode flag from environment variable


faboweb:
I am missing the option to set NODE_ENV=development/production from outside the app. So I can show both versions using one app version.

PR #220, opened 2017-12-11
added linux release possibility

faboweb:


faboweb:
Noticed again that the hardcoded devMode flag could be a source of future error. Pushing for #194

PR #224, opened 2017-12-11
Multiple accounts

mappum:
This change was originally made by @jolesbi, but I tweaked a few things.

Now when users sign in, they can choose their account instead of always using "default". I'd also be open to the account selection being a row of buttons, so that all choices are exposed without having to dropdown.


nylira:
Made some slight alterations for better usability:
* signUpName has working form validations, minLength: 5
* forms have a flexible height if they're too long
* dimmer dim color because it looks too similar to normal text

screen shot 2017-12-12 at 12 53 41 pm
screen shot 2017-12-12 at 12 53 34 pm

app/src/renderer/vuex/modules/user.js
```diff
@@ -62,41 +63,42 @@ export default ({ commit, node }) => {
     // to create a temporary seed phrase, we create a junk account with name 'trunk' for now
     async createSeed ({ commit }) {
       let JUNK_ACCOUNT_NAME = 'trunk'
+      let TRUNK_PASSWORD = '1234567890'
```
faboweb: We need to get rid of this. A malicious user could access the last generated trunk account with the standard password.

faboweb: Prerequesit: https://github.com/cosmos/cosmos-sdk/issues/301

jolesbi: do you mean in this PR or an another PR? this is not new — this is currently how it works...

faboweb: Just a note. I commented on that, because I implemented it like that and wanted to share the thought.

jolesbi:
@nylira the tests are still broken 😅

test/unit/specs/NISessionWelcome.spec.js
```diff
@@ -23,18 +25,32 @@ describe('NISessionWelcome', () => {
...
     wrapper.findAll('.ni-session-header a').at(1).trigger('click')
     expect(store.commit.mock.calls[0]).toEqual(['setModalHelp', true])
   })
 
+  it('should not show sign-in link since we have no accounts', () => {
+    wrapper.find(LiSession).trigger('click')
+    expect(store.commit.mock.calls[0][1]).not.toEqual('sign-in')
+  })
+
+  it('(set up accounts for following tests)', () => {
```
faboweb: 'it' tests should be independent

app/src/renderer/components/wallet/PageBalances.vue
```diff
@@ -19,7 +19,7 @@ page(title='Balance')
       :key="i.denom"
       :dt="i.denom.toUpperCase()"
       :dd="i.amount")
-    list-item(v-if='wallet.balances.length === 0' dt="N/A" dd="None Available")
+    list-item(v-if='wallet.denoms.length === 0 && wallet.balances.length === 0' dt="N/A" dd="None Available")
```
faboweb: why not use allDenomBalances.length?

test/unit/specs/NISessionSignIn.spec.js
```diff
@@ -70,9 +73,10 @@ describe('NiSessionSignIn', () => {
   })
 
   it('should show a notification if signin failed', async () => {
-    store.dispatch = jest.fn(() => Promise.reject('Planed rejection'))
```
faboweb: Thx for correcting always my spelling. ^^ Will get a spellchecker for VSCode

Issue #51, opened 2017-11-10
Building the app on Windows


faboweb:
Currently the wallet doesn't run on windows.

PR #234, opened 2017-12-13
added windows build

faboweb:
Implements: #51 

We can merge this as is.

Also implements: #236 

Currently the installer looks ugly and could benefit from a cosmos gif: https://github.com/cosmos/cosmos-ui/issues/235


app/src/main/index.js
```diff
@@ -130,31 +130,7 @@ function createWindow () {
   webContents.on('will-navigate', handleRedirect)
   webContents.on('new-window', handleRedirect)
 
-  // setup menu to handle copy/paste, etc
-  var template = [
```
jolesbi: why are you removing this?

faboweb: This removes the menu: https://github.com/cosmos/cosmos-ui/issues/236

PR #240, opened 2017-12-13
Fix app footer to bottom of screen

mappum:
I fixed the app footer to the bottom of the screen, since on long pages like the balance page it gets hidden and nobody ever sees it. If the screen is very small, the footer hides to show the action buttons instead.

@nylira, let me know what you think.

https://gfycat.com/GreatOrganicLhasaapso


nylira:
Made a tweak to make sure the right side of the footer doesn't extend past the screen, and made sure it doesn't overlap with the app menu.

PR #250, opened 2017-12-18
Show Candidates

faboweb:
Real candidates can now be queried

image

### DIFF
--- a/app/src/renderer/vuex/modules/delegates.js
+++ b/app/src/renderer/vuex/modules/delegates.js
@@ -1,17 +1,12 @@
-function pubkeyToString (pubkey) {
-  // go-wire key encoding,
-  // ed25519 keys prefixed w/ 01, secp256k1 w/ 02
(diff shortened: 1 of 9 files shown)


app/src/renderer/vuex/modules/delegates.js
```diff
@@ -30,7 +25,9 @@ export default ({ dispatch, node }) => {
       }
     },
     async getDelegate ({ commit }, pubkey) {
-      let delegate = (await node.candidate(pubkeyToString(pubkey))).data
+      let delegate = (await axios.get('http://localhost:8998/query/stake/candidate/' + pubkey.data)).data.data
```
jolesbi: `pubkey.data)).data.data` 🥁

faboweb: :D :dancer:

jolesbi:
@nylira have you noticed that San Francisco is not on Windows... maybe we should consider loading SF instead of relying on system fonts...?

PR #257, opened 2017-12-19
Hooking up bonding/unbonding to wallet

mappum:
Still needs unbonding, which we can't do yet since the data about which candidates we have bonded to doesn't seem to be working properly in gaia.


mappum:
This PR is functional now: users can go through the staking flow, restart the app, and then edit their existing delegations. But before we merge, it would be nice to rethink the current UX.

Until now, we made the staking flow only keeping in mind the situation where a user has never bonded before. With the changes in this PR, users can go back to PageBond and see their current delegation amounts, edit them (increasing or decreasing delegations) and hit 'Bond' again. Any amount decreased from previous delegations are now in the unbonding process and aren't liquid for a month.

So basically what I'm thinking is the language in PageBond just needs to be changed to explain about coins going into the unbonding period, and that the amounts aren't the amount being deducted but the balance of bonded coins after the user submits.

@jolesbi @nylira @faboweb Any ideas about how to communicate this clearly?

faboweb:
I think we need to show Atoms in their tristate like: Total [Available, Bonded, Unbonding]

faboweb:
For current error 'Cannot set property '_error' of undefined' merge #270

jolesbi:
i think we should merge this (if it's ready) and make another ticket to improve the UX.

mappum:
This PR is ready

app/src/renderer/components/staking/PageBond.vue
```diff
@@ -66,65 +70,59 @@ export default {
...
-    atomsMin: 1,
-    reservedAtomsMin: 0,
+    atomsMin: 0,
     fields: {
-      reservedAtoms: 0,
       delegates: []
     }
   }),
   methods: {
     fillAtoms (delegateId) {
```
jolesbi: what is `fillAtoms`?

app/src/renderer/components/staking/PageBond.vue
```diff
@@ -66,65 +70,59 @@ export default {
...
     clearAtoms (delegateId) {
-      if (delegateId === 'unreserved') {
-        console.log('clearing reserved atoms')
-        this.fields.reservedAtoms = 0
-      } else {
-        console.log('clearing atoms for', delegateId)
-        let delegate = this.fields.delegates.find(c => c.delegateId === delegateId)
-        delegate.atoms = 0
-      }
+      console.log('clearing atoms for', delegateId)
```
jolesbi: remove console.log

app/src/renderer/components/staking/PageBond.vue
```diff
@@ -144,11 +142,7 @@ export default {
...
     },
     async onSubmit () {
-      if (this.unbondedAtoms === this.user.atoms) {
-        this.$store.commit('notifyError', { title: 'Unbonded Delegates',
-          body: 'You either haven\'t bonded any atoms yet, or some delegates have 0 atoms bonded.' })
-        return
-      } else if (this.unbondedAtoms < 0) {
+      if (this.unbondedAtoms < 0) {
         this.$store.commit('notifyError', { title: 'Too Many Allocated Atoms',
           body: `You've bonded ${this.unbondedAtoms * -1} more atoms than you have.`})
```
jolesbi: "You've tried to bond {x} more atoms than you have"

app/src/renderer/components/staking/PageBond.vue
```diff
@@ -159,7 +153,7 @@ export default {
         try {
           await this.$store.dispatch('submitDelegation', this.fields)
           this.$store.commit('notify', { title: 'Atoms Bonded',
-            body: 'You have successfully bonded your atoms. You can rebond after the  30 day unbonding period.' })
+            body: 'You have successfully updated your delegations.' })
```
jolesbi: "delegated your atoms"?
"bonded your atoms"?

app/src/renderer/components/staking/PageBond.vue
```diff
@@ -205,13 +199,6 @@ export default {
   },
   validations: () => ({
     fields: {
-      reservedAtoms: {
```
jolesbi: i like that we're getting rid of the concept of "reserved" 👍

app/src/renderer/components/staking/PageDelegates.vue
```diff
@@ -69,15 +69,14 @@ export default {
...
-        // { id: 1, title: 'Keybase ID', value: 'keybaseID', initial: true },
-        { id: 1, title: 'Public Key', value: 'id', initial: true },
-        { id: 2, title: 'Country', value: 'country' },
-        { id: 3, title: 'Voting Power', value: 'voting_power' },
-        { id: 4, title: 'Bonded Power', value: 'shares' },
-        { id: 5, title: 'Commission', value: 'commission' }
+        { id: 1, title: 'Name', value: 'description.moniker' },
+        { id: 2, title: 'Public Key', value: 'id' },
+        { id: 3, title: 'Voting Power', value: 'shares', initial: true },
+        { id: 4, title: 'Bonded Atoms', value: 'voting_power' },
```
jolesbi: why not...?
```
{ id: 3, title: 'Voting Power', value: 'voting_power', initial: true },
{ id: 4, title: 'Bonded Atoms', value: 'bonded_atoms' },
```

app/src/renderer/vuex/getters.js
```diff
@@ -5,7 +5,8 @@ export const connected = state => state.node.connected
 export const lastHeader = state => state.node.lastHeader
 
 export const delegates = state => state.delegates
-export const shoppingCart = state => state.shoppingCart.delegates
+export const shoppingCart = state => state.delegation.delegates
+export const committedDelegations = state => state.delegation.committedDelegates
```
jolesbi: how does `committedDelegations` differ from `bondedAtoms`? also, should this be...
```
export const committedDelegations = state => state.delegation.committedDelegations

File: app/src/renderer/vuex/modules/config.js [hunk 8]
```diff
@@ -5,6 +5,8 @@ export default ({ commit, basecoin }) => {
     activeMenu: '',
     desktop: false,
     devMode: process.env.PREVIEW !== undefined ? JSON.parse(process.env.PREVIEW) : process.env.NODE_ENV === 'development',
+    // TODO: change to atom
+    bondingDenom: 'fermion',
```
jolesbi: can we check which network we're using to determine this instead of using a TODO?

app/src/renderer/vuex/modules/index.js
```diff
@@ -5,7 +5,7 @@ export default (opts) => ({
   node: require('./node.js').default(opts),
   notifications: require('./notifications.js').default(opts),
   proposals: require('./proposals.js').default(opts),
-  shoppingCart: require('./shoppingCart.js').default(opts),
+  delegation: require('./delegation.js').default(opts),
```
jolesbi: good call 👍

app/src/renderer/vuex/modules/wallet.js
```diff
@@ -64,10 +64,16 @@ export default ({ commit, node }) => {
...
       dispatch('queryWalletHistory')
     },
-    async queryWalletBalances ({ state, commit }) {
+    async queryWalletBalances ({ state, rootState, commit }) {
       let res = await node.queryAccount(state.key.address)
       if (!res) return
       commit('setWalletBalances', res.data.coins)
+      for (let coin of res.data.coins) {
+        if (coin.denom === rootState.config.bondingDenom) {
+          commit('setAtoms', coin.amount)
```
jolesbi: why are atoms treated differently than other denoms here?

test/unit/specs/__snapshots__/PageDelegates.spec.js.snap
```diff
@@ -487,50 +487,50 @@ exports[`PageDelegates should filter the delegates 1`] = `
...
             <i
               class="fa fa-square-o"
             />
             <a
               class=""
-              href="#/staking/delegates/pubkeyX"
+              href="#/staking/delegates/pubkeyY"
             >
-              pubkeyX
+               undefined
```
jolesbi: this ok?

mappum: Good catch

app/src/renderer/vuex/modules/wallet.js
mappum: Could be cleaned up later, but we're using the 'atoms' value elsewhere in the app for bonding code. This is just a product of the original bonding code using a an atom balance separate from the wallet.

app/src/renderer/components/staking/PageDelegates.vue
mappum: It makes more sense to call the percentage of power "Voting Power", since that's how it's been referred to outside of the code. And what's called 'voting_power' in the code isn't necessarily voting power, it's more accurate to refer to it as the number of bonded tokens.

app/src/renderer/vuex/getters.js
mappum: `committedDelegations` are our delegations which have been finalized on chain, the other values are just our local values before we have submitted.

That's actually the right name, called it `...Delegates` to match the `delegates` array inside the module

app/src/renderer/vuex/modules/config.js
mappum: That would be ideal, but I'm not sure of a good heuristic for checking if a network is a testnet (and I'm not sure what the mainnet will be called in the code). We can do this once we have our first mainnet and need the bonding denom to be 'atom'

app/src/renderer/components/staking/PageBond.vue
mappum: I figured it might be weird if you only unbond and then see "you have bonded your atoms". What do you think?

app/src/renderer/components/staking/PageBond.vue
mappum: :+1:

app/src/renderer/components/staking/PageBond.vue
mappum: I'm not really sure, but it's not being used anymore

nylira: `fillAtoms` was used in the Equalize button to automatically dump all your atoms to the delegation. It was mainly for testing. We don't need it anymore.

app/src/renderer/components/staking/PageBond.vue
jolesbi: oh i see! ok. makes sense.

app/src/renderer/components/staking/PageDelegates.vue
jolesbi: > percentage of power "Voting Power"

is percentage of power === voting power / bonded atoms?

> 'voting_power' in the code isn't necessarily voting power

😕  why not?

i'm probably just misunderstanding something...

app/src/renderer/vuex/modules/wallet.js
```diff
@@ -160,7 +161,17 @@ export default ({ commit, node }) => {
...
+        // TODO fix in cosmos-sdk-js
+        let tx = await (async function () {
+          switch (args.type) {
+            case 'buildDelegate': return axios.post('http://localhost:8998/build/stake/delegate', args)
+              .then(res => res.data)
+            case 'buildUnbond': return axios.post('http://localhost:8998/build/stake/unbond', args)
+              .then(res => res.data)
+            default: return node[args.type](args)
+          }
+        })()
```
mappum: Oops, I had just forgotten to publish the changes to `cosmos-sdk-js`

jolesbi:
tests for `PanelSort` too!

mappum:
Now PageBond has 96% coverage, this PR is ready (sorry for the delay).

I'll revert back to using the `cosmos-sdk` for the API calls in a later PR (
see #301)

Issue #242, opened 2017-12-14
Add time to transactions


faboweb:
Time is currently not delivered for transactions. We need to query the block-header for each transaction and enrich the transactions with the discovery time of the block-header.

PR #259, opened 2017-12-19
Fabo/242 add time to transactions

faboweb:
Implements #242


app/src/renderer/vuex/modules/wallet.js
```diff
@@ -65,10 +73,37 @@ export default ({ commit, node }) => {
...
+      let block = await dispatch('queryBlock', blockHeight)
+      commit('setTransactionTime', { blockHeight, block })
+    },
+    async queryBlock ({ state, commit }, height) {
+      let block = state.blocks.find(b => b.block_meta.header.height === height)
+      if (block) {
+        return block
+      }
+      block = await new Promise((resolve, reject) => {
+        node.rpc.block({height}, (err, block) => {
```
mappum: We should probably just get the block header, instead of the whole thing (the difference is that the block contains all the transaction data, which may grow to be large in the future).

We can get only the header by using `node.rpc.blockchain({ minHeight: height, maxHeight: height }, cb)` (looks a little less clean, but probably worth it). The result format is similar but contains `block_metas` as an array.

faboweb:
Awesome. Thx for the inside! Will change this tomorrow!

Matt Bell <notifications@github.com> schrieb am Di., 19. Dez. 2017, 18:23:

Issue #209, opened 2017-12-08
Real Validator Profile Data


nylira:
We need real validator profile data from the LCD.

AC (profile shows):
- [] Validator Name
- [] Validator Website
- [] Validator Description (this can hold links to external ressources like an audit)
- [x] atoms self-bonded
- [x] atoms delegated to them
- [] commission rate
  -https://github.com/cosmos/gaia/blob/proposals/spec.md#fee-pool-commission
- [] how long they have been validator
- [] max commission rate (the rate that it will never go higher than)
- [] max commission increase rate (the max amount it can increase over time)

- [x] Public Key
- [x] Address

jolesbi:
let's close this in favour of #309

jolesbi:
referencing #135 for info about how to calculate commissions.

PR #271, opened 2017-12-27
Fabo/209 candidate profiles

faboweb:
Implements #209 

Currently missing commision data as this data is not yet provided by Gaia.


nylira:
`Failure message: ENOMEM: not enough memory, write` error again on CircleCI despite the tests running fine locally.

PR #276, opened 2017-12-28
Restore Account -> Import Account

### LINKED ISSUE #247: change "Restore From Seed" to "Import Account"
as discussed in the #security channel... let's remove seeds as an option until deterministic + verifiable builds are in

@ebuchman

nylira:
* Renames all instances of Restore to Import
* Hides the Import Account option via `config.devMode`

screen shot 2017-12-28 at 12 58 31 pm
screen shot 2017-12-28 at 12 58 24 pm

Closes #247


jolesbi:
is "Generate" the right word?

nylira:
@jolesbi I think so. Generate means creating something (an account) from something else (a seed). Do you have a better word in mind?

faboweb:
How about: `Recreate an account with an existing seed`? (just an idea)

PR #278, opened 2017-12-28
Use z-index layers instead of hardcoded values

### LINKED ISSUE #243: Use z-index layers instead of hard coded values
Based on: https://www.sitepoint.com/better-solution-managing-z-index-sass/

```js
$z-layers: (
  "goku":            9001,
  "shoryuken":       8000,
  "default":            1,
  "below":             -1,
  "bottomless-pit": -9000
);

@function z($layer) {
  @return map-get($z-layers, $layer);
}
```

nylira:
Closes #243

### DIFF
--- a/app/src/renderer/styles/variables.styl
+++ b/app/src/renderer/styles/variables.styl
@@ -82,3 +82,21 @@ h3 = 1.25rem
 h4 = 1.125rem
 h5 = 1rem
 h6 = 0.75rem
+
+//==============================================================================
+// z-index
+
+z-layers = {
+  modalHelp: 1001,
+  modal: 1000,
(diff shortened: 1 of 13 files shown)


app/src/renderer/styles/variables.styl
```diff
@@ -82,3 +82,21 @@ h3 = 1.25rem
 h4 = 1.125rem
 h5 = 1rem
 h6 = 0.75rem
+
+//==============================================================================
+// z-index
+
+z-layers = {
```
faboweb: :heart:

faboweb: This reads so nice.

PR #279, opened 2017-12-28
remove file menu for Windows builds

### LINKED ISSUE #236: Remove electron menu under windows
Windows shows a menu:

image

nylira:
Didn't get to test it yet with a real build, but it should work. Ref: https://github.com/electron/electron/blob/master/docs/api/browser-window.md#winsetmenumenu-linux-windows

Closes #236

### DIFF
--- a/app/src/main/index.js
+++ b/app/src/main/index.js
@@ -130,7 +130,11 @@ function createWindow () {
   webContents.on('will-navigate', handleRedirect)
   webContents.on('new-window', handleRedirect)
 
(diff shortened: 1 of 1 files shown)


faboweb:
Does this target Mac or Windows? The filemenu was already removed on current develope by the line https://github.com/cosmos/cosmos-ui/pull/279/files#diff-6f1bf0a59b91e2427edb5a476a341460R134

nylira:
Ah it targets Windows. I guess it's unnecessary then?

PR #299, opened 2017-12-29
clicking a balance will go to the send page with that denom selected...

jolesbi:
... balance page displays balances and denominations on the network as separate lists

part of #128


app/src/renderer/routes.js
```diff
@@ -68,7 +68,7 @@ export default [
 
   // WALLET
   { path: '/', name: 'balances', component: wallet('Balances') },
-  { path: '/wallet/send', name: 'send', component: wallet('Send') },
+  { path: '/wallet/send', name: 'send', props: true, component: wallet('Send') },
```
faboweb: Nice learning. :)

app/src/renderer/components/wallet/PageSend.vue
```diff
@@ -8,7 +8,7 @@ page(title='Send')
         field-id='send-denomination' field-label='Denomination')
         field#send-denomination(
           type="select"
-          v-model="fields.denom"
+          v-model="fields.denom || denom"
```
faboweb: I am not sure if this works properly. This will set the model to use the property "denom". onSubmit uses "fields.denom". Is it tested, that this works? I proposed another way to set the denom.

app/src/renderer/components/wallet/PageSend.vue
```diff
@@ -121,9 +121,10 @@ export default {
     },
     ...mapActions(['walletSend'])
   },
+  props: ['denom'],
   mounted () {
-    if (this.denominations.length === 1) {
-      this.fields.denom = this.denominations[0].value
+    if (this.denom) {
```
faboweb: I set the denom from the property here. So the variable "fields.denom" stays the one truth for the currently selected denom.

jolesbi: i think this is better. thanks for the improvement.

PR #303, opened 2018-01-02
0.3.0

mappum:
Just switched the default network to `gaia-2-dev` and bumped both package.json versions.


jolesbi:
aren't we on `gaia-2` now?

mappum:
Well it didn't exist at the time when I made the release, but we should definitely start testing with it as we develop with `npm run testnet gaia-2`. The default shouldn't matter anymore after #304.

PR #311, opened 2018-01-03
set state.address from key instead of state.accounts

### LINKED ISSUE #290: sign in on account creation doesn't work as expected
app menu still says "Sign In" and errors below

```
user.js?dff5:88 Uncaught (in promise) TypeError: Cannot read property 'address' of undefined
    at Store._callee7$ (webpack-internal:///508:292:19)
    at tryCatch (webpack-internal:///487:65:40)
    at Generator.invoke [as _invoke] (webpack-internal:///487:299:22)
    at Generator.prototype.(anonymous function) [as next] (webpack-internal:///487:117:21)
    at step (webpack-internal:///17:17:30)
    at eval (webpack-internal:///17:35:14)
    at Promise (<anonymous>)
    at F (webpack-internal:///32:35:28)
    at eval (webpack-internal:///17:14:12)
    at Store.signIn (webpack-internal:///508:309:10)
```

jolesbi:
closes #290

### DIFF
--- a/app/src/renderer/vuex/modules/user.js
+++ b/app/src/renderer/vuex/modules/user.js
@@ -88,10 +88,11 @@ export default ({ commit, node }) => {
     async signIn ({ state, dispatch }, { password, account }) {
       state.password = password
       state.account = account
-      state.address = state.accounts.find(_account => _account.name === account).address
       state.signedIn = true
 
       let key = await node.getKey(account)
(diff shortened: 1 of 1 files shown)


faboweb:
I think the problem was, that 'loadAccounts' wasn't done before signIn. This PR makes signIn more atomic and we don't have to care about requests. 👍

Issue #307, opened 2018-01-02
Reconnect logic not being triggered


mappum:
Meant to leave this as a review on #283, but wasn't quite fast enough 😛 

For this code:
https://github.com/cosmos/cosmos-ui/pull/283/files#diff-d2589d9a81753ea8fa56e1202bff83f6R41
> 
> I tried to reproduce a node disconnect by turning off my wifi and reloading my wallet balance. But it looks like this error handler never gets called in that case since the refresh makes its request to the relayServer which is still alive. So I think the reconnect logic won't ever get triggered, @faboweb were you able to trigger it outside of the tests?
> 
> The RPC client gets a response of 500 Internal Server Error with the error in the body: Error: connect ECONNREFUSED 127.0.0.1:8998. We'll need a way of having the app detect when it gets a 500 response (assuming the LCD only gives 500s when it can't connect to the node), then it could do the same as what's in this error handler.
>

faboweb:
I triggered the status by connecting to a local node and turning that one on and off. I will follow up on this.

PR #313, opened 2018-01-03
Reconnect if connection drops

faboweb:
Solves: #307

### DIFF
--- a/app/src/renderer/vuex/modules/node.js
+++ b/app/src/renderer/vuex/modules/node.js
@@ -1,3 +1,5 @@
+import { setTimeout } from 'timers'
+
 'use strict'
 
 export default function ({ node }) {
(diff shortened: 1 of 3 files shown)


app/src/renderer/vuex/modules/node.js
```diff
@@ -67,6 +72,25 @@ export default function ({ node }) {
...
+      state.nodeTimeout = setTimeout(() => {
+        // clear timeout doesn't work
+        if (state.nodeTimeout) {
+          state.nodeTimeout = null
+          dispatch('reconnect')
+        }
+      }, timeout)
+      node.rpc.status((err, res) => {
+        if (!err) {
+          state.nodeTimeout = null
```
mappum: Should this call `clearTimeout`?

mappum: Oh, I see the comment now. Why doesn't it work? I didn't think nulling the return value would work 🤔

mappum: Nvm, missed the `if (state.nodeTimeout)`. :+1:

PR #314, opened 2018-01-03
Deterministic gaia binaries

### LINKED ISSUE #245: deterministic, verifiable builds
this is vert important. but can wait until next sprint.

JB:
> i'd like to hold off on implementing deterministic builds (with signed releases) until at least dec 28

Jae:
> as long as we don’t put blog posts out, and we put all the disclaimers, that sounds good.

related to #232 

@jaekwon

mappum:
Closes #245 

So far, I've only made the gaia binary builds deterministic. You'll now need to install [Docker](https://store.docker.com/search?type=edition&offering=community) to make production builds, but luckily its usage can be fully automated so we won't have to do anything special in the build process.

Will now have to investigate all the other sources of non-determinism in Electron app bundles and platform-specific packages.

### DIFF
--- a/config.js
+++ b/config.js
@@ -24,7 +24,7 @@ let config = {
     ignore: /^\/(src|index\.ejs|icons)/,
     out: path.join(__dirname, 'builds'),
     overwrite: true,
-    platform: process.env.PLATFORM_TARGET || 'darwin,linux',
+    platform: process.env.PLATFORM_TARGET || 'darwin,linux,windows',
     packageManager: 'yarn'
   }
 }
--- a/tasks/release.js
+++ b/tasks/release.js
@@ -17,7 +17,7 @@ process.argv.forEach(function (val) {
   }
   if (val.startsWith('--binary')) {
     binaryPath = val.replace('--binary=', '')
-    console.log('Using build binary', binaryPath)
+    console.log('Using prebuilt binary', binaryPath)
   }
 })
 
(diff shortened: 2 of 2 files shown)


mappum:
Good news, I've investigated and our webpack output is deterministic, even after reinstalling the dependencies with `yarn` 👍 So now I'll just have to look into whatever is different inside the electron bundle output.

mappum:
Known sources of non-determinism:

- [ ] Native module binaries - May be able to just get rid of all native deps so that we don't have to deal with them, (`keccak`, `leveldown`, `secp256k1`, `ed25519-supercop`)
- [ ] Release packages (e.g. `.dmg`, Windows installers, etc.)

faboweb:
I like the idea of getting rid of the native binaries. But is this possible?

mappum:
>I like the idea of getting rid of the native binaries. But is this possible?
  
Yes, should be easy in the short-term since we actually aren't using a lot of these deps right now 🤦‍♂️. We'll have to use them eventually for deriving fundraiser seeds, but we can deal with that before launch.

mappum:
I was thinkign we'd have all the deteminism changes in this PR, but now it makes more sense to merge this as-is to get deterministic go binaries (and also a more convenient go compiler script for the build step). Then we can figure out determinism on the next layer (the `electron-builder` layer, #329) and merge that separately.

Will also have to remove the native module dependencies outside of this PR, which goes along with #315.

nylira:
Yeah let's keep the PR smaller and merge what we have. Good work @mappum!

Issue #47, opened 2017-11-09
Use only one source for node_modules


faboweb:
We have node_modules in root and in app/. This creates overhead while installing and is unpractically for testing as testing is done from root.

faboweb:
This was more work then expected. I will defer that to a later time.

jolesbi:
since @jaekwon and @Greg-Szabo were discussing deterministic builds and human audit-able dependencies — this ticket might be worth doing — for the sake of simplicity?

faboweb:
We can take it into the next sprint. This should be easier now as everything is a lot more stable.

PR #315, opened 2018-01-03
Only one package src

faboweb:
Implements: #47 
  
Depends on: https://github.com/mappum/js-tendermint/pull/3
Change back to using js-tendermint package after merge.


jolesbi:
great PR. love when we delete stuff!

PR #317, opened 2018-01-03
simplify app menu / architecture

### LINKED ISSUE #128: Simplify NiAppMenu

jolesbi:
closes #128

### DIFF
--- a/app/src/renderer/components/common/NiUserPane.vue
+++ b/app/src/renderer/components/common/NiUserPane.vue
@@ -6,6 +6,12 @@ list-item.ni-li-user(
   @click.native="close"
   icon="face"
   :title="user.account")
+div(v-else-if="user.signedIn")
+  list-item.ni-li-user(
+  type="link"
+  icon="exit_to_app"
+  @click.native="signOut"
+  title="Sign Out")
 </template>
 
 <script>
--- a/app/src/renderer/components/common/NiUserPane.vue
+++ b/app/src/renderer/components/common/NiUserPane.vue
@@ -8,10 +8,10 @@ list-item.ni-li-user(
   :title="user.account")
 div(v-else-if="user.signedIn")
   list-item.ni-li-user(
-  type="link"
-  icon="exit_to_app"
-  @click.native="signOut"
-  title="Sign Out")
+    type="link"
+    icon="exit_to_app"
+    @click.native="signOut"
+    title="Sign Out")
 </template>
 
 <script>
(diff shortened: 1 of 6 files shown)


app/src/renderer/components/common/NiUserPane.vue
```diff
@@ -5,18 +5,7 @@ list-item.ni-li-user(
...
-    type="link"
-    icon="exit_to_app"
-    @click.native="signOut"
-    title="Sign Out")
-list-item.ni-li-user(
-  v-else
-  @click.native="openSession"
-  icon="exit_to_app"
-  title="Sign In")
+  :title="user.account")
```
faboweb: Users in production mode will not be able to sign out right?

jolesbi: users should be able to sign out from their profile page in the top bar. this is left over from when you could use the app before signing in. IMO this area should never change. should just be user's name.

faboweb: User profile is not production ready I think. (May be wrong) This was the original idea behind leaving the sign-out in the menu for now.

jolesbi: yep you're right. user profile is not ready for production (missing a lot of data). i created a ticket for this (#322). 

i'll bring back the sign out button.

nylira:
This is really nice!

PR #329, opened 2018-01-05
Switch to electron-builder

faboweb:
AC:
- [x] Working on Windows
- [x] Working on Linux
  ~~- [ ] Make icons work on Linux~~
- [x] Working on Mac

Optional:
~~- [ ] Use docker to build cross-platform~~


mappum:
Tested on mac, and it works in the `.dmg`, although the one in the `.zip` does not. This might be OK since we only want to release with a dmg anyway, but hopefully we can make this dmg deterministic.

faboweb:
Awesome! Thx Matt! I will limit the build to .DMG only.
I will test determinism next. Maybe we need docker for that.

Matt Bell <notifications@github.com> schrieb am Fr., 5. Jan. 2018, 20:31:

mappum:
>I will test determinism next. Maybe we need docker for that.

OK, before then we'll have to merge #314 to get deterministic go binaries first and remove the native node module dependencies, then merge those changes into this.

I'm not sure docker will fix things for this layer of the app. For example the dmg format probably has a timestamp in it somewhere, which is what's breaking determinism. We might have to do tricky stuff to modify the output (e.g. writing over the timestamps), and we'll probably have to do it for each platform.

faboweb:
Investigated a lot today for the deterministic builds. Most of the stuff i.e. building the js application even with native modules seems to run deterministic. The problem arises when everything is put into an .exe or .dmg. Matt already mentioned it, but there are probably timestamps in there, that are not deterministic. (https://github.com/electron-userland/electron-builder/issues/2454)

As the building itself works, should we merge and investigate further?

mappum:
>Most of the stuff i.e. building the js application even with native modules seems to run deterministic.

Maybe you didn't reinstall the dependencies? When we install dependencies, it compiles the native modules which gives a different output every time.

>As the building itself works, should we merge and investigate further?

Sure, the determinism stuff can be in a different PR.

tasks/release.js
```diff
@@ -49,88 +46,113 @@ function pack () {
...
     if (platform === 'win32') platform = 'windows'
     if (platform === 'mas') platform = 'darwin'
     if (GOARCH[arch]) arch = GOARCH[arch]
 
     console.log(`\x1b[34mBuilding gaia binary (${platform}/${arch})...\n\x1b[0m`)
 
-    let output = 'gaia'
-    if (platform === 'windows') output += '.exe'
-
-    let cmd = `
```
mappum: Looks like the deterministic gaia build from #314 got removed (I could make a PR on top of this PR to re-enable it).

faboweb: Some merge conflicts probably. Sorry. Readded your 2(?) commits. But it fails on Windows:
image

faboweb: I removed the commits again. Could create a PR like you proposed? My thoughts: Do you think this is actually necessary? In development we don't need deterministic builds right? And in production I propose we always use a distributed binary with the "--binary" flag.

mappum: >But it fails on Windows:

Ah OK I'll fix it, just need the Windows equivalent to `/tmp`.

>My thoughts: Do you think this is actually necessary? In development we don't need deterministic builds right? And in production I propose we always use a distributed binary with the "--binary" flag.

It's the difference between manually downloading the gaia binary every time we build, or instead having our build system make the binary for us. This way, we will all be independently building gaia from source, instead of maintaining a single point of failure (plus we don't have to wait for someone else to make binaries for us).

jolesbi: i see both sides. not sure i can vote one way or another. sorry guys.

nylira: I vote for using a distributed Gaia binary for simplicity. We can add multiplatform Gaia binary building in another PR, but I would love to get this `electron-builder` PR merged.

mappum:
I discovered that the mac builds are actually not working yet, I had a false positive when building before because I still had leftover output files from the old build system. I'm working on fixes.

nylira:
Regarding the question whether to switch from `electron-packager` or not: I trust in Matt and Fabo's opinions on this issue. If the complexity of `electron-builder` and the sorry state of its documentation isn't worth the hassle, we should cancel this PR and continue using what we already have. We're prioritizing deterministic zips instead of installers, so one of my primary reasons for going with `electron-builder` is gone.

jolesbi:
hey ya'll! 🤠 

have we come to a decision about this?

mappum:
I voted for abandoning this change, not sure if consensus was reached.

nylira:
Conclusion of Tuesday's meeting: We're closing this PR and @faboweb can reopen it if he wants next week.

PR #330, opened 2018-01-05
Update README.md

faboweb:
added env variable to select a certain node


jolesbi:
thanks for adding this!

PR #331, opened 2018-01-05
bonding + unbonding ux / ui

jolesbi:
hey ya'll! i tried to improve the look and feel of the whole bonding / unbonding experience. i'm not entirely satisfied yet, but i think this is an improvement. bonding / unbonding can be very complex — we haven't accounted for all the use cases yet...

delegates page
- [x] checkbox in own column
- [x] obvious button to next step
- [x] no column sorted by default
- [x] candidate list order should be randomized
- [x] clean up column alignment
  
bonding page
- [x] add total, bonded, and unbonded atoms
- [x] show previously bonded (when possible)


jolesbi:
screen shot 2018-01-05 at 4 41 10 pm
screen shot 2018-01-05 at 4 44 32 pm

main points
- real checkboxes instead of switching icons (have to fix `v-model` + vuex bug)
- checkbox in its own column (column widths could still use some work)
- obvious link to next step (consider using pluralize `delegate(s)`)

jolesbi:
before:
screen shot 2018-01-05 at 4 42 31 pm

after:
screen shot 2018-01-05 at 4 45 09 pm

main points
- user's Atom information at top to help inform decision making
- fixed width form for improved readability
- including committed bonds for context and clarity
- cleaned up messaging and adjusted language
- using moniker, removed address / ID
---
- add link to candidate profile
- add alert after submit including a summary of transaction and warning about unbonding again
- would like to discuss using component other than inputs for allocation

nylira:
Good work. I think the "Split Allocation Equally" and "Reset Allocation" buttons can be removed. Most likely our users will not want to split their allocations equally -- they'll favor specific validators or they may want to reserve some atoms. And reset buttons on forms are bad for usability - most likely the user will press it by accident.

faboweb:
I like the structuring of your current atoms instead of the text. Could there maybe be sth. similar for the last 3 lines?

faboweb:
Are you sure about native check-boxes? I am not a fan as it usually looks bad cross-platform.

It looks like this on windows:
image

faboweb:
Apparently there is also a bugs:
image

nylira:
Agree with @fabo that we shouldn't use native check boxes. I like how they have their own column though

jolesbi:
@mappum i fixed up the mobile issues (peng will be redoing the UI here anyways) 📱 
@faboweb fixed the bugs 🐛 
@nylira went back to icons instead of checkboxes ✅ 

any remaining issues can be addressed in #332 

i think having the buttons for equal allocation and reseting is nice. can be discussed further in #332

nylira:
Fixed some duplicate id errors:

screen shot 2018-01-12 at 11 40 48 am
screen shot 2018-01-12 at 11 38 52 am

Also I ran into this when I tried to delegate.

screen shot 2018-01-12 at 12 00 45 pm

What could be causing it?

faboweb:
@nylira Looks like a Gaia error. Could you check the network transaction if all the data is there? Maybe some property isn't set and "null" is not a nominated account or so.

jolesbi:
@nylira that's strange. i can successfully bond.

mappum:
I went through the delegation/unbonding flow and everything worked well, I have a few comments but they can be addressed as separate issues.

PR #333, opened 2018-01-08
Peng/123 vuex tests

### LINKED ISSUE #123: unit tests for /vuex (peng)
```
validators.js
wallet.js
```
  

not sure if these can be unit tested...
```
actions.js
getters.js
```

nylira:
* created tests for `vuex/modules/validators.js`
* `actions.js` only has deprecated actions, so I'm deleting the file.
* `getters.js` doesn't have to be tested because we don't have any complicated computation in that file.

Still need to check tests for `wallet.js`.

Closes #123


nylira:
@mappum or @faboweb could I get your help on `wallet.js` action tests? I'm not clear on how I should test these actions:

https://github.com/cosmos/cosmos-ui/pull/333/files#diff-8966235b1c8735bfa970472f398c21eaR128

I marked them with TODO.

faboweb:
I will give it a look!

nylira:
@faboweb Thank you Fabo! It looks great.

@jolesbi @mappum I can't review it because it's my own PR. But's ready to be merged.

app/src/renderer/components/wallet/PageSend.vue
```diff
@@ -105,27 +105,25 @@ export default {
...
 
           // resets send transaction form
-          this.resetForm()
+        this.resetForm()
 
           // refreshes user transaction history
-          this.$store.dispatch('queryWalletHistory')
-        }
+        this.$store.dispatch('queryWalletHistory')
+      }, err => {
```
jolesbi: is `err => {` the same as `.catch()`?

app/src/renderer/node.js
```diff
@@ -61,6 +62,28 @@ module.exports = function (nodeIP) {
 
       node.rpcConnecting = false
       return nodeIP
+    },
+    async sendTx (args, account, password) {
```
jolesbi: why is this in `node.js`?

mappum: I feel like this might not be the right module, although I understand why `wallet` might not have felt right either. Maybe this could just be its own `tx` module, with only this action and the queue of transactions plus the sequence number?

app/src/renderer/vuex/modules/delegation.js
```diff
@@ -66,14 +66,14 @@ export default ({ commit }) => {
       await dispatch('walletTx', args)
     },
     async submitDelegation ({ state, dispatch }, delegation) {
-      for (let delegate of delegation.delegates) {
+      return Promise.all(delegation.delegates.map(delegate => {
```
mappum: Running these in parallel does not help since we have to sequentially increment the nonce and check that the tx went through. Really this should all happen in an atomic operation, but the LCD does not support that. Will open a ticket.

app/src/renderer/vuex/modules/wallet.js
```diff
@@ -39,7 +38,7 @@ export default ({ commit, node }) => {
       state.sendQueue.push(sendReq)
     },
     shiftSendQueue (state) {
-      state.sendQueue = state.sendQueue.shift(1)
+      state.sendQueue = state.sendQueue.length > 1 ? state.sendQueue.shift(1) : []
```
mappum: Hm, looks like I had a bug when I originally implemented this. This line should be `state.sendQueue = state.sendQueue.slice(1)`, to shift off the first element (if any). (Also triggers a render since we're setting `sendQueue` to a different object.

app/src/renderer/node.js
```diff
@@ -61,6 +62,28 @@ module.exports = function (nodeIP) {
...
+            .then(res => res.data)
+          default: return node[args.type](args)
+        }
+      })()
+      let signedTx = await node.sign({
+        name: account,
+        password: password,
+        tx
+      })
+      let res = await node.postTx(signedTx)
```
mappum: It's important that this goes through the same wallet sending code that it was separated from in this PR, to properly process multiple transactions serially and set the correct account sequence number on each transaction. Otherwise, transactions created here will only suceed if it is an account's first transaction.

app/src/renderer/components/wallet/PageSend.vue
faboweb: Yes, but the `then(_, onReject)` will get called instead of `.catch`if available. See: https://codepen.io/faboweb/pen/WdmRoO

jolesbi: cool! thanks for the example. still not sure when to use one vs the other...?

mappum:
This should be good to go now, maybe needs one last look.

app/src/renderer/vuex/modules/send.js
```diff
@@ -0,0 +1,106 @@
...
+      let res = await node.queryNonce(address)
+      if (!res) return
+      commit('setNonce', res.data)
+    },
+
+    // builds, signs, and broadcasts a tx of any type
+    sendTx ({ state, dispatch, commit, rootState }, args) {
+      // wait until the current send operation is done
+      if (state.sending) {
+        args.done = new Promise((resolve, reject) => {
```
faboweb: Thx for picking up my promise over callback solution

PR #337, opened 2018-01-09
implemented blocks page

### LINKED ISSUE #29: Network Monitoring
As a user
I want to be able to see the status of the blockchain network
So that [I can see if there are enough nodes to securely facilitate my transaction?]

AC:
- [ ] User is able to see the node up time for every node in the network for last [day, week, month, year?]
- [ ] User is able to see the amount of currently active node?
- [ ] ~The network monitor is accessible from outside as a standalone website~
- [x] The network monitor is accessible from inside of the wallet

Notes:
- What is the value proposition?

jolesbi:
closes #29 

>User is able to see the node up time for every node in the network for last [day, week, month, year?]

i don't think this info is particularly important as part of the netmon. this is info that should matter for a validator / candidate. i'd propose to cross this off the list above.

>User is able to see the amount of currently active node?

i'd propose this be moved to the footer — also, is there any way we can actually get this data? right now it's just `validators.length` which is not the same as "active nodes".

1. i'd like to propose that we switch the word in the sidebar from 'Monitor' to 'Blocks'. this page will then show a list of blocks, each one clickable, taking the user to the PageBlock for that block. this would be fairly simple to implement and should be called the "block explorer" not "network monitor".

2. "the network monitor" is the graphical interface with fancy visualizations which really has nothing to do with this PR (and as discussed requires some re-designing). i'd like to propose that we leave this visualization tool (AKA Netmon) until post launch. i would love to have it but it does not seem vital for pre-launch.

3. if we feel we want to show the version numbers in the app as seen below, i'd recommend that we put them in the native menu under an 'about' section. if we show any version numbers, we ougght to show all of them (tendermint, abci?, gaia / hub, ui) but arguably, we don't need to share version numbers. i'd propose to remove them altogether.

```
ABCI Version
0.7.1
Tendermint Version
0.13.0
```

### DIFF
--- a/app/src/renderer/components/common/NiConnectivity.vue
+++ b/app/src/renderer/components/common/NiConnectivity.vue
@@ -16,9 +16,8 @@ export default {
     ListItem
   },
   computed: {
-    ...mapGetters(['lastHeader', 'nodeIP', 'connected']),
+    ...mapGetters(['lastHeader', 'nodeIP', 'connected', 'validators']),
     blockString () {
-      //- TODO: calculate actual number of active nodes, not just validators
       return `${this.lastHeader.chain_id} (#${num.prettyInt(this.lastHeader.height)})`
     }
   },
--- a/app/src/renderer/components/common/NiConnectivity.vue
+++ b/app/src/renderer/components/common/NiConnectivity.vue
@@ -16,10 +16,10 @@ export default {
     ListItem
   },
   computed: {
-    ...mapGetters(['lastHeader', 'nodeIP', 'connected', 'validators']),
+    ...mapGetters(['lastHeader', 'nodeIP', 'connected']),
     blockString () {
       //- TODO: calculate actual number of active nodes, not just validators
-      return `${this.lastHeader.chain_id} (#${num.prettyInt(this.lastHeader.height)}) (${this.validators.length} Nodes)`
+      return `${this.lastHeader.chain_id} (#${num.prettyInt(this.lastHeader.height)})`
     }
   },
   data: () => ({
--- a/app/src/renderer/components/common/NiConnectivity.vue
+++ b/app/src/renderer/components/common/NiConnectivity.vue
@@ -16,9 +16,10 @@ export default {
     ListItem
   },
   computed: {
-    ...mapGetters(['lastHeader', 'nodeIP', 'connected']),
+    ...mapGetters(['lastHeader', 'nodeIP', 'connected', 'validators']),
     blockString () {
-      return `${this.lastHeader.chain_id} (#${num.prettyInt(this.lastHeader.height)})`
+      //- TODO: calculate actual number of active nodes, not just validators
+      return `${this.lastHeader.chain_id} (#${num.prettyInt(this.lastHeader.height)}) (${this.validators.length} Nodes)`
     }
   },
   data: () => ({
--- a/app/src/renderer/components/monitor/PageBlock.vue
+++ b/app/src/renderer/components/monitor/PageBlock.vue
@@ -39,7 +39,7 @@ page(:title="pageBlockTitle")
 
   part(title='Transactions')
     list-item(v-if="block.header.num_txs > 0" v-for="tx in block.data.txs" :key="tx.id" dt="Transaction" :dd="TODO")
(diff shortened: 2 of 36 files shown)


app/src/renderer/vuex/modules/blockchain.js
```diff
@@ -19,27 +19,22 @@ export default ({ commit }) => {
...
-        }
-        console.log('abci_info', JSON.stringify(res.body.result, null, 2))
-        state.abciInfo = res.body.result.response
+      axios(url + '/abci_info').then((res) => {
+        state.abciInfo = res.data.result
       })
     }
   }
 
+  setTimeout(() => {
```
jolesbi: i don't know how to tell if i can use sockets despite @mappum sharing this link... https://github.com/tendermint/tendermint/blob/master/types/events.go#L12-L32

app/src/renderer/components/monitor/PageBlockchain.vue
```diff
@@ -1,25 +1,18 @@
...
-    a(@click='setSearch(true)')
-      i.material-icons search
-      .label Search
-
   template(v-if="blockchain")
-    part(title='Metadata')
-      list-item(dt='Network Name' :dd='status.node_info.network')
-      list-item(dt='App Version' :dd='version')
-      list-item(dt='Tendermint Version' :dd='status.node_info.version')
+    part(title='Versions')
```
jolesbi: i'd like to remove this section and turn this page into a block explorer

app/src/renderer/components/common/NiConnectivity.vue
```diff
@@ -16,9 +16,10 @@ export default {
     ListItem
   },
   computed: {
-    ...mapGetters(['lastHeader', 'nodeIP', 'connected']),
+    ...mapGetters(['lastHeader', 'nodeIP', 'connected', 'validators']),
     blockString () {
-      return `${this.lastHeader.chain_id} (#${num.prettyInt(this.lastHeader.height)})`
+      //- TODO: calculate actual number of active nodes, not just validators
```
jolesbi: is it possible to calculate active nodes? if not, let's just remove this.

app/src/renderer/components/monitor/PageBlock.vue
```diff
@@ -38,29 +38,41 @@ page(:title="`Block ${block.header.height}`")
     :dd="p.signature.data")
 
   part(title='Transactions')
-    list-item(v-for="tx in block.data.txs" :key="tx.id"
-      dt="Transaction" :dd="TODO")
+    list-item(v-if="block.header.num_txs > 0" v-for="tx in block.data.txs" :key="tx.id" dt="Transaction" :dd="TODO")
```
jolesbi: use `li-transactions`?

app/src/renderer/components/monitor/PageBlockchain.vue
faboweb: I am pro keeping the versions in there. Could be helpful for developers (us) to give an easy entry point for submitting errors. BUT: if we don't have the versions yet, let's add it as an idea and do it later.

faboweb: We also get a whole bunch of versions for the connected node via `node_info.other` in `gaia client rpc status`.

app/src/renderer/components/monitor/PageBlock.vue
faboweb: It should probably show more information at some point, but so should the information in the transactions list.

faboweb:
I am OK with doing the block explorer first and the Netmon later. Could we maybe limit the Netmon to show only the node_info (`gaia client rpc` > node_info) for the connected node when we click on the connection info (bottom left)?
The versions shown are specific to the connected node and not to the installed application. But they could still go in the future Netmon.

jolesbi:
@nylira @mappum @faboweb this should be good to go. 

does not include #344 of course. will add tests there.

nylira:
Love it! Works well. This is great to have in a release for end-users to try out.

faboweb:
Cool thing!
Some notes: Screenshot would have been nice.

PR #358, opened 2018-01-12
Peng/357 fix double self tx

### LINKED ISSUE #357: Transaction history shows two negative transactions after sending tokens to your own address
I just tried to send myself 13 fermions and this is how it is logged in transaction history:

screen shot 2018-01-12 at 1 31 54 pm

Not sure why this happens, and it should be considered 100% wrong because I didn't lose 26 fermions from my balance. I think the optimal tx history is showing both a negative tx for 13 fermions and a positive tx for 13 fermions.

- [x] Show both positive and negative tx when sending coins to yourself

nylira:
## Old

screen shot 2018-01-12 at 1 31 54 pm

# New

screen shot 2018-01-12 at 2 04 17 pm

Closes #357


jolesbi:
this is interesting! are we sure we don't just want to show the positive and the negative as opposed to a special kind of transaction?

nylira:
Hah, the reason I implemented it this way is because it's easier. In the JSON representation, both the receive and send transactions are exactly the same, down to the timestamp. I figured showing both in one line made more sense because it is a special sort of transaction.

app/src/renderer/components/wallet/PageTransactions.vue
```diff
@@ -43,7 +43,8 @@ export default {
   computed: {
     ...mapGetters(['filters', 'transactions', 'wallet', 'config']),
     orderedTransactions () {
-      return orderBy(this.transactions, [this.sort.property], [this.sort.order])
+      let list = orderBy(this.transactions, [this.sort.property], [this.sort.order])
+      return uniqBy(list, 'time') // filter out duplicate tx to self
```
faboweb: Correct me, but I think this would squash all tx to yourself by block (which might be ok, as it's a rare case and blocks are proposed every 1s?).

faboweb:
On the design question:
I think it looks good and is understandable. It creates a little more code to handle the corner case. I am pro this change but would like to hear more arguments by @jolesbi why we shouldn't do it.

jolesbi:
reasons i like it 
- it's clever
- like an easter egg 🥚 (maybe could be more fun!)
- better than two negative transactions (which is just incorrect, we can all agree)

reasons i don't like it
- it's not as true as showing a positive and negative transaction happening at the same timestamp (how can this even be possible?! there isn't even a ms difference?!)
- might be confusing for people (atypical transaction type) looks like a swap of some kind...
- this is an edge case and i wish we didn't have to do anything to account for it 😞 

i think in the end, i'd prefer only two transaction types in the transaction history (+ / -).

we should hear from @mappum on this too!

are we sure this is not a bug? it seems like there should be a positive and negative transaction, not two negative transactions.

nylira:
@jolesbi it's not a bug. Right now sending tokens to yourself results in two identical transactions. The faulty logic in the UI made them both look negative because they both involve sending coins from your own address. The duplicated tx states that tokens are exiting and entering the same wallet.

app/src/renderer/components/wallet/PageTransactions.vue
nylira: Ah you're right. That's not intended. Will fix.

nylira: I looked into resolving this and I think we need to wait until transactions have uids before this can be correctly fixed. Any sort of uniqueness filtering is stymied by the fact that the use can send two self transactions with exactly the same amount of tokens in the same block.

jolesbi:
thanks for the explanation @nylira!

> I argue that it should stand out.

what is your thinking here? 

my preference would be to show self tx's as one positive and one negative instead of including an additional transaction type with corresponding icon. i don't want to confuse users with an atypical transaction type. there are only two types of transactions — sends and receives. let the user use the addresses and timestamps to see it was a self tx.

nylira:
> my preference would be to show self tx's as one positive and one negative instead of including an additional transaction type with corresponding icon. i don't want to confuse users with an atypical transaction type. there are only two types of transactions — sends and receives. let the user use the addresses and timestamps to see it was a self tx.

I think txs should only appear as a positive if the user gains coins from another address, and only show as a negative if coins are sent to another address. If the user decides to send their coins to the same address they're being clever and they should be rewarded with a different visual. 

There are already going to be more than two transaction types anyway, because we're going to implement bonding transactions in the transaction log. Later we will need to implement a transaction filter in the tx log, where they may be check boxes for self-txs, bonding txs, sends, and receives. 

Showing self-txs as send and receive pollutes the user's transaction log with false volatility. Later if we want to generate charts from the transaction data, self-tx events will result in an instantaneous decrease and increase in the user's balance when nothing of the sort happened in reality.

@mappum what do you think?

nylira:
@jolesbi yeah, we have an issue #366 to track real uids

PR #364, opened 2018-01-12
added send button to balances page

### LINKED ISSUE #346: Improve UX for sending
As the menu item for sending is gone a user has to click on a denom on the balances page to send coins. But the action is not clear. It doesn't state "send" or sth. similar. It just has an arrow that indicates actionability. I was lost there for a moment and propose we quickly add the "Send" button.

- [x] add send button to `ni-li` component for balances

image

jolesbi:
closes #346 

screen shot 2018-01-12 at 6 00 48 pm


nylira:
Looks good!

PR #369, opened 2018-01-15
importing from seed copy changes 

### LINKED ISSUE #310: importing from seed copy improvements
it's not obvious that when you are importing from seed, the seed is the only thing being used from the past. both name and password are actually being set for the first time. the form should be re-arranged to make this more obvious.

it might even be nice for this to be a 2 step process. 

step 1 - enter the seed
-- next screen --
step 2 - name your account and enter a password (twice - to confirm)

- [x] change language on form to provide more clarity
- [ ] ~implement two step import process from above~

jolesbi:
unfortunately, we can't make this a 2 step process right now because `/keys` expects a name and password. i think this is an ok solution.

closes #310

### DIFF
--- a/app/src/renderer/components/common/NiFormMsg.vue
+++ b/app/src/renderer/components/common/NiFormMsg.vue
@@ -62,7 +62,10 @@ export default {
   padding 0.25rem 0 0
   display flex
   margin 0 !important
- 
+
+  &.sm
+    font-size sm
+
 .ni-form-msg:before
   content ''
   font-family 'Material Icons'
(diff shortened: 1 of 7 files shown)


nylira:
I like all of the changes except for renaming "Import Account" to "Create Account with Seed". There are good reasons to use "Import" instead:

* Seed = Account. Perhaps the seed is being used on the user's other computer, it was generated during the fundraiser, or it's hasn't been used at all, but this seed can conceivably contain an address with tokens. That's the definition of an account, so we don't have to 'create' it again.
* "Create new account" and "Create account with seed" can be confusing for new users. With the options next to each other
    * "Create new account" could mean "create an account (not including a seed)"
    * "Create account with seed" can be interpreted as "create an account (including a seed)".
* "Import account" vs "Create new account" is clearly distinct without having the user read subtitles.
* Other cryptocurrency wallets use the terms "restore" and "import" for generating accounts from existing seed phrases.

jolesbi:
i don't think seed = account. seed = address. accounts are local and include a name and a password for the UI.

that being said, i switched it back to "import" to help with the distinction.

PR #385, opened 2018-01-23
Deterministic zip/tar

### LINKED ISSUE #342: Deterministic ZIP instead of Setup
The current develop branch delivers deterministic builds. I checked this also with removing node_modules and reinstalling them with yarn. The setup files on the other hand are not deterministic. We could instead ship a [deterministic zip](https://github.com/bitgenics/deterministic-zip). This would have the same Checksum every time.

faboweb:
Closes #342


jolesbi:
what’s our strategy for releasing / verifying hashes? is that something we should or should not document somewhere?

faboweb:
I output the MD5 hashsum. We could post this one in the channel to check determinism and also put it on the release page.

image

mappum:
Nice work Fabo! I switched the hash function from MD5 to SHA256 (MD5 can be brute forced in a matter of seconds now!).

Although I'm having problems running the app after extracting it from the zip, seems like the deterministic zip may have messed up some file permissions. It runs fine before the zipping step.

faboweb:
Thx for the test! Could you maybe assist me on this. I will test on Ubuntu. It runs fine on Windows.

faboweb:
:/ It actually neither works on Ubuntu.

mappum:
I looked into it and `zip` doesn't contain some file metadata that is useful for unix filesystems (permissions, etc.). Maybe for Mac and Linux we can just use `.tar.gz` since it preserves that info and we can still make it deterministic: https://github.com/dominictarr/deterministic-tar (at least the .tar part, hopefully we can make the .gz part deterministic).

faboweb:
Sounds great. I experimented with another zipping lib which also didn't work on linux (checkout latest commit). I am down for `.tar.gz` do you want to take car of that? If you have to much to do. I continue on that tomorrow. (E2E tests might suffer but this has higher prio)

faboweb:
`.tar.gz` results in the same behaviour

faboweb:
found it: it is missing executable permissions. will try to leave those in.

Issue #89, opened 2017-11-22
Mock FS in tests


faboweb:
Checking FS operations like if a log is created is difficult if the tests run in parallel on one FS. An idea is to use sth. like MemFS to create a FS in memory for each test.

- [ ] mock FS in tests 😛

PR #391, opened 2018-01-24
Fabo/fs mock

faboweb:
implements #89 

The tests are not deterministic as the tests UI writes to the actual filesystem. This could cause errors when tests ran in parallel influenced each other.
Know every tests gets an own "virtual" filesystem he works on so we can assume deterministic behavior of our code.


.vscode/launch.json
```diff
@@ -0,0 +1,18 @@
+{
+  // Verwendet IntelliSense zum Ermitteln möglicher Attribute.
```
jolesbi: ?

jolesbi:
@faboweb for my own learning, what was the problem and what is the solution? 😆

faboweb:
Sorry. Updated the description.

PR #392, opened 2018-01-24
update readme with code coverage tip

nylira:


faboweb:
This is already in the README unter testing ;)

nylira:
@faboweb ah I see it!

PR #393, opened 2018-01-25
Peng/345 eliminate flashing app

### LINKED ISSUE #345: eliminate flashing app before showing session modal
sometimes you can see the app for a split second before the modal appears. let's try to eliminate this.

- [x] Should not flash app on startup

nylira:
* eliminates flashing app on startup
* eliminates flashing welcome screen if user has an account
* creates a new NiSessionLoading.vue component 
* created test for above component

Closes #345

### DIFF
--- a/app/src/renderer/App.vue
+++ b/app/src/renderer/App.vue
@@ -1,12 +1,13 @@
 <template lang="pug">
 #app
-  app-header
-  #app-content
-    router-view
-    app-footer
-  notifications(:notifications='notifications' theme='cosmos')
-  modal-help
-  session
+  template(v-if="!config.modals.session.active")
+    app-header
+    #app-content
+      router-view
+      app-footer
+    notifications(:notifications='notifications' theme='cosmos')
+    modal-help
+  session(v-else)
 </template>
 
 <script>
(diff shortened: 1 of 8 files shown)


app/src/renderer/App.vue
```diff
@@ -1,12 +1,13 @@
 <template lang="pug">
 #app
-  app-header
-  #app-content
-    router-view
-    app-footer
-  notifications(:notifications='notifications' theme='cosmos')
-  modal-help
-  session
+  template(v-if="!config.modals.session.active")
```
faboweb: Cool! I like the switching of components in the container more then in the component itself. This makes it better composable.

test/unit/specs/components/common/NISession.spec.js
```diff
@@ -30,13 +31,20 @@ describe('NiSessionWelcome', () => {
     })
   })
 
-  it('should not show by default', () => {
-    expect(wrapper.isEmpty()).toBe(true)
+  it('should show by default', () => {
```
faboweb: it is soooo much easier reading changes if the tests are done right. :)

PR #395, opened 2018-01-25
Matt/158 - Show error for baseserver init failure

### LINKED ISSUE #158: No error shown if baseserver fails
There is no error shown in the UI if the baseserver fails.

- [ ] show error in UI if baseserver fails

image

mappum:
Closes #158 

In the case that baseserver fails to start up (rare, but has happened occasionally when the data directory was corrupted), we now show an error notification which stays present in the window. It could look better, but I'm not expecting this to happen much in the wild.

### DIFF
--- a/app/src/renderer/vuex/modules/node.js
+++ b/app/src/renderer/vuex/modules/node.js
@@ -74,11 +74,19 @@ export default function ({ node }) {
       dispatch('pollRPCConnection')
     },
     async checkConnection ({ commit }) {
+      let error = () => commit('notifyError', {
+        title: 'Critical Error',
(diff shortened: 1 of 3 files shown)


app/src/renderer/vuex/modules/node.js
```diff
@@ -74,11 +74,19 @@ export default function ({ node }) {
       dispatch('pollRPCConnection')
     },
     async checkConnection ({ commit }) {
+      let error = () => commit('notifyError', {
+        title: 'Critical Error',
+        body: `Couldn't initialize the blockchain client. If the problem persists, please contact Cosmos support.`
```
jolesbi: "If the problem persists, please make a ticket on <a>Github</a>"

PR #396, opened 2018-01-25
Matt/348 - Native menu bar

### LINKED ISSUE #348: Enable menu bar
The menubar was removed, but it would be nice to have (on mac at least), could potentially keep it disabled on Windows.

- [ ] add native menu bar back for MacOS and possibly Windows
- [ ] add relevant version info in new native menu area

mappum:
Closes #348 

Puts the native menu bar back in.

Currently disabled on Windows, since Fabo said he liked not having it. Although I think I remember there's a convention on Windows to enable it when pressing `alt`?

Right now it's pretty bare, just adding the `About Cosmos UI` menu option, and `Edit` since that's how some users would probably like to copy addresses. Should we add some more just so it looks more complete, like `Window`, `Help`, etc?

Before:

After:

### DIFF
--- a/app/src/main/index.js
+++ b/app/src/main/index.js
@@ -1,6 +1,6 @@
 'use strict'
 
-let { app, BrowserWindow, Menu } = require('electron')
+let { app, BrowserWindow } = require('electron')
 let fs = require('fs-extra')
 let { join } = require('path')
 let { spawn } = require('child_process')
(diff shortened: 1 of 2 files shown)


jolesbi:
@nylira i think there is a way to style it? #darktheme maybe?

nylira:
@jolesbi you mean for Windows? The macOS one is controlled by the user account right?

faboweb:
It just looked ugly and I then thought it was desired to remove it. In the current concept I would add it also on Windows to be consistent. I will create a PR.

Another thing: Do we need the "Edit" part?

PR #397, opened 2018-01-25
WIP: Jordan/121 block tests

jolesbi:
- add block hash to page block 
- use subscription for current block data


nylira:
After taking way too much time trying to get CircleCI to return the same timezone and date as the local `yarn test`, I decided it's best to just  the request the date to be "Loading..." in the test.

jolesbi:
going to start fresh because this PR is causing problems

PR #407, opened 2018-01-29
Fabo/62 dev prod ports

### LINKED ISSUE #62: The process is not quitting if the electron window gets closed

faboweb:
Closes #62 

Ports are now configured in the config.js. There are also different port sets for development and production.


app/src/main/index.dev.js
```diff
@@ -8,7 +8,6 @@
 /* eslint-disable no-console */
 
 // Set babel `env` and install `babel-register`
-process.env.NODE_ENV = 'development'
```
faboweb: removed to enable setting of node_env from the outside

Issue #124, opened 2017-11-27
establish e2e testing strategy


jolesbi:
- [ ] The critical functionalities are tested E2E via some automated browser tests
- [ ] break up e2e into component tickets

faboweb:
Most UI functionality is tested via unit tests.

What is not covered:
- [ ] test if app actually starts
  - included by following tests
- [ ] test if app builds and runs
  - we can have a feedback call in the relay server and have the main process return a promise, that is resolved when the actual vue application lunched without errors
- [ ] test if communicating with the network actually works
  - @mappum how about a TEST_BOT account we persist in the genesis.  we then use this account to send some "test_bot_coins" to himself on the actual network?

faboweb:
Investigating on the e2e tests Matt wrote, I see, that those already cover core topics. I will try to bring those to the latest gaia and UI.

mappum:
>how about a TEST_BOT account we persist in the genesis. we then use this account to send some "test_bot_coins" to himself on the actual network?

I like this idea. This will be easiest with a local testnet, and it will be nice to add other tests here such as delegating.

We should be able to use Spectron to automate the app, I had a good experience with that on the old e2e tests.

faboweb:
Currently blocked on the Linux package of Gaia not being properly installable

PR #410, opened 2018-01-29
Test build executables

faboweb:
USES #409 

Added a test for build executables `yarn run test:exe {path to exe}`

How:
The vue app calls the main thread when connected to the LCD. The main thread will console.log a success message. The test just reads the log and waits for the success message.


jolesbi:
what does this have to do with e2e?

faboweb:
This tests if the app actually works in comparison to only testing parts of the application. Let's discuss the 
 testing strategy in the issue here #124

PR #411, opened 2018-01-29
Check Gaia version at runtime

### LINKED ISSUE #387: switch to gaia-2
- [ ] also check the gaia version in the runtime
- [ ] the chain should state which gaia version it needs

mappum:
Closes #387 

This PR adds a gaia version check at startup, and changes the testnet script to download the `gaiaversion.txt` file.

@faboweb For some reason, many unrelated tests are now failing, I thought it was from updating to gaia `v0.5.0` but the same tests fail on `v0.4.0`. Any guidance would be appreciated.


faboweb:
child_process you used is mocked and never returned resulting in timeouts.

Please merge #416 first

PR #420, opened 2018-01-30
WIP: Add coverage to PRs

faboweb:
Purpose adding coverage reports to PRs so it is easier accessible if enough coverage was reached with a PR


faboweb:
Will merge this to test the integration.

PR #421, opened 2018-01-30
Release documentation

mappum:


jolesbi:
nice! waiting on @faboweb and @peng for approval but this has my 👍

PR #426, opened 2018-02-01
Peng/203 improve buttons

### LINKED ISSUE #203: Improve button design
We need a good primary button design with (most likely) a full color background using our `link` color.

We need an danger! button design using the `danger` color for serious actions like deleting an account.

The default button height of 2rem looks too small on desktop, consider increasing button height to 2.5rem by default?

- [x] primary button color
- [x] success button color
- [x] warning button color
- [x] danger button color
- [x] fix button overflow issue

nylira:
* improves primary button style
* adds support for image icons @faboweb 
* adds success, warning, danger button styles
* fixed the cut-off text bug
* improves the `hover-bg` variable color 

screen shot 2018-02-01 at 1 40 28 pm
screen shot 2018-02-01 at 1 47 46 pm
screen shot 2018-02-01 at 1 40 23 pm

Closes #203

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,3 +31,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Removed
 * Removed unused `sass-loader` node dependency. @nylira
+
+## [0.4.0] - 2018-02-01
+### Added
+* Added button styles for Success, Warning, and Danger states. @nylira
+* Added support for image icons. @faboweb 
+
+### Changed
+* Improved primary button style. @nylira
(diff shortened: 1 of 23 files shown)


nylira:
Instead of adding the `.btn__primary` class to the element, you now have to add the attribute `type="primary"`. This is less prone to breaking CSS changes.

faboweb:
Thx for adding it in the end ^^ Let's see if we need them anywhere.

One note: The colors look very (maybe too) intense. @jolesbi what do you think? But nothing blocking for me.

jolesbi:
they are really intense. but i think it's for the best. they are much brighter than the rest of the app... but i think this means we need to brighten more things as opposed to keeping everything so dark!

PR #427, opened 2018-02-01
Peng/390 combine delegates and validators

### LINKED ISSUE #390: combine validators and delegates
- [x] remove validators index in favour of delegates index
- [x] rename "delegates" in sidebar to "validators"
- [x] add differentiator for validators in index (background color for eg.)
- [x] make sure candidate profile and validator profile are the same component (with flag for candidates and validators!)

nylira:
* Removes the Validators page
* Renamed the Delegates page Validators
* Made validators in the list stand out with a background color and icon
* Added some text on the Delegate profile page to indicate whether its a validator or candidate
* Currently hardcodes validators to the top 10 atom holding/bonded accounts, accessible in the `vuex/modules/config.js` file.

## Screenshots

screen shot 2018-02-06 at 4 21 54 pm
screen shot 2018-02-06 at 4 22 12 pm
screen shot 2018-02-06 at 4 21 58 pm

Closes #390


test/unit/specs/components/staking/__snapshots__/PageDelegateIndex.spec.js.snap
```diff
@@ -213,11 +227,139 @@ exports[`PageDelegate has the expected html structure 1`] = `
...
+              >
+                <div
+                  class="ni-li-dt"
+                >
+                  Max Commission Increase
+                </div>
+                <div
+                  class="ni-li-dd"
+                >
+                  NaN%
```
faboweb: Is this intended?

nylira:
The idea with the file renaming is that we keep all the routes and filenames as "Delegates", not "Validators", because all delegates are not validators. This keeps the naming convention clear for developers. But we rename the sidebar item Delegates -> Validators because it's more important to showcase the term "Validators" than "Delegators" to our users.

test/unit/specs/components/staking/__snapshots__/PageDelegateIndex.spec.js.snap
```diff
@@ -18,7 +18,7 @@ exports[`PageDelegate has the expected html structure 1`] = `
           class="ni-page-header-title"
         >
           <h2>
-            Delegate
+            Validator - undefined
```
faboweb: Is "undefined" intended?

test/unit/specs/components/staking/PageDelegateIndex.spec.js
```diff
@@ -24,22 +24,30 @@ describe('PageDelegate', () => {
...
     expect(wrapper.vm.inCart).toBe(false)
     wrapper.vm.add(wrapper.vm.delegate)
     expect(wrapper.vm.inCart).toBe(true)
     wrapper.vm.rm('pubkeyX')
     expect(wrapper.vm.inCart).toBe(false)
   })
 
+  /*
+  it('should show indicate validator/candidate', () => {
+    expect(wrapper.find('#delegate-country .ni-li-dd').text().trim())
```
faboweb: The name of the test sounds useful, should we maybe add it?

test/unit/specs/components/staking/__snapshots__/PageDelegateIndex.spec.js.snap
nylira: It's intended, because we still don't have commission data.

jolesbi:
> Made validators in the list stand out with a background color and icon

i don't think the background color is going to work on mainnet. since there will be 100 validators, the user will have to scroll for quite a while to see a difference in background color. i think we have the same problem with the icon. especially now that the page is called "validators". 

maybe we need to add a status column that says "Validator" or "Candidate"?

maybe this doesn't matter if the list is randomized?

> Currently hardcodes validators to the top 10 atom

i don't think we should be doing any hardcoding at this point. if `delegate.voting_power > 0` they are a validator.

app/src/renderer/components/staking/LiDelegate.vue
```diff
@@ -1,9 +1,13 @@
...
 .li-delegate(:class='styles'): .li-delegate__values
-  .li-delegate__value.name
-    span
-      router-link(v-if="config.devMode" :to="{ name: 'delegate', params: { delegate: delegate.id }}")  {{ ' ' + delegate.moniker }}
-      a(v-else) {{ ' ' + delegate.moniker }}
+  .li-delegate__value.name: span
+    router-link(v-if="config.devMode"
+      :to="{ name: 'delegate-index', params: { delegate: delegate.id }}")
+      .li-delegate__icon(v-if="delegate.isValidator")
+        img(src="~assets/images/cosmos-validator.png")
```
jolesbi: i don't know how i feel about the icon in the table. maybe on validator profile pages? maybe both? not sure...

app/src/renderer/vuex/modules/config.js
```diff
@@ -2,6 +2,7 @@ import noScroll from 'no-scroll'
 
 export default ({ commit }) => {
   const state = {
+    maxValidators: 10,
```
jolesbi: why is this necessary?

jolesbi:
> The idea with the file renaming is that we keep all the routes and filenames as "Delegates", not "Validators", because all delegates are not validators. This keeps the naming convention clear for developers. But we rename the sidebar item Delegates -> Validators because it's more important to showcase the term "Validators" than "Delegators" to our users.

I used to agree. but now i think that we made the right call with the term "Delegates". if the page is called Validators, everyone will ask, "Where are the candidates"... i don't think we've solved this yet.

jolesbi:
i opened this pr on #463 because i didn't want to mess things up on this branch. merge conflicts resolved over there.

Issue #130, opened 2017-11-28
Use variables in initial splash screen


faboweb:
I as a user
want to always have the initial splash screen using the same styling the rest of the application uses 
so that the styling is consistent.

Idea:
- Render styles to css
- Render css into index.html at build time

PR #431, opened 2018-02-01
calculating splash styles with variables

faboweb:
implements #130

I replaced the existing colors and some sizes with the variables available. Those where no 100% match so I post the comparison below.

Before:
image

After:
image

One thing: I couldn't replace the font-family with the one from the variables (sans). If I do I get following output:
image

Which one is the correct one? (maybe @nylira)

### DIFF
--- a/app/src/renderer/styles/index.styl
+++ b/app/src/renderer/styles/index.styl
@@ -1,0 +1,42 @@
+body { margin: 0; }
+#load {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  width: 100vw;
+  height: 100vh;
+  background: app-bg;
+}
+#load-container {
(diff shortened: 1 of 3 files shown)


app/src/renderer/styles/index.styl
```diff
@@ -0,0 +1,42 @@
...
+#load {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  width: 100vw;
+  height: 100vh;
+  background: app-bg;
+}
+#load-container {
+  font: 16px/1.5 system, -apple-system, BlinkMacSystemFont, "Roboto", "Seogue UI", "Helvetica Neue", "Lucida Grande", sans-serif;
```
nylira: Hardcoding the font family is okay, we won't change this often.

nylira:
Thanks @faboweb!

Issue #380, opened 2018-01-22
let's show a loading graphic when appropriate


jolesbi:
right now, the app defaults to the no-data screen. better than blank, but it should show a loading graphic while loading.

- [x] balances
- [x] tx history
- [x] delegates index
- [x] validators index
- [x] proposals index
- [x] page block

... other places?

jolesbi:
closed by #439

PR #439, opened 2018-02-02
Fabo/380 loading spinners

faboweb:
implements #380

### DIFF
--- a/app/src/renderer/components/staking/LiDelegate.vue
+++ b/app/src/renderer/components/staking/LiDelegate.vue
@@ -42,13 +42,13 @@ export default {
       return value
     },
     vpMax () {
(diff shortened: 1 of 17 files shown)


app/src/renderer/components/staking/LiDelegate.vue
```diff
@@ -42,13 +42,13 @@ export default {
       return value
     },
     vpMax () {
-      if (this.delegates.length > 0) {
-        let richestDelegate = maxBy(this.delegates, 'voting_power')
+      if (this.delegates.delegates.length > 0) {
```
jolesbi: why `delegates.delegates`?

faboweb:
the delegates modules now holds the delegates and the loading state resulting.

PR #448, opened 2018-02-05
Changes to data directory

### LINKED ISSUE #372: address having to delete folders often
increment or label `~/.Cosmos-dev` folder based on network so we can stop deleting this folder. also, should this have a capital letter in it?

- [x] remove capital letter from folder
- [x] use network in name of folder or increment in some other way so we don't have to delete all the time

mappum:
Closes #372 

- Now data is kept in `~/.cosmos[-dev]/<networkName>` to prevent replacing the data when changing networks
- Since we don't replace data anymore, we don't need to copy old data to `X_backup`. Removed all the code for replacing/backing up data, and instead just made the app fail with an error if the data is incompatible/corrupted


mappum:
Added a test to pass the code coverage checks, this is ready as soon as someone reviews it

app/src/main/index.js
```diff
@@ -266,36 +262,6 @@ async function initBaseserver (chainId, home, node) {
   await expectCleanExit(child, 'gaia init exited unplanned')
 }
 
-async function backupData (root) {
```
jolesbi: why don't we need to backup anymore?

mappum: We were previously backing up data when we switched networks, since we had to re-init and we didn't want to just delete the old data. Now we use separate folders for different networks, so we shouldn't ever have to do this.

jolesbi: makes sense!

PR #449, opened 2018-02-05
Jordan/388 profile

### LINKED ISSUE #388: validator profile design improvements
- [x] title should say "Validator Profile" or "Candidate Profile" when appropriate
- [x] add pubkey (should be available)
- [x] remove dummy data

jolesbi:
this is a PR that simplifies the user profile page and the delegate profile page. both candidates and validators can now use `PageDelegate` as their profile pages. i also removed a lot of unused code.

closes #388


nylira:
Ah this may have been better made on top of my https://github.com/cosmos/cosmos-ui/pull/427 PR. A lot of the changes are duplicated.

app/src/renderer/components/common/PageProfile.vue
```diff
@@ -1,48 +1,41 @@
...
     Page,
     Part,
     ToolBar
   },
-  created () {
-    // if (!this.user.signedIn) { this.$store.commit('notifyAuthRequired') }
+  computed: {
+    ...mapGetters(['user']),
+    userInfo () {
+      console.log(this.user)
```
faboweb: was it intended to leave this in? userInfo will be null I guess and is not used in the component

app/src/renderer/components/monitor/PageValidatorIndex.vue
```diff
@@ -1,26 +1,9 @@
 <template lang="pug">
-page(icon="storage" :title="validator.pub_key.data")
+page(title="Validator Profile")
   div(slot="menu"): tool-bar
     router-link(to="/validators" exact)
       i.material-icons arrow_back
       .label Back
-
```
faboweb: so this will be a completely empty page? is this one of the components we don't use anymore?

test/unit/specs/components/staking/PageDelegate.spec.js
```diff
@@ -1,45 +1,53 @@
...
+  params: {
+    delegate: '1a2b3c'
+  }
+}
+
 describe('PageDelegate', () => {
-  let wrapper, store, router
-  let {mount} = setup()
+  let wrapper
+  let store = new Vuex.Store({
```
faboweb: why did you choose to remove the helper?

app/src/renderer/components/common/PageProfile.vue
jolesbi: no. i will remove.

app/src/renderer/components/monitor/PageValidatorIndex.vue
jolesbi: i will delete this file.

test/unit/specs/components/staking/PageDelegate.spec.js
jolesbi: assuming i understand correctly... because the helper installs `localVue.use(VueRouter)` — `$route` becomes read only and cannot be set manually in the test.

https://vue-test-utils.vuejs.org/en/guides/using-with-vue-router.html

faboweb: but before it worked with `router.push('/staking/delegates/pubkeyX')` or not?

faboweb: See: https://github.com/cosmos/cosmos-ui/pull/449/files#diff-b0c8aefb908fde6e011139a1c7f71df6L27

jolesbi: i think it worked before with `router.push` because there were no params involved.

jolesbi:
simple profile page where the word "Validator" will change to "Candidate" if the delegate is not a validator. removed all data that is not available. generic for all delegates.

screen shot 2018-02-07 at 11 16 29 am

test/unit/specs/components/staking/PageDelegate.spec.js
faboweb: The route is defined as `/staking/delegates/:delegate` `:delegate` is the param you want to get in the component. When you push the route `/staking/delegates/pubkeyX` it will set the param `delegate` to `pubkeyX`. This is how it worked before.
As already mentioned, I would like to change the test setup back to using vuex-setup for consistency and to also test vuex-actions existence (lite e2e testing).

jolesbi: OH! ok! let me try that out. that would be better!

PR #450, opened 2018-02-05
Jordan/180 back button

### LINKED ISSUE #180: it would be nice if the back button was on the left side instead of the right
what do you think @nylira ?

jolesbi:
improvements to the back button

closes #180


jolesbi:
closes #180

faboweb:
There are no screenshots of the change. Could you add examples?

Issue #474, opened 2018-02-13
Validator flags are wrong


faboweb:
> Please label the issue either bug or proposal according to its purpose.

### Version
> Please state the version of the UI.

0.4.1

### Description
> Please describe the bug or your proposed change to the software in as much detail as possible. When screenshots can be used to describe the bug/proposal, please include them.

In the "Validators" page, validators start with the label "Validator" but this gets updated to "Candidate" after some seconds.

### Log
> If you report a bug, please provide the log content and the steps required to reproduce the bug. The logs can be found at `%USER_HOME%/.Cosmos/main.log`
```
Tue Feb 13 2018 13:16:19 GMT+0100 (CET) Running Cosmos-UI
main-process: initializing data directory (/home/fabo/.cosmos-dev/gaia-2)
main-process: starting app
main-process: dev mode: true
main-process: winURL: http://localhost:9080
main-process: spawning /home/fabo/go/bin/gaia with args ""version""
main-process: gaia: v0.5.0

main-process: gaia version: "v0.5.0", expected: "v0.5.0"
main-process: Picked seed: 138.68.21.243:46656 of 178.62.35.120:46656,138.68.21.243:46656,159.203.88.224:46656,138.197.131.102:46656,95.85.37.158:46656,159.89.3.34:46656
main-process: Initializing baseserver with remote node 138.68.21.243:46657
main-process: spawning /home/fabo/go/bin/gaia with args ""client" "init" "--home" "/home/fabo/.cosmos-dev/gaia-2/baseserver" "--chain-id" "gaia-2" "--node" "138.68.21.243:46657""
main-process: gaia exited with code 0
main-process: gaia: Loading validator set from tendermint rpc...

main-process: gaia: 
Important: if this is incorrect, all interaction with the chain will be insecure!
  Given validator hash valid: AFAEF490C910CF6124D50EE2BA8FCF0D9883CD16
Is this valid (y/n)?

main-process: approving hash AFAEF490C910CF6124D50EE2BA8FCF0D9883CD16
main-process: gaia exited with code 0
main-process: starting gaia server with nodeIP 138.68.21.243:46657
main-process: startBaseserver /home/fabo/.cosmos-dev/gaia-2/baseserver
main-process: spawning /home/fabo/go/bin/gaia with args ""rest-server" "--port" 9070 "--home" "/home/fabo/.cosmos-dev/gaia-2/baseserver" "--node" "138.68.21.243:46657""
[...]

PR #478, opened 2018-02-14
Peng/bonding changes

### LINKED ISSUE #453: PageBond UI error
i experienced the same thing during my demo, but am unable to reproduce.

1_awgk_lowe4lr45oq-us-pq

https://www.youtube.com/watch?v=m7ARmElOhos&feature=youtu.be

### LINKED ISSUE #455: Unbonding should show different message than when bonding
i agree!

https://medium.com/@speculatecoin/when-i-unbond-atom-ui-shows-atoms-bonded-as-message-7dec047322e4

nylira:
### Your checklist for this pull request

- [x] Make sure you are requesting to **pull your feature/bugfix branch** (right side).
- [x] Make sure you are making a pull request against the **develop branch** (left side). Also make sure to start *your branch* off *our develop*.
- [x] If this PR produces a visible change, please provide screenshots showing these changes.
- [x] Confirm that your pull request will pass our linting and unit tests.
- [x] Make sure your code is properly tested, so that the code coverage is not decreasing.

### Changes checklist
- [x] on PageDelegates, delegates that the user have bonded to before are automatically checked and locked for use on PageBond. Because it doesn't make sense to allow the user to only bond a portion of their atoms, and it'll be confusing when it shows them that they only have a few unbonded atoms left.
- [x] bonding bar width is limited to remaining user atoms when increasing bonded atoms via the numerical input field
- [x] numerical input field for atoms per delegate has a min and max limit so that clicking the up and down arrows for the field don't pass 0 or max
- [x] resolve bonding bar width bug where it scales out of control
- [x] resolve lack of unbonding message

### Screenshots
The screenshot below shows what happens on PageDelegates on initial load, if the user has already delegated atoms in the past. Those validators are already checked and cannot be unchecked. If they want to 'uncheck' those validators they have to unbond from them first.

screen shot 2018-02-14 at 4 22 55 pm

### Issue
Closes #453 
Closes #455

### DIFF
--- a/app/src/renderer/components/staking/LiDelegate.vue
+++ b/app/src/renderer/components/staking/LiDelegate.vue
@@ -12,7 +12,9 @@
   .li-delegate__value.status
     span {{ delegateType }}
   template(v-if="userCanDelegate")
-    .li-delegate__value.checkbox#remove-from-cart(v-if="inCart" @click='rm(delegate)')
+    .li-delegate__value.checkbox(v-if="amountBonded > 0")
+      i.material-icons check_box
+    .li-delegate__value.checkbox#remove-from-cart(v-else-if="inCart" @click='rm(delegate)')
       i.material-icons check_box
     .li-delegate__value.checkbox#add-to-cart(v-else @click='add(delegate)')
       i.material-icons check_box_outline_blank
@@ -38,7 +40,7 @@ export default {
     },
     styles () {
       let value = ''
-      if (this.inCart) value += 'li-delegate-active '
+      if (this.inCart || this.amountBonded > 0) value += 'li-delegate-active '
       if (this.delegate.isValidator) value += 'li-delegate-validator '
       return value
     },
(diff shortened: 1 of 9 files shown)


nylira:
I'm out of time to work further on this. I have notes on #453, the out of control bonding bar problem.

This bug is caused by the fact that `this.fields.delegates` is either undefined or defined as an empty array on this line https://github.com/cosmos/voyager/blob/d947837db2c6df37145d6d83f1fc1ac6baae91ac/app/src/renderer/components/staking/PageBond.vue#L296. Because the function cannot find the correct delegate, the bar width has no limit, so it goes crazy.

I'm not sure why `this.fields.delegates` is undefined there, because when the page is mounted, `this.fields.delegates` seems to be already defined.

faboweb:
- fixed the issue with the bonding bar
- fixed issue #474

faboweb:
@jolesbi can u take "resolve lack of unbonding message"?

Issue #460, opened 2018-02-08
Add analytics to the UI


faboweb:
Analytics should only be available for certain builds that clearly indicate, that they use analytics.

faboweb:
We decided to add Google Analytics and Sentry.

faboweb:
Whitelist Gaia up to gaia-3

jolesbi:
closed by #482

PR #482, opened 2018-02-15
Add analytics

faboweb:

### Issue

#460

❤️ Thank you!


app/src/main/index.js
```diff
@@ -10,11 +10,12 @@ let semver = require('semver')
 let event = require('event-to-promise')
 let toml = require('toml')
 let axios = require('axios')
+let Raven = require('raven')
 
 let pkg = require('../../../package.json')
 let relayServer = require('./relayServer.js')
 let addMenu = require('./menu.js')
-let config = require('../../../config.js')
+let settings = require('../../../config.js')
```
jolesbi: requiring config.js and naming as settings?

app/src/network.js
```diff
@@ -1,9 +1,10 @@
 let { join } = require('path')
 let { readFileSync } = require('fs')
+let config = require('../../config.js')
 
 // this network gets used if none is specified via the
 // COSMOS_NETWORK env var
-let DEFAULT_NETWORK = join(__dirname, '../networks/gaia-2')
+let DEFAULT_NETWORK = join(__dirname, '../networks/' + config.default_network)
```
jolesbi: does the space matter here?

app/src/main/index.js
faboweb: because of a conflict with https://github.com/cosmos/voyager/blob/c37327e14f41732d6c2d15016fdf58febaba6bec/app/src/main/index.js#L478
I will rename the other instance then.

app/src/network.js
jolesbi: nvm. brain fart.

jolesbi:
small detail, but i much prefer `configTOML`!! thanks!

Issue #334, opened 2018-01-08
If gaia init fails app doesn't recover


faboweb:
On starting the gaia we have to init. currently init breaks in some cases with a hash mismatch. In this case we procede to starting gaia but get an error that we need to init gaia first. On consecutive start of the app we never init gaia again.

faboweb:
Just experienced it again:
![image](https://user-images.githubusercontent.com/5869273/36095560-f13ffb5a-0ff2-11e8-8bd3-57f63d89a9b5.png)

jolesbi:
how do we get around this in the mean time?

faboweb:
The user can remove the config folder. This "resets" the application.
My idea to simply solve this is to react to the specific error and init Gaia again.

jolesbi:
closed by #438

PR #483, opened 2018-02-16
Fix app not recovering from failed Gaia initialization

faboweb:

### Issue

#334 

❤️ Thank you!


app/src/main/index.js
```diff
@@ -328,6 +328,28 @@ function consistentConfigDir (appVersionPath, genesisPath, configPath, gaiaVersi
     exists(gaiaVersionPath)
 }
 
+// check if baseserver is initialized as the configs could be corrupted
+// we need to parse the error on initialization as there is no way to just get this status programmatically
```
jolesbi: could we ask gaia to return the status programatically?

faboweb: in theory this would be possible. but I think it doesn't have too high priority. Should I create an issue?

jolesbi: yeah, create an issue. just worries me a bit to depend on parsing the error.

PR #490, opened 2018-02-20
Made tar bundling not follow symlinks

### LINKED ISSUE #430: reduce size of mac distribution
When creating the deterministic tars (#422) we noticed, that apparently the mac `tar.gz` is very big as it pulls in or copies several symlinked files.

- [ ] reduce size of mac distribution to an acceptable level

mappum:
This took *way* more work to get working than I expected, but now our tarballs preserve symlinks instead of duplicating files (which brings our macOS builds from 150MB to 50MB).

### Issue

Closes #430

### DIFF
--- a/tasks/release.js
+++ b/tasks/release.js
@@ -151,34 +151,44 @@ function zipFolder (inDir, outDir, version) {
   })
 }
 
-function tarFolder (inDir, outDir, version) {
-  return new Promise(async (resolve, reject) => {
-    let name = path.parse(inDir).name
-    let outFile = path.join(outDir, `${name}_${version}.tar.gz`)
-    var pack = tar.pack()
+async function tarFolder (inDir, outDir, version) {
+  let name = path.parse(inDir).name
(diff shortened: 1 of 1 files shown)


ebuchman:
#hero

tasks/release.js
```diff
@@ -202,17 +212,17 @@ function deterministicTar () {
...
       .on('entry', function (header, stream, cb) {
-        if (header.type !== 'file') return cb()
-
         header.mtime = header.atime = header.ctime = UNIXZERO
         header.uid = header.gid = 0
 
         delete header.uname
         delete header.gname
 
-        header.mode = 0o777
```
faboweb: Execution on Linux still works without this?

mappum: Now we inherit the file mode from the `lstat`, so that we don't have to give full permissions to every file. https://github.com/cosmos/voyager/pull/490/files#diff-d95e7c916be2a81522fee0617e0d16a9R177

Issue #491, opened 2018-02-21
Block explorer doesn't use connected node


faboweb:
> Thank you for filing an issue!

### Version

Commit 8e3485367aaeaa2fe389cfa567024b60da5a94c6

### Description

The block explorer queries for block details at `https://gaia-2-node0.testnets.interblock.io/` it should use the connected node.
The loading indicator also doesn't hide when the request fails.

PR #496, opened 2018-02-23
Fabo/491 get block via node

faboweb:

### Issue

closes: #491 
Also removes some clutter like the url to the node in PageBlock. I think this delivers no value there.

### Screenshots

Before:
image

After:
image

❤️ Thank you!


jolesbi:
Thank you a lot for contributing to Cosmos Voyager! 😆

Issue #239, opened 2017-12-13
New component proposal: full page error


nylira:
For error states that block all app functionality. Example: LCD Failure (#228).

nylira:
Will work on this tomorrow.

PR #508, opened 2018-02-28
peng/create modal for fullscreen error

nylira:
- [x] create NiModalError component with editable icon, title, and body
- [x] allows user to click to create an issue from the error page
- [x] allows user to view their `~/.Cosmos/main.log` from the error page, no matter their OS
- [x] appears on top of everything with new z-index

### How to create a full page error
Somewhere in the app, run `this.$store.commit('setModalError', true)`. And then you can edit the error's title and message with props. For a quick demonstration of this error, set `config.modals.error.active = true` in vuex.

### Issue
closes: #239

### Screenshots
screen shot 2018-02-28 at 3 20 19 pm
screen shot 2018-02-28 at 3 20 30 pm
image


nylira:
@mappum @faboweb I need help implementing working tests for this new component. I created the tests but they're failing because Jest cannot figure out this line:

```
    this.logPath = remote.app.getPath('home') + '/.Cosmos/main.log'
```

Does `remote` from `electron` have be to imported into the test somehow? Or should I just simplify it and create some sort of local remote variable?

faboweb:
I didn't know electron came with `remote.app.getPath('home')`. Nice one.

Electron injects the `electron` package when running. My suggestion is to just mock it like here: https://github.com/cosmos/voyager/blob/develop/test/unit/specs/App-w-analytics.spec.js#L20

jolesbi:
really nice work @nylira !

Issue #498, opened 2018-02-23
The app crashes if node not available at init step


faboweb:
> Thank you for filing an issue!

### Version

0.4.3

### Description

The app crashes and shuts down if node not available at init step

### Log

```
Fri Feb 23 2018 17:46:55 GMT+0100 (Mitteleuropäische Zeit) Running Cosmos-UI
main-process: initializing data directory (C:\Users\Fabo\.cosmos-voyager\gaia-2)
Fri Feb 23 2018 17:47:51 GMT+0100 (Mitteleuropäische Zeit) Running Cosmos-UI
main-process: root exists (C:\Users\Fabo\.cosmos-voyager\gaia-2)
main-process: configs are compatible with current app version
main-process: starting app
main-process: dev mode: false
main-process: winURL: file://C:\Users\Fabo\Documents\tendermint\cosmos-ui\app\src\main/index.html
main-process: spawning C:\Users\Fabo\Documents\Entwicklung\Go\bin\gaia.exe with args ""version""
main-process: gaia.exe: v0.5.0

main-process: gaia version: "v0.5.0", expected: "v0.5.0"
main-process: Picked seed: 159.89.3.34:46656 of 178.62.35.120:46656,138.68.21.243:46656,159.203.88.224:46656,138.197.131.102:46656,95.85.37.158:46656,159.89.3.34:46656
main-process: spawning C:\Users\Fabo\Documents\Entwicklung\Go\bin\gaia.exe with args ""client" "init" "--home" "C:\\Users\\Fabo\\.cosmos-voyager\\gaia-2\\baseserver""
main-process: gaia.exe exited with code 0
main-process: gaia.exe: ERROR: "--chain-id" required

main-process: Trying to initialize baseserver with remote node 159.89.3.34:46657
main-process: spawning C:\Users\Fabo\Documents\Entwicklung\Go\bin\gaia.exe with args ""client" "init" "--home" "C:\\Users\\Fabo\\.cosmos-voyager\\gaia-2\\baseserver" "--chain-id" "gaia-2" "--node" "159.89.3.34:46657""
main-process: gaia.exe exited with code 1
main-process: gaia.exe: Loading validator set from tendermint rpc...

main-process: gaia.exe: ERROR: Status: Post http://159.89.3.34:46657: dial tcp 159.89.3.34:46657: connectex: Es konnte keine Verbindung hergestellt werden, da der Zielcomputer die Verbindung verweigerte.

main-process: gaia.exe exited with code 1
```

PR #536, opened 2018-03-06
Show full page error on crash

faboweb:

### Issue

closes: #498

This passes any error in the main thread to the view and shows the full page error, showing the error message.
This will not shutdown the application on an error anymore.

❤️ Thank you!


app/src/renderer/main.js
```diff
@@ -6,7 +6,7 @@ import Vuelidate from 'vuelidate'
 import shrinkStacktrace from '../helpers/shrink-stacktrace.js'
 import axios from 'axios'
 import Raven from 'raven-js'
-import {remote} from 'electron'
+import { remote } from 'electron'
```
jolesbi: i am in favour of these stylistic changes, but we should consider (agreeing on and) using a linter so no one changes this back / future styles are the same.

app/src/renderer/main.js
```diff
@@ -51,32 +54,37 @@ async function main () {
...
 
-  const store = Store({ node })
+  store = Store({ node })
 
-  let connected = await store.dispatch('checkConnection')
-  if (connected) {
-    store.dispatch('nodeSubscribe')
-    store.dispatch('showInitialScreen')
-    store.dispatch('subscribeToBlocks')
+  let error = getQueryParameter('error')
```
jolesbi: i find it odd that we have to use url's even in a desktop app. anyone have any thoughts on this? not a big deal, just a curiosity.

app/src/renderer/main.js
```diff
@@ -51,32 +54,37 @@ async function main () {
...
+  if (error) {
+    store.commit('setModalError', true)
+    store.commit('setModalErrorMessage', error)
+  } else {
+    let connected = await store.dispatch('checkConnection')
+    if (connected) {
+      store.dispatch('nodeSubscribe')
+      store.dispatch('showInitialScreen')
```
jolesbi: does this mean that if you get the full screen error, you have to sign back in 100% of the time?

app/src/renderer/vuex/modules/node.js
```diff
@@ -43,13 +43,14 @@ export default function ({ node }) {
...
     },
-    nodeSubscribe ({commit, dispatch}) {
+    async nodeSubscribe ({commit, dispatch}) {
       if (state.stopConnecting) return
 
       // the rpc socket can be closed before we can even attach a listener
       // so we remember if the connection is open
       // we handle the reconnection here so we can attach all these listeners on reconnect
       if (!node.rpcOpen) {
+        await sleep(500)
```
jolesbi: awesome

app/src/renderer/main.js
faboweb: 👍 
vs code was messed up for a moment and reformatted stuff. Wrote in the chat about agreeing on one rule.

app/src/renderer/main.js
faboweb: Currently this is the easiest solution. We can/should switch to a socket communication between the main thread and the browser.

app/src/renderer/main.js
faboweb: The full page error will block you from doing anything but close the application. So you have to sign back in the next time.

app/src/renderer/vuex/modules/node.js
faboweb: This was needed so the process isn't blocked while continuously  trying to reconnect.

PR #542, opened 2018-03-07
Peng/538 voyager mvp

### LINKED ISSUE #538: List Voyager MVP features

nylira:
I've started on the MVP feature list for Voyager. It's a markdown document in the `/docs` directory.

### Issue
Closes #538


docs/mvp-features.md
```diff
@@ -0,0 +1,50 @@
+# Voyager MVP Features
+These features of Voyager are considered essential for MVP. They are required to be completed before we can launch the Cosmos Hub.
+
+## Managing accounts
+
+The user should be able to:
+
+* Create an account with a new seed and address
+* Import an account with a Cosmos seed
+* Import an account with the seed contained in hardware (i.e. Ledger)
```
faboweb: I think this should be: Register a private key stored on hardware (i.e. Ledger)

As I think the key is not imported into the keystore.

faboweb:
Pulling in @ebuchman

nylira:
I'm not clear how this PR causes CI to fail. I didn't even touch any of the code! Something in the e2e test is too delicate and prone to breakage:

```
unhandledRejection Error: An element could not be located on the page using the given search parameters ("#sign-in-name select").
    at element("#sign-in-name select") - $.js:40:17
    at click() - common.js:78:34
```

nylira:
Thoughts on what else to add to this document? Or should we merge this and make issues for additions?

faboweb:
@nylira I would like to get an approval by @ebuchman. Are you OK with leaving the PR open until then?

NodeGuy:
It looks good but I'd like to discuss it during the team meeting before merging.  I've added it to the agenda for tomorrow.

NodeGuy:
In service of launching sooner and reducing the security audit cost I propose that we simplify this specification as much as possible for the first release.  Specifically, I propose that we cut:

- `Register a private key stored on hardware (i.e. Ledger)`
  - If not 100% of our users will be using hardware then it's not required for minimum viability.
- `Exploring blocks`
  - Even if this is already implemented it creates an attention cost and has already generated at least one issue.  Let's disable it with a feature toggle for the first release.

faboweb:
Summing up discussions I had:
- Multiple user support: This is already solid and there is no need to remove it.
- Register a private key stored on hardware: May be deferred.
- Filter and search through past transactions: May be deferred.
  - If we show all transactions, this may be enough for a version 1.
- Exploring blocks: Is pretty solid. Should be stable enough for launch.

nylira:
Ready to go.

Issue #464, opened 2018-02-10
PanelSort only sorts % of vote and total vote


jolesbi:
- [x] sort "name"
- [x] sort "your votes"
- [x] sort "status"

PR #549, opened 2018-03-08
Jordan/464 sorts

jolesbi:

### Issue

closes: #464 

### Screenshots

❤️ Thank you!


faboweb:
I forgot to ask yesterday:
Why did you rename "your bonds" to "your votes". Isn't that sth. different?

jolesbi:
the main reason was because the column title says "Your Votes" and i felt this was more consistent.

faboweb:
I think the column title was wrong. What do you think @nylira ?

jolesbi:
why? i think it make sense. it's the number of votes that the user has delegated to the validator. we can wait for @nylira though 👍

faboweb:
But it's not votes it's atoms right? Votes would be in the context of governance I thought.

mappum:
I agree that "votes" would get confused for the governance votes, but also saying "Your Bonds" maybe isn't perfect either. Maybe we already moved away from this, but should it not be "Your Delegations"?

faboweb:
What about "Your Stake"?

nylira:
I like "Your Stake". It's clearer than delegations, bonds, and votes.

jolesbi:
though i understand where you guys are coming from, i think the rest of the columns and the context leave "Your Votes" as the best choice.

```
makes sense to me
...
- % of Vote
- Total Votes
- Your Votes
```

```
makes less sense to me
...
- % of Vote
- Total Votes
- Your Stake
```

though this is maybe a bad time to bring this up again, the purple bar doesn't make any sense to me and i wish it would be on the "% of Vote" column as a visual representation of the percentage.

screen shot 2018-03-12 at 12 16 16 pm

Issue #534, opened 2018-03-06
Use docker the build


faboweb:
> Thank you for filing an issue!

### Version

0.4.3

### Description

We still have problems to create deterministic builds cross platforms. I propose readding docker to the build process.

PR #563, opened 2018-03-20
Docker

NodeGuy:
### Issue

closes: #534


faboweb:
Would `cd tasks/build && sh ./build.sh` also work on Unix as a build command?

faboweb:
I get some strange issues, can someone help me with them?
image

NodeGuy:
> Would cd tasks/build && sh ./build.sh also work on Unix as a build command? Without sh it doesn't work on Windows.

Yes.

faboweb:
It works!!! :)
Would it be possible to remove the default `--binary` so one can set a specific binary?

NodeGuy:
> Would it be possible to remove the default --binary so one can set a specific binary? This way we can easily compare build hashes.
>
> One idea: Maybe we can make the script take a git-tag for Voyager and then it pulls the code according to the tag. This could reduce confusion about which commit are we on right now.

How about replacing `--binary` with `--gaia-commit` to specify a commit for Gaia and introducing `--commit` for specifying a Voyager commit?  We shouldn't allow the use of tags because that would open a security hole until we start signing them.

faboweb:
Didn't Greg push for using built binaries? I am OK with `--commit` for Voyager if that is securer. 

Note: I put signing commits on the agenda for next Tuesday. ;)

NodeGuy:
Yes but we can specify which prebuilt binary to use via the commit.

faboweb:
But you would still build Gaia inside the docker?

NodeGuy:
> But you would still build Gaia inside the docker?

No, I'll download the prebuilt version similar to this:

```
wget --output-document=/usr/local/bin/gaia \
  https://tendermint-packages.interblock.io/binaries/gaia_linux_amd64/gaia_${commit}
```

NodeGuy:
OK it's not that pretty but I don't think we want to spend a lot of time on this.  To build you now specify Gaia and Voyager commits on the command line before other arguments (Gaia's must be a hash and Voyager's can be any Git reference):

```sh
yarn run build decb23bed0179244e4f42d697dcd4bf759200a2d 0.4.3 --platform={darwin|win32|linux}
```

One thing I'm not thrilled about is that we're trusting Interblock and HTTPS with the integrity of Gaia:

```sh
wget --output-document=/usr/local/bin/gaia \
  https://tendermint-packages.interblock.io/binaries/gaia_linux_amd64/gaia_$1
```

Maybe this won't matter when it's all on the same build machine or maybe we want to do something fancy with signing the binaries.

TBD: See if we can add the build step to the CI script.

faboweb:
I didn't knew we have Gaia builds for every commit. That is great! Let's do it like this then.

@mappum already implemented building the binaries in docker before. I am not sure if building Gaia is deterministic.

NodeGuy:
Sorry, I was unclear.  We don't have Gaia builds for every commit; we have exactly one provided by @greg-szabo.  If he builds another one then we can specify that too.  Originally I was building Gaia in the container and we could have done it for any arbitrary commit but @greg-szabo objected.

faboweb:
Then I suggest we go with `--binary` first and migrate when we have "Gaia" (or what will follow) on every commit. What do you think?

NodeGuy:
So you want the user to be able to provide their own arbitrary version of Gaia?  I think that would go in the opposite direction of making the build more deterministic.

Let's get @greg-szabo into this conversation.  Greg, what are your thoughts?

faboweb:
Not an arbitrary Gaia build but a Gaia build that is shareable and has a fixed checksum. The purpose is, that different people on different computers can build Voyager and still create the same hash in the end so we can prove that the builds we ship have no malicious code in it.

For development we can always fall back to the installed Gaia/SDK.

NodeGuy:
Ah, OK.  That's exactly how it is now in this pull request.

greg-szabo:
Ok, I was able to get through the conversation. Great work. Some notes for clarity:
1. interblock.io is ours, I manage the DNS, the certificate, and the website. We don't have to worry about trust there (apart from regular paranoia). Eventually, we could move it to packages.tendermint.com or Nexus/Artifactory for better artifact integrity. (And zip it too.)
2. I think the `--binary` option makes sense. It should accept a binary on your local machine (built by the dev or automation). In that case, it's easy for developers to add any random gaia binary for testing. Automation deals with downloading binaries from the repository separately. (So, automation will also have a local version of the binary.) For what it's worth, the whole wget part should not be needed or be done separately outside of the container.
3. I still don't really get in a lot of cases why we build Docker containers though. The binary could be added as part of a volume so we don't rebuild the container for each and every new binary. I think we're abusing docker in a lot of cases. (Thoughts for a later discussion.)
4. Automation is coming along well, I should be able to create gaia binaries for the repository quite easily. Eventually, it will be automated for every PR and we can have a manual trigger for any branches that you come up with.
5. Voyager CI/CD is not that far along, yet. :(

faboweb:
@greg-szabo 
1) I don't think we should trust anything, not even ourselves.
3) For deterministic builds we need deterministic environments. Docker is the logical solution I guess. Do you have another solution?
5) Not a problem. If this will get important to use, we will push you and/or help you implement this.

.circleci/config.yml
```diff
@@ -31,11 +31,11 @@ jobs:
             - node_modules
           key: v1-dependencies-root-{{ checksum "package.json" }}
 
-      - run: 
+      - run:
           name: Add Gaia
           command: |
-            sudo wget https://www.dropbox.com/s/vkyuut8hnzse614/gaia-0.5.0-linux-32?dl=1 -O ~/repo/gaia
+            sudo wget https://tendermint-packages.interblock.io/binaries/gaia_linux_amd64/gaia_decb23bed0179244e4f42d697dcd4bf759200a2d -O ~/repo/gaia
```
faboweb: Thanks. I uploaded this myself as there was nothing there when I implemented it.

tasks/release.js
```diff
@@ -7,38 +7,40 @@ const packager = require('electron-packager')
...
+  { default: { platform: `darwin,linux,win32` } }
+)
+
+if (binaryPath) {
+  fs.accessSync(binaryPath)
+  console.log('Using prebuilt binary', binaryPath)
+} else {
   console.error(`\x1b[31mPlease specify a gaia binary for this platform using the "--binary" flag
-    Example: npm run build:darwin -- --binary=./gaia
+    Example: node tasks/release.js --binary=./gaia
```
faboweb: shouldn't this be `yarn run build --platform=darwin --binary=./gaia`?

tasks/build/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+    wine \
+    wine32 \
+    libwine
+
+# Use Gaia v0.5.0 by default.
+ARG GAIA_COMMIT=decb23bed0179244e4f42d697dcd4bf759200a2d
+
+# Download Gaia.
+RUN wget --output-document=/usr/local/bin/gaia \
+  https://tendermint-packages.interblock.io/binaries/gaia_linux_amd64/gaia_$GAIA_COMMIT
```
faboweb: So we still download Gaia by commit?

README.md
```diff
@@ -93,19 +93,33 @@ $ COSMOS_NODE=localhost yarn testnet
...
+Build the app:
+
+The first argument is a commit for Voyager (e.g., the tag `0.4.3`).  Optional
+arguments (such as `--platform`) must come afterwards.
+
+```bash
+yarn run build 0.4.3 --platform={darwin|win32|linux}
+```
+
+You can specify a commit hash for the version of Gaia to use if it has been prebuilt:
```
faboweb: Didn't we agree, that devs need to provide a build Gaia binary? Shouldn't the documentation reflect that? Maybe we can also add why we chose to do it that way.

tasks/build/Dockerfile
faboweb: It still downloads if I specify the binary at least and this is only the linux binary. I think this could confuse devs.

README.md
```diff
@@ -93,19 +93,33 @@ $ COSMOS_NODE=localhost yarn testnet
...
+Run the app.
 ```bash
-yarn run build --platform={darwin|win32|linux} -- --binary={path to the gaia binary}
 open builds/Cosmos-{platform}-x64/Cosmos.app
 ```
 
 When you are testing the build system you can skip the repackaging of the JS files.
 ```bash
-$ yarn run build --platform= --{darwin|win32|linux} --skip-pack --binary=...
+$ yarn run build 0.4.3 --platform=darwin --skip-pack
```
faboweb: Is the here specified version used somewhere? What do you think about taking it from the package.json?

README.md
```diff
@@ -93,19 +93,33 @@ $ COSMOS_NODE=localhost yarn testnet
...
 
-Install [Docker](https://docs.docker.com/get-started/).
+Building requires that [Docker](https://www.docker.com/get-docker) is installed
+on your system.
 
-Build and run the app.
+Build the app:
+
+The first argument is a commit for Voyager (e.g., the tag `0.4.3`).  Optional
+arguments (such as `--platform`) must come afterwards.
```
faboweb: if `--platform` is optional, then we need to detect the OS of the dev I think. I would prefer making it mandatory so the dev knows for what he is building. The build process will in most cases be cross-platform I guess.

package-lock.json
```diff
@@ -104,6 +104,33 @@
         "js-tokens": "3.0.2"
```
faboweb: What do you think about removing the `package-lock.json`? Having it could create confusion for devs as we have the `yarn.lock` as our primary lock file. I think a `yarn` would not update the `package-lock.json`, am I right?

faboweb:
`yarn run build --platform=win32 --binary="C:\Users\Fabo\Downloads\gaia-0.5.0-windows-32.exe"` doesn't currently work:

image

greg-szabo:
Uh, I find it hard to discuss these through a message board. I was under the assumption that the question is about building gaia for voyager versus bringing in a binary that was built earlier. If it's about using Docker vs not using Docker, there are pros and cons for both. I can summarize them, if necessary. Just one note in advance: although Docker helps, it doesn't solve the problem of deterministic builds by itself.

I was also under the assumption that David was writing Dockerfiles, so I thought that's what the team is going with, for Voyager. In this regard, bringing in a pre-built binary is possible and encouraged, because the pre-built binary went through unit-testing and integration testing. (Whereas if you build it, it didn't.)

tasks/build/Dockerfile
NodeGuy: Yes, the commit is used to identify the Gaia version.

We want only the Linux binary because it's executed in the (Linux) Docker container.

tasks/release.js
NodeGuy: No, `yarn run build` and `node tasks/release.js` are two different processes with two different sets of arguments.  `yarn run build` creates a Docker container and runs `node tasks/release.js` within it.

README.md
NodeGuy: It's not a version, it's a commit reference.  It's explained a few lines earlier in the README:

> The first argument is a commit for Voyager (e.g., the tag `0.4.3`).  Optional
arguments (such as `--platform`) must come afterwards.

I added this in response to your request:

> Maybe we can make the script take a git-tag for Voyager and then it pulls the code according to the tag.

README.md
NodeGuy: If `--platform` isn't specified then it uses the default that was already in the code: `darwin,linux,win32`

package-lock.json
NodeGuy: Oops, that was a mistake.

README.md
faboweb: It is not possible to have a default platform. The binary is OS specific. I think it is a mistake it is still in there.

tasks/build/Dockerfile
faboweb: I am not sure I understand. I was under the impression, that the docker container isn't executing Gaia but it is just adding it to the files that get bundled into Voyager for each specified platform.

tasks/release.js
faboweb: Ok cool. But wouldn't the dev be confused, if he did a mistake in `yarn run build` and gets a recommendation about `node tasks/release.js`? How about catching this error in the build script then?

README.md
faboweb: Thanks for clarifying this.

tasks/build/entrypoint.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+set -o errexit
+
+git clone /.git .
+git checkout $1
```
faboweb: How about adding comments about which parameters are passed around? This could increase the readability.

NodeGuy:
> I think the --binary option makes sense. It should accept a binary on your local machine (built by the dev or automation). In that case, it's easy for developers to add any random gaia binary for testing. Automation deals with downloading binaries from the repository separately. (So, automation will also have a local version of the binary.) For what it's worth, the whole wget part should not be needed or be done separately outside of the container.

> I was also under the assumption that David was writing Dockerfiles, so I thought that's what the team is going with, for Voyager. In this regard, bringing in a pre-built gaia binary is possible and encouraged, because the pre-built binary went through unit-testing and integration testing. (Whereas if you build it, it didn't.)

@greg-szabo, I'm confused because these sound contradictory to me.  I had originally designed the build process to allow a random Gaia build but after our conversation I thought you wanted me to download the pre-built binary only.  Are you softening that stance?

> I still don't really get in a lot of cases why we build Docker containers though. The binary could be added as part of a volume so we don't rebuild the container for each and every new binary. I think we're abusing docker in a lot of cases. (Thoughts for a later discussion.)

I'm building Voyager in a container to control the build environment.  The image is built only once per Gaia version and reused for every Voyager build.

tasks/build/entrypoint.sh
NodeGuy: Done.

tasks/build/Dockerfile
NodeGuy: Oops, you're right of course!  I'll fix that.

README.md
NodeGuy: Right, that makes sense.

greg-szabo:
> I thought you wanted me to download the pre-built binary only.
The emphasis is on the `pre-built`, not on the `download`. Don't build the gaia binary in your voyager build process, always use a pre-built one. If - in a separate step - you want to build a specific branch, then it should be ok for developer use. (With the addition that those voyager binaries should never be published, since they are untested.) If you want to publish voyager, always download gaia from the binary repository (because those are tested).

I think I didn't understand the question about the `-binary` parameter, I have no idea what it does in yarn. If it builds a gaia binary, don't use it. If it just points to the (pre-built) gaia binary on your local machine, that should be ok.

> The container (image) is built only once per Gaia version and then reused for each Voyager build.

I don't understand why there needs to be a Docker container built for each gaia version. The gaia binary can be fed into a Docker container using the docker volume option. It does not need to be baked into the docker image. I think we (as a team, not just in this case) are abusing Docker compose which slows down automation for no real benefit. Docker is fine to use but most probably you can use the standard `yarn` image (if there is one) and feed your script and the files necessary through volume. This way you don't need to build any docker images at all.

NodeGuy:
> The emphasis is on the pre-built, not on the download. Don't build the gaia binary in your voyager build process, always use a pre-built one. If - in a separate step - you want to build a specific branch, then it should be ok for developer use. (With the addition that those voyager binaries should never be published, since they are untested.) If you want to publish voyager, always download gaia from the binary repository (because those are tested).

Ah OK, now I understand.  Thanks for explaining it to me.

@faboweb What are the use cases for developers building their own Voyager production applications?  Is there any reason to create a separate process outside of the official build-and-test process of the binary repository?

> I think I didn't understand the question about the -binary parameter, I have no idea what it does in yarn. If it builds a gaia binary, don't use it. If it just points to the (pre-built) gaia binary on your local machine, that should be ok.

The `--binary` parameter is one that we're making up and discussing.  It has nothing to do with Yarn.

> I don't understand why there needs to be a Docker container built for each gaia version. The gaia binary can be fed into a Docker container using the docker volume option. It does not need to be baked into the docker image. I think we (as a team, not just in this case) are abusing Docker compose which slows down automation for no real benefit. Docker is fine to use but most probably you can use the standard yarn image (if there is one) and feed your script and the files necessary through volume. This way you don't need to build any docker images at all.

You're right, Gaia could be fed in through a volume.  I was doing it in the image because of two assumptions a) the cost would be low because we won't be changing the Gaia version often and 2) we should take control of it out of the developer's hands because we're trying to to create predictable builds.  After the above discussion about binaries, however, I'm going to remove Gaia from the build process and treat it as an input from the developer.

We can't just use the standard Node.js image because we need to install Wine as well for the Windows builds.

greg-szabo:
Ok, cool, I like the way this is going. As for the Node+Wine image, I get the point, that needs to be built. To make the process faster and easier, I propose that an docker image like that is managed separately (I can help adding an image like that to the tendermint repo on hub.docker.com). This way the build process is literally going to be `docker pull` + `docker run -v` which seems a more streamlined process.

Also, I assume the Node+Wine image doesn't change much, that's why I think it's worth separating it. If that's not the case, then let's discuss further.

faboweb:
Sounds great :)

nylira:
I got a warning building for darwin. Not sure if it's important.

```
Step 7/7 : ENTRYPOINT [ "/entrypoint.sh" ]
 ---> Running in 53cf666c702e
Removing intermediate container 53cf666c702e
 ---> 05fbe9685625
Successfully built 05fbe9685625
Successfully tagged cosmos/voyager-builder:latest
(node:33369) UnhandledPromiseRejectionWarning: TypeError: Expected a string, got undefined
    at module.exports.str (/Users/pz/Projects/voyager/node_modules/untildify/index.js:7:9)
    at cli (/Users/pz/Projects/voyager/tasks/build/build.js:16:43)
    at cli (/Users/pz/Projects/voyager/node_modules/@nodeguy/cli/lib/index.js:12:24)
    at Object.<anonymous> (/Users/pz/Projects/voyager/tasks/build/build.js:9:1)
    at Module._compile (module.js:649:30)
    at Object.Module._extensions..js (module.js:660:10)
    at Module.load (module.js:561:32)
    at tryModuleLoad (module.js:501:12)
    at Function.Module._load (module.js:493:3)
    at Function.Module.runMain (module.js:690:10)
(node:33369) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:33369) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
✨  Done in 336.42s.
```

nylira:
Ok, I tried running:

```
yarn run build --platform=darwin --gaia $GOPATH/src/github.com/cosmos/gaia/gaia-mac
```

And I continue to get that error above. Nothing appears in my `./builds/` folder.

nylira:
I've built Voyager for all three platforms using this Docker branch and gaia-0.5.0 binaries. Here are their hashes:

```
Cosmos Voyager-darwin-x64_0.4.3.tar.gz SHA256: 
b7602e20b483d088456bc83dd5261c277d43f49c58fec24ad43a51a33e0008ad

Cosmos Voyager-win32-x64_0.4.3.zip SHA256: 7bfdb84278cdd26fb110bc9dc31489461b1060ed00c637be01dd761c9e7c7f6e

Cosmos Voyager-linux-x64_0.4.3.tar.gz SHA256: 
c995e44dc972a35bd6229583b7943ccae36f1a943b8cedea5a90482cc8798a7c
```

If we can reproduce these hashes -- sweet. I ran the darwin version and it works. I did not have time to test running the Linux and Windows versions.

nylira:
@faboweb I just posted links to Gaia 0.5.0 in Slack.

NodeGuy:
Thank you, @peng.  These are the Gaia binaries that Greg wants us to use and they're different than the ones you uploaded to Slack:

https://tendermint-packages.interblock.io/binaries/gaia_darwin_amd64/gaia_decb23bed0179244e4f42d697dcd4bf759200a2d

https://tendermint-packages.interblock.io/binaries/gaia_linux_amd64/gaia_decb23bed0179244e4f42d697dcd4bf759200a2d

https://tendermint-packages.interblock.io/binaries/gaia_windows_amd64/gaia_decb23bed0179244e4f42d697dcd4bf759200a2d

NodeGuy:
Here are my hashes, using the Greg-approved Gaia binaries:

```
$ yarn run build --commit=4f23fdc03b1a2055bd42e2ce431442d822bdd935 --gaia=../gaia/darwin --platform=darwin

**Comment by nylira** [comment]
Here are my hashes, using the Greg-approved Gaia binaries:

```
Zip successful! /usr/src/app/builds/Cosmos Voyager-darwin-x64_0.4.3.tar.gz SHA256: 5abe9966a6b60cdbead281e3488d5ad87c1a4073c6ca730f16a42427ee9175c8

Zip successful! /usr/src/app/builds/Cosmos Voyager-linux-x64_0.4.3.tar.gz SHA256: fa7e2fe767f6e7f3f0855ba2e227623fbb8e81b5da848ef6cd0b474317988be1

Zip successful! /usr/src/app/builds/Cosmos Voyager-win32-x64_0.4.3.zip SHA256: 53f6402578c918d3d6a80166f80db9b772926ce2c883db4ad21310a151232358
```

The hashes are exactly the same. Awesome job @NodeGuy 🥇

faboweb:
Amazing!!! :)

Peng Zhong <notifications@github.com> schrieb am Mi., 4. Apr. 2018, 22:35:

faboweb:
I get the same hash! :)

faboweb:
It works on Windows.

NodeGuy:
Yay yay yay!

Issue #375, opened 2018-01-22
Secure web-view node-process communication


faboweb:
We should check the security of this communication.

Electron has some option that limits the web-view: https://electronjs.org/docs/api/sandbox-option
Communication to the node-process should probably be via IP: https://github.com/electron/electron/blob/master/docs/api/ipc-main.md

PR #572, opened 2018-03-23
IPC

faboweb:

### Issue

closes: #375

Missing:
- [x] rewrite tests

❤️ Thank you!


app/src/renderer/connectors/node.js
```diff
@@ -2,17 +2,17 @@
 const RestClient = require('./lcdClient.js')
 const mockedRestClient = require('./lcdClientMock.js')
 const RpcWrapper = require('./rpcWrapper.js')
-const mockedRpcWrapper = require('./rpcWrapperMock.js')
+const MockedRpcWrapper = require('./rpcWrapperMock.js')
```
jolesbi: what is our standard for capital vs camel casing?

app/src/renderer/connectors/node.js
```diff
@@ -2,17 +2,17 @@
...
 
-module.exports = function (nodeIP, relayPort, mocked = false) {
-  const RELAY_SERVER = 'http://localhost:' + relayPort
+module.exports = function (nodeIP, lcdPort, mocked = false) {
+  const LCD_SERVER = 'http://localhost:' + lcdPort
```
jolesbi: 👍

app/src/renderer/connectors/node.js
```diff
@@ -24,7 +24,6 @@ module.exports = function (nodeIP, relayPort, mocked = false) {
   }
   // TODO: eventually, get all data from light-client connection instead of RPC
 
-  connector.setMocked(mocked)
-  connector.initRPC(nodeIP)
```
jolesbi: why don't we need to call initRPC anymore?

faboweb: init is called only after the main process signals, that it has connected to a node

app/src/renderer/connectors/node.js
faboweb: I used capital here, as it is a constructor-like function.

app/src/renderer/connectors/node.js
jolesbi: is it because we're using IPC instead?

app/src/renderer/connectors/node.js
faboweb: In general only classes and constants should have capital letters

app/src/renderer/main.js
```diff
@@ -46,23 +45,9 @@ Vue.use(Router)
...
-
-  node.lcdConnected()
-    .then(connected => {
-      if (connected) {
-        axios.get(`http://localhost:${relayPort}/startsuccess`)
-      }
-    })
+  let lcdPort = getQueryParameter('lcd_port')
+  console.log('Expecting lcd-server on port:', lcdPort)
+  const node = Node('localhost', lcdPort)
```
jolesbi: isn't this defaulting to a localhost now instead of the actual node we want to connect to?

faboweb: This is confusing I will change this.

app/src/renderer/node.js
```diff
@@ -0,0 +1,59 @@
...
+        node.rpc.on('error', () => { })
+
+        node.rpc.ws.destroy()
+      }
+
+      console.log('init rpc with', nodeIP)
+      let newRpc = new RpcClient(`ws://${nodeIP}`)
+      node.rpcOpen = true
+      node.rpcConnecting = false
+      // we need to check immediately if he connection fails. later we will not be able to check this error
```
jolesbi: "if the connection"

app/src/renderer/node.js
```diff
@@ -0,0 +1,59 @@
+'use strict'
+const { ipcRenderer } = require('electron')
+const RpcClient = require('tendermint')
+const RestClient = require('./lcdClient.js')
+
+module.exports = function (nodeIP, relayPort, lcdPort) {
+  const RELAY_SERVER = 'http://localhost:' + relayPort
```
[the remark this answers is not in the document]
faboweb: This was dead code. I deleted the file.

PR #573, opened 2018-03-23
Prettier spike

NodeGuy:


NodeGuy:
Hmm, that test isn't failing on my machine.  This is going to take some digging.

faboweb:
@jolesbi what do you think about the eslint files?

NodeGuy:
I consciously chose to keep the `.eslint.rc` file because it will be useful for adding non-style rules such as:

```
    "no-undef": "error",
    "no-unused-vars": "warn"
```

> @NodeGuy usually we name our branches something like david/123-prettier where 123 is the number of the associated ticket / issue.

Got it, thanks.

faboweb:
Do we then need Eslint and Prettier? Or does Prettier read the `.eslint.rc`?

NodeGuy:
We need both.

faboweb:
Do you think it's worth using 2 tools instead of one, that basically do the same thing? Currently the only real change I see, is that Prettier uses `"` instead of `'`. Is prettier actually linting the stylus parts in Vue? That was the main advantage right? I probably miss the main plus for Prettier, would you mind selling it to me again? :)

nylira:
I agree with Fabo that adding Prettier seems to be adding extra complexity to the project without discernible benefit. If we have to use eslint anyway, let’s use eslint only.

NodeGuy:
I misspoke.  We don't need both.  ESLint can serve two functions:

* enforcing style rules (and automatically fixing a few of them)
* warning about problems such as references of undefined variables

Prettier does a much better job of handling style because instead of warning about style problems it simply fixes all of them automatically.  _It saves a huge amount of time._

I was suggesting that we keep ESLint only for the second function of warning about problems.  Although it's also a large benefit, we don't need it, and if I had to choose only one of the two I would choose Prettier.

jolesbi:
this is interesting. i'm interested in trying out Prettier - sounds neat. if we can't agree on needing this or it being unwanted complexity, let's try it in our code editors for a bit and see what it's like?

Issue #561, opened 2018-03-19
Fix the MaxListenersExceededWarning spam


nylira:
### Version
`develop`

### Description
We're getting spammed with the following line in CircleCI builds. It comprises nearly 50% of the build log which makes it harder to find out what is wrong.

```
Remove (node:215) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 unhandledRejection listeners added. Use emitter.setMaxListeners() to increase limit
```

Let's increase the max listeners.

### Screenshot

![circleci](https://user-images.githubusercontent.com/172531/37609294-08c8fd38-2b9d-11e8-9400-e75c3c274b60.png)

faboweb:
You are so right! We need to kill those damn messages :)

mappum:
I would say that we should actually diagnose and ensure there isn't a memory leak, but it looks like it never goes above 11 so it's probably OK to just increase the limit 👍

PR #584, opened 2018-03-28
Raise event listener limit on process to prevent warnings

mappum:

I spent a while digging into where all these listeners were coming from, but found in the end that it was cleaner to just raise the limit. It should never be an issue since the listeners only accumulate in the tests (normal app will only register one listener), and the number of listeners is small enough that the memory usage is negligible.

### Issue

closes: #561 

### Screenshots

❤️ Thank you!


mappum:
> Just a thought, but would this also work, if we apply this only for the test process? So as a preparation in jest?

Good idea, I'll change this.

mappum:
Now the limit is only changed in tests, so we won't hide any listener leaks that happen in the actual app.

Issue #553, opened 2018-03-13
block explorer during fast sync is unusable


jolesbi:
while a node is in fast-sync, the block explorer can become unusable because of how quickly data is being updated. 

we should consider adding a "currently catching up to the future" image or something fun.

PR #587, opened 2018-03-31
Handle Node Sync in Block Explorer

mappum:

This changes the block subscription to first check if the node is still syncing, and if so it waits another 30 seconds before trying again. If the node is syncing, a "Node is syncing blockchain..." dialog is shown in the explorer.

### Issue

closes: #553

### Screenshots

❤️ Thank you!


app/src/renderer/components/common/NiDataLoading.vue
```diff
@@ -1,13 +1,19 @@
...
   div(slot="subtitle") Please wait a moment.
 </template>
 
 <script>
 import DataMsg from 'common/NiDataMsg'
 export default {
   name: 'ni-data-loading',
-  components: { DataMsg }
+  components: { DataMsg },
+  props: {
```
faboweb: good idea

app/src/renderer/vuex/modules/blockchain.js
```diff
@@ -69,21 +70,39 @@ export default ({ commit, node }) => {
...
+
+      node.rpc.status((err, status) => {
+        if (err) return error(err)
+
+        if (status.syncing) {
+          // still syncing, let's try subscribing again in 30 seconds
+          state.syncing = true
           state.subscription = false
-          commit('notifyError', { title: `Error subscribing to new blocks`, body: err.message })
+          setTimeout(() => dispatch('subscribeToBlocks'), 30e3)
```
jolesbi: 30e3? never seen that before. can't find anything on google about it.

mappum: Scientific notation for 30000, maybe I should do `30 * 1000` so it's more obvious.

PR #590, opened 2018-04-02
Peng/onboarding mockup

### LINKED ISSUE #359: OS default scrollbars don't fit the app's theme
We should use `perfect-scrollbar` for `#app-content` for a better looking scrollbar.

screen shot 2018-01-12 at 2 43 36 pm

34558901-1acc2b28-f140-11e7-99a4-d72ff1873848

### LINKED ISSUE #539: Design the onboarding wizard

nylira:
### Issue
Closes #539 

- [x] created mockups for the onboarding process.
- [x] move design files out of git repo

### Mockup Screenshots
screen shot 2018-04-02 at 10 44 58 pm

You can view the PDF too for a larger version.


faboweb:
This looks like a good first version for the wizard. Nice @nylira! How about full-buttons in the footer in comparison to the flat-buttons you put there? With the low contrast they don't feel very actionable to me.

jolesbi:
looks great peng. 

agreed on the buttons.

how does this address sign up / sign in / seeds / hardware? would this lil wizard come before or after authentication?

part of what @NodeGuy and I were working on was a graduated system where you'd go from local node, to testnet, to mainnet through participation and gamification. is this out of scope for our onboarding v1? (it very well could be). our goal was to convey real value (which i think this has) and also encourage the user to actually go for it (send tokens for eg...).

ps. i don't think we should include sketch files or PDF's in the repo. let's use design software (Zeplin or others). Sketch isn't open source so some won't be able to open it - it doesn't really have any value for our users, and they don't need to download this extra stuff when trying to run / fork voyager.

faboweb:
I think we aligned on a basic first version for the onboarding. A sophisticated version is desirable but will cost a lot of time and should be split up in smaller units.

For files:
I like that we have everything on GitHub. Maybe we can use the Wiki for that?

nylira:
I was assuming the onboarding would appear after the user signs in. I think the sign in process is standard enough for users to understand without requiring onboarding. We defined our customer as a fundraiser contributor or developer, and they should already know about usernames, passwords, and seed phrases from other cryptocurrency wallets.

nylira:
I also prefer having the mockup files stored in GitHub. Sketch is not open source, but neither is Zeplin. Users will not be cloning this repo because they have binaries, only developers will be here. The PDF version should definitely be in here for third party developers to see the future direction of Voyager. The Sketch file is more arguable. But I think we are the maintainers, so I think this repo should serve our needs first--which is to version control and manage everything related to Voyager.

I considered storing them in Google Docs but that's not public. If we use the wiki, it's one more source we'd have to maintain (I also don't have experience with the version control aspect of the wiki).

jolesbi:
i'm onboard with the onboarding related stuff. 👍 

> Sketch is not open source, but neither is Zeplin. 

i was trying to say that we should use something like zeplin so we can collaborate more effectively on visual stuff. github comments won't work. the note about sketch not being open source was a separate point.

> The PDF version should definitely be in here for third party developers to see the future direction of Voyager. 

i am all for sharing the future of voyager - but i'm not convinced that PDFs in this repo is the best way to do this.

---

i am against sketch and design files in this repo for a number of reasons.

1. design files can become quite large and i don't want to bloat repo's (for eg. my sketch file for the cosmos-website is 270mb)
2. designers who join our team in the future shouldn't have to know how to use github
3. there are tools designed specifically for sharing design files, managing version history, and that promote collaboration in a way that github really does not.
4. i really don't like the idea of sharing proprietary files (sketch) in an open source repo.
5. if we want to "open source" our design process, we need to establish what this really looks like - but i think it's a lot more than sharing sketch files and PDFs in this repo. i'd love to have a conversation about this outside of this ticket.

nylira:
@jolesbi I agree with your points - let’s discuss after the presentation sprint  a place to store and collaborate design files out of this repo.

nylira:
I'm getting an odd error when I run `yarn lint`

```
pz@mbp ~/P/voyager> yarn lint
yarn run v1.5.1
$ yarn lint:eslint && yarn lint:format
$ eslint -f ./node_modules/eslint-friendly-formatter {,**/}*.{js,vue}
$ prettier --list-different {,**/}*.{css,js,json,md,vue}
docs/onboarding-outline.md
error An unexpected error occurred: "Command failed.
Exit code: 1
Command: sh
Arguments: -c prettier --list-different {,**/}*.{css,js,json,md,vue}
Directory: /Users/pz/Projects/voyager
Output:
".
info If you think this is a bug, please open a bug report with the information provided in "/Users/pz/Projects/voyager/yarn-error.log".
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

faboweb:
Should we close this or keep it open until we have the onboarding wizard?

nylira:
I think we can merge this and use this ticket https://github.com/cosmos/voyager/issues/527 to implement it in code.

Issue #304, opened 2018-01-02
Pass COSMOS_NETWORK env var through in production build


mappum:
When building, we should be able to specify the `COSMOS_NETWORK` env var and have the built app bundle use this network. Instead, we currently have to change the default in `app/src/main/index.js`.

dependencies: #534

NodeGuy:
If we're going to start the app with a local network by default then isn't this feature superseded by #27?

faboweb:
This is a good point. The purpose of this issue was, to create a quick way to build Voyager for different networks without changing config files. I.e. for our demo in April or for developers that test their own blockchain. Do you think, that we will be able to integrate #27 and the whole process of switching from local to online before then?

mappum:
Closed via #604

PR #604, opened 2018-04-04
Allow bundling of network in production build. #304

NodeGuy:
…cally.

### Issue

closes: ISSUE

### Screenshots

❤️ Thank you!


config.js
```diff
@@ -1,44 +1,9 @@
...
-  google_analytics_uid: 'UA-51029217-3',
-  sentry_dsn: 'https://4dee9f70a7d94cc0959a265c45902d84:cbf160384aab4cdeafbe9a08dee3b961@sentry.io/288169',
-  sentry_dsn_public: 'https://4dee9f70a7d94cc0959a265c45902d84@sentry.io/288169'
-}
-
-config.building.name = config.name
-
-module.exports = config
+module.exports = toml.parse(
+  fs.readFileSync(path.join(__dirname, `config.toml`), { encoding: `utf8` })
```
faboweb: Good idea!

NodeGuy:
The work is complete but the local network functionality is currently broken so I don't know how to test it properly.

tasks/build/index.js
```diff
@@ -1,35 +1,52 @@
...
+cli(specification, options => {
+  const { commit, gaia, network } = options
+
   shell.exec(`docker build --tag cosmos/voyager-builder .`, {
     cwd: __dirname
   })
 
-  const builds = path.resolve(__dirname, "../../builds")
-  shell.mkdir(`-p`, builds)
+  const resolved = R.map(R.pipe(untildify, path.resolve), {
```
faboweb: Do we need a plugin just for these lines?

NodeGuy: [Ramda](http://ramdajs.com/) is the best and most popular functional programming library for JavaScript.  It will find its way into every one of our projects and should probably just be included automatically by default.

faboweb: I usually try to avoid libraries because they are almost always an overhead to understand the code. 
Either way I think the line could benefit from some comments. I.e. what does untildify do?

NodeGuy: I also see the use of libraries as a cost but the value of Ramda is well worth it.  Many of its methods (such as `map` and `pipe` above) are the same in all functional programming languages and so are universally understandable.

I added more comments to the whole file.

faboweb: Note: We also talked about already having Lodash as a library in place. I guess there it would be `_.mapValues({ ... }, _.compose(untildify, path.resolve)`. Would this also be feasible?

faboweb: More important on the other hand side is merging this. I just want to reduce the time spent on cleaning up code when we decide on one functional library.

NodeGuy: Ramda is superior to Lodash.  See [The Philosophy of Ramda](http://fr.umio.us/the-philosophy-of-ramda/#whynotjustuse) (or for an in-depth explanation [Hey Underscore, You're Doing It Wrong!](https://www.youtube.com/watch?v=m3svKOdZijA)).  We should switch from Lodash to Ramda.

nylira: I don't want to waste too time debating which lib is better--let's be practical about this. Lodash is imported a grand total of six times into Voyager, so it's easy to replace with Ramda if we want to use the latter instead. I've never used Ramda before but the docs make it look simple enough. I'm fine with switching.

Next time we want to replace a library in Voyager, we need to make a separate issue for the the lib switch instead of trying to do two things in one PR. This slows down reviews and merges. We can put this in the coding guide.

NodeGuy: I understand and agree with your concern but I'm not trying to replace a library in this PR.  I wasn't even aware that we were using Lodash before I wrote this.

Do you want me to rewrite the PR using Lodash instead of Ramda?  To me that feels like the slower path.

NodeGuy: I rewrote the PR using Lodash.

faboweb:
I can't build on Windows. I use Docker Quickstart Terminal and do `yarn build --platform=win32 --commit={commit} --gaia="C:\Users\Fabo\Downloads\gaia.exe"`.

I get `invalid mount config for type bind "bind": invalid mount path: 'C:\Users\Fabo\Downloads\gaia.exe' mount path must be absolute.`

faboweb:
Build works on Windows

mappum:
@NodeGuy thanks for doing this, it looks like it was difficult since the change affected so many pieces across the build system.

I'm trying to build from this branch, but it's failing when I specify the new `--network` flag (it builds OK if I don't specify `--network`, although then the `defaultNetwork` is set incorrectly):

```
❯ COSMOS_NETWORK=~/.cosmos-voyager-dev/basecoind-demo1 yarn build --gaia=`which basecli` --platform=darwin --network=app/networks/basecoind-demo1

yarn run v1.5.1
$ node tasks/build --gaia=/Users/mappum/go/bin/basecli --platform=darwin --network=app/networks/basecoind-demo1
Sending build context to Docker daemon  15.36kB
Step 1/7 : FROM node:9.4.0
 ---> c888d933885c
Step 2/7 : RUN dpkg --add-architecture i386   && apt-get update   && apt-get --yes install     wine     wine32     libwine
 ---> Using cache
 ---> c4dcdbbb4149
Step 3/7 : RUN touch /usr/local/bin/gaia
 ---> Using cache
 ---> f365029e00d6
Step 4/7 : RUN mkdir -p /usr/src/app
 ---> Using cache
 ---> 9736223a8323
Step 5/7 : WORKDIR /usr/src/app
 ---> Using cache
 ---> dbb3fa584eba
Step 6/7 : COPY entrypoint.sh /
 ---> Using cache
 ---> 4a74ab477594
Step 7/7 : ENTRYPOINT [ "/entrypoint.sh" ]
 ---> Using cache
 ---> 9d0d8d923651
Successfully built 9d0d8d923651
Successfully tagged cosmos/voyager-builder:latest
docker: Error response from daemon: invalid mount config for type "bind": bind source path does not exist.
See 'docker run --help'.
✨  Done in 1.75s.
```

How can I debug this further?

Issue #596, opened 2018-04-03
Check config file (ports / chain)


faboweb:

faboweb:
Done I guess

PR #606, opened 2018-04-05
WIP: Check config file (ports / chain) #596

NodeGuy:
Depends on #604 and is untested.

### Issue

closes: #596


tasks/build/index.js
```diff
@@ -1,35 +1,52 @@
 "use strict"
 
 const { cli } = require(`@nodeguy/cli`)
-const options = require(`./options.json`)
+const specification = require(`./options.json`)
```
jolesbi: what is your thinking here @NodeGuy ?

NodeGuy: I just clarified it.

NodeGuy: I mean I pushed a new commit that renames `specification` to `optionsSpecification`.  Does that help?

config.js
```diff
@@ -2,10 +2,20 @@
...
-  })
-)
+const readToml = directory =>
+  toml.parse(
+    fs.readFileSync(path.join(directory, `config.toml`), {
+      encoding: `utf8`
+    })
+  )
+
+// Try to read the configuration file in the root directory first.  If that
```
faboweb: A thought: Would it make sense to put the config file in the app folder? This way we could also define in there, what the config folders name is. This could reduce problems running 2 Voyagers in parallel.

NodeGuy: You mean you want to configure `COSMOS_HOME` in the `config.toml` file?

faboweb: Yes, that was the idea.

NodeGuy: OK.  It sounds like you're asking for two configuration files: one in the `app` directory that will be read first which points to the second.  Let's split up the keys; which would you like in each file?

```toml
# Name of electron app
# Will be used in production builds
name = "Cosmos Voyager"

# webpack-dev-server port
wds_port = 9080
lcd_port = 9070
lcd_port_prod = 9071
relay_port = 9060
relay_port_prod = 9061

default_network = "gaia-2"
google_analytics_uid = "UA-51029217-3"
sentry_dsn = "https://4dee9f70a7d94cc0959a265c45902d84:cbf160384aab4cdeafbe9a08dee3b961@sentry.io/288169"
sentry_dsn_public = "https://4dee9f70a7d94cc0959a265c45902d84@sentry.io/288169"
```

NodeGuy: Based on your comment in the meeting today I'm pushing back on this request and suggest you create a new issue instead.  :-)

faboweb: Let's do it your way first so we can get this merged. I want to add an explanation why I still think a local config file would have made more sense and I therefor worked on #620 
The purpose for this issue was to have 2 Voyager instances running in parallel. With the config file sitting in the global config folder, both instances would use the same config file ergo would not have different ports and so not be able to run in parallel.

NodeGuy: I think we've had a giant miscommunication about the requirements for this feature.

Do you want a user-modifiable `config.toml` that lives in `~/.cosmos-voyager` or do you want a developer-modifiable `config.toml` that is bundled into the Electron app?

NodeGuy:
Closing this PR because the requested functionality is provided by https://github.com/cosmos/voyager/pull/604.

Issue #608, opened 2018-04-05
Docker doesn't build on Windows


faboweb:
> Thank you for filing an issue!

### Version

--

### Description

`yarn run build ...` outputs errors on Windows.

PR #609, opened 2018-04-05
Docker build on Windows

faboweb:

### Issue

closes: #608 

❤️ Thank you!


NodeGuy:
Unless I'm not understanding something correctly, `path.resolve` should do the right thing on Windows automatically.  The bug here may be that not all of the paths are being sent through `path.resolve` (such as `builds`).  I tightened this up in another PR: https://github.com/cosmos/voyager/blob/436119a146005eb50de57701625d840a1afd0aac/tasks/build/index.js#L17-L22

faboweb:
It resolves to a Windows path. But Docker takes Unix paths.

PR #612, opened 2018-04-05
added polling for balances

### LINKED ISSUE #595: Polling balances

jolesbi:
closes #595


test/unit/specs/components/wallet/__snapshots__/PageWallet.spec.js.snap
```diff
@@ -192,103 +192,39 @@ exports[`PageWallet has the expected html structure 1`] = `
         <main
```
faboweb: I guess the test should also catch the loaded state

nylira: We want this for the demo, so let's make an issue for it and merge this one :)

PR #619, opened 2018-04-06
WIP: merge develop for demo

faboweb:
Bringing in the IBC design by Peng

- [ ] reduce technical dept -> tests


faboweb:
@mappum do we want this branch still, or do we continue from yours?

PR #626, opened 2018-04-10
peng/528 safer looking unbonding

### LINKED ISSUE #528: Make staking appear less dangerous

### Version

0.4.3

### Description
The colors for staking are off-putting to users. Can we make them less aggressive. The one in question is the bar of unbonded atoms (scary red).

nylira:
- [x] use --input-bc instead of --warning as unbonding color

### Issue
Closes #528 

### Screenshots
I can't view the bonding page right now because we don't have mock server in yet. But the code is only changing two variables, so I think it's clear enough?

### DIFF
--- a/app/src/renderer/components/staking/PageBond.vue
+++ b/app/src/renderer/components/staking/PageBond.vue
@@ -424,9 +424,9 @@ export default {
 
   &.bond-group--unbonding
     .bond-bar__inner
-      background var(--warning)
+      background var(--input-bc)
     .bond-delta
-      color var(--warning)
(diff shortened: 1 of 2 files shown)


faboweb:
If you add a second parameter `true` to the node initialization here: https://github.com/cosmos/voyager/blob/develop/app/src/renderer/main.js#L42
The node connector is mocked.

nylira:
@faboweb Ok that's awesome! Well done :)

### new unbonding color
screen shot 2018-04-10 at 1 16 10 pm

PR #633, opened 2018-04-12
peng/518 hide assets not holding

### LINKED ISSUE #518: Hide assets not holding

### Version

0.4.3

### Description

We want to hide tokens, that have a balance of 0 so the user can focus on things that are important.

nylira:
# Issue
Closes #518

### Screenshots
screen shot 2018-04-12 at 12 07 40 pm


faboweb:
In the mocks by Gautier there was a button to show all assets. Do we want that?

faboweb:
Your are right. Let's keep it like this then.

jolesbi:
i just want to flag that i believe we agreed to hide _and_ optionally reveal all the denominations on the network.

i'm bringing it up so we can improve our processes - not to revert any changes or propose i'm unsatisfied with this decision.

jolesbi:
well, i would have preferred to still show all the denoms on the network. and i worked on a ticket to make this happen. so this is ok - there are good reasons on both sides. i'm just saying we already decided on this a while back, so i'd like us to note that we have an opportunity to improve the way we log and discuss our product design decisions.

nylira:
@jolesbi I missed the original conversation around the ticket #518. I followed the description on the ticket ("We want to hide tokens, that have a balance of 0 so the user can focus on things that are important."). I didn't know that toggling the tokens was part of the spec, so I didn't implement it to save me some work.

That said, I still believe there's no point in showing tokens to the users they cannot access. On Ethereum wallets and blockchain explorers, they do not show that you have 0 balances of every ERC20 token. On websites that do show 0 balances of supported tokens (like the Poloniex crypto exchange), the page ends up being very long and overloaded with information.

polo-step2a-big

faboweb:
I added a ticket so we don't forget to talk about this later on.

PR #635, opened 2018-04-12
peng/517 show all tokens on wallet

### LINKED ISSUE #517: Show note about only showing liquid assets

### Version

0.4.3

### Description

Voyager will only show liquid assets (not staked). We want to show a note so the user is aware of this.

Alternatively we can also show staked assets!

nylira:
- [x] show all tokens associated with address on wallet page
- [x] use `config.bondingDenom` from vuex instead of hardcoding Atom on `PageBond.vue` and `PageWallet.vue`
- [x] update tests and snapshots
- [x] renamed Denomination Balances to Available Balances, and then added Staked Balances
- [x] e2e test fix
- [x] improve test coverage

### Issue
Closes #517 

### Screenshots
screen shot 2018-04-12 at 1 53 09 pm
screen shot 2018-04-12 at 1 55 36 pm
screen shot 2018-04-12 at 1 52 40 pm


nylira:
I could use help on the e2e test issue. Sorry for not notifying earlier, been busy.

nylira:
Looks like--at some point--merging from develop has fixed the e2e test. Will improve test coverage asap.

app/src/renderer/components/wallet/PageWallet.vue
```diff
@@ -83,6 +90,17 @@ export default {
...
       }
+    },
+    stakedTokens() {
+      let tokens = 0
+      let values = Object.values(this.committedDelegations)
+      if (values.length > 0) {
+        tokens = values.reduce((sum, d) => sum + d)
+      }
+      return tokens
+    },
```
NodeGuy: Let's tighten that up:

```JavaScript
import _, { includes, orderBy } from "lodash"

stakedTokens() {
  return _.sum(Object.values(this.committedDelegations))
}
```

test/unit/specs/components/govern/LiProposal.spec.js
```diff
@@ -0,0 +1,58 @@
+import { shallow } from "@vue/test-utils"
+import LiProposal from "renderer/components/govern/LiProposal"
+
+describe("LiProposal", () => {
+  let wrapper
+  let propsData = {
```
NodeGuy: It's better to use `const` by default and `let` only when necessary because `const` preserves [referential transparency](https://en.wikipedia.org/wiki/Referential_transparency).

app/src/renderer/components/wallet/PageWallet.vue
nylira: Cool, I didn't realize there was a sum function :)

test/unit/specs/components/govern/LiProposal.spec.js
nylira: Makes sense, `propsData` is entirely static.

nylira:
This is ready to be approved.

PR #662, opened 2018-05-07
solve the install problem 'engine 'node' is incompatible with this module'

MrXJC:

### Issue

closes: ISSUE

### Screenshots

❤️ Thank you!


NodeGuy:
Thank you, that fix doesn't solve the problem but merely suppresses the warning.  The solution is to use the correct version of Node.js.

nylira:
Please use Node 9.4 and above. We're not responsible if you use an unsupported version of Node.

PR #650, opened 2018-04-30
Changes to allow mocked builds

mappum:
> Thank you a lot for contributing to Cosmos Voyager!

Here are the changes I needed to be able to make our mocked release build. It adds a `mocked` option in config.toml which disables the LCD and enables the mock renderer connector.

### Issue

closes: ISSUE

### Screenshots

❤️ Thank you!

PR #670, opened 2018-05-08
Revert "Changes to allow mocked builds"

nylira:
Reverts cosmos/voyager#650


nylira:
I merged this too early, there are still some conflicts regarding #650 and running a local testnet.

PR #673, opened 2018-05-08
Changes to allow mocked builds

nylira:
I merged #650 too early, there are still some conflicts with these changes and running a local testnet. So right now, I'm able to run a local testnet with `gaia-2`, and run Voyager, using the following commands:

```
COSMOS_NODE=localhost yarn testnet
```
```
gaia node start --home/$HOME/.gaia2
```

But within this PR, I am no longer able to do so. The app says that `config.toml` cannot be found. @mappum can you help me figure it out?


nylira:
There's an issue with this PR where it cause this problem:

screen shot 2018-05-09 at 3 01 58 pm

Any ideas?

mappum:
I diagnosed the issue, looks like somewhere down the line our dev builds switched to packing the files into an electron `asar` archive, which prevents us from loading files from the bundle at runtime. So in `app/src/config.js`, it tries to load `../config.toml` and it doesn't exist since all the files are in that archive.

I'm looking but I can't find where this changed or how to change it back.

NodeGuy:
I'll start working on the remove Webpack spike which has the potential to obviate this whole mess.

faboweb:
Fixed. Note: The renderer should not look for files. It should get all information through the main thread. The main thread is better aware of the file structure. Information should get passed to the view thread via `electron.remote.setGlobal` as seen here: https://github.com/cosmos/voyager/blob/develop/app/src/main/index.js#L16

nylira:
Thank you @faboweb it looks great. Feel free to review + merge this.

faboweb:
Blocked on E2E tests not being able to send coins.

faboweb:
Not able to send coins :/

Issue #665, opened 2018-05-08
Multiple back buttons


okwme:
### Multiple back buttons

Click 'Staking'
Click 'mr_mounty'
Two back buttons present

OR

Click 'Proposals'
Click first proposal
Two back buttons present

nylira:
Closed by #674

nylira:
@okwme next time write "closes #665" to automatically close an issue in your PR.

okwme:
@nylira 👍

PR #674, opened 2018-05-08
issue #665, removed back buttons

okwme:
`yarn run lint` and `yarn run test` both fail with exit code 1

### Issue 665

closes: ISSUE #665 

### Screenshots

not sure why the package.json formatting got changed. thought this might get reset when lint is run, but as noted above lint and test did not work. is that error a common sight / easy fix?

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,8 +2,7 @@
   "name": "cosmos-voyager",
   "productName": "Cosmos Voyager",
   "version": "0.4.3",
-  "description":
-    "Voyager is an electron-based user interface for the Cosmos Network.",
+  "description": "Voyager is an electron-based user interface for the Cosmos Network.",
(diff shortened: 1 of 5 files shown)


nylira:
Yeah, the team uses [Prettier](https://github.com/prettier/prettier) for formatting our code. You should have it run in your editor after every save and it should solve it.

NodeGuy:
BTW, `package.json` gets updated by Yarn when adding dependencies so Prettier is configured to ignore it.

package.json
```diff
@@ -15,6 +15,8 @@
     "node": ">=9.4.0"
   },
   "scripts": {
+
```
jolesbi: can you remove these two new lines?

okwme: 👍

jolesbi:
welcome to the team @okwme 🎉

PR #675, opened 2018-05-09
WIP: peng/gaia-5000 support

nylira:
**This PR contains #591. We can merge #591 first.**

I have been running a full node on the last several public testnets: `gaia-3008`, `gaia-4000`, and now `gaia-5000`. There have been updates to the testnet joining process that we have to adapt for in Voyager.

First of all, we can no longer download the genesis file from [tendermint/testnets](https://github.com/tendermint/testnets). Testnet data has been moved to [cosmos/testnets](https://github.com/cosmos/testnets). You'd think it'd be easy to just change our download paths, but no--the genesis.json file now has to be generated locally. I've done this already to run my own `gaia-5000` node, so I copied that genesis file to `./app/networks/gaia-5000`.

Note that you also need to install this version of Cosmos SDK locally:

```
$ gaiacli version
0.16.0-dev-7ef5e90f
```

`gaiacli` is also now being used instead of `basecli`, so I had to rename those instances in Voyager.

### gaiacli up and running in voyager's electron process
screen shot 2018-05-09 at 2 55 41 pm


nylira:
Getting further! Woo.

screen shot 2018-05-09 at 3 30 23 pm

New error:

screen shot 2018-05-09 at 3 30 30 pm

mappum:
Interesting, wasn't expecting the SDK team to change the testnet config file formats again.

nylira:
Deprecated by #591

Issue #624, opened 2018-04-09
Spike: Remove unnecessary Webpack code.


NodeGuy:
`/webpack.main.config.js` looks like a good candidate for removal.

PR #680, opened 2018-05-09
David/624 webpack spike

NodeGuy:

### Issue

closes: #624 

### Screenshots

❤️ Thank you!


faboweb:
The spike is again running for several days. ;)

NodeGuy:
No, I started it while waiting to be unblocked by Matt's merge and then quickly stopped working on it to switch to building the SDK in Docker.  I'm not sinking much time into it.

faboweb:
Ok, never mind then. :)

NodeGuy:
The spike demonstrates that Webpack is unnecessary for some of the application.  `yarn testnet` runs without using `webpack.main.config.js`.

I did not attempt to make `yarn test` work.

faboweb:
So yarn test doesn't work without it? Does building the application work
without it?

David Braun <notifications@github.com> schrieb am Fr., 18. Mai 2018, 00:03:

jolesbi:
what was the original intention here? we could remove the webpack file and the code would be simpler?

NodeGuy:
> So yarn test doesn't work without it? 

`yarn test` doesn't work without further changes but I thought you wanted me to keep the spike small.  Do you want me to make `yarn test` work?

> Does building the application work without it?

I haven't tried; see above.

> what was the original intention here? we could remove the webpack file and the code would be simpler?

That we can stop bundling a bunch of files into `main.js` which will:

* make the code simpler
* shorten the code-build-test cycle
* make it easier to debug errors because the source code files won't be squished together

faboweb:
Let's discuss this in the Tuesday meeting

NodeGuy:
OK, I added it to the agenda.

Issue #690, opened 2018-05-14
The receive button doesn't open the receive modal


faboweb:
> Thank you for filing an issue!

### Version

Commit a30f6494d19a621092fceb7e9ad5e8e2f90dfe88

### Description

The receive button doesn't open the receive modal

PR #691, opened 2018-05-14
fixed receive modal not showing

faboweb:

### Issue

closes: #690 

❤️ Thank you!


nylira:
I think your linter altered the Stylus when it shouldn’t have. I prefer the no commas no brackets style, but we can discuss that later.

faboweb:
fixed

Issue #615, opened 2018-04-05
PageWallet.spec.js should test for loaded state


nylira:
### Version
`develop`

More information in https://github.com/cosmos/voyager/pull/612

### Description

The wallet unit test doesn't test for loaded state.

faboweb:
#698 is missing a snapshot of PageWallet in a loaded state @nylira @mappum

mappum:
Closed in #698

PR #698, opened 2018-05-15
Wallet load test

mappum:

Just a simple test case, with a bonus test for IBC transactions going through the `send` module.

### Issue

closes: #615 

### Screenshots

❤️ Thank you!


nylira:
Thanks @mappum!

test/unit/specs/store/__snapshots__/send.spec.js.snap
```diff
@@ -1,5 +1,26 @@
...
+  Array [
+    "zone/address",
+    Object {
+      "amount": Array [
+        Object {
+          "amount": 123,
+          "denom": "mycoin",
+        },
+      ],
+      "chain_id": "",
```
faboweb: is it correct that `chain_id` and `src_chain_id` are empty?

PR #702, opened 2018-05-16
added prettier precommit hook

faboweb:
Source: https://prettier.io/

Will always apply prettier to every commit. No more `linted` commits!


NodeGuy:
Good idea.  Do you use an [editor plugin](https://prettier.io/docs/en/editors.html)?  I never need to run it manually because my editor does it automatically on saves.

faboweb:
I do, but sometimes in some configurations it doesn't format it the same way. :/ Which causes me to get back and forth in some PRs. This should solve this.
But I am not feeling too strongly about it.

nylira:
I think we should do this. The easier it is for developers to contribute (without forcing them to install Prettier) the better.

faboweb:
Feel free to approve and merge

nylira:
Changelog needs an update!

PR #713, opened 2018-05-17
WIP: fix back buttons

faboweb:

### Issue

The back buttons only moved back in history which could be leading the user to an undesired page i.e. for blocks where the user would want to go back to the blocks overview.

NiToolbar takes now a `backPath` which is required to enable the back button and is used to determine the path where to go back to.

❤️ Thank you!


jolesbi:
this isn't really a back button anymore...

what about routes with an unspecified `backPath`?

what was the issue with going back in history?

faboweb:
Not in the context of a browser back button but in the terms of a folder
"go one level up" button. Our navigation is designed to only have two
levels. We once agreed that going back would go to the top level. Changing
the top level can be done via the menu.
The problem occured in the block explorer in the Szenario: click on block
-> click next -> click back button. Voyager now showed the previous visited
block instead of the block overview which is inconsistent with the behavior
on other pages.

Pages with no specified backPath do not show the back button. I guessed
this is desired for top level pages.

Is this reasonable?

Jordan Bibla <notifications@github.com> schrieb am Do., 17. Mai 2018, 23:40:

jolesbi:
hmmm no i don’t think this button is useful if it only takes you back to “one folder” up. this is basically always possible from the sidebar. 

the back button is specifically there to let you go back like you would expect in the browser. i would find another interaction confusing.

the bug billy reported wasn’t unexpected behaviour but rather some kind of bug with the block explorer (i think).

i would much prefer we leave the back button as history -1 and address the issue with the block explorer.

nylira:
This is a continuation of a long discussion we've been having about the purpose of the back button in our app.

I agree with Jordan that the back button should behave exactly like a browser back button. We should keep a standard back button because Voyager is a browser for Cosmos. If we later switch to a web app only, we can ditch in-app the back button.

However, I also agree with Fabo that we need some sort of navigation element to go up one level in the app hierarchy. For example, if the user clicks on a transaction within a block, or the bonded atom history on a validator, it's not clear where they are in the app anymore. I proposed an app breadcrumb five months ago: https://github.com/cosmos/voyager/issues/280 It may be time to implement it.

In conclusion, I think the back button should be history(-1) again, but we still do need to disable it if there's no history to go back to. Just like the browser back button.

faboweb:
Do we need a history go back button? In which case would I want that? Could you guys test my case manually to feel the pain please. ;D

jolesbi:
the button that makes the most sense to me is one that takes you back from where you came. a "dumb" back button so to speak. i believe this is the expected behaviour. i don't think the button should know what the parent of the route is (`backPath`). 

there may be situations where you don't want to go to the parent, but would rather just go back to the previous route. also, the parent routes are always available in the sidebar. a historical back button (which we already have implemented) is nice for when the user ends up more than 2 levels deep.

what do you mean "test my case manually"?

nylira:
Okay, I've just tested this PR, and I don't think Voyager needs this change.

If the user visits a block a page, presses "Newer Block" and then presses "Back", we want the user to go from block #214,800 to #214,799. If the user wants to go back to the Blocks index page, they can click on the link in the sidebar.

I vote for closing without merging.

faboweb:
Did you also test in mobile mode? There this would mean opening the menu and then clicking on the menu item again. Also for going back one block there is an extra button only for that purpose.

faboweb:
Ok, I give up.

PR #715, opened 2018-05-18
Fix for reconnect causing a full page error

faboweb:

### Issue

reconnect causes a full page error

❤️ Thank you!


okwme:
full page error does not happen on disconnect / reconnect.
however after first disconnect and reconnect, subsequent disconnect / reconnects fail

PR #721, opened 2018-05-19
Send fix

### LINKED ISSUE #694: Voyager doesn't recover from an error while sending coins

### Version

Commit 220397c93f1555e89971181e4c1a2cc929954de3

### Description

When a sending transaction fails, all following transactions do not produce REST requests.

@mappum maybe?

mappum:

Fixed the issue where the send lock wouldn't get unset when an error occurred, causing future sends to fail. Also added a test case for this to prevent regressions.

### Issue

closes #694 

### Screenshots

❤️ Thank you!

### DIFF
--- a/app/src/renderer/vuex/modules/send.js
+++ b/app/src/renderer/vuex/modules/send.js
@@ -45,6 +45,10 @@ export default ({ commit, node }) => {
   }
 
   let actions = {
+    // `lock` is a Promise which is set if we are in the process
+    // of sending a transaction, so that we can ensure only one send
+    // happens at once. otherwise, we might try to send 2 transactions
+    // using the same sequence number, which means 1 of them won't be valid.
     async sendTx(...args) {
       // wait to acquire lock
       while (lock != null) {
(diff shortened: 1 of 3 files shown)


app/src/renderer/vuex/modules/send.js
```diff
@@ -45,21 +45,27 @@ export default ({ commit, node }) => {
...
-      let res = await lock
-      lock = null
-
-      return res
+      try {
+        // send and unlock when done
+        lock = doSend(...args)
+        // wait for doSend to finish
+        let res = await lock
+        return res
```
NodeGuy: This can be `return await lock`.

test/unit/specs/store/send.spec.js
```diff
@@ -58,5 +58,28 @@ describe("Module: Send", () => {
       await store.dispatch("sendTx", args)
       expect(node.ibcSend.mock.calls).toMatchSnapshot()
     })
+
+    it("should send a transaction after failing", async () => {
```
faboweb: Good job adding the test for this problem. :) 👍

app/src/renderer/vuex/modules/send.js
faboweb: I think the solution is readable enough. I will approve. If @NodeGuy still insists we should improve the code here, I am happy to create another PR.

NodeGuy: OK but in the future if the intention is to clarify then `res` is insufficient.  An English word is always preferable.

Issue #720, opened 2018-05-18
fresh 0.6.0 bugs 🐛🐜🐞


jolesbi:
> Thank you for filing an issue!

### Version

v0.6.0

### Description

noticed a few issues right off the bat

- first time i opened the app, created a new account, was still on gaia-5000, had to quit and restart app to get to gaia-5001
- fresh install, create new account, sign out - account will not be there to sign back in until you restart the app
- clicking on the mocked option in settings worked nicely, but unclicking does not set network back to gaia-5001
- sign out while in mocked mode goes to sign in screen - but account doesn't exist on mocked server (confusing ux) - also shows `network error: keys` on sign out
- if available balance is 0 - loading screen persists unless you hit refresh
- link to staking is still visible - and route still works (supposed to be hidden)
- onboarding tutorial checkbox takes you back to onboarding modals when trying to mark it as completed

NodeGuy:
> * onboarding tutorial checkbox takes you back to onboarding modals when trying to mark it as completed

I couldn't reproduce this.

nylira:
> * onboarding tutorial checkbox takes you back to onboarding modals when trying to mark it as completed

The bug where the onboarding tutorial shows up after every app restart has been fixed in #717 This resolves the root of this problem.

The settings checkbox for onboarding isn't for marking it completed. It's for toggling it, like the other checkboxes. Checkboxes are not the correct UX, which is mentioned in #669

faboweb:
Staking doesn't show:
![image](https://user-images.githubusercontent.com/5869273/40361066-5d090d2c-5dc8-11e8-9799-358b274211e3.png)

faboweb:
Thank you guys for finding and following up on all those bugs! Great teamwork :)

faboweb:
Do any of these bugs remain?

jolesbi:
i tested these bugs out and all look cleared up except for the one i didn't describe very well. i'll fix it now.

PR #723, opened 2018-05-21
Fix 0.6.0 bugs.

NodeGuy:

### Issue

closes: #720 

### Screenshots

❤️ Thank you!

### DIFF
--- a/app/src/renderer/components/common/AppMenu.vue
+++ b/app/src/renderer/components/common/AppMenu.vue
@@ -3,7 +3,10 @@ menu.app-menu
   .app-menu-main
     list-item(to="/" exact @click.native="close" title="Wallet")
     list-item(to="/wallet/transactions" exact @click.native="close" title="Transactions" v-if="config.devMode")
-    list-item(to="/staking" exact @click.native="close" title="Staking" v-bind:class="{ 'active': isValidatorPage }" v-if="config.devMode")
+
+    // Hide staking for now (see https://github.com/cosmos/voyager/issues/720).
+    //- list-item(to="/staking" exact @click.native="close" title="Staking" v-bind:class="{ 'active': isValidatorPage }" v-if="config.devMode")
+
     list-item(to="/proposals" exact @click.native="close" title="Proposals" v-if="config.devMode")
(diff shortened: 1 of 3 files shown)


app/src/renderer/vuex/modules/node.js
```diff
@@ -145,14 +145,7 @@ export default function({ node }) {
...
       node.rpcReconnect()
-
-      if (mocked) {
-        // if we run a mocked version only, we don't want the lcd to run in the meantime
-        ipcRenderer.send("stop-lcd")
-
-        // the mocked node is automatically connected
-        dispatch("reconnected")
-      }
+      dispatch("reconnected")
```
NodeGuy: @faboweb Please review this change; it fixed the bug but it's unclear to me why we originally wanted to stop the LCD.

app/src/renderer/components/common/AppMenu.vue
```diff
@@ -3,7 +3,10 @@ menu.app-menu
   .app-menu-main
     list-item(to="/" exact @click.native="close" title="Wallet")
     list-item(to="/wallet/transactions" exact @click.native="close" title="Transactions" v-if="config.devMode")
-    list-item(to="/staking" exact @click.native="close" title="Staking" v-bind:class="{ 'active': isValidatorPage }" v-if="config.devMode")
+
+    // Hide staking for now (see https://github.com/cosmos/voyager/issues/720).
+    //- list-item(to="/staking" exact @click.native="close" title="Staking" v-bind:class="{ 'active': isValidatorPage }" v-if="config.devMode")
```
faboweb: we have a development mode flag `v-if="config.devMode"` that is already applied. If the staking option is still visible, the user has `NODE_ENV=development` set globally.

app/src/renderer/vuex/modules/node.js
faboweb: Do you let your car running while you use your bike? ;)

faboweb:
I fixed the reconnection bug in #726 while shutting down the lcd process. If this is acceptable let's use that fix.

PR #735, opened 2018-05-23
peng/release 0.6.2

nylira:
### Issue
Closes #734 

### Release Notes

# Cosmos Voyager Alpha 0.6.2 (UNSAFE)

NOTE: DO NOT ENTER YOUR FUNDRAISER SEED. THIS SOFTWARE HAS NOT BEEN AUDITED. NEVER ENTER YOUR FUNDRAISER SEED 12 WORDS ONTO AN ONLINE COMPUTER.

Even when we do start supporting fundraiser seeds, don't use it except for testing or with small amounts. We will release a CLI to use for offline signing of transactions, and we will also add hardware support for this UI.

### Fixed

* Re-enabled staking and transactions pages for mocked network. @nylira
* Revealed the password for the default account on the mocked network. @nylira
* Fixed not being able to reconnect to a live network if started in mocked mode. @faboweb

Issue #741, opened 2018-05-23
The build script doesn't output the same hashes for everybody


faboweb:
If @okwme runs the build script with commits specified, he gets different hashes.

NodeGuy:
I'm hoping that #742 fixes this.

faboweb:
Could you maybe give an insight into what caused this?

faboweb:
Which hashes do you guys (@okwme i.e.) get for `yarn build --commit=e1658685 --network=app/networks/gaia-5001 --sdk-commit=187be1a5`?

I have:
Cosmos Voyager-darwin-x64_0.6.2.tar.gz SHA256: 4de52ab5d8e025046b69961a8f75d45c7f2a20b919de089a2b3458fc57884db6
Cosmos Voyager-linux-x64_0.6.2.tar.gz SHA256: cce127af72400357a6ac69ec95058d6b38598d10b62258f7b49cb7ddde8177fa
Cosmos Voyager-win32-x64_0.6.2.zip SHA256: f51f24c1e5ef3f8ed77e5076ce13fa6969091865f36c216c2e529dd63d378f88

nylira:
I can match your hashes with that command @faboweb 

Cosmos Voyager-darwin-x64_0.6.2.tar.gz SHA256: 4de52ab5d8e025046b69961a8f75d45c7f2a20b919de089a2b3458fc57884db6

Cosmos Voyager-linux-x64_0.6.2.tar.gz SHA256: cce127af72400357a6ac69ec95058d6b38598d10b62258f7b49cb7ddde8177fa

Cosmos Voyager-win32-x64_0.6.2.zip SHA256: f51f24c1e5ef3f8ed77e5076ce13fa6969091865f36c216c2e529dd63d378f88

faboweb:
Nice!

Issue #740, opened 2018-05-23
Specifying a sdk-commit in the build script doesn't work


faboweb:
Commit: 22e9bf5f6fd78866038be774b01d4a6590d79e38

![image](https://user-images.githubusercontent.com/5869273/40420820-e4abf4ee-5e89-11e8-8184-e2bc729acfc1.png)

NodeGuy:
I'm hoping that #742 fixes this.

faboweb:
This still fails (for different reasons):
![image](https://user-images.githubusercontent.com/5869273/40472636-4dacad0c-5f3a-11e8-9303-dcf0a0a29c93.png)

nylira:
Building works for me with a specific SDK commit. I'm using the short form for the commit ID. Can you try this command?

```
yarn build --commit=e1658685 --network=app/networks/gaia-5001 --sdk-commit=187be1a5
```

faboweb:
Sorry wrong screenshot. :/ Updated it. But for reasons it works now.

nylira:
Should we close this then?

faboweb:
Ok.

PR #742, opened 2018-05-23
tasks/build: Redo build process.

NodeGuy:
This is intended to fix #735, #740, and #741.

Build SDK from source every time we build Voyager because it's changing so fast.
Build all three platforms at once because it doesn't save much time to build them individually.


tasks/build/Dockerfile
```diff
@@ -2,84 +2,95 @@ FROM golang:1.10.1
...
+RUN apt-get update && \
+  apt-get install --yes xz-utils
+
+# from https://raw.githubusercontent.com/nodejs/docker-node/773be66016fd1ff4a860529a80af61277f75c7e2/9/Dockerfile
+
+RUN groupadd --gid 1000 node \
+  && useradd --uid 1000 --gid node --shell /bin/bash --create-home node
+
+# gpg keys listed at https://github.com/nodejs/node#release-team
+RUN set -ex \
```
nylira: This is cool, I like how we're verifying node and yarn versions through gpg keys.

nylira:
I'm able to build and run the built Voyager apps with this PR (macOS and Windows).

faboweb:
Nice work! 🚀 
Could you give an insight what the issue was?

tasks/build/Dockerfile
NodeGuy: Thanks, although I can't take credit for it.  That code is copied-and-pasted from the official Node.js Dockerfile.

FWIW I also question how valuable the exercise is given that it's not leveraging PGP's Web of Trust.

NodeGuy:
I don't know exactly what the issue was.

I designed two phases into the build process: the first runs once and the results are cached to be re-used in further builds, and the second phase runs on every build.  When I originally designed the process we were using a long-lived version of Gaia so I put the downloading of Gaia into the first phase (and later replaced it with building the Cosmos SDK from source).  The issue appeared somewhere in there, including once on my machine, but I thought it was a fluke because I resolved it with `docker system prune`.

When I saw it happening to other people I decided to move the building of the Cosmos SDK into the second phase so that it happens on every build.  This also makes more sense now because we're changing the version of the Cosmos SDK we're using more rapidly than we used to.

faboweb:
Ah nice. Thanks for the update.

PR #745, opened 2018-05-24
Fabo/log user out on live switch

faboweb:
Description:

When a user switched to a live connector, he would still see the mocked address and the other way around. This way the user can be sure about which address he sees.

❤️ Thank you!


app/src/renderer/vuex/modules/node.js
```diff
@@ -152,7 +152,14 @@ export default function({ node }) {
 
         // the mocked node is automatically connected
         dispatch("reconnected")
+      } else {
+        // if we switch to a live connector, we need to wait for the process to have started up again so we can access the KMS
+        commit("setModalSession", "loading")
+        await new Promise(resolve => ipcRenderer.once("connected", resolve))
```
NodeGuy: Nice use of a promise here.

PR #755, opened 2018-05-28
Billy/1051 12 seed 16 seed

okwme:
Closes #ISSUE
https://github.com/cosmos/cosmos-sdk/issues/1051

Description: 12 word recovery phrases cause an error in the cosmos SDK.

I've added form validation to prevent 12 word phrases from being entered until an SDK fix is made.

screenshot 2018-05-28 13 50 08


okwme:
https://github.com/cosmos/voyager/issues/724

faboweb:
@okwme could you add some more tests for NiFormMsg to match coverage?

PR #757, opened 2018-05-28
update logo on readme

jolesbi:
the old image was bothering me - and we have a fresh new logo we otta show off  🎉


faboweb:
I liked to spanning one :X
@nylira what do you think?

PR #770, opened 2018-05-30
Billy/454 tx info

### LINKED ISSUE #454: PageBlock isn't displaying Tx data (when there is supposed to be data)
https://medium.com/@speculatecoin/with-blocks-that-have-transactions-the-section-at-bottom-on-blocks-page-shows-no-transaction-data-aa163de6afcb

okwme:
Closes #454 

Description:
the block explorer doesn't show any details of transactions.

followed the same method of querying txs from the explorer and added an rpc handler to get them.
need help with layout design of how these should look @nylira @jolesbi 

screenshot 2018-05-30 17 24 47

will add tests next as this is still WIP


app/src/renderer/connectors/lcdClient.js
```diff
@@ -89,6 +89,8 @@ Object.assign(Client.prototype, {
   },
   coinTxs: argReq("GET", "/tx/coin"),
 
+  txs: argReq("GET", "/txs"),
```
faboweb: 🥇

jolesbi:
i'm not sure about what data is actually available - but if it could match the tx history component that might be nice.

app/src/renderer/components/monitor/PageBlock.vue
```diff
@@ -44,13 +44,23 @@ page(:title="pageBlockTitle")
     part(title='Transactions')
       data-loading(v-if="blockchain.blockLoading")
       data-empty(v-else-if="block.header.num_txs === 0" title="Empty Block" subtitle="There were no transactions in this block.")
-      list-item(v-else v-for="tx in block.data.txs" :key="tx.id" dt="Transaction" :dd="TODO")
+      template(
```
faboweb: @jolesbi this is what you were looking for, right?

app/src/renderer/components/monitor/PageBlock.vue
```diff
@@ -68,12 +79,15 @@ export default {
     Page
   },
   computed: {
-    ...mapGetters(["blockchain"]),
+    ...mapGetters(["blockchain", "blocTxInfo", "config"]),
```
faboweb: typo?

app/src/renderer/vuex/modules/blockchain.js
```diff
@@ -61,6 +81,66 @@ export default ({ commit, node }) => {
         })
       })
     },
+    async queryTxInfo({ state, dispatch }, height) {
+      let blockTxInfo = state.blockTxs.find(
```
faboweb: why not store them per height?

app/src/renderer/vuex/modules/blockchain.js
```diff
@@ -61,6 +81,66 @@ export default ({ commit, node }) => {
...
+              : 0,
+          txs:
+            state.block && state.block.data && state.block.data.txs
+              ? state.block.data.txs.slice(0)
+              : []
+        })
+        blockTxInfo && state.blockTxs.push(blockTxInfo)
+        return blockTxInfo
+      } catch (error) {
+        return Promise.reject(error)
```
faboweb: isn't that the default for async functions rejecting on error?

app/src/renderer/vuex/modules/blockchain.js
```diff
@@ -61,6 +81,66 @@ export default ({ commit, node }) => {
...
+      let tmp = new Uint8Array(varintlen.byteLength + txbytes.byteLength)
+      tmp.set(new Uint8Array(varintlen), 0)
+      tmp.set(new Uint8Array(txbytes), varintlen.byteLength)
+      return createHash("ripemd160")
+        .update(Buffer.from(tmp))
+        .digest("hex")
+    },
+    getTxs({ state, commit, dispatch }, { key, len, txs }) {
+      return new Promise(async (resolve, reject) => {
+        if (key >= len) return resolve(txs)
```
faboweb: Could you add a comment stating what this statement accomplishes?

app/src/renderer/vuex/modules/blockchain.js
```diff
@@ -61,6 +81,66 @@ export default ({ commit, node }) => {
...
+              ? state.block.data.txs.slice(0)
+              : []
+        })
+        blockTxInfo && state.blockTxs.push(blockTxInfo)
+        return blockTxInfo
+      } catch (error) {
+        return Promise.reject(error)
+      }
+    },
+    convertTx({ state }, txstring) {
```
faboweb: this doesn't do any manipulation on the store and should therefor maybe be in a script or someplace

app/src/renderer/vuex/modules/wallet.js
```diff
@@ -84,9 +85,9 @@ export default ({ commit, node }) => {
     },
     async queryWalletHistory({ state, commit, dispatch }) {
       commit("setHistoryLoading", true)
-      // let res = await node.coinTxs(state.address)
+      let res = await node.coinTxs(state.address)
```
faboweb: does history work again?!

app/src/renderer/vuex/modules/blockchain.js
okwme: was following method used by `queryBlockInfo()` but you're right keying the height is smart i'll update both of em

app/src/renderer/vuex/modules/blockchain.js
okwme: i was having trouble getting the error reporting come out correctly in the tests so went through and made each error very explicit until i found the culprit, i'll walk it back

app/src/renderer/vuex/modules/blockchain.js
okwme: 👍

app/src/renderer/vuex/modules/blockchain.js
okwme: is there a clever place to put it in or just make a new file under /src/scripts?

app/src/renderer/vuex/modules/wallet.js
okwme: oh i uncommented this to see what `NiLiTx` looked like and it seemed to work with mock network. i'll remove tho since that's not the point of this pr

app/src/renderer/vuex/modules/blockchain.js
okwme: actually just updating the blockchain one for sake of PR scope

okwme: actually i think both of these need to stay as arrays as they are used in for loops

faboweb: Can you point me to where they are used in loops? What I have found, there is always a selection by height happening when blockTxs appear in code.

app/src/renderer/scripts/utils.js
```diff
@@ -0,0 +1,14 @@
+import createHash from "create-hash"
+import varint from "varint"
+import b64 from "base64-js"
+
+export default function(txstring) {
```
faboweb: The script is called utils and exports a function that converts tx strings into json? I think this is not really understandable.
How about a script called tx-utils.js that exports a function named parseTx?

app/src/renderer/vuex/modules/blockchain.js
```diff
@@ -30,21 +36,33 @@ export default ({ commit, node }) => {
...
+          dispatch("queryBlock", height).then(block =>
+            commit("setBlock", block)
+          ),
+          dispatch("queryBlockInfo", height).then(blockMetaInfo =>
+            commit("setBlockMetaInfo", blockMetaInfo)
+          )
+        ])
+          .then(() => {
+            state.blockLoading = false
+            dispatch("queryTxInfo", height)
```
faboweb: If you return this, I think you save yourself the inner .catch

app/src/renderer/vuex/modules/blockchain.js
```diff
@@ -61,6 +79,60 @@ export default ({ commit, node }) => {
...
+      //  this function is recursice promie used as an async loop in order to query all tx
+      // found in a block. it's made similarly to queryBlockInfo only there
+      // is more than one async call to make. the txstring is included but might not
+      // actually be useful. etherscan.io includes something similar but it's seldom helpful
+      return new Promise(async (resolve, reject) => {
+        if (key >= len) return resolve(txs)
+        let txstring = atob(txs[key])
+        let hash = await convertTx(txs[key])
+        node
+          .txs(hash)
```
mappum: Why mix `await` and `.then` syntax?

app/src/renderer/vuex/modules/blockchain.js
```diff
@@ -61,6 +79,60 @@ export default ({ commit, node }) => {
...
+        b => b.length && b[0].height === height
+      )
+      if (blockTxInfo) {
+        return blockTxInfo
+      }
+      try {
+        blockTxInfo = await dispatch("getTxs", {
+          key: 0,
+          len:
+            state.block && state.block.data && state.block.data.txs
```
mappum: What's the purpose of these chains of &&'s? Seems like it makes things harder to debug because if the format changes or data is missing, it will silently pass this call and have undefined behavior inside getTxs.

app/src/renderer/scripts/utils.js
```diff
@@ -0,0 +1,14 @@
+import createHash from "create-hash"
```
mappum: Couldn't we just get this from the Node crypto module to avoid adding a dependency?

app/src/renderer/scripts/utils.js
mappum: I agree it should be named, could be something like `getTxHash`

app/src/renderer/vuex/modules/blockchain.js
okwme: oops i totally misinterpreted the unit test error relating to `should disable the next block button if last block`. thought that info was counting an array length for looping and failed when it became an obj (as per a similar error in `blockchain.spec.js`). i've gone back and updated to a keyed storage method for `blockTxs` as well as `blockMetas`

app/src/renderer/scripts/utils.js
okwme: makes sense, i think `getTxHash` is an accurate name

app/src/renderer/vuex/modules/blockchain.js
okwme: refactored all the promises to be more concise

app/src/renderer/scripts/utils.js
```diff
@@ -0,0 +1,14 @@
...
+import varint from "varint"
+import b64 from "base64-js"
+
+export default function(txstring) {
+  let txbytes = b64.toByteArray(txstring)
```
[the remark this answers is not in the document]
okwme: didn't realize it was that ez!

app/src/renderer/vuex/modules/blockchain.js
okwme: refactored all the promises to be more concise

app/src/renderer/vuex/modules/blockchain.js
okwme: this was to allow the tests that had very minimal dummy blocks to pass. but you're right if the block is malformed it should throw an error. i've updated the other tests to use a well formed block dummy

app/src/renderer/scripts/utils.js
okwme: nice, didn't realize it was also that ez

PR #790, opened 2018-06-05
Poll until LCD has started to prevent timing issues

### LINKED ISSUE #706: "Couldn't read keys Network Error"

### Version

0.4.3

### Description

When I launch the release version of the app for gaia-5000 on macOS it reports the error in the attached screenshot.

screen shot 2018-05-16 at 2 04 13 pm

I'm able to connect to `http://localhost:9071/keys` manually so it looks like a spurious/timing error.

### Log

[main.log](https://github.com/cosmos/voyager/files/2010325/main.log)

mappum:
Closes #706 

Description:

I made a quick change to make Voyager poll the LCD until it responds to ensure it has successfully started. This is a simple change although for some reason it seems to have broken a lot of tests and I haven't resolved that yet.


faboweb:
We have a callback now: https://github.com/cosmos/cosmos-sdk/pull/1131
I don't know if the LCD really is up at that point in time though.

mappum:
This is ready for review/merge

test/unit/specs/main.spec.js
```diff
@@ -244,12 +251,12 @@ describe("Startup Process", () => {
       expect(fs.existsSync(testRoot)).toBe(true)
     })
 
-    xit("should init lcd server with correct testnet", async function() {
+    it("should init lcd server with correct testnet", async function() {
```
faboweb: is the test description intended like this?

app/src/main/index.js
```diff
@@ -237,12 +238,20 @@ async function startLCD(home, nodeIP) {
...
       "--chain-id",
       chainId
-      // '--trust-node'
     ])
     logProcess(child, join(home, "lcd.log"))
 
-    // XXX: should wait for server to be ready
-    setTimeout(() => resolve(child), 1000)
+    // poll until LCD is started
+    let client = new LcdClient(`http://localhost:${LCD_PORT}`)
```
faboweb: nice and simple solution.
would in a perfect world it be better if the LCD properly tells us when it is ready to serve?

faboweb:
ping @mappum

test/unit/specs/main.spec.js
mappum: You're right, that test was intended for initting the data directory and not starting the LCD. I added the testnet check to the LCD start test instead.

Issue #762, opened 2018-05-29
Implement Bech32Cosmos


faboweb:
UI Version: 0.6.2

Description:

Implement the spec in the view and the REST server. https://github.com/cosmos/bech32cosmos

faboweb:
Places this probably effects:
- Wallet view
- Transaction view
- Staking view
- user module
- wallet module
- delegates module
- delegation module
- LCD kms
- LCD txs
- LCD staking
- LCD accounts

nylira:
From the repo: DEPRECATED: We are abandoning the bech32cosmos experiment and will use standard bech32 in cosmos

faboweb:
I already converted to bech32. Currently there is a discussion, if we should only use bech32 at the view layer or if this should be the only exchange format. And that we should still do a format check on the bech32 addresses in the view layer: https://github.com/cosmos/cosmos-sdk/pull/1086
Is this a valid plan? @nylira @okwme

okwme:
ok so same plan, just bech32 instead of bech32cosmos 👍

faboweb:
yep :)

PR #791, opened 2018-06-05
Billy/762 bech32

okwme:
Closes #ISSUE #762 

Description:
make all addresses bech32 compliant and checksummed


okwme:
pinging @fabo or @mappum the e2e tests fail because the users are generated via lcd command line and their addresses are stored in non-bech32 format—therefor rejected when attempting to store them in the state. I wasn't sure where I could intervene to transform them or rather that should happen from the beginning and we just need to wait for the new sdk release?

faboweb:
If the bech32 change is in the SDK and the testnet supports that version we can upgrade and use it in Voyager. 
For now the change is merged to develop on the SDK so we could test with a custom build.

faboweb:
I found some bugs in my bech32 implementation in the SDK. I need to fix those first.

faboweb:
This is now mostly working. @okwme could you fix the e2e tests?

okwme:
think the tests are working except those that fail because the testnet

faboweb:
After intense debugging we found that probably the signing in the LCD is broken. We will fix this now.

okwme:
so the error message says `encoding/hex: invalid byte: U+006F 'o'` when sending to address `cosmosaccaddr1pxdf0lvq5jvl9uxznklgc5gxuwzpdy5ynem545`

app/src/renderer/connectors/lcdClientMock.js
```diff
@@ -375,7 +385,11 @@ module.exports = {
   },
   async candidates() {
     return state.candidates
-  }
+  },
+  // exports to be used in tests
```
okwme: nice

PR #801, opened 2018-06-06
Fabo/remove ci from coverage

faboweb:
Remove ci files from code coverage. As they are not really testable.

❤️ Thank you!


.circleci/config.yml
```diff
@@ -44,7 +44,7 @@ jobs:
       - attach_workspace:
           at: builds/gaia
 
-      - run: bash ./ci/changelog-changed-check.sh
+      - run: bash ./tasks/changelog-changed-check.sh
```
NodeGuy: The Debian Policy Manual mandates that shell scripts conform to the [POSIX standard](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/contents.html), i.e., don't use bash.  This is a good policy to follow for all shell scripts.

Best practice is to start your script with the [shebang](https://en.wikipedia.org/wiki/Shebang_(Unix)) `#!/bin/sh` and then in `config.yml` you can write simply `tasks/changelog-changed-check.sh`.

faboweb: This didn't work properly: https://circleci.com/gh/cosmos/voyager/3196

PR #809, opened 2018-06-07
Fabo/776 fallback to mocked connection

### LINKED ISSUE #804: Check node version against gaiacli version
UI Version: 0.6.2

Description:
The node we are connecting to might have another version of the SDK then we are using and therefor the LCD might produce unexpected results. We should check the version and display an error if the node is on another version.

### LINKED ISSUE #851: Close menu on route change
UI Version: 0.7.0

Description:

If I click on the network indicator -> change network the menu is still open and I need to manually close it.

faboweb:
Closes #804
Closes #851

Description:

Do round robin on all known nodes. If there are no available nodes, show an error and let the user decide to either try all nodes again or switch to a mocked connection.

image

Also has code for #804 already

❤️ Thank you!


faboweb:
Wait for SDK 0.20.0

app/src/main/index.js
```diff
@@ -500,11 +500,14 @@ function pickNode(nodes) {
 }
 
 async function checkNodeSDKVersion() {
-  const nodeSDKVersion = await axios(
-    `http://localhost:${LCD_PORT}/version`
-  ).then(res => res.data)
+  const nodeSDKVersion = await axios
+    .get(`http://localhost:${LCD_PORT}/node_version`)
+    .then(res => res.data)
```
NodeGuy:
```JavaScript
(await axios.get(
    `http://localhost:${LCD_PORT}/node_version`
  )).data
```

test/unit/specs/main.spec.js
```diff
@@ -367,9 +368,9 @@ describe("Startup Process", () => {
 
       registerIPCListeners(registeredIPCListeners)
 
-      // axios is used to ping nodes for the reconnection intent
+      // axios is used to ping nodes and get the SDK version of the node
       let axios = require("axios")
-      axios.get = () => Promise.resolve()
+      axios.get = () => Promise.resolve({ data: "0.19.0" })
```
NodeGuy:
```JavaScript
axios.get = async () => ({ data: "0.19.0" })
```

test/unit/specs/main.spec.js
```diff
@@ -693,6 +695,11 @@ function resetModulesKeepingFS() {
   // we want to keep Raven quiet
   const Raven = require("raven")
   Raven.disableConsoleAlerts()
+
+  // axios is used to ping nodes and get the SDK version of the node
+  // we need to keep this mocked, if not, the process fails
+  let axios = require("axios")
+  axios.get = () => Promise.resolve({ data: "0.19.0" })
```
NodeGuy:
```JavaScript
axios.get = async () => ({ data: "0.19.0" })
```

jolesbi:
looks great to me. my only comment is about introducing the word “demo” - usually we use “mock” or “mocked”.

faboweb:
Do users understand the term "mocked"? This is why I used another terminology.

faboweb:
I also got rid of mocked mode flags as well. Voyager will always prompt now to switch to mocked mode which is more usable should avoid confusion.

faboweb:
Motherf***ing big PR. I apologize to reviewers in advance.

NodeGuy:
> Do users understand the term "mocked"? This is why I used another terminology.

No, they don't.  "demo" is better for users.

jolesbi:
the question of demo or mocked is a good one. we tend to assume more technical knowledge with our users than in a conventional consumer product by requiring keys, showing IP addresses, and referring to testnets. so mocked fits in with this technical user profile in my mind. though, demo would be ok too. the important thing for me is just consistency. 

i don't think the user should click the "try demo" button and then see the reference to a mocked server. consistency would be better here - no matter the choice of words.

faboweb:
I think "mocked" is a term mostly familiar with software developers.
Technically interested people will still get "test network" and "ip
address". I propose we switch to "demo" network everywhere. What do you
think?

Jordan Bibla <notifications@github.com> schrieb am Di., 19. Juni 2018,
08:54:

jolesbi:
demo sounds good to me (we discussed in the team meeting today)!

nylira:
Let's clarify it further and rename it "offline-demo".

PR #811, opened 2018-06-11
Peng/775 mocked indicator

### LINKED ISSUE #775: Have a better visual indicator about the mocked connector
UI Version: 0.6.2

Description:

Users might mistake the mocked connection for a real connection. An idea is to have a big visual warning, that this is a mocked connection.

nylira:
Closes #775 

Description: It's not clear to users which network they're connected to, especially now that we start with mocked by default.

Solution: Increase the readability of the network connection component, highlight the `mock-chain` with a special color, and make it easier for users to change networks.

@jolesbi I was tempted to go with a colored border or background, but I found it to be too visually distracting. So I kept the highlight minimal.

Screenshots:

screen shot 2018-06-11 at 5 28 53 pm
screen shot 2018-06-11 at 5 28 41 pm

screen shot 2018-06-11 at 5 37 52 pm
screen shot 2018-06-11 at 5 37 39 pm


nylira:
I could use help on the e2e test. It seems to run fine locally. Thanks!

faboweb:
Do we need the IP there?

faboweb:
Looks fantastic by the way.

nylira:
I think the IP is nice to have. Later the user may want to know more about the node they're connected to, including privacy implications--sometime in the future we may also implement the node switcher.

jolesbi:
this looks great peng. my only request is to use a darker yellow for the light version for increased readability. it looks like the darker yellow is already being used as the background for the icon. if that's the same as the text, perhaps you can darken it slightly or add some text shadow.

🎨

nylira:
We use one yellow for both light and dark themes. I think a good way of making yellow readable is to not use yellow, and to use orange instead:

screen shot 2018-06-13 at 8 14 55 am
screen shot 2018-06-13 at 8 14 48 am

jolesbi:
orange is the new yellow

nylira:
I'm stuck on test coverage. I cannot get the router to navigate to the Preferences page, to test for the existence of a link. Can you help me @faboweb?

PR #834, opened 2018-06-14
David/build fixes

NodeGuy:
Fix bug in which the network configuration isn't copied to the build.

Build Gaia automatically if it isn't found.

Fix a new bug in release caused by moving it from ci/ to tasks/.

### DIFF
--- a/tasks/build/runBuild.js
+++ b/tasks/build/runBuild.js
@@ -51,7 +51,7 @@ cli(optionsSpecification, options => {
       --mount type=bind,readonly,source=${resolved.git},target=/mnt/.git \
       --mount type=bind,readonly,source=${
         resolved.network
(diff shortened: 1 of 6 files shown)


tasks/build/runBuild.js
```diff
@@ -18,23 +18,30 @@ const execSync = (command, options) => {
   childProcess.execSync(command, Object.assign({ stdio: `inherit` }, options))
 }
 
-cli(optionsSpecification, options => {
+cli(optionsSpecification, async options => {
+  const builds = path.join(__dirname, `../../builds`)
+
+  if (!(await fs.pathExists(path.join(builds, `Gaia`)))) {
```
faboweb: ❤️

PR #865, opened 2018-06-20
loading improvements

### LINKED ISSUE #765: let's use an animated loader so everyone knows we care 😍
we tend to default to `loading...` but a small loading image or animated html element will give our app a nice sense of aliveness 🚤

jolesbi:
- changed loading icon
- added loading icon to some missing spots
- removed content from initial html
- set background color for electron

loading-screenshot

closes #765


app/index.ejs
```diff
@@ -1,25 +1,19 @@
...
+    </script>
+  <% } %>
 
-<style>
-<%= htmlWebpackPlugin.options.styles %>
-</style>
+  <body style="background: #15182d">
+    <div id="app" style="background: #15182d">
+      <span>Loading...</span>
+      <!-- <img src="imgs/loader.svg" alt="spinner"> -->
```
jolesbi: does anyone know how to get an image to load here?

faboweb:
Can we get some screenshots of the visual changes?

app/index.ejs
faboweb: What is the problem? Maybe the path "imgs/loader.svg" is not correct in production?

jolesbi:
> Can we get some screenshots of the visual changes?

yes of course! sorry! i took them but forgot to include them! they are now above.

app/index.ejs
jolesbi: yeah i think this file ends up in the dist folder

jolesbi:
ok this is ready to go!

faboweb:
Summing up: This removes the preloader showing before the app is actually loaded. We don't need the preloader anymore, as initialization is done asynchronous via IPC.

PR #872, opened 2018-06-22
style updates and contrast improvements

jolesbi:
what started out as a simple "improve accessibility by improving contrast" quickly turned into "make minor improvements in 30+ files". sorry about that. i think i like where i ended up - there is still a lot to do - but if ya'll like these changes then i can continue applying these conventions throughout. if not, i can peel some back and just focus on the contrast as originally intended.

some of the (existing) conventions include:
- pink for selected state
- blue buttons as primary actions
- all squares / rectangles / elements have 2px border-radius
- tighter use of grid system
- contrast for text passes AA and usually AAA standards
- better use of font hierarchy

screen shot 2018-06-21 at 11 29 33 pm

screen shot 2018-06-22 at 10 13 54 am

screen shot 2018-06-22 at 10 23 26 am

screen shot 2018-06-22 at 10 22 40 am

✌️


NodeGuy:
I like it!

PR #882, opened 2018-06-25
Fabo/857 addressbook

### LINKED ISSUE #857: Create a peer store
UI Version: 0.7.0

Description:

Relying on peers in the config.toml is error prone. Especially on testnets. I think we should store discovered peers and cycle through them when trying to connect.

faboweb:
Closes #857

Description:

introduce an addressbook that stores found peers so we do not use the hard coded ips everytime and therefor also avoid crashing voyager if those go down.

- [x] write tests

❤️ Thank you!


app/src/main/addressbook.js
```diff
@@ -0,0 +1,134 @@
...
+
+    subPeersHostnames
+      // check if we already know the peer
+      .filter(subPeerHostname => !this.peerIsKnown(subPeerHostname))
+      // add new peers to state
+      .forEach(subPeerHostname => {
+        this.addPeer(subPeerHostname)
+      })
+
+    this.persistToDisc()
```
okwme: don't need to `persistToDisc()` if `!subPeersHostnames.length`

app/src/main/addressbook.js
```diff
@@ -0,0 +1,134 @@
...
+    this.peers = []
+
+    this.loadFromDisc()
+
+    // add persistent peers to already stored peers
+    persistent_peers
+      .filter(peer => !this.peerIsKnown(peer))
+      .forEach(peer => this.addPeer(peer))
+
+    this.persistToDisc()
```
okwme: don't need to `persistToDisc()` if `!persistent_peers.length`

test/unit/specs/main.spec.js
```diff
@@ -173,26 +173,32 @@ describe("Startup Process", () => {
...
+          }
+      )
+
       Object.assign(process.env, {
         COSMOS_NODE: "123.456.789.123"
       })
       // run main
-      main = await require(appRoot + "src/main/index.js")
-
-      expect(
```
okwme: why would these not pass anymore?

test/unit/specs/main.spec.js
```diff
@@ -173,26 +173,32 @@ describe("Startup Process", () => {
...
-        childProcess.spawn.mock.calls.find(
-          ([path, args]) =>
-            path.includes("gaiacli") &&
-            args.includes("rest-server") &&
-            args.includes("123.456.789.123:46657")
-        )
-      ).toBeDefined()
```
okwme: why would these not pass anymore?

faboweb:
All very good comments @okwme. Made me rethink some things. :)

test/unit/specs/main.spec.js
faboweb: I put this logic now into the addressbook

Issue #888, opened 2018-06-26
New components don't look right


okwme:
Preferences page doesn't look right w select dropdowns
are they supposed to be full width?

![screenshot 2018-06-26 17 46 52](https://user-images.githubusercontent.com/964052/41923998-34cb07cc-7969-11e8-8773-59750f647b12.png)

jolesbi:
- full width pho sho
- the little bar on the right should be height 100%
- ideally, the height of the dropdowns and the buttons would be the same
- ideally, skinny buttons wouldn't be necessary :trollface: 
- ideally, the border colors would match our button styles or vice versa

nylira:
The height of `select` elements looked fine before the app color/contrast PR by @jolesbi. They used to be the same height as the buttons, so let's go back to that. (see picture in #864)

nylira:
I like having two button sizes. Look at GitHub: there's a separate button size for "Close and comment"/"Comment" and the "Edit"/"New Issue" buttons in the top right of this page. I don't think this has to be changed.

faboweb:
Is the distinction: primary action buttons / secondary action buttons?

NodeGuy:
Yo yo Mr. Picnic, do you need help getting this done for tomorrow?

PR #894, opened 2018-06-28
Billy/864 preferences pages

### LINKED ISSUE #864: preference page components
screen shot 2018-06-20 at 11 26 29 am

1. excellent page!
2. i don't think the "off" state needs to be red or "danger" - something like `app-fg` would be suitable
3. the circle on the toggle (the part that you click) should have a border like our buttons
4. the dropdowns should match the button styling or vice-versa (the borders are different)
5. i think my preference would be for the in-button-icons to be on the right of the text

okwme:
Closes #864 , #888 

Description: field styles were wrong because the tendermint/ui component TmField was still using ni- prefix on the select dropdown. This fix is actually blocked by an update to the npmjs.com repo of the Tm components. When using the updated version locally voyager looks like this:

screenshot 2018-06-28 18 35 04


nylira:
Just updated @tendermint/ui to 0.2.8 with your fixes!

PR #895, opened 2018-06-28
David/fix release

NodeGuy:
Fixes bugs in the automated release process.


package.json
```diff
@@ -36,6 +36,7 @@
   },
   "devDependencies": {
     "@nodeguy/cli": "0.2.1",
+    "@octokit/rest": "15.9.4",
```
faboweb: can we remove simple-git then?

NodeGuy: No, simple-git is an interface to Git and @octokit/rest is an interface to GitHub.

faboweb: We are also using this tool only to create one PR, right? Wouldn't this have been possible just using the API and REST calls?
I am fine with merging this for now, but we discussed preferring fewer dependencies before. Let's keep this in mind for the next dependency cleanup session.

PR #905, opened 2018-07-02
Fabo/local testnet tutorial

faboweb:
Added a tutorial how to setup a local testnet.


mappum:
To fix that issue, there will have to be one step added:
```
./builds/Gaia/{OS}/gaiad version > ./app/networks/local-testnet/basecoindversion.txt
```

(And note that you'll have to `rm -rf /Users/david/.cosmos-voyager-dev/test-chain-zpvZU5` before trying again since it was initialized without the `basecoindversion.txt`)

faboweb:
Thanks for fixing the bugs @mappum @NodeGuy :)

2 questions:
- should we maybe rename basecoindversion to sdkversion?
- should basecoindversion always be copied to the config folder if not found?

PR #910, opened 2018-07-03
Update wallet when a transaction is detected

### LINKED ISSUE #679: Wait for sent transaction to be committed before updating wallet
See https://github.com/cosmos/voyager/pull/591/files/7138a811263287af901e4b8376e2761feeb807aa#diff-f8cf2faad79bb70cb93dee868aa131aaR45

We can probably check for the tx to go through somehow through RPC (block events or polling).

### LINKED ISSUE #847: Wallet Balance doesn't update
UI Version: 0.7.0

Description: After getting voyager up for the first time on a fresh install and importing my keys i was told i held no tokens. I went to a faucet and requested some but still no tokens. Went to the block explorer and looked for a recent block with any transactions and found mine (i think it was mine). Went back to wallet and still no balance and then remembered there's a refresh button. maybe wallet balance should be automatically refreshed on the mounted event?

mappum:
Closes #847 
Closes #679 

Description:

This is working pretty well, now the wallet doesn't have to poll and will update automatically when an incoming or outgoing tx is detected.

[🎥  PREVIEW](https://i.imgur.com/jCo1SnE.gifv)


faboweb:
Very nice PR. Amazing video showing it works. :)
I did some lite refactoring, I hope that was OK.
While doing so I figured, that the LCD doesn't return bech32 addresses for key creation: https://github.com/cosmos/cosmos-sdk/pull/1522

E2E tests currently fail until this is resolved.

mappum:
>While doing so I figured, that the LCD doesn't return bech32 addresses for key creation

I'm confused about this, the LCD binary I have from `yarn build:gaia` does give me bech32 addresses:

faboweb:
Not for the `GET /keys` endpoint but for the `POST /keys` endpoint

mappum:
Ah I see, so creating a new account attempts to decode the non-bech32 address but my try/catch block was obscuring that.

faboweb:
Correct

Matt Bell <notifications@github.com> schrieb am Di., 3. Juli 2018, 19:16:

PR #911, opened 2018-07-03
Hide block subscription errors

### LINKED ISSUE #844: Error subscribing to new blocks
0.7.0

Description: I'm getting a lot of "Error subscribing to new blocks" error messages, both during account import and while logged in. It may be because Voyager is trying to switch to different full nodes in the background. The app continues to work, and the block count is increasing despite these errors. I vote for disabling these error messages because they are very user unfriendly.

screen shot 2018-06-14 at 2 28 06 pm

mappum:
Closes #844 

Description:

We sometimes can't subscribe to blocks (e.g. when we can't connect to a node), and we shouldn't spam the user with errors that don't really make sense to them. This just switches from an error notification to `console.error`.


faboweb:
Nice and simple solution.

PR #937, opened 2018-07-06
David/744 build qol

### LINKED ISSUE #744: QOL improvements for building
UI Version: 0.6.2

Description: When we build for all three operating systems, it would be nice if we get a final summary of (1) all three SHA256 hashes for the three zip files, and (2) how long it took to build for each OS, and (3) a combined time taken total, presented in minutes and seconds.

Also, let's discuss adding a `--mocked` flag to build mocked releases as default.

NodeGuy:
Closes #744

Description:

* No `--mocked` flag because we're going to test for unstable networks automatically instead.
* Currently hashes network name but not whole network config because we're moving away from providing whole network config in another PR.


tasks/build/build.js
```diff
@@ -50,19 +52,23 @@ const pack = () => {
   childProcess.execSync(`npm run pack`, { stdio: `inherit` })
 }
 
-function sha256File(path) {
-  let hash = createHash("sha256")
-  fs.createReadStream(path).pipe(hash)
-  return new Promise((resolve, reject) => {
-    hash.on("data", hash => resolve(hash.toString("hex")))
-  })
+const sha256 = async data => {
```
faboweb: wouldn't `async function sha256(data) {...}` tell us better what this statement does?

tasks/build/build.js
```diff
@@ -50,19 +52,23 @@ const pack = () => {
...
+  const hash = createHash(`sha256`).setEncoding(`base64`)
+
+  return data instanceof stream.Readable
```
faboweb: I think overloading the input to this function doesn't make it very readable.

tasks/build/build.js
NodeGuy: That question is blog-worthy!

https://www.nodeguy.com/javascript-function-definitions/

tasks/build/build.js
NodeGuy: The purpose is to make the calling code to the function more readable.  The alternative would look like this:

```JavaScript
const inputsHash = await sha256Array([gaiaVersionHash, options])
const outputsHash = await sha256String(buildHashes)
const gaiaVersionHash = await sha256File(path.join(__dirname, `Gaia/COMMIT.sh`))
```

In the future there would need to be `sha256Object`, etc., for every new type that could be hashable.  It's better to make a more complex implementation in service of a simpler interface.

faboweb: The other way would be to only do sha256 of string to keep the interface clean and use `JSON.stringify` on arrays and objects to reduce those types to string. For files it is possible to just read the file content as string and hash this. Doing so makes it very clear what happens. In comparison to the developer needing to dive into a function to understand its functionality.

If you insist to keeping it this way, I don't want to block this from progressing.

tasks/build/build.js
faboweb: Unification of the syntax is not always better for readability: https://en.wikipedia.org/wiki/Brainfuck
Hoisting sounds like a good reason to use this syntax. Which is to be valued higher, I don't know.
Let's keep it this way.

NodeGuy: Your Brainfuck point is an excellent rebuttal.

I still prefer the expression syntax because it makes it easier to modify functions later with higher-order modifiers.

tasks/build/build.js
NodeGuy: I'd like to understand your confusion better.  You're saying it's unclear how to use the function?

Would you make a similar argument about `JSON.stringify` and that it should be like this instead?

```JavaScript
JSON.stringifyArray
JSON.stringifyObject
JSON.stringifyString
JSON.stringifyNumber
JSON.stringifyBoolean
etc
```

tasks/build/build.js
faboweb: You have an example?

tasks/build/build.js
faboweb: You are right. I would probably hate that. :)
Good explanation. Thank you.

tasks/build/build.js
NodeGuy: Not off the top of my head.  I'll let you know the next time I run across one.

PR #941, opened 2018-07-08
Fabo/show connection infos

faboweb:
Added information about the current connection as the addressbook might attempt to connect to a lot of offline nodes in some cases.

Video:
out

❤️ Thank you!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Added
+
+* Information about current connection attempts @faboweb
+
 ### Changed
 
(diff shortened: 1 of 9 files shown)


faboweb:
Sorry for the bad screencast :D anyone has nice tooling for that?

jolesbi:
can you also share a static screenshot? i can't tell what i'm supposed to be reviewing 🎥😆

faboweb:
This is the initial "Loading..." view before the login screen.

faboweb:
image

okwme:
hard to test without a testnet. using localnode i connect, using localnode without actually running localnode i hang on "connecting to 'gaia-6002'" without any ip addresses. Maybe good to have a final message when you run out of ip addresses to try?

okwme:
ah it did work! forgot that testnet is up just no new blocks coming in

PR #956, opened 2018-07-11
SDK 0.21.1

### LINKED ISSUE #953: Move to SDK 0.20.0

faboweb:
Closes #953 

Description:

❤️ Thank you!


faboweb:
Blocked on https://github.com/cosmos/cosmos-sdk/pull/1683

app/src/renderer/vuex/modules/wallet.js
```diff
@@ -1,5 +1,6 @@
 let fs = require("fs-extra")
 let { join } = require("path")
+let { uniqBy } = require("lodash")
```
NodeGuy: Please use `const` whenever possible.

test/e2e/launch.js
```diff
@@ -235,22 +230,78 @@ function startLocalNode() {
...
+        seed ? "--recover" : null,
+        "--home",
+        join(cliHome, "lcd"),
+        "--output",
+        "json"
+      ].filter(x => x !== null)
+    )
+
+    child.stdout.once("data", data => {
+      let msg = data.toString()
```
NodeGuy: Please use `const`.

test/e2e/launch.js
```diff
@@ -235,22 +230,78 @@ function startLocalNode() {
...
-    ])
+function initLocalNode() {
+  return new Promise((resolve, reject) => {
+    const command = `${nodeBinary} init --home ${nodeHome} --name local --owk`
+    console.log(command)
+    const localnodeProcess = spawn(command, { shell: true })
+    localnodeProcess.stderr.pipe(process.stderr)
+
+    localnodeProcess.stdout.once("data", data => {
+      let msg = data.toString()
```
NodeGuy: Please use `const`.

test/e2e/launch.js
```diff
@@ -235,22 +230,78 @@ function startLocalNode() {
...
+    })
+
+    localnodeProcess.once("exit", reject)
+  })
+}
+
+// save the version of the currently used gaia into the newly created network config folder
+function saveVersion() {
+  return new Promise((resolve, reject) => {
+    let versionFilePath = join(nodeHome, "config", "basecoindversion.txt") // nodeHome/config is used to copy created config files from, therefor we copy the version file in there
```
NodeGuy: Please use `const`.

test/e2e/launch.js
```diff
@@ -235,22 +230,78 @@ function startLocalNode() {
...
+    const command = `${nodeBinary} version`
+    console.log(command, ">", versionFilePath)
+    let child = spawn(command, { shell: true })
```
NodeGuy: Please use `const`.

test/e2e/launch.js
```diff
@@ -235,22 +230,78 @@ function startLocalNode() {
...
+    child.stderr.pipe(process.stderr)
+    child.stdout.once("data", data => {
+      let msg = data.toString()
```
NodeGuy: Please use `const`.

test/e2e/launch.js
```diff
@@ -235,22 +230,78 @@ function startLocalNode() {
...
+      } else {
+        reject(msg)
+      }
+    })
+  })
+}
+
+function createAccount(name, seed) {
+  return new Promise((resolve, reject) => {
+    let child = spawn(
```
NodeGuy: Please use `const`.

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 * Copy the network configuration only during local builds. @NodeGuy
 * Ignoring the potential local-testnet folder that is used in the README for local testnets @faboweb
+* Moved to SDK v0.20.0 @faboweb
```
okwme: or v.0.21.1?

test/unit/specs/lcdClient.spec.js
```diff
@@ -97,6 +97,32 @@ describe("LCD Client", () => {
...
+
+  it("does not throw error for empty results", async () => {
+    axios.get = jest.fn().mockReturnValueOnce(
+      Promise.reject({
+        response: {
+          data: "account bytes are empty"
+        }
+      })
+    )
+    let res = await client.queryAccount("address")
```
NodeGuy: Please use `const`.

app/src/renderer/components/wallet/PageSend.vue
```diff
@@ -148,7 +148,7 @@ export default {
         await this.sendTx({
           type,
           to: address,
-          amount: [{ denom, amount }]
+          amount: [{ denom, amount: amount.toString() }]
```
okwme: was this causing the weird rounding errors before?

app/src/renderer/connectors/lcdClient.js
```diff
@@ -95,9 +95,7 @@ Object.assign(Client.prototype, {
   candidates: req("GET", "/stake/validators"),
   getValidators: req("GET", "/validatorsets/latest"),
   queryDelegation: function(delegator, validator) {
-    return req("GET", `/stake/${delegator}/bonding_status/${validator}`).call(
-      this
-    )
+    return req("GET", `/stake/${delegator}/delegation/${validator}`).call(this)
```
okwme: will this same endpoint keep us from DDoSing ourselves?

app/src/renderer/connectors/lcdClientMock.js
```diff
@@ -40,89 +40,78 @@ let state = {
...
+                ]
+              }
             }
-          }
+          ]
         }
       },
+      hash: "999ADECC2DE8C3AC2FD4F45E5E1081747BBE504A",
       height: 1
     },
```
okwme: worried this tx format might not be in sync w new tm components... might have to update my open PR if tx format is updated again

test/e2e/wallet.js
```diff
@@ -41,7 +41,7 @@ test("wallet", async function(t) {
     async function goToSendPage() {
       await navigate(app, "Wallet")
       await $("#part-available-balances")
-        .$(".tm-li-dt=FERMION")
+        .$(".tm-li-dt=LOCALTOKEN")
```
okwme: nice, where did fermion come from actually?

test/e2e/wallet.js
```diff
@@ -52,15 +52,15 @@ test("wallet", async function(t) {
     let sendBtn = () => $(".tm-form-footer button")
     let addressInput = () => $("#send-address")
     let amountInput = () => $("#send-amount")
-    let defaultBalance = 9007199254740992
```
okwme: RIP big boy

test/unit/specs/addressbook.spec.js
```diff
@@ -48,7 +48,7 @@ describe("Addressbook", () => {
       persistent_peers: ["123.456.123.456"]
     })
     let node = await addressbook.pickNode()
-    expect(node).toBe("123.456.123.456:46657")
+    expect(node).toMatchSnapshot()
```
okwme: curious as to the benefit of this way over the previous. previous is more explicit no? whereas matching snapshot is more easily overwritten with the -u tag meant for another test. are there other benefits?

app/src/renderer/vuex/modules/wallet.js
mappum: IMO we should talk about our style for this. I don't think using `const` is usually worth it since I don't think it's ever prevented a bug for me.

It actually doesn't even error when re-assigned outside of strict mode, which seems like it could even hide some bugs:
```js
let foo = 5
foo = 6
console.log(foo) // => 5
```
So maybe it makes sense if we also make a rule of enabling strict mode for each file, but I'm not convinced going through and changing all that is worth it.

test/e2e/wallet.js
faboweb: I guess something something atomic: https://de.wikipedia.org/wiki/Fermion
;)

test/unit/specs/addressbook.spec.js
faboweb: While refactoring I noticed the benefit of updating snapshot with `-u`. A downside could be that one could miss a  `undefined` snapshot

app/src/renderer/vuex/modules/wallet.js
faboweb: Let's not use const everywhere then. We can discuss in our team meeting tomorrow if we want to enforce this.

CHANGELOG.md
faboweb: 👍

app/src/renderer/connectors/lcdClient.js
faboweb: I don't think so. This was just an API rename.

NodeGuy:
GitHub won't let me review this again.

PR #959, opened 2018-07-12
Fabo/detect halted node

faboweb:
Description:

Show a modal when a node doesn't signal that it produces blocks.

image

❤️ Thank you!


app/src/renderer/components/common/TmModalNodeHalted.vue
```diff
@@ -0,0 +1,105 @@
+<template lang="pug">
+.ni-modal-error__wrapper
```
jolesbi: are we using `ni-modal` still?

faboweb: https://github.com/cosmos/voyager/pull/967

PR #963, opened 2018-07-13
Fabo/vue jest

faboweb:
Proposal for another vue transpiler in tests that has actual sourcemaps showing the original file. and is better maintained.

❤️ Thank you!


faboweb:
Feel free to merge.

PR #964, opened 2018-07-13
Billy/943 block explorer tm

### LINKED ISSUE #943: consolidate explorer components into TM library
Description: follow up actionables from https://github.com/cosmos/voyager/issues/887

- [x] move all logic out of components, accept props instead
- [x] move to tm library
- [x] update tm version and import into voyager
- [ ] update tm version and import into explorer

okwme:
Closes #943

Description: 
 - Moved explorer dependent components into tendermint/ui
 - Update tx format


app/networks/local-testnet/genesis.json
```diff
@@ -0,0 +1,102 @@
+{
```
faboweb: checked in local-testnet

okwme: oops, must have added it before your .gitignore addition!

test/unit/specs/components/monitor/__snapshots__/PageBlock.spec.js.snap
```diff
@@ -30,219 +30,12 @@ exports[`PageBlock has the expected html structure 1`] = `
...
-                        <!---->
-                     </div>
-         <div class=\\"tm-li-dl\\">
-            <div class=\\"tm-li-dt\\">
-               Block Hash
-            </div>
-         <div class=\\"tm-li-dd\\">
-            hash
-         </div>
+   <tm-block blockmeta=\\"[object Object]\\" block=\\"[object Object]\\"></tm-block>
```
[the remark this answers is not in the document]
okwme: true... maybe better to tackle the time problem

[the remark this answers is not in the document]
faboweb: Maybe together on Monday

okwme:
Closing until I can figure out the time problem

NodeGuy:
What's the time problem?

faboweb:
The block explorer component is affected by the timezone of the CI environment.

NodeGuy:
Can you refactor it in accordance with the principles of my talk?

okwme:
@NodeGuy I like the idea of moving moment out of the tm library!

I was able to do it easily with `TmBlock.vue` by passing the time as a pre-formatted string, but if I were to do the same with `TmLiTransaction.vue` then the times would need to be formatted on each Tx at an earlier point. 

This could also be done with a mapping inside of the `PageBlock.vue` where it was done for `TmBlock.vue` (the order goes `TmLiTransaction.vue` < `TmBlock.vue` < `PageBlock.vue`).  But `TmLiTransaction.vue` is also used in `PageTransactions.vue`.

Now you've got two different components that pre-format an array of times to be passed down to children components. Shouldn't that format be consistent across parents? To do so in this situation you'd need to look at where the txs are originally getting fetched inside the vuex module.

At that point I'd be applying template level logic inside of vuex, but that doesn't feel right. If I want to later change the date format, I would expect to do so at the rendering level, not the vuex module. Really I would expect to do that in the component library, but that would violate the principals you've described.

Do you have a recommendation for following the principals of your talk, while not hiding render level login inside vuex, or is that actually an ok place to decide formatting?

CHANGELOG.md
```diff
@@ -19,9 +19,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 * Copy the network configuration only during local builds. @NodeGuy
 * Ignoring the potential local-testnet folder that is used in the README for local testnets @faboweb
+* Extracted components into @tendermint/ui for use in explorer @okwme
 * Moved the changelog check to a GitHub status check @faboweb
 * Moved to SDK v0.21.1 @faboweb
 * Using vue-jest for transpiling in tests for better code coverage output @faboweb
+* Moved to SDK v0.21.1 @faboweb
```
faboweb: dupe

NodeGuy:
Let's try talking about this on Slack tomorrow.

faboweb:
A dependency is missing. Maybe cache bust the dependencies?

PR #974, opened 2018-07-17
Make command line programs behave better during exceptions.

NodeGuy:
There was some boilerplate code in our CLI tools that looked like this:

```JavaScript
main().catch(reason => {
  console.error(reason)
  process.exit(1)
})
```

This PR adds the above code to the CLI argument parsing library used by some of the tools.


faboweb:
Could you give a little more insight about what the problem was and how the new version fixes it?

faboweb:
+ please do not skip push hooks, you missed the changelog update again ;P

NodeGuy:
> Could you give a little more insight about what the problem was and how the new version fixes it?

Done.

faboweb:
Is it correct that there is no code change on our side then?

NodeGuy:
Correct and I don't have permission to merge.

PR #976, opened 2018-07-18
Billy/920 yarn start

### LINKED ISSUE #920: Rename yarn testnet to yarn start
UI Version: 0.7.0

Description:

The default npm and yarn start cmd is `yarn start` / `npm start`. I think we should move to this default. Specifying a testnet still works like this: `yarn start gaia-6002`.

okwme:
Closes #920 

Description: Rename yarn testnet to yarn start


package.json
```diff
@@ -17,7 +17,8 @@
     "build": "node tasks/build/runBuild",
     "build:gaia": "cd tasks/build/Gaia; ./runBuild.sh",
     "local": "yarn run testnet local",
-    "testnet": "node tasks/testnet.js",
+    "start": "node tasks/testnet.js",
+    "testnet": "yarn start",
```
faboweb: Nice! :) Maybe add `echo "'yarn testnet' will be deprecated. Please use 'yarn start'"`?

okwme: gets lost if `yarn testnet` falls back to `yarn start`, think maybe  it's better to just print the message

PR #977, opened 2018-07-18
gaia 7001

### LINKED ISSUE #973: Gaia-7001

okwme:
Closes #973 

Description: Voyager defaults to gaia 7001 with sdk commit hash for 0.22.0
genesis file taken from instructions at https://github.com/cosmos/cosmos-sdk/blob/master/cmd/gaia/testnets/README.md
added nodes from the explorer to the seeds list

having very unstable experience w voyager, but those might all be separate issues


app/config.toml
```diff
@@ -16,4 +16,4 @@ sentry_dsn = "https://4dee9f70a7d94cc0959a265c45902d84:cbf160384aab4cdeafbe9a08d
 sentry_dsn_public = "https://4dee9f70a7d94cc0959a265c45902d84@sentry.io/288169"
 
 # time to wait for a block until node is declared halted
-node_halted_timeout = 10000
+node_halted_timeout = 600000
```
faboweb: This is now 10mins until the user is prompted that the node has halted. I would propose max 5min maybe 2mins.

okwme: 120,000 = 2 min now

jolesbi:
despite the existing problems, i think we should merge this and open issues for the bugs

PR #1014, opened 2018-07-25
Fabo/986 fix subscription

### LINKED ISSUE #986: Fix subscription to txs for balance updating
UI Version: 0.8.0

Description:

Currently the balance is not updated when a tx for your account is detected. The subscription is never called.

faboweb:
Closes #986 

Description:

The tags now use the bech32 addresses instead of the hex addresses.

❤️ Thank you!


jolesbi:
why don't we have to decode addresses anymore?

faboweb:
because tags now use the bech32 addresses

PR #1030, opened 2018-07-27
Fabo/1005 page views

### LINKED ISSUE #1005: google analytics is not tracking pageviews
investigate if this is possible in an electron app or this is misconfigured

faboweb:
Closes #1005 

Description:

❤️ Thank you!

### DIFF
--- a/app/src/renderer/vuex/modules/user.js
+++ b/app/src/renderer/vuex/modules/user.js
@@ -149,9 +149,10 @@ export default ({ node }) => {
       if (state.errorCollection) {
         console.log("Analytics enabled in browser")
         enableGoogleAnalytics(config.google_analytics_uid)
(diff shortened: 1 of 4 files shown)


app/src/renderer/vuex/modules/user.js
```diff
@@ -27,6 +27,10 @@ export default ({ node }) => {
     },
     addHistory(state, path) {
       state.history.push(path)
+      window.analytics &&
+        window.analytics.send("pageview", {
```
jolesbi: what else could we be tracking that we aren't?

faboweb: Let's move this here: https://github.com/cosmos/voyager/issues/1032

PR #1050, opened 2018-07-30
automatic release created for v0.9.2

faboweb:
### Changed

* Switched to Gaia-7004 @faboweb
* showing warning about out of range of atoms in bonding UI only once on top and renaming it @faboweb

### Fixed

* slider in staking UI @faboweb
* error with empty accounts @faboweb


NodeGuy:
Did the automatic release process finally work?

faboweb:
Apparently :D

PR #1056, opened 2018-07-31
Fabo/982 fix login not working

### LINKED ISSUE #982: Password input fails in the beginning
UI Version: 0.7.1

Description:

When I first enter my password it fails. Maybe the binary has not started yet?

faboweb:
Closes #982 

Description:

This disabled the signin if not connected to a node. I hope this fixes the problem.

❤️ Thank you!


app/src/renderer/components/common/TmSessionImport.vue
```diff
@@ -73,6 +74,9 @@ export default {
     TmFormMsg,
     TmFormStruct
   },
+  computed: {
+    ...mapGetters(["connected"])
```
NodeGuy: I think we should stick to using only standard features of JavaScript.  We want our code to be used by other developers.

app/src/renderer/components/common/TmSessionSignIn.vue
```diff
@@ -52,7 +53,7 @@ export default {
     }
   }),
   computed: {
-    ...mapGetters(["user", "mockedConnector", "lastHeader"]),
+    ...mapGetters(["user", "mockedConnector", "lastHeader", "connected"]),
```
NodeGuy: same as above

app/src/renderer/components/common/TmSessionSignUp.vue
```diff
@@ -77,6 +79,9 @@ export default {
     TmFormMsg,
     TmFormStruct
   },
+  computed: {
+    ...mapGetters(["connected"])
```
NodeGuy: same as above

app/src/renderer/components/common/TmSessionImport.vue
faboweb: ?? This is a feature of vuex. Could you elaborate what you mean and what you propose as an alternative?

NodeGuy: I mean the [object spread property is only a proposal](https://github.com/tc39/proposal-object-rest-spread) and is not part of JavaScript.

As an alternative you can write the following in JavaScript:

```JavaScript
computed: mapGetters(["connected"]),
```

faboweb: We would need to change this in the whole application, as it is used like this in almost every component. I think this change doesn't belong into this PR.
Let's discuss this in the next meeting, ok?

NodeGuy: I don't understand why it needs to be changed all at once but I'm happy to discuss in the meeting if you like.

PR #1062, opened 2018-08-01
automatic release created for v0.9.3

faboweb:
### Fixed

* builds not updating testnet configurations if the gaia version doesn't change @faboweb


NodeGuy:
Hmm, why aren't the tests running?

PR #1067, opened 2018-08-02
Fabo/960 cache state

### LINKED ISSUE #960: Cache wallet balance locally
UI Version: 0.7.1

Description:

Currently the wallet shows always loading in the beginning. How about caching the wallet balance locally so the user sees their last balance until it updates? Action buttons like sending / staking could be disabled until the balance is confirmed. The locally cached balance could be encrypted with the user password.

faboweb:
Closes #960 

Description:

❤️ Thank you!


fedekunze:
@faboweb Approved but some of the required checks failed. Once addressed we can merge to develop

faboweb:
Thank you for the quick review. Checks are passing now @fedekunze

PR #1082, opened 2018-08-07
WIP: Create Pact tests for Gaia-Lite staking endpoints.

NodeGuy:
Closes #ISSUE

Description:

❤️ Thank you!

### DIFF
--- a/app/src/renderer/connectors/lcdClient.js
+++ b/app/src/renderer/connectors/lcdClient.js
@@ -33,14 +33,21 @@ class Client {
 
   async request(method, path, data) {
     try {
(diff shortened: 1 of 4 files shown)


NodeGuy:
This is taking a bigger investment into learning the SDK than I thought it would.  I created a Pact test on both sides (Voyager & SDK) for `/stake/validators`.  The other end points will require more learning on my part.

Since this was started to support https://github.com/cosmos/voyager/issues/1010, which is now done, I'm deferring the rest of this until a time in the future when it makes sense for me to continue learning about the SDK.

PR #1097, opened 2018-08-08
Billy/1073 bar bond

### LINKED ISSUE #1073: Bond Bar out of margin
UI Version: 0.9.3

Description:

Disable or hide bond bar if the user is revoked.
See picture below.

screen shot 2018-08-02 at 1 26 22 pm

okwme:
Closes #1073 

Description: added more checks since it's possible to type in numbers outside of the input range and removed the draggable handle.


app/src/renderer/components/staking/PageBond.vue
```diff
@@ -125,7 +126,6 @@ tm-page.page-bond(:title="`Bond ${denom}`")
 import { between, numeric, required } from "vuelidate/lib/validators"
 import { mapGetters } from "vuex"
 import num from "scripts/num"
-import interact from "interactjs"
```
faboweb: do we still need the interactjs dependency?

okwme: good call! i'll remove

app/src/renderer/components/staking/PageBond.vue
```diff
@@ -286,6 +286,7 @@ export default {
     bondBarPercent(dividend) {
       let divisor = this.totalAtoms
       let ratio = Math.round(dividend / divisor * 100)
+      if (isNaN(ratio)) ratio = 0
```
NodeGuy: This is OK but next time it's better to write like this:

```
if (isNan(ratio)) { ratio = 0 }
```

That calls out more clearly that there's a side effect being caused.

app/src/renderer/components/staking/PageBond.vue
```diff
@@ -294,6 +295,7 @@ export default {
       let divisor = this.totalAtoms
       let ratio = Math.round(dividend / divisor * 100) / 100
       let width = ratio * (maxWidth - offset) + offset
+      if (width > maxWidth) width = maxWidth
```
NodeGuy: Same as above.

app/src/renderer/components/staking/PageBond.vue
okwme: 👍

app/src/renderer/components/staking/PageBond.vue
okwme: 👍

PR #1101, opened 2018-08-08
Billy/1096 votes sorting

### LINKED ISSUE #1096: Your Votes not sorting
UI Version: 0.9.3
Description: on staking page your votes doesn't sort

okwme:
Closes #1096 

Description: Sorting by "your votes" didn't work. This is because they are added after the sorting mechanism occurs. Actually they are the only thing that occurs in `enrichedAndFilteredData` as opposed to `filteredDelegates`. 

I couldn't see a reason why they should be calculated in their own computed value. If I'm missing something we'll need another solution.

### DIFF
--- a/app/src/renderer/components/staking/PageStaking.vue
+++ b/app/src/renderer/components/staking/PageStaking.vue
@@ -84,12 +84,6 @@ export default {
         v.percent_of_vote = num.percent(v.voting_power / this.vpTotal)
         v.your_vote = this.num.prettyInt(this.committedDelegations[v.id])
       })
-      console.log(
-        this.delegates.delegates,
-        [this.sort.property, "small_moniker"],
-        [this.sort.order, "asc"]
-      )
-      console.log(this.delegates.delegates)
       let delegates = orderBy(
         this.delegates.delegates,
         [this.sort.property, "small_moniker"],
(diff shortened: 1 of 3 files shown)


okwme:
can't grog these e2e failures...
unhandled error from a file called api.js in one of the first setup tasks.
can't have anything to do with the files I changed.
Also getting chromedriver errors locally when i try to run e2es

faboweb:
There is probably an error while starting Voyager (in the main thread). Therefor Electron is not started and can't be found by ChromeDriver.

app/src/renderer/components/staking/PageStaking.vue
```diff
@@ -76,19 +76,13 @@ export default {
...
-        })
-      )
-    },
     filteredDelegates() {
       let query = this.filters.delegates.search.query || ""
 
       forEach(this.delegates.delegates, v => {
         v.small_moniker = v.moniker.toLowerCase()
         v.percent_of_vote = num.percent(v.voting_power / this.vpTotal)
+        v.your_vote = this.num.prettyInt(this.committedDelegations[v.id])
```
faboweb: you are mutating the delegates object at this place. We don't know if this could cause side effects. How about a computed property `enrichedDelegates` similar to `enrichedAndFilteredDelegates` you then can use in `filteredDelegates`?

okwme:
😱
how could i have?!?
hopefully this is a bit better

faboweb:
Naming got very long ^^ but let's go with it.

okwme:
sortedFilteredEnrichedReviewedApprovedMergedDelegates 😜

PR #1122, opened 2018-08-13
Billy/1072 min max

### LINKED ISSUE #1072: Add min and max values to input number
UI Version: 0.9.3

Description:

Input type number should have:
min: 0 
max: User balance

I'd keep the error message just in case.

screen shot 2018-08-02 at 1 31 49 pm

okwme:
Closes #1072 

Description: Made changed to tendermint/ui repo in order to fix this. Need to include it as github reference in the package.json until tendrmint/ui is updated on npmjs.com


package.json
```diff
@@ -97,7 +97,7 @@
     "webpack-dev-server": "2.11.2"
   },
   "dependencies": {
-    "@tendermint/ui": "0.2.13",
+    "@tendermint/ui": "https://github.com/tendermint/ui.git",
```
faboweb: Should we go with this? Or should we wait with this PR until a release?

okwme: @nylira could we update the npmjs package?

nylira: Just published 0.2.15!

okwme:
🎉

PR #1129, opened 2018-08-14
Billy/1125 build local

### LINKED ISSUE #1125: automate build local network process
UI Version: 0.X.X

Description: so that if you run `yarn start local` and get `The network configuration for the network you want to connect to doesn't exist. Have you run `yarn build:testnets` to download the latest configurations?` it could prompt you to run `yarn build:local` or sth

okwme:
Closes #1125 

Description: adds a build command for making a local node when the testnet is down

❤️ Thank you!


faboweb:
Should we handle that case, this already exists:
image

Also I think we should update the README and add messages to `yarn start local` and `yarn start local-testnet`.

PR #1136, opened 2018-08-15
Fabo/1112 cache per network

### LINKED ISSUE #1112: Vuex store needs to be cached per network
UI Version: 0.9.3

Description:

Currently the Vuex store is cached in localStorage independent of which network is connected.  This means if the network changes (e.g., from gaia-7004 to gaia-7005) then the cached balances, etc., are carried over inappropriately from the old network.

We can use a storage key that incorporates `state.lastHeader.chain_id` but this may not yet be set by sign in so we should modify the [`signIn` mutator](https://github.com/cosmos/voyager/blob/9c1c014724812f62d84e2afe4b548d4859799509/app/src/renderer/vuex/modules/user.js#L104) to require the chain_id:

```JavaScript
async signIn({ state, commit, dispatch }, { password, account, chain_id }) {
```

faboweb:
Closes #1112 

Description:

❤️ Thank you!


NodeGuy:
Nice use of the user address.

PR #1139, opened 2018-08-15
WIP: Add shell script diagnostics to testnets build script.

### LINKED ISSUE #1126: Downloading testnet config script doesn't work
UI Version: 0.9.4

Description:

Running `yarn build:testnets` doesn't generate any testnet files in `./builds/testnets`

NodeGuy:
Closes #1126

Description:

❤️ Thank you!


faboweb:
I pushed a solution proposal. Can you verify this works? And maybe explain to me, why it works this way? ^^

NodeGuy:
That is SUPER mysterious!  We're running in Docker so shouldn't be getting a "works on my machine" error like this.  I can't think of any hypothesis that explains the problem.  Your workaround is not a good one and should be unnecessary.

Please try collaborating with someone local (Billy or Fede) and see if they notice anything, otherwise we can schedule a meeting to debug it together.

faboweb:
> Your workaround is not a good one and should be unnecessary.

Can you elaborate why this is a bad solution?

NodeGuy:
Closed in favor of https://github.com/cosmos/voyager/pull/1176.

PR #1144, opened 2018-08-16
Fabo/997 already subscribed to transactions

### LINKED ISSUE #997: error subscribing to transactions
UI Version: 0.7.1

Description:

Upon app loading:

screen shot 2018-07-20 at 7 17 15 pm

faboweb:
Closes #997 

Description:

We need to subscribe on two ways:
- after signing in
- after reconnecting to another node

On start sometimes both ways would try to subscribe. Doing a check now, if already subscribed to that rpc object.

❤️ Thank you!


test/unit/specs/store/wallet.spec.js
```diff
@@ -14,16 +14,7 @@ describe("Module: Wallet", () => {
...
-      balances: [],
-      balancesLoading: true,
-      history: [],
-      historyLoading: false,
-      denoms: [],
-      address: null,
-      zoneIds: ["basecoind-demo1", "basecoind-demo2"]
-    }
-    expect(store.state.wallet).toEqual(state)
+    expect(store.state.wallet).toMatchSnapshot()
```
okwme: this is cool

PR #1148, opened 2018-08-16
R4R: Fixed fractions in tokens

### LINKED ISSUE #1042: Validators with non-whole fractions of steaks showing incorrectly in staking window
UI Version: 0.9.1 Windows

Description:

Validators with non-whole fractions of steaks are showing the incorrect amount in the Staking window of Voyager.

image

Querying the above two validators in the Explorer shows that they should have 135.9 and 126.9 steaks respectively.

image

image

While this looks like the decimal place is not being handled correctly, another validator (#kingnet.com) is showing as 432 steak when according to the Explorer they have 86.4 (out by factor of 5 instead of 10).

The percent of vote is being calculated off the incorrect Total Votes, so I'm guessing this will be resolved once the votes is fixed.

Feel free to open Voyager v0.9.1 Windows on live testnet (gaia-7003) to view the same results.

fedekunze:
Fixes #1042 

Description:

❤️ Thank you!


app/src/renderer/components/staking/PageStaking.vue
```diff
@@ -1,7 +1,7 @@
 <template lang="pug">
 tm-page(title='Staking')
   div(slot="menu"): tool-bar
-    a(@click='connected && updateDelegates()' v-tooltip.bottom="'Refresh'" :disabled="!connected")
```
faboweb: bad resolve on a merge conflict?

okwme: what should it be? it looks like your disabled PR work @fabo but i don't see the error.

fedekunze: fixed @faboweb

app/src/renderer/vuex/modules/delegates.js
```diff
@@ -17,7 +17,16 @@ export default ({ node }) => {
       Object.assign(delegate, delegate.description)
 
       // TODO: calculate voting power
-      delegate.voting_power = delegate.tokens
+      let divIdx = delegate.tokens.indexOf("/")
```
faboweb: If there is still a '/' in the tokens, we are not converting them correctly in the LCD. :/ let's fix this over there.

test/unit/specs/components/staking/__snapshots__/LiDelegate.spec.js.snap
```diff
@@ -28,7 +28,7 @@ exports[`LiDelegate has the expected html structure 1`] = `
       class="li-delegate__value number_of_votes num bar"
     >
       <span>
-        14
+        14.00
```
faboweb: Are we sure on always showing fractions?

fedekunze: the goal is to show the most accurate way possible the amount of tokens that a candidate has. The explorer already shows fractions (specially in the ones that got slashed)

app/src/renderer/vuex/modules/delegates.js
fedekunze: you mean that this would require us to update the response from the SDK, right ?

fedekunze: @faboweb If so I'd merge this and open another issue for that in both the SDK and voyager

faboweb: Let's do that.

test/unit/specs/components/staking/__snapshots__/LiDelegate.spec.js.snap
faboweb: OK

app/src/renderer/vuex/modules/delegates.js
faboweb: Can you create the issues?

PR #1150, opened 2018-08-16
Billy/1105-staking-denom

### LINKED ISSUE #1105: "You do not have any ATOMs to delegate." confusing while using STEAK
UI Version: 0.9.3

Description:

I've been staking with STEAK but just got the error message, "You do not have any ATOMs to delegate."  The terminology is confusing.

okwme:
Closes #1105

Description: uses config.bondingDenom instead of "Atom" hardcoded

❤️ Thank you!


okwme:
oops the title of the PR?
it just took my last commit message...

okwme:
was getting e2e errors on my local version because of the onboarding popup, i remember we had this issue before which was supposed to be solved by upgrading electron but seems it doesn't happen on circle...

or was it something else i should change? @faboweb

faboweb:
Let's keep an eye out for the onboarding issue

Issue #856, opened 2018-06-15
Refactor main process


faboweb:
Scope:
- [ ] Design of the refactor
- [ ] Implemantation

NodeGuy:
Split out [refactor of Addressbook](https://github.com/cosmos/voyager/issues/1195).

NodeGuy:
Proposed design:

* remove all global variables

faboweb:
> move all code related to Gaia Lite into main/gaiaLite.js

created this already in the SDK 0.26 PR

faboweb:
There is none no more

PR #1167, opened 2018-08-21
WIP: David/856 refactor main by checking nodes for compatibility before connection

NodeGuy:
Improves #856.

Description:

The main change is that nodes are checked for version compatibility before connecting to one and that the checking has been moved from main.js into the addressbook.


CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 * changed ATOM to bondingDenom getter @okwme
 * more bondingDenom and copy updates @jbibla
 * restored loading animation on start @faboweb
+* Nodes are now checked for version compatibility before connection. @NodeGuy
```
jbibla: 👍

NodeGuy:
I need to figure out why the E2E tests are failing.  Running them three times in a row didn't work as usual.  ;-)

app/src/main/addressbook.js
```diff
@@ -1,50 +1,57 @@
...
-    let nodeAlive = await axios
-      .get(pingURL, { timeout: 3000 })
-      .then(() => true, () => false)
+
+    let nodeVersion
+
+    try {
+      nodeVersion = (await this.fetch(pingURL, {
+        timeout: 3000
+      })).data.result.node_info.version
```
faboweb: Isn't that the Tendermint version?

NodeGuy: Ah yes, I just learned more about how Tendermint works.  Thanks, I'll fix it.

faboweb: I think your approach might not work because of this. We get the SDK version via the LCD we need to start by connecting to a node. So we can get the SDK version only after having connected to a node.

NodeGuy: Yes, I agree.

NodeGuy:
Closing in favor of https://github.com/cosmos/voyager/issues/1195.

Issue #1169, opened 2018-08-21
Gaiacli version error makes Voyager crash


fedekunze:
UI Version: 0.X.X

Description:

I had the v0.24.0 installed of the SDK and I got an error when running `yarn start`. Also `yarn build:gaia` wasn't very helpful.

```
electron  gaiacli version: "0.24.0", expected: "0.23.1"

electron  [Unhandled Promise Rejection] Error: Requires gaia 0.23.1, but got 0.24.0.
                Please update your gaiacli installation or build with a newer binary.
              at _callee12$ (/Users/federico/Projects/voyager/app/src/main/index.js:654:11)
              at tryCatch (/Users/federico/Projects/voyager/node_modules/regenerator-runtime/runtime.js:62:40)
              at Generator.invoke [as _invoke] (/Users/federico/Projects/voyager/node_modules/regenerator-runtime/runtime.js:296:22)
              at Generator.prototype.(anonymous function) [as next] (/Users/federico/Projects/voyager/node_modules/regenerator-runtime/runtime.js:114:21)
              at step (/Users/federico/Projects/voyager/node_modules/babel-runtime/helpers/asyncToGenerator.js:17:30)
              at /Users/federico/Projects/voyager/node_modules/babel-runtime/helpers/asyncToGenerator.js:28:13
              at <anonymous>
              at process._tickDomainCallback (internal/process/next_tick.js:228:7)
```

faboweb:
Which network are you trying to connect to? Probably gaia-7005 which expects `0.23.1`. Maybe delete the config folder of that network under `~/.voyager-dev/`.

fedekunze:
```
federicos-mbp-2:voyager federico$ gaiacli version
0.23.1-23e3d5ac
```
Still happening

fedekunze:
It's probably because the 7005 is down

faboweb:
The Voyager Gaia binaries are in `{voyager-dir}/builds/gaia/darwin.../` try getting the version from that binary.

fedekunze:
The main issue that I had with this bug is that it didn't allow me to change to the local testnet setting. I had to quit Voyager and run the CLI with `yarn build:local`

fedekunze:
@faboweb : So `gaiacli version: "0.24.0", expected: "0.23.1"` has nothing to do with remote nodes. This just means `gaiacli version` is different then the content of the `gaiaversion.txt` that lives in `~/.voyager-dev/{NETWORK}`

[...]

PR #1187, opened 2018-08-28
Fabo/1169 improve version mismatch err

faboweb:
New message (2nd phrase is in development only):
image

❤️ Thank you!

### DIFF
--- a/app/src/main/index.js
+++ b/app/src/main/index.js
@@ -653,8 +653,13 @@ const checkGaiaCompatibility = async gaiacliVersionPath => {
     semver.minor(gaiacliVersion) == semver.minor(expectedGaiaCliVersion)
 
   if (!compatible) {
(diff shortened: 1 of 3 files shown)


app/src/main/index.js
```diff
@@ -739,7 +744,7 @@ async function main() {
     fs.writeFileSync(appVersionPath, pkg.version)
   }
 
-  checkGaiaCompatibility(gaiacliVersionPath)
+  await checkGaiaCompatibility(gaiacliVersionPath)
```
NodeGuy: Thanks, I missed that.

faboweb: Finding on that: if you call async functions without await, errors are not passed to the wrapping function. ergo they never surface. :/

NodeGuy: Yes, I know.  :-(

fedekunze:
Closes #1169

PR #1197, opened 2018-08-28
Create Git tag during publish instead of during the release pull requ…

### LINKED ISSUE #1165: Tags should be generated in tasks/publish.js instead of tasks/releasePullRequest.js
UI Version: 0.9.4

Description:

Tags shouldn't be pushed to GitHub until the release is published.

NodeGuy:
…est.

Closes #1165 

Description:

Creates a Git tag during publish instead of when creating the release pull request.  This will be fully tested only when we publish the next release.


NodeGuy:
I just noticed that this PR contains a lint error.  How did it pass the tests?  I don't see any record on CircleCI.

NodeGuy:
Done: https://github.com/cosmos/voyager/issues/1217

I was confused because `yarn test` tests for it but I see now that we don't use that on CircleCI.

PR #1208, opened 2018-08-29
Fabo/1059 test exe

### LINKED ISSUE #1059: Create test if build Voyager starts
UI Version: 0.9.2

Description:

It may happen that production builds behave different then development builds. We might miss some configuration. We should not be releasing bad builds. Therefor we should test in the CI if a build Voyager actually starts.

Plan:
Voyager outputs `[START SUCCESS]` to console if it can start the view thread without errors. In the CI we create a check that starts the actual build Linux version and listens for the success message on the console. This check is already implemented with `yarn test:exe`.
This is not a test for all build versions for all OSs. Testing this is currently not possible and we need to defer this until we have a proper multi OS CI setup.

faboweb:
Closes #1059

Description:

❤️ Thank you!


faboweb:
Tested here: https://circleci.com/gh/cosmos/voyager/6719

app/networks
```diff
@@ -1 +1 @@
-../builds/testnets
\ No newline at end of file
+/Users/fabo/Development/voyager/builds/testnets
```
fedekunze: Double checking if `/Users/fabo/` is ok or not

faboweb: XD thx

tasks/publish.js
```diff
@@ -36,8 +35,6 @@ const publishRelease = ({ notes, tag, token }) =>
   })
 
 async function main() {
-  await build.buildAllPlatforms()
```
NodeGuy: How is this program going to publish a release on GitHub it's not longer building it?

tasks/releasePullRequest.js
```diff
@@ -22,7 +22,7 @@ function updateChangeLog(changeLog, newVersion, now) {
 const updatePackageJson = (packageJson, version) =>
   Object.assign({}, packageJson, { version })
 
-const pushCommit = async ({ token, tag, head }) => {
```
NodeGuy: That should have been caught earlier on CircleCI by the linter; weird.

tasks/publish.js
faboweb: As of: https://github.com/cosmos/voyager/issues/1059#issuecomment-416956483

NodeGuy: I see.

Issue #1211, opened 2018-08-29
TypeError: Cannot read property 'keybase' of undefined


faboweb:
https://sentry.io/tendermint/cosmos-voyager/issues/667195282/

```
TypeError: Cannot read property 'keybase' of undefined
  at Store.setKeybaseIdentity (/private/var/folders/m9/14ny26h57vj1kt0_twc07ydc0000gn/T/AppTranslocation/472A41FE-40CF-4336-B823-C9C428E4EAD8/d/Cosmos%20Voyager.app/Contents/Resources/app/dist/renderer.js:106905:22)
  at wrappedMutationHandler (/private/var/folders/m9/14ny26h57vj1kt0_twc07ydc0000gn/T/AppTranslocation/472A41FE-40CF-4336-B823-C9C428E4EAD8/d/Cosmos%20Voyager.app/Contents/Resources/app/dist/renderer.js:6646:13)
  at commitIterator (/private/var/folders/m9/14ny26h57vj1kt0_twc07ydc0000gn/T/AppTranslocation/472A41FE-40CF-4336-B823-C9C428E4EAD8/d/Cosmos%20Voyager.app/Contents/Resources/app/dist/renderer.js:6339:7)
  at Array.forEach (<anonymous>)
  at None (/private/var/folders/m9/14ny26h57vj1kt0_twc07ydc0000gn/T/AppTranslocation/472A41FE-40CF-4336-B823-C9C428E4EAD8/d/Cosmos%20Voyager.app/Contents/Resources/app/dist/renderer.js:6338:11)
...
(10 additional frame(s) were not displayed)
```

NodeGuy:
What are the steps to reproduce this?

faboweb:
Check Sentry. I have no idea. But it was reported by our error logger.

faboweb:
Just happened to me. Apparently happens when you sign out while the keybase data is still loading. When signing out the delegate list is cleared. So it can't find the delegate anymore, it wanted to load the keybase data for.

faboweb:
Will be fixed with #1216

sabau:
![screenshot 2019-03-01 at 12 59 47](https://user-images.githubusercontent.com/1656471/53636930-0b70b080-3c22-11e9-8cd4-e8d6a819cf39.png)

Issue #1080, opened 2018-08-06
Error fetching validator set


NodeGuy:
UI Version: 0.9.3

Description:

Before I sign in, the error message "Error fetching validator set: Internal error" is displayed.

<img width="899" alt="screen shot 2018-08-06 at 12 14 06 pm" src="https://user-images.githubusercontent.com/2148537/43728237-6feaa1f4-9972-11e8-8923-2b7a05b61758.png">

I don't see anything significant in either console.

jbibla:
it's odd that we are making requests before the user is signed in

nylira:
Can we turn session modals into pages in `vue-router` instead of a z-index overlay on top of the wallet?

faboweb:
@nylira can you create an issue recommending this change?

faboweb:
I think this is a strange typo issue `/stake/validators/` vs `/stake/validators` and I experienced it as well in #1068 . It will hopefully be resolved after this is merged.

faboweb:
Assumed fixed

faboweb:
Still an issue

NodeGuy:
What's blocking it?

faboweb:
the blocking label was still on there

NodeGuy:
I can't reproduce it.  @faboweb are you able to reproduce it consistently?

NodeGuy:
So what's the desired behavior?  Should `getValidators` be called only when connected to a node or should we suppress the error message?

faboweb:
Both seem to be valid options. Maybe we need to defer REST calls a little bit, so we are sure the connected node is working and we wont reconnect. This would solve the root cause.

faboweb:
I handle the case "skipping it" in #1216

fedekunze:
Created #1313 to turn the session modals to pages

faboweb:
The rpc subscription starts before the REST client is up. The headers come in and trigger a load of the validators before the the REST server is up. This creates this issue.

Current patch: Do not load data before the user is not signed in (which should be after the REST server is up).

PR #1216, opened 2018-08-30
Fabo/1199 improve jank at start

### LINKED ISSUE #1199: Jank while loading validator profiles
UI Version: 0.9.4

Description:

When starting the applications the app freezes for some seconds. I assume this is while downloading the validator profiles.

Chrome performance profile:
[Profile-20180829T120817.json.zip](https://github.com/cosmos/voyager/files/2331719/Profile-20180829T120817.json.zip)

faboweb:
Closes #1199, #1211, #1080 

Also fixes a bug on the PageBond where the atoms would not have been updated after staking (optimistic update).

❤️ Thank you!


app/src/renderer/vuex/modules/wallet.js
```diff
@@ -100,7 +100,7 @@ export default ({ node }) => {
         transactions.map(async t => {
           let blockMetaInfo = await dispatch("queryBlockInfo", t.height)
           t.time = blockMetaInfo && blockMetaInfo.header.time
-          return transactions
+          return t
```
jbibla: i would vote for `transactions` over `t`

faboweb: Adding: I would try to avoid refactoring existing untouched code in PRs so we do not bloat the PRs.

jbibla: ah i see - ok makes sense. thank you for clarifying.

app/src/renderer/components/staking/PageBond.vue
```diff
@@ -38,8 +38,8 @@ tm-page.page-bond(title="Staking")
       :class="bondGroupClass(delta(d.atoms, d.oldAtoms))")
       .bond-group__fields
         .bond-bar
-          label.bond-bar__label(v-if="!d.delegate.revoked") {{ d.delegate.moniker }}
-          label.bond-bar__label.revoked(v-if="d.delegate.revoked") {{ d.delegate.moniker }}
+          label.bond-bar__label(v-if="!d.delegate.revoked") {{ d.delegate.description.moniker }}
```
faboweb: fixed monikers not showing

app/src/renderer/vuex/modules/node.js
```diff
@@ -44,7 +44,11 @@ export default function({ node }) {
         rootState.wallet.zoneIds.unshift(header.chain_id)
       }
 
-      await dispatch("maybeUpdateValidators", header)
+      // updating the header is done even while the user is not logged in
```
jbibla: why do we do this?

test/unit/specs/components/staking/PageBond.spec.js
```diff
@@ -34,23 +36,24 @@ describe("PageBond", () => {
...
-          Object.assign(
-            {
-              id: "pubkeyZ",
-              voting_power: 20000,
-              shares: 75000,
-              moniker: "aChileanValidator"
-            },
-            candidates[2] // this is the revoked one
-          )
+        let chileanValidator = Object.assign(
```
jbibla: 🇨🇱

test/unit/specs/components/staking/PageBond.spec.js
```diff
@@ -34,23 +36,24 @@ describe("PageBond", () => {
           },
           voting_power: 30000,
           shares: 10000,
-          description: "descriptionY",
-          country: "Canada",
-          moniker: "someOtherValidator"
+          description: {
+            description: "descriptionY",
+            country: "Canada",
```
jbibla: 🇨🇦

app/src/renderer/vuex/modules/node.js
faboweb: yes, we bind to the "new header" events before signing in. the event would then trigger an update of the validatorset which is skipped here. This will also change when we react to the validatorset updated event.

faboweb:
> side note: i would LOVE if we refactored just the language because delegates and delegators feels very old and unfamiliar. what is committedDelegates?

Love the idea. Can you create an issue for that?

> not sure i understand why / how separating keybase out from the validator getter is a good thing but i believe you.

keybase is a completely different ressource then where we get our validators from. so I thought we should split these concerns.

> also not sure i understand why the atomDiff needs to be calculated the way it does.

how would you do it? maybe my approach was too much

faboweb:
> side note: i would LOVE if we refactored just the language because delegates and delegators feels very old and unfamiliar. what is committedDelegates?

Love the idea. Can you create an issue for that?

> not sure i understand why / how separating keybase out from the validator getter is a good thing but i believe you.

keybase is a completely different ressource then where we get our validators from. so I thought we should split these concerns.

> also not sure i understand why the atomDiff needs to be calculated the way it does.

how would you do it? maybe my approach was too much

jbibla:
done https://github.com/cosmos/voyager/issues/1283

PR #1223, opened 2018-08-31
Billy/1191 tx designs

### LINKED ISSUE #1191: Add new transaction designs
We want the new design of the transactions to appear in the tx history list of Voyager. This probably means using the https://github.com/tendermint/ui/tree/develop/src/components/TmLiAnyTransaction component.

okwme:
Closes #1191 

Description: Added the new transactions from tendermint/ui (updated it to fix a moniker bug and layout error: https://github.com/tendermint/ui/pull/63/commits/dd88c5d141a19e8b89771499405f64ffcf0787be)

Refactored `modules/blockchain.js` and added `modules/transactions.js` for all current transaction info and ready for future transaction formats.


app/src/renderer/vuex/modules/transactions.js
```diff
@@ -0,0 +1,123 @@
...
+        default:
+          throw new Error("Unknown transaction type")
+      }
+      if (!res) return []
+      res = res.map(r => {
+        r.type = type
+        return r
+      })
+      return uniqBy(res, "hash")
+    },
```
NodeGuy: See https://www.nodeguy.com/p/87512fce-e9ec-4402-8f18-7e914d93a3d7/

```JavaScript
const fp = require(`lodash/fp`)

async getTx(
  {
    rootState: {
      user: { address }
    }
  },
  type
) {
  let response

  switch (type) {
    case "staking":
      response = await node.getDelegatorTxs(address)
      break
    case "wallet":
      response = await node.txs(address)
      break
    default:
      throw new Error("Unknown transaction type")
  }

  return response ? uniqBy(response.map(fp.set(`type`, type)), "hash") : []
}
```

app/src/renderer/vuex/modules/transactions.js
```diff
@@ -0,0 +1,123 @@
...
+        if (!blockHeights.find(h => h === t.height)) {
+          blockHeights.push(t.height)
+        }
+      })
+      await Promise.all(
+        blockHeights.map(blockHeight =>
+          dispatch("queryTransactionTime", { blockHeight, scopes })
+        )
+      )
+    },
```
NodeGuy:
```JavaScript
async enrichTransactions({ dispatch }, { transactions, scopes = [] }) {
  const blockHeights = new Set(transactions.map(({ height }) => height))

  await Promise.all(
    [...blockHeights].map(blockHeight =>
      dispatch("queryTransactionTime", { blockHeight, scopes })
    )
  )
},
```

test/unit/specs/store/blockchain.spec.js
```diff
@@ -11,14 +11,14 @@ describe("Module: Blockchain", () => {
...
   }
-  const block = {
-    test: "test",
-    height: 42,
-    time: "Good Morning",
-    data: {
-      txs: []
-    }
-  }
+  // const block = {
```
fedekunze: delete code commented

app/src/renderer/components/wallet/PageTransactions.vue
```diff
@@ -49,28 +42,21 @@ export default {
...
       "config",
       "delegation",
-      "connected"
+      "connected",
+      "validators"
     ]),
-    somethingToSearch() {
-      return !this.wallet.historyLoading && !!this.allTransactions.length
+    validatorURL() {
+      return "https://explorecosmos.network/validators"
```
faboweb: The link should be to the validators page in Voyager I think. Do you think linking to an outside page makes more sense?

app/src/renderer/vuex/modules/blockchain.js
```diff
@@ -164,27 +168,27 @@ export default ({ node }) => {
...
 
         // only subscribe if the node is not catching up anymore
         node.rpc.subscribe({ query: "tm.event = 'NewBlock'" }, (err, event) => {
-          state.subscription = true
+          commit("setSubscription", true)
 
           if (err) return error(err)
-
-          state.blocks.unshift(event.data.value.block)
+          commit("unshiftBlock", event.data.value.block)
```
faboweb: Do we still need blocks? We don't have a block explorer anymore.

app/src/renderer/vuex/modules/transactions.js
```diff
@@ -0,0 +1,123 @@
+let { uniqBy } = require("lodash")
```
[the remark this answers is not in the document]
faboweb: We didn't define a rule for this.

app/src/renderer/vuex/modules/transactions.js
faboweb: short and functional, but I had my problems reading this. how about making it a bit more descriptive:

```
const enrichedTransactions = response.map(fp.set(`type`, type))
return response ? uniqBy(enrichedTransactions, "hash") : []
```

Apart from that: Maybe let's avoid demanding refactors of copy pasted code as it bloats PRs.

okwme: woa i didn't realize you could set default arguments as object children!

faboweb:
I agree with removing the dead code. It bloats the files and we can get the code from history again. What do you think?

app/src/renderer/components/wallet/PageTransactions.vue
okwme: when i pass a `validatorURL` like `/staking/validators/` the way our voyager routing works, the `tendermint/ui` component forces the link into a new window as if it were `target="_blank"`. Probably fixed by using `vue-router` in the subcomponent but should we be using `vue-router` there?

app/src/renderer/components/wallet/PageTransactions.vue
```diff
@@ -85,7 +82,8 @@ export default {
     sort: {
       property: "height",
       order: "desc"
-    }
+    },
+    validatorURL: "https://explorecosmos.network/validators"
```
faboweb: How about `${window.location.hostname}/stake/validators`? Will this also open a new window?

okwme:
@fabo sounds good to me
thanks for all the js tips @NodeGuy !

this is still blocked by `tendermint/ui` PRs and npmjs.com updates

also unresolved how to handle the validator link. Should we just use `vue-router` in the component library?

okwme:
sorry can you elaborate?

okwme:
yup, they both go to `
http://localhost:9080/localhost/staking/validators/{address}`

app/src/renderer/components/wallet/PageTransactions.vue
okwme: yup, they both go to `http://localhost:9080/localhost/staking/validators/{address}`

faboweb: OK. let's go with this version first and then switch to vue-router later?

faboweb:
> Should we just use vue-router in the component library?

I would vote yet. The explorer also uses the vue-router.

app/src/renderer/vuex/modules/transactions.js
NodeGuy: @faboweb I like your improvement.

> Apart from that: Maybe let's avoid demanding refactors of copy pasted code as it bloats PRs.

I don't want to demand refactors (I'm still figuring out how to use GitHub properly here) but I don't see how the source of the code makes any difference—it should be improved regardless.

> woa i didn't realize you could set default arguments as object children!

This is [object destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Object_destructuring).

app/src/renderer/vuex/modules/transactions.js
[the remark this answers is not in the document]
NodeGuy: True, in my mind we're waiting for https://github.com/cosmos/voyager/issues/1051 because of Matt's comment.

app/src/renderer/vuex/modules/transactions.js
faboweb: @NodeGuy valid point. I add an agenda point for tomorrows meeting to discuss a strategy.

faboweb:
Follow up #1275

faboweb:
Unbonding delegations are not shown correctly -> blocked on https://github.com/tendermint/ui/pull/64

okwme:
screenshot 2018-09-11 18 29 55

PR #1267, opened 2018-09-03
R4R: Skip non indexing nodes

### LINKED ISSUE #988: Identify and avoid connecting to non-indexing nodes
UI Version: 0.7.1

Description:

Usually nodes are not indexing transactions. This results in falsely txs not showing.

Tendermint shows us if a node is indexing via the status endpoint and the property `node_info.other.tx_index`

fedekunze:
Closes #988 

Description:

Skips non indexing nodes when discovering peers

❤️ Thank you!

### DIFF
--- a/app/src/main/addressbook.js
+++ b/app/src/main/addressbook.js
@@ -64,6 +64,11 @@ module.exports = class Addressbook {
     })
   }
 
+  isIndexingNode(peer) {
+    let tx_index = peer.node_info.other[4].split("=")[1]
+    return tx_index === "on"
+  }
+
   loadFromDisc() {
     // if there is no address book file yet, there are no peers stored yet
     // the file will be created when persisting any peers to disc
@@ -139,8 +144,12 @@ module.exports = class Addressbook {
     let subPeers = (await axios.get(
       `http://${peerIP}:${this.config.default_tendermint_port}/net_info`
     )).data.result.peers
-    let subPeersHostnames = subPeers.map(peer => peer.node_info.listen_addr)
-
+    let subPeersHostnames = subPeers.map(peer => {
+      // skip non-indexing nodes
+      if (this.isIndexingNode(peer)) {
+        return peer.node_info.listen_addr
+      }
+    })
     subPeersHostnames
       // check if we already know the peer
       .filter(subPeerHostname => !this.peerIsKnown(subPeerHostname))
--- a/app/src/main/addressbook.js
+++ b/app/src/main/addressbook.js
@@ -65,7 +65,7 @@ module.exports = class Addressbook {
   }
 
   isIndexingNode(peer) {
-    let tx_index = peer.node_info.other[4].split("=")[1]
+    let tx_index = peer.node_info.other && peer.node_info.other[4].split("=")[1]
     return tx_index === "on"
   }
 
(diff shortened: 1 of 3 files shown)


faboweb:
As discussed here https://github.com/cosmos/voyager/issues/988#issuecomment-416896199 this flag doesn't serve it's purpose.

app/src/main/addressbook.js
```diff
@@ -139,7 +144,11 @@ module.exports = class Addressbook {
     let subPeers = (await axios.get(
       `http://${peerIP}:${this.config.default_tendermint_port}/net_info`
     )).data.result.peers
-    let subPeersHostnames = subPeers.map(peer => peer.node_info.listen_addr)
+
+    let subPeersHostnames = subPeers
```
faboweb: this is only excluding non indexing nodes in the discovery phase. the flag could skip in the meantime until we connect to that node. I suggest we check the flag when we are trying to connect to the node.

fedekunze:
@faboweb I agree, but we the decision of skipping non indexing nodes is independent of the bug on the SDK

app/src/main/addressbook.js
fedekunze: you mean when adding the peer ?

faboweb:
Ok, sure we can go that way. Did you test if this works if this works for a local nodes that has `index_all_tags = false` in `config.toml`?

fedekunze:
@faboweb No. I forgot about that. I'll update the PR accordingly

app/src/main/addressbook.js
faboweb: No, when connecting to a peer. Probably here https://github.com/cosmos/voyager/blob/35fa2baa86e9943097d43b5507616273bffeb2a1/app/src/main/addressbook.js#L98

fedekunze: `this.peers` is just the `hostname` and a `state`, it doesn't show the full peer `node_info`. Maybe we should consider adding a new `boolean` attribute called `indexing` to each `peer` Object and pass it to the `addPeer` function

fedekunze: So this are the options:
1) leave the `isIndexing` filter on the `discoverPeers` function
2) change the whole Addressbook logic to be able to filter peers on the `pickNode` function

faboweb: As I wrote before: the indexing behavior could change in the time between adding the peer and connecting to the peer.

1) therefor is not an option

would you disagree?

test/unit/helpers/node_mock.js
```diff
@@ -42,14 +42,152 @@ module.exports = {
     on: () => {},
     subscribe: () => {},
     validators: cb => cb(null, { validators: mockValidators }),
-    block: (args, cb) => cb({}),
-    blockchain: (args, cb) => cb(null, { block_metas: {} }),
+    block: (args, cb) =>
```
faboweb: this is out of scope of this PR I would say. :/

test/unit/specs/addressbook.spec.js
```diff
@@ -217,4 +280,15 @@ describe("Addressbook", () => {
       .then(done.fail)
       .catch(() => done())
   })
+
+  it("should flag nodes that are not indexing transactions", async done => {
```
faboweb: as discussed, I think this is not the right way to go. the flag could change after we have discovered the peer the first time.

test/unit/specs/addressbook.spec.js
```diff
@@ -12,8 +12,78 @@ describe("Addressbook", () => {
     Addressbook = require("src/main/addressbook.js")
 
     jest.mock("axios", () => ({
-      get: async () => {
-        return { data: { result: { peers: [] } } }
+      getNetInfo: async () => {
```
faboweb: imo a lot of clutter here, that is not interesting for the test. what do you think?

PR #1286, opened 2018-09-05
WIP: David/1195 refactor addressbook

### LINKED ISSUE #1195: refactor Addressbook proposal
The main process [is horribly complex](https://github.com/cosmos/voyager/issues/856).

Addressbook is a class in the main process that discovers and keeps a stateful collection of node addresses.  A node is chosen randomly from the pool when starting Gaia Lite.

The usage of the address book is essentially as follows:

1.  Load seed nodes from the network's `config.toml` file.
2.  Load more nodes that were saved to disk during the previous session.
3.  Filter out nodes that don't respond to a simple query.
4.  Discover new nodes by asking a node for its peers.
5.  Save nodes to disk for the next session.

I propose that we replace this complex, stateful class with a single stateless function.  The function will:

* take a set of seed nodes, discover new ones, and return a combined set of nodes known to respond
* combine the aliveness query with peer discovery to reduce the amount of network traffic
* make the network queries in parallel for increased performance
* not talk to the disk to eliminate the need to mock file system calls in tests

NodeGuy:
Closes #1195 

Description:

TDB


NodeGuy:
Closing in favor of #1291.

Issue #1219, opened 2018-08-30
Delegation / Redelegation


NodeGuy:
Jordan, please link your design files here.

jbibla:
https://zpl.io/bWL0krn - Staking
https://zpl.io/VxlvdWk - Restaking

NodeGuy:
Please remove issues from the backlog when moving them into Scrumptious.

NodeGuy:
Hold on cowboys, this is dependent on #1171 which is also part of this sprint.

jbibla:
it doesn't need to be dependent on #1171 - we can add a button to the existing profile page and work on the new modal ...

faboweb:
I agree with Jordan. I would see this as a replacement for PageBond. We can add the 2 buttons to the old PageValidator for now and replace the checkboxes (and shoppingCart logic) already.

jbibla:
yes, let's please get rid of `shoppingCart`.

🛒 🛒 🛒

NodeGuy:
OK

NodeGuy:
After looking at the designs it seems this isn't about redelegation—it's about delegation and undelegation, right?  I'm going to rename the title.

jbibla:
we have to account for (`delegation / staking / bonding`), (`undelegation / unstaking / unbonding`), (`redelegation / restaking / rebonding`).

faboweb:
Follow up for redelegation: https://github.com/cosmos/voyager/issues/1293

NodeGuy:
Our standard modals have a darker background for the header:

<img width="581" alt="screen shot 2018-09-06 at 3 37 52 pm" src="https://user-images.githubusercontent.com/2148537/45180976-0985e180-b1eb-11e8-9c6d-67e80f12df42.png">

There's no such distinction in the design.  Do you care or should I reimplement `TmModal` for this one?

NodeGuy:
Split out #1294 and #1295.

faboweb:
the design for staking is actually not the same modal we have already. It is more a pop-over on the bottom like when you write a new email in google inbox. maybe it should still have the same layout. @jbibla can you help out?

NodeGuy:
Oh, OK.  I see another difference: With `TmModal` the rest of the window is darkened but that's not happening here.  I won't use it.

fedekunze:
> close window when clicking outside of it

@NodeGuy do we need this point if we are already adding the "Close" and `X` buttons ?

[...]

PR #1290, opened 2018-09-05
new delegation window

NodeGuy:
Make progress against #1219.

Description:

Implements the new design for staking to a validator.  It does not remove the old one.

Still to be done:

## Jordan

- [ ] add `X` and `Cancel` buttons to design

## David & Fede

- [ ] write "Atoms" within `Amount` field
- [ ] allow selection of previously-staked validators in `From` field
- [ ] close window when clicking outside of it
- [ ] don't change layout when invalid amount error message is displayed

❤️ Thank you!

### DIFF
--- a/app/src/renderer/components/staking/ModalStake.vue
+++ b/app/src/renderer/components/staking/ModalStake.vue
@@ -1,0 +1,112 @@
+<template lang="pug">
+  .modal-stake
+    .stake-header
+      img.icon(class='stake-atom' src="~assets/images/cosmos-logo.png")
+      span.tm-modal-title Stake
+
+    tm-form-group.stake-form-group(
+      :error='$v.amount.$invalid'
+      field-id='amount'
+      field-label='Amount'
+    )
+      tm-field#amount(
+        type="number"
+        v-model="amount")
+      tm-form-msg(name='Amount' type='between' :min='1' :max='maximum'
+        v-if='$v.amount.$invalid')
+
+    tm-form-group.stake-form-group(
+      field-id='to' field-label='To')
+      tm-field#to(v-model="to")
+
+    tm-form-group.stake-form-group(
+      field-id='from' field-label='From')
+      tm-field#from(
+        type="select"
+        :options="fromOptions"
+        v-model="fromIndex"
+      )
+
+    .stake-footer
+      tm-btn(
+        @click.native="onStake"
+        color="primary"
+        :disabled="$v.amount.$invalid"
+        size="lg"
+        value="Stake"
+      )
+</template>
+
+<script>
+import { between } from "vuelidate/lib/validators"
+import Modal from "common/TmModal"
+import { TmBtn, TmField, TmFormGroup, TmFormMsg } from "@tendermint/ui"
+
+export default {
+  props: [`fromOptions`, `maximum`, `to`],
+  components: {
+    Modal,
+    TmBtn,
+    TmField,
+    TmFormGroup,
+    TmFormMsg
+  },
+  data: () => ({
+    amount: 1,
+    fromIndex: 0
+  }),
+  validations() {
+    return {
+      amount: {
+        between: between(1, this.maximum)
+      }
+    }
+  },
+  methods: {
+    close() {
+      this.$emit("update:showModalStake", false)
+    },
+    onStake() {
+      this.$emit(`stake`, {
+        amount: this.amount,
+        from: this.fromOptions[this.fromIndex]
+      })
+
+      this.close()
+    }
+  }
+}
+</script>
+
+<style lang="stylus">
+@import '~variables'
+
+.modal-stake
+  background: var(--app-nav)
+  display: flex
+  flex-direction: column
+  height: 50%
+  justify-content: space-between
+  left: 50%
+  padding: 3%
+  position: fixed
+  top: 50%
+  width: 40%
+  z-index: z(modal)
+
+  .stake-header
+    align-items: center
+    display: flex
+
+    .stake-atom
+      height: 3em
+      width: 3em
+
+  .stake-form-group
+    display: block
+    padding: 0
+
+  .stake-footer
+    display: flex
+    justify-content: flex-end
+</style>
--- a/app/src/renderer/components/staking/ModalStake.vue
+++ b/app/src/renderer/components/staking/ModalStake.vue
@@ -1,24 +1,28 @@
 <template lang="pug">
   .modal-stake
+    //- Header
     .stake-header
       img.icon(class='stake-atom' src="~assets/images/cosmos-logo.png")
       span.tm-modal-title Stake
 
+    //- Amount
     tm-form-group.stake-form-group(
-      :error='$v.amount.$invalid'
       field-id='amount'
       field-label='Amount'
     )
       tm-field#amount(
+        :max="maximum"
+        :min="1"
         type="number"
+        v-focus
         v-model="amount")
-      tm-form-msg(name='Amount' type='between' :min='1' :max='maximum'
-        v-if='$v.amount.$invalid')
 
+    //- To
     tm-form-group.stake-form-group(
       field-id='to' field-label='To')
-      tm-field#to(v-model="to")
+      tm-field#to(readonly v-model="to")
 
+    //- From
     tm-form-group.stake-form-group(
       field-id='from' field-label='From')
       tm-field#from(
@@ -27,18 +31,17 @@
         v-model="fromIndex"
       )
 
+    //- Footer
     .stake-footer
       tm-btn(
         @click.native="onStake"
         color="primary"
-        :disabled="$v.amount.$invalid"
         size="lg"
         value="Stake"
       )
 </template>
 
 <script>
-import { between } from "vuelidate/lib/validators"
 import Modal from "common/TmModal"
 import { TmBtn, TmField, TmFormGroup, TmFormMsg } from "@tendermint/ui"
 
@@ -55,19 +58,12 @@ export default {
     amount: 1,
     fromIndex: 0
   }),
-  validations() {
-    return {
-      amount: {
-        between: between(1, this.maximum)
-      }
-    }
-  },
   methods: {
     close() {
       this.$emit("update:showModalStake", false)
     },
     onStake() {
-      this.$emit(`stake`, {
+      this.$emit(`submitDelegation`, {
         amount: this.amount,
         from: this.fromOptions[this.fromIndex]
       })
@@ -82,31 +78,31 @@ export default {
 @import '~variables'
 
 .modal-stake
-  background: var(--app-nav)
-  display: flex
-  flex-direction: column
-  height: 50%
-  justify-content: space-between
-  left: 50%
-  position: fixed
-  top: 50%
-  width: 40%
-  z-index: z(modal)
+  background var(--app-nav)
+  display flex
+  flex-direction column
+  height 50%
+  justify-content space-between
+  left 50%
   padding 2em
+  position fixed
+  top 50%
+  width 40%
+  z-index z(modal)
 
   .stake-header
-    align-items: center
-    display: flex
+    align-items center
+    display flex
 
     .stake-atom
-      height: 3em
-      width: 3em
+      height 3em
+      width 3em
 
   .stake-form-group
-    display: block
-    padding: 0
+    display block
+    padding 0
 
   .stake-footer
-    display: flex
-    justify-content: flex-end
+    display flex
+    justify-content flex-end
 </style>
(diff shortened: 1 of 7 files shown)


app/src/renderer/components/staking/ModalStake.vue
```diff
@@ -0,0 +1,112 @@
...
+@import '~variables'
+
+.modal-stake
+  background: var(--app-nav)
+  display: flex
+  flex-direction: column
+  height: 50%
+  justify-content: space-between
+  left: 50%
+  padding: 3%
```
okwme: interesting choice of percentage on padding, i usually think of this being fixed around a scaling element. we've probably never established a design standard for styling but curious about your decision, penny for your thoughts?

app/src/renderer/components/staking/PageValidator.vue
```diff
@@ -34,26 +34,56 @@ tm-page(:title='validatorTitle(this.validator)')
...
+import { TmBtn, TmListItem, TmPage, TmPart, TmToolBar } from "@tendermint/ui"
 import { TmDataError } from "common/TmDataError"
+import ModalStake from "staking/ModalStake"
 import numeral from "numeral"
 import AnchorCopy from "common/AnchorCopy"
 export default {
   name: "page-validator",
   components: {
     AnchorCopy,
+    "modal-stake": ModalStake,
```
okwme: this should auto convert from camelCase to that format so no need to list it explicitly like this afaik

app/src/renderer/components/staking/PageValidator.vue
```diff
@@ -68,6 +98,45 @@ export default {
...
+
+      try {
+        await this.$store.dispatch("submitDelegation", delegation)
+
+        this.$store.commit("notify", {
+          title: "Successful Staking!",
+          body: `You have successfully staked your ${this.bondingDenom}s.`
+        })
+      } catch ({ message }) {
+        let errData = message.split("\n")[5]
```
okwme: is this from @fedekunze's error catch PR from a while ago? Wasn't this supposed to be handled by the SDK eventually?

test/unit/specs/components/staking/ModalStake.spec.js
```diff
@@ -0,0 +1,63 @@
...
+  localVue.use(Vuelidate)
+
+  return mount(ModalStake, {
+    localVue,
+    propsData: {
+      fromOptions: [
+        `My Wallet - cosmosaccaddr15ky9du8a2wlstz6fpx3p4mqpjyrm5ctpesxxn9`
+      ],
+      maximum: 100,
+      to: `cosmosvaladdr15ky9du8a2wlstz6fpx3p4mqpjyrm5ctplpn3au`
```
okwme: and this one?

test/unit/specs/components/staking/PageValidator.spec.js
```diff
@@ -1,5 +1,50 @@
 import setup from "../../../helpers/vuex-setup"
 import PageValidator from "renderer/components/staking/PageValidator"
+import { mount } from "@vue/test-utils"
+
+// Create a getters object from an object of simple values.
+const mockGetters = values =>
```
okwme: i like it

app/src/renderer/components/staking/PageValidator.vue
```diff
@@ -91,9 +160,9 @@ export default {
 <style lang="stylus">
 @media screen and (min-width: 640px)
   #validator-profile .tm-part-main
-    display flex
-    flex-flow row-reverse nowrap
+    display: flex
```
faboweb: Is your linter working?

test/unit/specs/components/staking/PageValidator.spec.js
faboweb: 👍

app/src/renderer/components/staking/PageValidator.vue
```diff
@@ -34,26 +34,56 @@ tm-page(:title='validatorTitle(this.validator)')
...
     TmToolBar,
     TmDataError
   },
+  data: () => ({
+    showModalStake: false
+  }),
   computed: {
-    ...mapGetters(["delegates", "config", "keybase"]),
+    ...mapGetters([
+      `bondingDenom`,
```
faboweb: Do we need a linter rule for strings without interpolation?

app/src/renderer/components/staking/ModalStake.vue
```diff
@@ -0,0 +1,112 @@
...
+      img.icon(class='stake-atom' src="~assets/images/cosmos-logo.png")
+      span.tm-modal-title Stake
+
+    tm-form-group.stake-form-group(
+      :error='$v.amount.$invalid'
+      field-id='amount'
+      field-label='Amount'
+    )
+      tm-field#amount(
+        type="number"
```
fedekunze: let's add the following here as well:
```
:max="this.maximum"
:min="0"
v-focus
```

app/src/renderer/components/staking/ModalStake.vue
```diff
@@ -0,0 +1,112 @@
...
+        v-model="amount")
+      tm-form-msg(name='Amount' type='between' :min='1' :max='maximum'
+        v-if='$v.amount.$invalid')
+
+    tm-form-group.stake-form-group(
+      field-id='to' field-label='To')
+      tm-field#to(v-model="to")
```
fedekunze: `tm-field#to(v-model="to" readonly)`

app/src/renderer/components/staking/ModalStake.vue
```diff
@@ -0,0 +1,112 @@
...
+    TmFormMsg
+  },
+  data: () => ({
+    amount: 1,
+    fromIndex: 0
+  }),
+  validations() {
+    return {
+      amount: {
+        between: between(1, this.maximum)
```
fedekunze: Imo it should be `between(0.0000000001, this.maximum)` we should allow for decimals as well

app/src/renderer/components/staking/ModalStake.vue
fedekunze: (check [here](https://github.com/cosmos/voyager/pull/1310/files#diff-83b38b5ba2b68f6f0e53f186c743d3b8) for how to add the `v-focus` )

app/src/renderer/components/staking/ModalStake.vue
faboweb: Apparently we don't allow decimals. Decimals will be `sub-atoms`.

app/src/renderer/components/staking/ModalStake.vue
NodeGuy: My mind is an empty canvas on this one and wide open to suggestions.

app/src/renderer/components/staking/PageValidator.vue
NodeGuy: Indeed you're right!

app/src/renderer/components/staking/PageValidator.vue
NodeGuy: This is copied-and-pasted code from PageBond.vue.  I don't like the code but felt that fixing it was out of scope for this PR.

app/src/renderer/components/staking/PageValidator.vue
NodeGuy: That was my editor automatically adding its opinion to Stylus formatting.  I see now that we're not using colons so I've changed it's settings.

app/src/renderer/components/staking/PageValidator.vue
NodeGuy: What would the rule do?

app/src/renderer/components/staking/ModalStake.vue
NodeGuy: Wow, that's a great improvement to add the min and max to the field!

app/src/renderer/components/staking/PageValidator.vue
faboweb: Decide on single quotes on strings without interpolation?

app/src/renderer/components/staking/ModalStake.vue
faboweb: I would go for `padding 1rem` or something like that.

NodeGuy:
@faboweb Why hasn't the code coverage test run?  Is it being blocked by Fede's changes request?

faboweb:
@NodeGuy I don't. You can always go to CircleCI to rerun the workflow, which I did now. Let's see if it works.

NodeGuy:
Coverage is failing because of main.js.  It's not clear to me why, though I did add this new code:

```JavaScript
directives: {
  focus: {
    // directive definition
    inserted: function (el) {
      el.focus()
    }
  }
}
```

It's straight out of the [Vue manual](https://vuejs.org/v2/guide/custom-directive.html) and is an integration so I don't think it needs to be tested.

@faboweb What would you like to do?

faboweb:
Apparently PageValidator is not fully covered. Clicking on the CodeCov report shows you which lines are missing coverage.

app/src/renderer/components/staking/PageValidator.vue
NodeGuy: As a matter of habit I always use backquotes, otherwise I have to go through this less efficient process:

1.  Writing a string with single quotes.
1.  Realize that I want to add an interpolation.
1.  Add the interpolation.
1.  Change the single quotes to backquotes.

Starting with backquotes means I can skip step #4.  It happens often enough to save me significant time.

NodeGuy:
It's strange—I tried clicked on PageValidator before but it read `File not found in report`.  I was able to see the problem in the diff, however.

I added another test to coverage a missing branch but it's also complaining about this function not being covered:

```JavaScript
    closeCannotStake() {
      this.showCannotStake = false
    },
```

I don't think it makes sense to require 100% coverage, or always increasing coverage, unless we adopt the functional core/imperative shell paradigm I referenced in my talk.  What do you think?

NodeGuy:
> How about making a rule to always use backquotes then?

I would be tempted but there are a few edge cases in which they don't work, e.g.,

```JavaScript
const objectLiteral = {
  "unusual-key": value
}
```

app/src/renderer/components/staking/PageValidator.vue
NodeGuy: Aww!  Now I want to give you colons!

app/src/renderer/components/staking/PageValidator.vue
faboweb: How about this rule: https://eslint.org/docs/rules/quotes which would auto fix all quotes to backticks where possible?

app/src/renderer/components/staking/PageValidator.vue
```diff
@@ -34,26 +34,68 @@ tm-page(:title='validatorTitle(this.validator)')
...
+    modal-stake(
+      v-if="showModalStake"
+      v-on:submitDelegation="submitDelegation"
+      :showModalStake.sync="showModalStake"
+      :fromOptions="[{ key: `My Wallet - ${this.wallet.address}`, value: 0 }]"
+      :maximum="availableAtoms()"
+      :to="validator.owner"
+    )
+
+    tm-modal(:close="closeCannotStake" icon="warning" v-if="showCannotStake")
```
faboweb: Are you sure, we want this to be a modal? I thought it is discouraged to use blocking modals in modern UX. (Not a blocker, we can have follow up on this)

app/src/renderer/main.js
```diff
@@ -39,6 +39,13 @@ Vue.use(Router)
 Vue.use(Tooltip, { delay: 1 })
 Vue.use(Vuelidate)
 
+// directive to focus form fields
+Vue.directive("focus", {
```
faboweb: 👍 
If we use more directives, maybe we can put them into a separate file.

app/src/renderer/components/staking/PageValidator.vue
fedekunze: I personally prefer to have double quotes everywhere and use backquotes only when using interpolation

app/src/renderer/components/staking/PageValidator.vue
fedekunze: yeah, this should be a notification

test/unit/specs/components/staking/ModalStake.spec.js
```diff
@@ -0,0 +1,49 @@
...
+import ModalStake from "staking/ModalStake"
+
+// Create an example stake modal window.
+const Wrapper = () => {
+  return mount(ModalStake, {
+    propsData: {
+      fromOptions: [
+        `My Wallet - cosmosaccaddr15ky9du8a2wlstz6fpx3p4mqpjyrm5ctpesxxn9`
+      ],
+      maximum: 100,
```
fedekunze: this should be the total amount of unstaked tokens

PR #1299, opened 2018-09-07
Fabo/fix delegation update issue

faboweb:
Description:

delegations request would update state of next user after sign out sign in

❤️ Thank you!

### DIFF
--- a/app/src/renderer/vuex/modules/delegation.js
+++ b/app/src/renderer/vuex/modules/delegation.js
@@ -58,6 +58,7 @@ export default ({ node }) => {
     },
     resetSessionData({ rootState }) {
       rootState.delegation = JSON.parse(JSON.stringify(emptyState))
+      console.log("rest", rootState.delegation)
     },
     // load committed delegations from LCD
     async getBondedDelegates(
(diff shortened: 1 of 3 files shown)


app/src/renderer/vuex/modules/delegation.js
```diff
@@ -58,6 +58,7 @@ export default ({ node }) => {
     },
     resetSessionData({ rootState }) {
       rootState.delegation = JSON.parse(JSON.stringify(emptyState))
+      console.log("rest", rootState.delegation)
```
NodeGuy: Was this intended for temporary debugging?

test/unit/specs/store/delegation.spec.js
```diff
@@ -3,11 +3,12 @@ import setup from "../../helpers/vuex-setup"
 let instance = setup()
 
 describe("Module: Delegations", () => {
-  let store
+  let store, node
```
NodeGuy: It's better form to use separate lines for each variable:

```JavaScript
let store
let node
```

app/src/renderer/vuex/modules/delegation.js
faboweb: Removed it

test/unit/specs/store/delegation.spec.js
faboweb: To be consistent, I would need to change this everywhere. Let's not do it until we come up for a rule about it, ok?

PR #1314, opened 2018-09-13
Fabo/1189-catch reconnect errors

### LINKED ISSUE #1189: Surface errors from the main thread better
UI Version: 0.9.4

Description:

`main-process: gaiacli: ERROR: unknown command "rest-server" for "gaiacli"` did not surface to the UI.

faboweb:
Closes #1189 

Description:
the async call for reconnection wasn't handling errors. now in the connect function handles errors directly.

❤️ Thank you!

### DIFF
--- a/app/src/main/index.js
+++ b/app/src/main/index.js
@@ -274,11 +274,13 @@ async function startLCD(home, nodeIP) {
     ])
     logProcess(child, join(home, "lcd.log"))
 
-    child.on("exit", () => {
-      let msg = `The ${LCD_BINARY_NAME} rest-server (LCD) exited unplanned`
+    child.stderr.on("data", error => {
+      let errorMessage = `The gaiacli rest-server (LCD) experienced an error:\n${error.toString(
+        "utf8"
+      )}`
       lcdStarted
-        ? handleCrash(msg) // if fails later
-        : reject(msg) // if fails immediatly
+        ? handleCrash(errorMessage) // if fails later
+        : reject(errorMessage) // if fails immediatly
     })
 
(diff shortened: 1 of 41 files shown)


app/src/main/index.js
```diff
@@ -98,7 +98,11 @@ function handleCrash(error) {
   afterBooted(() => {
     if (mainWindow) {
       mainWindow.webContents.send("error", {
-        message: error ? error.message : undefined
+        message: error
+          ? error.message
+            ? error.message
+            : error
+          : "An unspecified error occurred"
```
jbibla: can we give any more info?

app/src/main/index.js
```diff
@@ -269,6 +274,13 @@ async function startLCD(home, nodeIP) {
     ])
     logProcess(child, join(home, "lcd.log"))
 
+    child.on("exit", () => {
+      let msg = `The ${LCD_BINARY_NAME} rest-server (LCD) exited unplanned`
```
jbibla: same here?

app/src/main/index.js
fedekunze: ++

faboweb: If there is no error, we can not guess the error. ;)
We can change the error message to sth. More meaningful maybe? Do u guys have sth in mind?

jbibla: what are the circumstances where this could happen?

faboweb: i.e. an uncaught `Promise.reject` without error message. We should avoid having those, but still have the fallback if we forget it.

PR #1317, opened 2018-09-14
Fabo/1171 Validator Profile

### LINKED ISSUE #1171: new UI For validator profile
- [x] Design
- [ ] Implementation

https://zpl.io/VQEQ6zn - Validator Profile

faboweb:
Closes #1171 

New:
image

❤️ Thank you!

### DIFF
--- a/app/src/renderer/components/staking/PageValidator.vue
+++ b/app/src/renderer/components/staking/PageValidator.vue
@@ -1,14 +1,75 @@
 <template lang="pug">
+mixin dl(label, value, color)
+  dl.colored_dl
+    dt= label
+    dd(style={borderColor: color})= value
+mixin info(label, value)
+  dl.info_dl
+    dt= label
+    dd= value
 tm-page(:title='validatorTitle(this.validator)')
   div(slot="menu"): tm-tool-bar
     router-link(to="/staking" exact): i.material-icons arrow_back
     anchor-copy(v-if="validator" :value="validator.owner" icon="content_copy")
-
+  template(slot="menu-body", v-if="config.devMode"): tm-balance(:unstakedAtoms="user.atoms")
 
   tm-data-error(v-if="!validator")
   template(v-else)
     tm-part(title='!!! CRITICAL ALERT !!!' v-if="validator.revoked")
       tm-list-item(title="This validator is revoked!" subtitle="Are you the owner? Go fix it!" type="anchor" href="https://cosmos.network/docs/validators/validator-setup.html#common-problems")
+
+    .validator-profile__header.container
+      .column
+        img.avatar(v-if="validator.keybase" :src="validator.keybase.avatarUrl")
+        img.avatar(v-else src="~assets/images/validator-icon.svg")
+      .column.validator-profile__header__info
+        .row.validator-profile__header__name
+          .column
+            .validator-profile__header__name__title {{ validator.description.moniker || 'Anonymous' }}
+            //- TODO replace with address component when ready
+            anchor-copy.validator-profile__header__name__address(:value="validator.owner" :label="shortAddress(validator.owner)")
+          .column.validator-profile__header__actions
+            tm-btn(value="Stake" color="primary")
+            tm-btn(value="Unstake" color="secondary")
+        .row.validator-profile__header__data
+          +dl('My Stake', 12300)
+          +dl('My Rewards', 3456)
+          .validator-profile__header__data__break
+          +dl('Voting Power', 12300)
+          +dl('Uptime', 3456)
+          +dl('Commission', 12300)
+          +dl('Slashes', 3456)
+        //- .LiValidator__break: span
+        //-   .LiValidator__value.your-votes
+        //-     span {{ yourVotes }}
+        //-   .LiValidator__value.your-rewards
+        //-     span {{ yourRewards }}
+        //-   .LiValidator__value.percent_of_vote
+        //-     span {{ validator.percent_of_vote }}
+        //-   .LiValidator__value.uptime
+        //-     // add .green .yellow or .red class to this span to trigger inidication by color
+        //-     span {{ uptime }}
+        //-   .LiValidator__value.commission
+        //-     // add .green .yellow or .red class to this span to trigger inidication by color
+        //-     span {{ commission }}
+        //-   .LiValidator__value.slashes
+        //-     // add .green .yellow or .red class to this span to trigger inidication by color
+        //-     span {{ slashes }}
+
+    .container.validator-profile__details
+      .row
+        .column
+          +info('Owner / Address', 'shortAddress(validator.owner)')
+          +info('Keybase ID', 'B536B81DDA4323BB')
+          +info('First Seen', 'December 16, 2018')
+          +info('Keybase ID', 'B536B81DDA4323BB')
+          +info('Details', 'Hello from Staking Facilities!')
+        .column
+          +info('Commission', '1.3%')
+          +info('Max Commission Rate', '1.3%')
+          +info('Max Daily Commission Change', '1.3%')
+          +info('Commission Change Today', '1.3%')
+
 
     tm-part#validator-profile(title='Validator Profile')
       img.avatar(v-if="validator.keybase" :src="validator.keybase.avatarUrl" width="192" height="192")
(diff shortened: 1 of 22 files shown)


app/src/renderer/components/staking/PageValidator.vue
```diff
@@ -1,89 +1,129 @@
...
+          dl.colored_dl(v-if="config.devMode")
+            dt Slashes
+            dd n/a
+
+    .validator-profile__details.validator-profile__section
+      .row
+        .column
+          dl.info_dl
+            dt Public Key
+            dd {{validator.owner}}
```
fedekunze: This should be `validator.pub_key`

app/src/renderer/components/staking/PageValidator.vue
```diff
@@ -1,89 +1,129 @@
...
+        img.avatar(v-if="validator.keybase" :src="validator.keybase.avatarUrl")
+        img.avatar(v-else src="~assets/images/validator-icon.svg")
+      .column.validator-profile__header__info
+        .row.validator-profile__header__name
+          .column
+            div.validator-profile__status-and-title
+              span.validator-profile__status(v-bind:class="statusColor" v-tooltip.top="status")
+              .validator-profile__header__name__title {{ validatorTitle(validator) }}
+            //- TODO replace with address component when ready
+            anchor-copy.validator-profile__header__name__address(:value="validator.owner" :label="shortAddress(validator.owner)")
```
fedekunze: Let's display the full address here, we have enough space to show it

app/src/renderer/components/staking/PageValidator.vue
```diff
@@ -1,89 +1,129 @@
...
+    .validator-profile__header.validator-profile__section
+      .column
+        img.avatar(v-if="validator.keybase" :src="validator.keybase.avatarUrl")
+        img.avatar(v-else src="~assets/images/validator-icon.svg")
+      .column.validator-profile__header__info
+        .row.validator-profile__header__name
+          .column
+            div.validator-profile__status-and-title
+              span.validator-profile__status(v-bind:class="statusColor" v-tooltip.top="status")
+              .validator-profile__header__name__title {{ validatorTitle(validator) }}
```
fedekunze: Can we get rid of the `... - (address)` please ? I don't see a point on keeping it since we're already showing the validator's owner address

app/src/renderer/components/staking/PageValidator.vue
```diff
@@ -1,89 +1,129 @@
...
+            dd {{validator.owner}}
+          dl.info_dl
+            dt Keybase ID
+            dd {{validator.description.identity || 'n/a'}}
+          dl.info_dl(v-if="config.devMode")
+            dt First Seen
+            dd n/a
+          dl.info_dl
+            dt Details
+            dd.info_dl__text-box {{validator.description.details || 'n/a'}}
```
fedekunze: @faboweb I think we should definitely add the validator `website` on another box as well

app/src/renderer/components/staking/PageValidator.vue
```diff
@@ -247,7 +247,6 @@ export default {
       }
     },
     validatorTitle(validator) {
```
jbibla: shouldn't this be a computed property instead as opposed to a method?

app/src/renderer/components/staking/PageValidator.vue
fedekunze: also let's change all the `[do-not-modify]` cases to `n/a`

app/src/renderer/components/staking/PageValidator.vue
```diff
@@ -105,19 +146,60 @@ export default {
...
+      // status: candidate
+      if (parseFloat(this.validator.voting_power) === 0)
+        return "This validator has declared candidacy but does not have enough voting power yet"
+
+      // status: validator
+      return "This validator is actively validating"
+    },
+    statusColor() {
+      // status: jailed
+      if (this.validator.revoked) return "red"
```
fedekunze: can we change the background color of the validator header as well (cc: @jbibla )? I spent a couple minutes trying to figure out where I could check the validator status

app/src/renderer/components/staking/PageValidator.vue
```diff
@@ -1,89 +1,129 @@
...
+            dt Commission
+            dd(id="validator-profile__commission" v-bind:class="[commissionLevel]") {{pretty(validator.commission)}} %
+          dl.colored_dl(v-if="config.devMode")
+            dt Slashes
+            dd n/a
+
+    .validator-profile__details.validator-profile__section
+      .row
+        .column
+          dl.info_dl
```
fedekunze: maybe for another PR/issue, but can we change the box background color to gray? It looks like an input field rn... (cc: @jbibla)

app/src/renderer/components/staking/PageValidator.vue
fedekunze: or border line color

jbibla: i think we should go ahead with this for now - but i am very happy to play around with more ways of making status more obvious. good feedback.

app/src/renderer/components/staking/PageValidator.vue
jbibla: let's merge it for now, i will keep iterating

app/src/renderer/components/staking/PageValidator.vue
jbibla: i don't think we should add the website as a form element in the second box. the designs have an icon for visiting the website beside the name.

app/src/renderer/components/staking/PageValidator.vue
jbibla: no i don't think that's a good idea. the validator profile is like a blown up version of the validator Li - i want to keep it for consistency and familiarity. the form below (validator details) should show the full address.

app/src/renderer/components/staking/PageValidator.vue
jbibla: no i don't think that's a good idea. the validator profile is like a blown up version of the validator Li - i want to keep it for consistency and familiarity. the form below (validator details) should show the full address.

app/src/renderer/components/staking/PageValidator.vue
jbibla: i think owner is the same as pubkey (at least in the mocked response)

app/src/renderer/components/staking/PageValidator.vue
faboweb: I don't think an icon alone is self explanatory. We can have both. I will add it and we can remove it later again.

fedekunze: > the designs have an icon for visiting the website beside the moniker.

Ohh that's the website ? can we change it for this icon then ? 

image

app/src/renderer/components/staking/PageValidator.vue
fedekunze: @jbibla we're not showing the address below, only the `pub_key`

app/src/renderer/components/staking/PageValidator.vue
faboweb: Fede meant the suffix after the title. For consistency with the LiValidator I removed it.

fedekunze: Well according to the [design](https://app.zeplin.io/project/59e66beff6721ea8bf562e5b/screen/5b8d49761076bc0ce6e551b1) we shouldn't be showing the `- (address)` ...

app/src/renderer/components/staking/PageValidator.vue
fedekunze: no they are different and they also have different `bech32` prefixes

faboweb: Those should not be the same. One is the owner and one is the pub_key. But we use the owner address to delegate. Currently I don't see a need for the pub_key anywhere. So let's not show the pub key.

app/src/renderer/components/staking/PageValidator.vue
faboweb: Maybe have a color label?

Can one of you please create an issue for this?

app/src/renderer/components/staking/PageValidator.vue
faboweb: @fedekunze can you create an issue for this please?

faboweb:
Updated screenshot

app/src/renderer/components/staking/PageValidator.vue
fedekunze: See #1324

app/src/renderer/components/staking/PageValidator.vue
fedekunze: I actually think the problem relies in that we haven't yet implemented the same dots on the validators page. When I saw the original designs it was more clear that they were jailed. I'll make an issue for this

fedekunze: See #1325

Issue #1278, opened 2018-09-04
Create My Stake tab.


okwme:
Originally part of https://github.com/cosmos/voyager/issues/1054 this was blocked because it was non-trivial to acquire a list of all validators to which the current user is currently unbonding from. Once that data is available it is possible to use the tab "My Stake" as a trigger for filtering the validators into sections of bonding and unbonding.

Scope:

- [x] create my stake page
- [x] link my stake page from the tabs menu
- [x] create list of validators you have bond with (based on `committedDelegations`)
- [x] create list of validators you are unbonding from (based on `unbondingDelegations`)
- [x] add link to all validators

faboweb:
This should be possible right now. We are already querying this data. The data lives in `delegations.js -> state.unbondingDelegations`.

jbibla:
`unbondingDelegations` is not the same as "bonded delegators" by which i think @okwme means "currently bonded validators".

faboweb:
> all validators to which the current user is currently unbonding from

sounds like `unbondingDelegations`. maybe the title is wrong, maybe I am wrong. 🤷‍♂️

faboweb:
accessing the current state of the my stake page:
the distinction in bonded / unbonding validators is missing.

siman:
Is it possible to see your designs before implementing them or maybe even take some part in discussions?

jbibla:
that sounds like a great idea.

[here is a link](https://scene.zeplin.io/project/59e66beff6721ea8bf562e5b) to some designs we're working on.

faboweb:
Scope:
- [ ] create my stake page
- [ ] link my stake page from the tabs menu
- [ ] create list of validators you have bond with (based on `committedDelegations`)
- [ ] create list of validators you are unbonding from (based on `unbondingDelegations`)
- [ ] add link to all validators

NodeGuy:
Your scope expands the issue beyond bonded delegators.  I've copied your scope additions to the main body and changed the title.

NodeGuy:
I see now there was already an issue for [implement "My Stake" page](https://github.com/cosmos/voyager/issues/1054) which was closed even though it wasn't done.  Man, this is confusing.

NodeGuy:
I split out the completion of LiDelegate to #1319.

PR #1318, opened 2018-09-14
WIP: Create My Stake tab.

NodeGuy:
Makes progress against #1278.

Description:

Adds a new tab, My Stake, to the Staking page.


app/src/renderer/components/staking/PageStaking.vue
```diff
@@ -7,12 +7,40 @@ tm-page(title='Staking')
...
   .delegates-tabs
     .tab(v-for="(tab, i) in tabs",
       :key="'tab-' + i",
       :class="{'tab-selected': i === tabIndex}",
-      @click="tabIndex = 1") {{tab}}
-  .delegates-container
+      @click="tabIndex = i") {{tab}}
+
+  //- My Stake
+  template(v-if="tabIndex === 0")
```
jbibla: do we want to implement tabs this way? i think we should use vue-router to switch routes.

NodeGuy: OK, I'll look into it.

app/src/renderer/components/staking/PageStaking.vue
```diff
@@ -7,27 +7,14 @@ tm-page(title='Staking')
...
-  .fixed-button-bar(v-if="!delegates.loading")
-    template(v-if="userCanDelegate")
-      .label #[strong {{ shoppingCart.length }}] validators selected
-      tm-btn(id="go-to-bonding-btn" type="link" to="/staking/bond" :disabled="shoppingCart.length === 0" icon="chevron_right" icon-pos="right" value="Next" color="primary")
-    template(v-else)
-      .label(v-if="!delegation.loadedOnce && delegation.loading") Loading delegations...
-      .label(v-else) You do not have any {{bondingDenom}}s to stake.
-      tm-btn(disabled icon="chevron_right" icon-pos="right" value="Next" color="primary")
+      @click="tabIndex = i")
+        router-link(:to="tabs[i]") {{ tab }}
```
jbibla: `:to="My Stake"` will 404 because it doesn't exist in `routes.js` - try `/staking`

jbibla:
from https://router.vuejs.org/guide/essentials/nested-routes.html

> a rendered component can also contain its own, nested `<router-view>`
...
To render components into this nested outlet, we need to use the children option in VueRouter constructor config

basically, just have to include `router-view` on the page for `children` in `routes.js` and also use path friendly path names.

jbibla:
basically, just have to include `router-view` on the page when you use `children` in `routes.js` and also use path friendly path names.

app/src/renderer/components/staking/PageStaking.vue
NodeGuy: I implemented it with Vue Router and don't like it because if I selected the `Validators` tab, then click to `Wallet`, then click back to `Staking`, I would expect the `Validators` tab to remain open.  If we use routes then it will always open `My Stake` when going back to `Staking`.

Also, pressing the Back button doesn't update `tabIndex` so we'd have to solve that somehow.

@jbibla, what do you think?

jbibla: i don't think that's the expected behaviour. 

staking is the parent route but it only displays the children. therefore "staking" should always take the user to "my stake".

i guess the back button can be addressed by checking the `$route` and setting it accordingly.

NodeGuy: OK, then assuming I can fix the back button I'll keep the routed implementation.

NodeGuy: Done using your suggestions!

NodeGuy:
I used my admin privileges to merge this even though it doesn't yet have adequate coverage in order to unblock @jbibla.  I'll continue working on adding test coverage in another PR.

PR #1320, opened 2018-09-17
fedekunze/ implement pool & params tab

### LINKED ISSUE #1172: staking Params and Pool
Description:

Implement design to implement the view for querying `Params` and `Pool`
See PR: cosmos/cosmos-sdk#2099

- [x] Endpoints
- [x] [Design](https://app.zeplin.io/project/59e66beff6721ea8bf562e5b/screen/5bb26dc2cc296e3ffefa3046): Tab
- [ ] Implementation

fedekunze:
Closes #1172 

Description:

❤️ Thank you!


fedekunze:
marking as blocked because latest release of SDK (0.24.2) doesn't have the endpoints

jbibla:
share a screenshot! maybe we can find a way to merge this with a `v-if` so that when the endpoints exist they will display, but while they don't, it will still be ok?

no good for PRs to wait open for too long...

@faboweb think that would be ok?

fedekunze:
I disabled the tab and thus removing the `blocked` label

app/src/renderer/components/staking/TabParameters.vue
```diff
@@ -0,0 +1,215 @@
...
+    TmBtn,
+    TmListItem,
+    TmPage,
+    TmPart,
+    TmToolBar
+  },
+  data: () => ({
+    paramsTooltips: {
+      description: `Staking parameters define the high level settings for staking`,
+      inflation_rate_change: `maximum annual change in inflation rate`,
```
jbibla: these should all start with a capital letter

faboweb:
Next time please confirm that the data is available before starting on implementing views for it. So long I think mocked data is OK. The component should be hidden behind the development flag though.

PR #1345, opened 2018-09-21
David/1278 my stake

NodeGuy:
Make progress against #1278.

Close https://github.com/cosmos/voyager/issues/1343.

Fix recently introduced layout bugs and simplify two unit tests.

screen shot 2018-09-21 at 2 54 22 pm

### DIFF
--- a/app/src/renderer/components/staking/PageValidator.vue
+++ b/app/src/renderer/components/staking/PageValidator.vue
@@ -1,6 +1,6 @@
 <template lang="pug">
 tm-page
-  template(slot="menu-body", v-if="config.devMode"): tm-balance(:unstakedAtoms="user.atoms" :tabs="tabs")
+  template(slot="menu-body", v-if="config.devMode"): tm-balance(:unstakedAtoms="user.atoms")
   div(slot="menu"): tm-tool-bar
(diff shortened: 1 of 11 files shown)


app/src/renderer/components/staking/PageValidator.vue
```diff
@@ -379,60 +378,44 @@ export default {
       min-height 6.91rem
```
faboweb: @jbibla Imo we need a set of measures that are consistent throughout the application. Those should probably also be used as variables in Sketch. What do you think?

app/src/renderer/components/staking/TabMyStake.vue
```diff
@@ -119,12 +117,11 @@ export default {
...
-    unstakedValidators() {
-      const unbonding = new Set(
-        Object.keys(this.delegation.unbondingDelegations)
-      )
-
-      return this.delegates.delegates.filter(({ id }) => unbonding.has(id))
+    unstakedValidators(
+      { delegates: { delegates }, delegation: { unbondingDelegations } } = this
+    ) {
+      const unbonding = new Set(Object.keys(unbondingDelegations))
```
faboweb: 👍

PR #1348, opened 2018-09-21
Use gaia-seeds.interblock.io as seed node.

NodeGuy:
Description:

Previously we were using seed nodes from the documentation.  They proved unreliable so now we're trying gaia-seeds.interblock.io instead.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 * reconnection errors did not show up correctly in view @faboweb
 * fixed crash when reconnecting @faboweb
 * fixed crash when using an offline fixed node @faboweb
+* yarn build:testnets now uses gaia-seeds.interblock.io as seed node @NodeGuy
 
 ## [0.10.2] - 2018-08-29
 
--- a/tasks/build/testnets/build.sh
+++ b/tasks/build/testnets/build.sh
@@ -9,10 +9,11 @@ echo ###########################################################################
 ## Create config.toml file.
 builds/Gaia/linux_amd64/gaiad init --name Voyager
 
-### Add seed nodes from
-### https://github.com/cosmos/cosmos-sdk/tree/5c8791743411cac58c502f3b18bfe8a1970e830e/cmd/gaia/testnets#add-seed-nodes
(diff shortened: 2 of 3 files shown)


faboweb:
Was this tested? When I use this config I get can't connect to any node:

```
Starting electron...
  (network path: ./app/networks/gaia-8001/)

electron  raven@2.6.0 alert: no DSN provided, error reporting disabled

electron  (node:7840) [DEP0016] DeprecationWarning: 'root' is deprecated, use 'global'

electron  2018-09-24 13:56:16.184 Electron[7840:360685] *** WARNING: Textured window <AtomNSWindow: 0x7fdcaf78a780> is getting an implicitly transparent titlebar. This will break when linking against newer SDKs. Use NSWindow's -titlebarAppearsTransparent=YES instead.

electron  mainWindow opened

electron  Redirecting console output to logfile /Users/fabo/.cosmos-voyager-dev/gaia-8001/main.log

electron  root exists (/Users/fabo/.cosmos-voyager-dev/gaia-8001)

electron  configs are compatible with current app version

electron  spawning /Users/fabo/Development/voyager/builds/gaia/darwin_amd64/gaiacli with args ""version""

electron  gaiacli: 0.24.2-1c38c704

electron  gaiacli version: "0.24.2", expected: "0.24.1"

electron  Adding new peer: gaia-seeds.interblock.io

electron  Querying node: gaia-seeds.interblock.io

electron  gaiacli exited with code 0

electron  Unable to discover peers from node { host: 'gaia-seeds.interblock.io', state: 'available' }: Error: connect ECONNREFUSED 52.206.149.214:26657

electron  View has booted

electron  2018-09-24 13:56:18.654 Electron Helper[7845:360883] Couldn't set selectedTextBackgroundColor from default ()
```

NodeGuy:
Yes, I did test it and it worked for me.

The error you're reporting (`electron  Unable to discover peers from node { host: 'gaia-seeds.interblock.io', state: 'available' }: Error: connect ECONNREFUSED 52.206.149.214:26657`) shows that it attempted to use the correct new seed node (`gaia-seeds.interblock.io`) but that the node was unresponsive in that moment.  That's a different problem.

faboweb:
This still doesn't work. I suppose it can't work because `gaia-seeds.interblock.io` requires `https` where all the other nodes require `http`.

How to reproduce:
- run `yarn build:testnets`
- delete `~/.cosmos-voyager-dev/gaia-8001` (this also deletes the addressbook.json which would store already discovered peers)
- run `yarn start gaia-8001`

NodeGuy:
Ah, good catch.

Should we wait until we switch over to HTTPS with the new SDK release then?

faboweb:
Voyager currently doesn't work because of this. Fixed with #1360

faboweb:
Close in favor of #1360

PR #1374, opened 2018-09-28
David-Jordan/Network-Connectivity

### LINKED ISSUE #1322: move network connectivity bar to sidebar to match design

jbibla:
Closes #1322 

thanks @NodeGuy for the help with the tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 * Changed a bunch of JavaScript files to strict mode. @NodeGuy @faboweb
 * prefer backquotes in code @ƒaboweb
 * Refactord submitDelegation. @NodeGuy
+* Improved network connectivity bar @jbibla
 
 ### Fixed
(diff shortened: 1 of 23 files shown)


fedekunze:
what is this PR about ?

jbibla:
screen shot 2018-09-28 at 12 06 13 pm
screen shot 2018-09-28 at 12 06 20 pm
screen shot 2018-09-28 at 12 06 31 pm

jbibla:
as part of an effort to communicate trust and security.

when we implement https://github.com/cosmos/voyager/issues/1222 we can show this info in the tooltip on the lock.

screen shot 2018-10-01 at 8 59 56 am

screen shot 2018-10-01 at 8 54 55 am

faboweb:
When would this show an open lock?

faboweb:
So the icon has no purpose right now? As with code, should we maybe avoid implementing things we don't need yet?

faboweb:
I don't have the feeling that this symbol shows a "connection" maybe a "secure connection". But I don't feel strongly about it.

Issue #1312, opened 2018-09-13
Pick full node on UI


faboweb:
As a follow up for https://github.com/cosmos/voyager/issues/1291 we might allow the user to pick a node they trust.

Scope:
- component: ability to enter an IP of a trusted node
- logic: ability to connect to a full node based on an IP set in the view

jbibla:
a better ux might be to let the user pick from a list of full nodes including monikers and IPs. 

but i'm not sure we need to do any of this before launch.

faboweb:
agreed

jbibla:
> this will probably often be used by people how know the IP of their trusted node

even if you know the IP of the node, having a list would be nice / could help with verification etc...

faboweb:
Maybe make it a config setting.

jbibla:
> Maybe make it a config setting.

what do you mean?

faboweb:
Make it configurable in the config.toml.

jbibla:
isn't it already with `COSMOS_NODE`? i thought the point of this was to let users (not developers) pick a node...?

faboweb:
> Make it configurable in the config.toml.

This was an outcome of a discussion in the team meeting. We want to be able to define the set of trusted full nodes when shipping Voyager. A regular user will not use `COSMOS_NODE` as you pointed out. As we discussed that validators may ship a branded Voyager, validators may want to put in their preferred full node in the configuration easily.

PR #1394, opened 2018-10-02
Fabo/1291 query full nodes

### LINKED ISSUE #1291: Query against seed nodes
Description:

The new strategy pre launch is to solve performance problems by avoiding the client side LCD. This means we will query data against a full node directly. Our seed nodes i.e. `gaia-seed1.interblock.io` already serve a LCD interface at port 1317.

Scope:
- skip peer detection
- always connect to one of the seed nodes (full node)
- pass connected node to the view thread
- view thread querys (`GET`) directly against the seed (full node)

faboweb:
Closes #1291 

Description:

❤️ Thank you!


jbibla:
can we leave the addressbook in place, as a fallback for when the dedicated full nodes go down (which feels inevitable)?

faboweb:
I don't think we can. The full nodes are a trusted entity. Every other node isn't. We would need to design for the case that we connect to another node. I don't think we should handle this scenario right now.

NodeGuy:
Are we ready for this or should we wait for our node to be indexing?  I didn't receive the link to the issue about it from you.

fedekunze:
@faboweb is the node still not indexing txs?

faboweb:
Blocked until out node is indexing.

jbibla:
😭

app/src/renderer/connectors/rpcWrapper.js
```diff
@@ -24,15 +23,15 @@ module.exports = function setRpcWrapper(container) {
 
       rpcWrapper.rpcInfo.connected = false
     },
-    rpcConnect(nodeIP) {
-      rpcWrapper.rpcInfo.nodeIP = nodeIP
+    rpcConnect(rpcURL) {
+      let rpcHost = rpcURL.startsWith(`http`) ? rpcURL.split(`//`)[1] : rpcURL
```
fedekunze: this will throw if the URL starts with `http` but doesn't contain `//`, we should test both cases

faboweb: which url starts with `http` but doesn't contain `//`?

faboweb:
done @fedekunze

faboweb:
I would say we support at some point, that users can enter an URL to connect to: #1312

PR #1415, opened 2018-10-05
Jordan/1173 addresses

### LINKED ISSUE #1173: Cosmos address component
the UI can abbreviate addresses and pubkeys for a friendlier and less cluttered UI. 

though, bech32 prefixes have made our address _a bit_ more readable - they are still for computers (let's be honest).

i'd like to propose that we abbreviate addresses and pubkeys - the tooltip could display the full address. see below.

this will result in a more readable, less overwhelming, cleaner UI. 🥇 
```
cosmosvalpub12ztjt9nfj20cxw4w4fdg5g0lak83mc23csqzgx
cosmosvalpub...qzgx

cosmosaccaddr1qk2q3l28t6qhzyc7psdffsdf832ep8uwdaer
cosmosaccaddr...daer
```

### LINKED ISSUE #1437: Refactor ShortAddress
Currently the component handles other formats that are not bech32. I don't hink we need this and it also hides wrong addresses.
Additionally we want to rename ShortAddress to ShortBech32.

jbibla:
Closes #1173 
Closes #1437 

Description:
- new `ShortBech32` component that includes the short address script, copy ux, and a tooltip
- lots of updates to address format in tests
- improved `TmBalance` css 

❤️ Thank you!


test/unit/specs/components/staking/__snapshots__/LiValidator.spec.js.snap
```diff
@@ -26,12 +26,29 @@ exports[`LiValidator has the expected html structure 1`] = `
         <div
           class="top"
         >
-          mr_mounty
+          herr_schmidt_revoked
```
jbibla: 🐱

app/src/renderer/components/common/ShortAddress.vue
```diff
@@ -0,0 +1,79 @@
...
+<script>
+import { clipboard } from "electron"
+export default {
+  name: `short-address`,
+  props: [`address`],
+  data: () => ({
+    showSuccess: false
+  }),
+  computed: {
+    shortAddress({ address } = this, length = 4) {
```
fedekunze: question: do we always want to show the last 4 chars ? maybe getting the `length` value from the props is a better option ?

app/src/renderer/components/common/ShortAddress.vue
```diff
@@ -0,0 +1,79 @@
...
+  .address(@click.prevent.stop="copy" v-tooltip.top="address")#address {{ shortAddress }}
+  .success(:class="{active:showSuccess}")
+    i.material-icons check
+    span Copied
+</template>
+
+<script>
+import { clipboard } from "electron"
+export default {
+  name: `short-address`,
```
fedekunze: I'd rename the component to `short-bech32` since we are using it for addresses and pubkeys

app/src/renderer/components/common/ShortAddress.vue
jbibla: right now, we are going to show only the last 4 - but it will be easy to change it if we decide we should.

app/src/renderer/components/common/ShortAddress.vue
jbibla: i think this is a good suggestion!

fedekunze: we can change the name later tho, nbd

test/unit/specs/components/staking/LiValidator.spec.js
```diff
@@ -2,32 +2,50 @@ import setup from "../../../helpers/vuex-setup"
...
 
-  beforeEach(async () => {
-    let test = instance.mount(LiValidator, {
+  beforeEach(() => {
+    let instance = mount(LiValidator, {
       propsData: {
         validator: {
           id: `abc`,
-          description: {}
+          pub_key: `valpub123456789`,
```
fedekunze: you can use the lcd mocked validators instead 👍

app/src/renderer/components/common/ShortAddress.vue
faboweb: Good suggestion.

NodeGuy: Yes, I agree.

jbibla: agree that i should rename it or agree that it can wait until later? (or both)

fedekunze: I think they were talking about renaming the component

faboweb: I think we should rename it. We can so later.

app/src/renderer/components/common/ShortAddress.vue
```diff
@@ -0,0 +1,79 @@
...
+  props: [`address`],
+  data: () => ({
+    showSuccess: false
+  }),
+  computed: {
+    shortAddress({ address } = this, length = 4) {
+      if (!address) return `Address Not Found`
+      if (address.indexOf(`1`) === -1) {
```
faboweb: How can a bech32 address not have a 1 in it?

app/src/renderer/components/wallet/PageWallet.vue
```diff
@@ -14,8 +14,7 @@ tm-page(data-title="Wallet", :title="config.devMode ? '' : 'Wallet'")
     tm-list-item(
       :title="wallet.address"
       :btn="'Receive'"
-      :overflow="true"
-      @click.native="copy")
```
faboweb: Why was this removed?

test/unit/specs/components/common/ShortAddress.spec.js
```diff
@@ -0,0 +1,73 @@
...
+    wrapper.update()
+    expect(wrapper.vm.shortAddress).toBe(`add1asd`)
+
+    wrapper.setProps({ address: `add2asd` })
+    wrapper.update()
+    expect(wrapper.vm.shortAddress).toBe(`add2asd`)
+  })
+
+  it(`should return a short address with the first 4 letters`, () => {
+    wrapper.setProps({ address: `cosmosaddress2asdfasdfasdf` })
```
faboweb: shouldn't this fail as it is not bech32?

test/unit/specs/components/common/ShortAddress.spec.js
```diff
@@ -0,0 +1,73 @@
...
+        .includes(`active`)
+    ).toBe(false)
+    wrapper.find(`.address`).trigger(`click`)
+    expect(
+      wrapper
+        .find(`.success`)
+        .classes()
+        .includes(`active`)
+    ).toBe(true)
+    await sleep(3500)
```
faboweb: please do not use sleep, use `jest.useFakeTimers()`

test/unit/specs/components/common/TmBalance.spec.js
```diff
@@ -4,10 +4,10 @@ import TmBalance from "common/TmBalance"
 
 describe(`TmBalance`, () => {
   let wrapper
-  let { mount } = setup()
+  let { shallow } = setup()
 
   beforeEach(() => {
-    let instance = mount(TmBalance, {
+    let instance = shallow(TmBalance, {
```
faboweb: why was this changed?

test/unit/specs/components/staking/__snapshots__/TabValidators.spec.js.snap
```diff
@@ -89,21 +94,26 @@ exports[`PageStaking has the expected html structure 1`] = `
...
             <div class=\\"top\\">
                mr_mounty
             </div>
-      <div class=\\"bottom\\">
-         cosmosvaladdr…8yqw
+      <div class=\\"short-address\\">
+         <div id=\\"address\\" class=\\"address\\">
+            cosmosvalpub1234
+         </div>
+      <div class=\\"success\\">
```
faboweb: It is a little strange to have this DOM everywhere. How about using `v-if` instead of an `.active` class?

test/unit/specs/components/wallet/PageWallet.spec.js
```diff
@@ -6,7 +6,10 @@ describe(`PageWallet`, () => {
 
   beforeEach(async () => {
     let instance = mount(PageWallet, {
-      stubs: { "modal-search": `<modal-search />` }
+      stubs: {
+        "modal-search": `<modal-search />`,
+        "short-address": `<short-address />`
```
faboweb: this would also hide api incompatibility

app/src/renderer/components/common/ShortAddress.vue
jbibla: not sure. i copied this code from `scripts/common`.

app/src/renderer/components/wallet/PageWallet.vue
jbibla: the ux for copying form inputs was unsatisfactory. we can discuss ways of improving this.

the short address component is on every page and includes functionality for copying to clipboard (and the tooltip). no more copying to clipboard triggering notify.

test/unit/specs/components/common/ShortAddress.spec.js
fedekunze: why `cosm…asdf` ? shouldn't it show the whole prefix + the last 4 chars ?

faboweb: I created a follow up: https://github.com/cosmos/voyager/issues/1437

test/unit/specs/components/common/TmBalance.spec.js
faboweb: Example: If you use a property `address` for `ShortAdress` in `TmBalance` but the actually property used in `ShortAdress` is `bech32` the `TmBalance` test will pass even though the actual application will fail.

app/src/renderer/components/common/ShortBech32.vue
```diff
@@ -1,7 +1,7 @@
 <template lang="pug">
 .short-bech32
   .address(@click.prevent.stop="copy" v-tooltip.top="address")#address {{ shortBech32 }}
-  .success(:class="{active:showSuccess}")
+  .copied(:class="{active:showSuccess}")
```
faboweb: my comment wasn't about the class but about the DOM element being there all the time. did you check if this has sth. todo with the animation?

jbibla: ah i misunderstood. still, i think this is better. 

it _does_ have to do with the animation. currently, the element is `position: relative` and so if you use a `v-if` then it will bump things around it. also, transitions work when changing css properties, not inserting new elements into the DOM.

it could be changed to allow for a `v-if` but i'm not sure that would be _better_.

faboweb: hm ok

PR #1429, opened 2018-10-08
Fabo/1363 fix delegates test

### LINKED ISSUE #1363: Reconnection test for delegates is bad
Test test "should query for delegates on reconnection" in `delegates.spec.js` produces a snaphot that shows 0 calls to the endpoint where it should show more then one call.

faboweb:
Closes #1363

Description:
Refactoring of the delegates module tests

❤️ Thank you!

### DIFF
--- a/app/src/renderer/vuex/modules/delegates.js
+++ b/app/src/renderer/vuex/modules/delegates.js
@@ -16,7 +16,12 @@ export default ({ node }) => {
       state.loading = loading
     },
     setDelegates(state, delegates) {
-      state.delegates = delegates
+      state.delegates = delegates.map(delegate => {
+        delegate.id = delegate.owner
+        delegate.voting_power = ratToBigNumber(delegate.tokens)
+
+        return delegate
+      })
 
       // update global power for quick access
       state.globalPower = state.delegates
@@ -25,40 +30,15 @@ export default ({ node }) => {
         }, new BN(0))
         .toNumber()
     },
-    addDelegate(state, delegate) {
-      delegate.id = delegate.owner
-
-      // TODO: calculate voting power
-      let divIdx = delegate.tokens.indexOf(`/`)
-      let tokens
-      if (divIdx == -1) {
-        tokens = Number(delegate.tokens)
-      } else {
-        tokens =
-          Number(delegate.tokens.substring(0, divIdx)) /
-          Number(delegate.tokens.substring(divIdx + 1))
-      }
-      delegate.voting_power = tokens.toFixed(2)
-
-      // update if we already have this delegate
-      for (let existingDelegate of state.delegates) {
-        if (existingDelegate.id === delegate.id) {
-          Object.assign(existingDelegate, delegate)
-          return
-        }
-      }
-
-      state.delegates.push(delegate)
-    },
     setSelfBond(state, { validator, ratio }) {
       state.delegates.find(c => c.owner === validator.owner).selfBond = ratio
     }
   }
 
   const actions = {
-    reconnected({ state, dispatch }) {
+    async reconnected({ state, dispatch }) {
       if (state.loading) {
-        dispatch(`getDelegates`)
+        await dispatch(`getDelegates`)
       }
     },
     resetSessionData({ rootState }) {
@@ -70,8 +50,8 @@ export default ({ node }) => {
           validator.pub_key
         )
         if (!isEmpty(signing_info)) validator.signing_info = signing_info
-        commit(`addDelegate`, validator)
       }
+      commit(`setDelegates`, validators)
     },
     async getDelegates({ state, commit, dispatch }) {
       commit(`setDelegateLoading`, true)
@@ -82,7 +62,6 @@ export default ({ node }) => {
         if (validators.find(v => v.pub_key === delegate.pub_key)) {
           delegate.isValidator = true
         }
-        commit(`addDelegate`, delegate)
       }
 
       commit(`setDelegates`, candidates)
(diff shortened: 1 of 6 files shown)


app/src/renderer/vuex/modules/delegates.js
```diff
@@ -16,7 +16,12 @@ export default ({ node }) => {
       state.loading = loading
     },
     setDelegates(state, delegates) {
-      state.delegates = delegates
+      state.delegates = delegates.map(delegate => {
+        delegate.id = delegate.owner
+        delegate.voting_power = ratToBigNumber(delegate.tokens)
+
+        return delegate
+      })
```
NodeGuy: It's better not to mutate an input if you can avoid it:

```JavaScript
state.delegates = delegates.map(delegate =>
  Object.assign({}, delegate, {
    id: delegate.owner,
    voting_power: ratToBigNumber(delegate.tokens)
  })
)
```

test/unit/specs/store/delegates.spec.js
```diff
@@ -1,103 +1,185 @@
...
-      tokens: `10`
-    })
-    expect(store.state.delegates.delegates[0]).toEqual({
-      id: `foo`,
-      owner: `foo`,
-      tokens: `10`,
-      voting_power: `10.00`
-    })
-    expect(store.state.delegates.delegates.length).toBe(1)
+    let { mutations, state } = module
```
NodeGuy: It's good to be in the habit of using `const` always instead of `let` because it lowers the complexity of the code.  If `mutations` is declared `const` then the reader's mind knows that it's always the same no matter where it is in the function.  If it's declared `let` then it's not possible to know which value it's holding unless one also pays close attention to time and the order of assignments.  It creates more work for everyone who reads your code later and makes bugs more likely.

app/src/renderer/vuex/modules/delegates.js
```diff
@@ -82,15 +62,14 @@ export default ({ node }) => {
...
-        commit(`addDelegate`, delegate)
       }
 
       commit(`setDelegates`, candidates)
       commit(`setDelegateLoading`, false)
       dispatch(`getKeybaseIdentities`, candidates)
       dispatch(`updateSigningInfo`, candidates)
 
-      return state.delegates
+      return candidates
```
fedekunze: let's use validators to match the new terminology

app/src/renderer/vuex/modules/delegates.js
```diff
@@ -70,10 +50,10 @@ export default ({ node }) => {
           validator.pub_key
         )
         if (!isEmpty(signing_info)) validator.signing_info = signing_info
-        commit(`addDelegate`, validator)
       }
+      commit(`setDelegates`, validators)
     },
-    async getDelegates({ state, commit, dispatch }) {
+    async getDelegates({ commit, dispatch }) {
```
fedekunze: `getValidators`

test/unit/specs/store/delegates.spec.js
faboweb: We decided to go with const.

app/src/renderer/vuex/modules/delegates.js
faboweb: Vue doesn't deep watch arrays. If I use this pattern, Vue doesn't know the validators changed in PageValidator. I don't want to handle this issue in this PR.

app/src/renderer/vuex/modules/delegates.js
faboweb: I won't do API changes in this PR as it would increase the complexity of this PR immensely.

faboweb:
A follow up issue for renaming delegates to validators already exists: https://github.com/cosmos/voyager/issues/1381

app/src/renderer/vuex/modules/delegates.js
NodeGuy: OK.  In the future it would be better then to use `forEach` to signal that you're deliberately making side effects:

```JavaScript
delegates.forEach(delegate => {
  delegate.id = delegate.owner
  delegate.voting_power = ratToBigNumber(delegate.tokens)
})

state.delegates = delegates
```

test/unit/specs/store/delegates.spec.js
NodeGuy: Who decided that?

app/src/renderer/vuex/modules/delegates.js
faboweb: I completely agree. My mistake.

app/src/renderer/vuex/modules/delegates.js
```diff
@@ -15,8 +15,12 @@ export default ({ node }) => {
     setDelegateLoading(state, loading) {
       state.loading = loading
     },
-    setDelegates(state, delegates) {
-      state.delegates = delegates
+    setDelegates(state, validators) {
+      validators.forEach(validator => {
+        validator.id = validator.owner
```
jbibla: is there any reason we can't just use `validator.owner`?

jbibla: another way of asking this question: do we need `validator.id` for something?

app/src/renderer/vuex/modules/delegates.js
```diff
@@ -25,40 +29,23 @@ export default ({ node }) => {
...
-      }
-      delegate.voting_power = tokens.toFixed(2)
-
-      // update if we already have this delegate
-      for (let existingDelegate of state.delegates) {
-        if (existingDelegate.id === delegate.id) {
-          Object.assign(existingDelegate, delegate)
-          return
-        }
+    setSelfBond(
```
jbibla: better name!

app/src/renderer/vuex/modules/delegates.js
faboweb: I don't think we need validator.id anymore. It was a way to secure that a fixed property on the validator holds something like an id. We have the owner now which should never change.

PR #1434, opened 2018-10-08
Distinguish keys in validator lists.

### LINKED ISSUE #1364: Duplicate keys on mystake page
The keys need to be prefixed as a validator may be in both lists:

image

NodeGuy:
Closes #1364 

Description:

Use unique v-for keys in validator lists in My Delegations.  I couldn't reproduce the bug (#1364) but this should prevent it from recurring.

### DIFF
--- a/app/src/renderer/components/staking/TabMyDelegations.vue
+++ b/app/src/renderer/components/staking/TabMyDelegations.vue
@@ -8,7 +8,11 @@
     data-empty-search(v-if="yourValidators.length === 0")
     template(v-else)
       ol
-        li-validator(v-for='validator in yourValidators' :key='validator.id' :validator='validator')
+        li-validator(
+          v-for='validator in yourValidators'
+          :key='`Your/${validator.id}`'
+          :validator='validator'
+        )
     .check-out-message
       | Check out
       |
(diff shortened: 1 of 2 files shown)


app/src/renderer/components/staking/TabMyDelegations.vue
```diff
@@ -221,13 +228,13 @@ h3
 
 .check-out-message
   background var(--app-fg)
+  border 1px solid var(--bc-dim)
```
faboweb: using some style formatting plugin?

jbibla: yes. he auto-sorts it alphabetically. it's beautiful. 😢 <-- tears of joy

faboweb: can prettier do this?

NodeGuy: No, [Prettier doesn't support Stylus](https://github.com/prettier/prettier/issues/2465).

jbibla: what i'm hearing is that we all want to switch to plain css 😉 

no his editor has a plugin or something like that.

NodeGuy: It's [Stylus Supremacy](https://thisismanta.github.io/stylus-supremacy/#).  Good news @jbibla, there's an Atom plugin!

jbibla: the best way to format stylus is to write css :trollface:

PR #1456, opened 2018-10-11
1376/david locked keystore

### LINKED ISSUE #1376: Workaround for locked keystore
The keystore in gaia gets locked by the process while in use. On reconnection we stop the gaia process and restart a new process. In some cases, the keystore is not released yet when we start the new gaia process. In that case the process crashes with following message:
```
gaiacli: panic: resource temporarily unavailable
````

Let's create a simple workaround for handling this case.

Proposed workaround:
- detect `resource temporarily unavailable` error
- wait a second
- retry connecting

### PR DESCRIPTION
**Pull request by NodeGuy**
Closes #1376 

Description:

Currently Gaia Lite reports the error `gaiacli: panic: resource temporarily unavailable` if two instances are run at the same time.  This PR moves the error upstream into Voyager as an assertion in `startLCD` to make it easier to diagnose bugs (and has already caught three).

### DIFF
--- a/app/src/main/index.js
+++ b/app/src/main/index.js
@@ -1,5 +1,6 @@
 "use strict"
 
+const assert = require(`assert`)
 let { app, BrowserWindow, ipcMain } = require(`electron`)
 let fs = require(`fs-extra`)
 let { join, relative } = require(`path`)
@@ -117,14 +118,14 @@ function signalNoNodesAvailable() {
   })
 }
 
-function shutdown() {
+async function shutdown() {
   if (shuttingDown) return
 
   mainWindow = null
   shuttingDown = true
 
   if (lcdProcess) {
-    stopLCD()
+    await stopLCD()
   }
 
   return Promise.all(
@@ -257,6 +258,12 @@ app.on(`ready`, () => createWindow())
 
 // start lcd REST API
 async function startLCD(home, nodeIP) {
+  assert.equal(
+    lcdProcess,
+    null,
+    `Can't start Gaia Lite because it's already running.  Call StopLCD first.`
+  )
+
   let lcdStarted = false // remember if the lcd has started to toggle the right error handling if it crashes async
   return new Promise(async (resolve, reject) => {
     log(`startLCD`, home)
@@ -308,10 +315,13 @@ function stopLCD() {
     try {
       // prevent the exit to signal bad termination warnings
       lcdProcess.removeAllListeners(`exit`)
-      lcdProcess.on(`exit`, resolve)
+
+      lcdProcess.on(`exit`, () => {
+        lcdProcess = null
+        resolve()
+      })
+
       lcdProcess.kill(`SIGKILL`)
-      lcdProcess = null
-      resolve()
     } catch (err) {
       handleCrash(err)
       reject(`Stopping the LCD resulted in an error: ` + err.message)
@@ -583,6 +593,7 @@ async function pickAndConnect(addressbook) {
       err
     )
     addressbook.flagNodeIncompatible(nodeIP)
+    await stopLCD()
     return await pickAndConnect(addressbook)
   }
 
@@ -590,7 +601,7 @@ async function pickAndConnect(addressbook) {
     let message = `Node ${nodeIP} uses SDK version ${nodeVersion} which is incompatible to the version used in Voyager ${expectedGaiaCliVersion}`
     log(message)
     mainWindow.webContents.send(`connection-status`, message)
-
+    await stopLCD()
     return await pickAndConnect(addressbook)
   }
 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: app/src/main/index.js [hunk 1]
```diff
@@ -308,10 +336,14 @@ function stopLCD() {
     try {
       // prevent the exit to signal bad termination warnings
       lcdProcess.removeAllListeners(`exit`)
-      lcdProcess.on(`exit`, resolve)
+
+      lcdProcess.on(`exit`, () => {
+        lcdProcess = null
```
jbibla: what is the thinking re: `null` instead of `false`?

app/src/main/index.js
```diff
@@ -583,14 +615,15 @@ async function pickAndConnect(addressbook) {
       err
     )
     addressbook.flagNodeIncompatible(nodeIP)
+    await stopLCD()
     return await pickAndConnect(addressbook)
```
jbibla: i know this is unrelated to your PR @NodeGuy but why are we still "picking and connecting" CC @faboweb ?

test/unit/specs/main.spec.js
```diff
@@ -575,22 +576,18 @@ describe(`Startup Process`, () => {
...
-      })
+
+      childProcess.spawn = () =>
+        Object.assign(mockSpawnReturnValue(), {
+          stderr: {
+            on: (type, cb) => {
+              // type is always 'data'
+              errorCB = cb
+            },
+            pipe: () => {}
```
jbibla: what's the use?

jbibla:
cool PR!

app/src/main/index.js
faboweb: false would mix different types for one variable (Promise, Boolean). in JS everything is possible. but it's better to create predictable code I think.

app/src/main/index.js
faboweb: can be renamed/refactored imo. feel free to open an issue.

test/unit/specs/main.spec.js
```diff
@@ -508,8 +511,10 @@ describe(`Startup Process`, () => {
       expect(send.mock.calls[0][1]).toBeTruthy() // TODO fix seeds so we can test nodeIP output
     })
 
-    it(`should try another node if user disapproved the hash`, async () => {
-      main.shutdown()
+    // Skip for now because this test takes 2 seconds and we're not making use
+    // of it yet.
+    it.skip(`should try another node if user disapproved the hash`, async () => {
```
faboweb: remove it ;)

app/src/main/index.js
jbibla: if i understand you correctly... you are saying `null` and `true` are required - adding `false` to the mix would be unnecessary complication?

app/src/main/index.js
jbibla: ahh! your PR isn't in yet!

app/src/main/index.js
NodeGuy: `false` is traditionally used for boolean propositions (e.g., "There is an `lcdProcess` running.").

`null` is used to communicate absence of value ("`lcdProcess` contains an instance of `ChildProcess` and in this case there isn't one.")

app/src/main/index.js
```diff
@@ -257,6 +258,12 @@ app.on(`ready`, () => createWindow())
 
 // start lcd REST API
 async function startLCD(home, nodeIP) {
+  assert.equal(
+    lcdProcess,
+    null,
+    `Can't start Gaia Lite because it's already running.  Call StopLCD first.`
```
fedekunze: how do I call `StopLCD` ? though the Command line ? If so we'd need to add it to the README or add a bit more detail to the message.

 Also it should be called `StopGaiaLite` instead of `StopLCD`

NodeGuy: Sorry for the confusion; that error message is intended for developers.  It raises a good question:  Should we have a separate channel for developer errors vs. user errors?  Should the user error for this situation show something like, "An error has occurred.  Please restart Voyager and try again."

@faboweb Do we have any precedent for this?

jbibla: if it appears in a modal, it's a user error and should: 
1. be in plain english
2. provide an action for the user to move forward

if it appears in the console, it's for developers and should:
1. be in more detailed plain english
2. provide an action for the developer to move forward

:trollface:  🌮 👍

NodeGuy: I think it already satisfies the second case.

faboweb: @fedekunze does this satisfy your request for change?

fedekunze: ok resolving then

fedekunze:
@NodeGuy fix conflicts and we should be good to merge

PR #1462, opened 2018-10-11
David/1194 contract tests

### LINKED ISSUE #1194: Start writing contract tests.
We depend on the Gaia Lite REST API in the SDK.  Sometimes it changes, which breaks our code.

Both the SDK and Voyager have tests for the behavior of the API.  We can unify some of the two separate tests suites into contract tests, making sure the tests are the same on both sides, therefore ensuring agreement about what the behavior should be.  The contract will be written by Voyager so that it's a [consumer-driven contract](https://martinfowler.com/articles/consumerDrivenContracts.html).  This means that the contract cannot be changed without communicating with the Voyager team.

This will have the following benefits:

* We'll be reusing test code in both the SDK and Voyager, reducing complexity.
* Because the tests will be the same in both projects, divergent expectations about behavior will be prevented.
* The contract will be a natural forcing function to trigger communication about breaking changes between the two teams.
* We can use the contract tests in favor of some of our end-to-end tests (see [7 Reasons to Choose Consumer-Driven Contract Tests Over End-to-End Tests](https://reflectoring.io/7-reasons-for-consumer-driven-contracts/)), reducing the problems we've been having with the E2E tests (slowness, non-deterministic, difficulty in diagnosing failures).

We can start by creating a single, simple test and then adding more over time.  [Pact](https://docs.pact.io/readme) is the de facto standard tool for creating contract tests.  It has libraries for [JavaScript](https://github.com/pact-foundation/pact-js) and [Go](https://github.com/pact-foundation/pact-go/).

NodeGuy:
Closes #1194

Description:

Tests `keys` endpoints in lcdClient against a copy of the Swagger specification from the SDK.  The PR uses a copy because the version of the SDK that Voyager is using doesn't include the spec yet.


test/unit/specs/lcdClient.spec.js
```diff
@@ -1,229 +1,459 @@
...
-  it(`queries for staking pool`, async () => {
-    axios.get = jest.fn().mockReturnValue({})
-    await client.getPool()
-    expect(axios.get.mock.calls[0]).toEqual([
-      `${remoteLcdURL}/stake/pool`,
-      undefined
-    ])
-  })
+        expect(await client.queryAccount(`address`)).toEqual({
+          type: `auth/Account`,
```
faboweb: I think this is a bug in the LCD and should only return the value not the type. Not a problem here. I file an issue.

faboweb: https://github.com/cosmos/cosmos-sdk/issues/2503

test/unit/specs/lcdClient.spec.js
```diff
@@ -1,229 +1,459 @@
...
+            response: {
+              data: `account bytes are empty`
+            }
+          })
+        )
+        let res = await client.queryAccount(`address`)
+        expect(res).toBe(null)
+      })
+
+      it(`does not throw error for empty results`, async () => {
```
faboweb: dupe?

app/src/renderer/connectors/lcdClientMock.js
```diff
@@ -244,16 +234,21 @@ module.exports = {
...
+    let key = state.keys.find(k => k.name === name)
+    if (key.password !== password) {
+      throw new Error(`Passwords do not match`)
+    }
+    state.keys = state.keys.filter(k => k.name !== name)
   },
-  async updateKey(account, { name, old_password, new_password }) {
+
+  get: async name =>
+    name === `seed`
```
faboweb: I have the feeling that putting the request for a seed here will cause confusion. This is correctly mirroring the API but a developer wouldn't know of this `magic` key name. I would add another method to query for a seed. But not feeling too strongly.

fedekunze: this is handling the case when it calls `GET keys/{name}` where `{name} = seed`. I also think we should change it but on upstream (_i.e_ rename the endpoint on the LCD)

test/unit/helpers/Gaia-Lite.yaml
```diff
@@ -0,0 +1,910 @@
+swagger: "2.0"
```
fedekunze: Let's create an issue to use the same file as the SDK has

fedekunze: See #1467

test/unit/specs/lcdClient.spec.js
NodeGuy: Yes, nice catch!

app/src/renderer/connectors/lcdClientMock.js
NodeGuy: I agree with both of you:  We should fix it upstream and then create a new API to mirror the fixed LCD when that happens.

NodeGuy:
@fedekunze Your merge broke the unit tests.

app/src/renderer/connectors/lcdClientMock.js
faboweb: I don't think the local API needs to reflect the actual API 100%. The same way the actual API doesn't need to (more like shouldn't) reflect the DB or state machine imo. I think we should fix this here and adjust the endpoints when the upstream change is implemented.

NodeGuy:
I'm in the process of fixing it.

app/src/renderer/connectors/lcdClientMock.js
NodeGuy: OK, I'll do it in this one.

PR #1539, opened 2018-11-08
fedekunze/1131 show only err msgs on notifications

### LINKED ISSUE #1131: Parse LCD errors in the lcdClient
UI Version: 0.9.4

Description:

The LCD delivers errors as some complex string that includes the actual error message. We want to show only the final error to the user. In delegation we are parsing this error already once (#1094). But it would be nice, to parse any error that is delivered in that format and only return that error message to the rest of the system, abstracting the error format away from the Voyager logic.

fedekunze:
Closes #1131

_Description:_

❤️ Thank you!

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 * [\#1497](https://github.com/cosmos/voyager/issues/1451) Using an html table for table validators component @jbibla
 * [\#1496](https://github.com/cosmos/voyager/issues/1496) display validator pub_key instead of operator_address on livalidator and validator profile @jbibla
 * made running a local node easier by reducing it to 2 commands and presetting an account. @faboweb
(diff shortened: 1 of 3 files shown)


app/src/main/index.js
```diff
@@ -82,14 +82,14 @@ function sleep(ms) {
   return new Promise(resolve => setTimeout(resolve, ms))
 }
 
-function handleCrash(error) {
+function handleCrash(err) {
```
faboweb: going from a non abbreviated word to an abbreviation feels weird.

faboweb:
Do you think this is an advantage? I think the main problem was showing the correct errors from the LCD and not sth like `500 The server produced and error`.

app/src/main/index.js
fedekunze: `error` is a reserved word in some programming languages. Also it makes it easier to search retrieved error responses

fedekunze:
> Do you think this is an advantage? I think the main problem was showing the correct errors from the LCD and not sth like 500 The server produced and error.

The correct errors from the LCD are fixed on the SDK. The missing part was verifying that we didn't append any additional message on top of them

PR #1542, opened 2018-11-08
Fabo/1509 crash tests on warnings

### LINKED ISSUE #1509: Unit tests produce errors and warnings but do not fail
Warnings and errors should produce an exit code 1 to show the tests have failed.

faboweb:
Closes #1509 

_Description:_

readded the `console.error` and `console.warn` throw patches. but also added a quick way to opt out of them.

❤️ Thank you!

---


jbibla:
why do you want to add this back? it caused issues for us... maybe add a flag to opt in instead?

NodeGuy:
> why do you want to add this back? it caused issues for us... maybe add a flag to opt in instead?

Strong agreement.  This caused problems before but you haven't yet demonstrated a benefit for adding it back on.  So far the two examples we've discussed are currently passing tests that would be broken by this change.

I'd prefer that you discuss changes like this (undoing previously discussed and agreed upon work) before writing code.

faboweb:
My approach as discussed on Slack:
Apart from the two tests stated, there were warnings in TableValidators, where the used mock data didn't fit the components requirements and nobody noticed/cared.
I disagree that errors/warnings in the console is an acceptable behavior in the tests. Having them acceptable hides real problems. I would like us to have a safety net.
If you know, that your test will produce an error output you can always mock console.error / console.warning for this individual test.
The problem in the past was, that in the case of a console.error, you couldn't identify where it was called from so you would need to comment all the code out. I simplified this process for you so that you just need to call `ALLOW_CONSOLE=true jest {you file name}`. I thought about using the `CI` flag that is automatically set on the CI, so that only the CI runs conservative tests. But from experience in this team, it is better to have the local tests run exactly as the ones on the CI.

jbibla:
> I disagree that errors/warnings in the console is an acceptable behavior in the tests

i think i agree with you here @faboweb - can the tests fail without obscuring the logs though?

> If you know, that your test will produce an error output you can always mock console.error / console.warning

is there a situation where the expected behaviour is a console error?

>  it is better to have the local tests run exactly as the ones on the CI.

yep, i agree.

faboweb:
> can the tests fail without obscuring the logs though?

I didn't fine a possibility to do that.

> is there a situation where the expected behaviour is a console error?

yes in the case of the failing contract test David mentioned

Issue #1378, opened 2018-09-28
Steak is now called STAKE


jbibla:
find and replace `steak` with `stake`

faboweb:
This is blocked until the staking denom of the network actually says `stake`.

fedekunze:
@faboweb I think @jbibla refers to the mocked LCD mostly, since we can get the testnet bonding denom from staking `params`. So technically it's not blocked

faboweb:
If we change this, the staking denom we manually set will mostly say `stake` but in the wallet it will still show the actual denom `steak`.

jbibla:
when we release voyager for gaia-9000 - `steak` should no longer exist. in the mocked LCD and in the UI i suppose.

faboweb:
unblocking this as our next release will not be before gaia-9000

faboweb:
Can we make it both `stakes`?

zmanian:
I would need get @greg-szabo  to sign the `gen-txs` again.

I need to figure out how to get the validator bonding txs to be in `stake` before GoS but I don't want to block 9000 on it.

greg-szabo:
I ran the `testnet` command, apparently it's not reset there?

zmanian:
Depends on https://github.com/cosmos/cosmos-sdk/pull/2752/files

faboweb:
Doesn't work with SDK version 0.26.1. It still requires the staking denom to be "steak".

PR #1549, opened 2018-11-11
WIP: fedekunze/1548 mocked submit deposit

### LINKED ISSUE #1548: add submit deposit to lcdClientMock
Description:

It's not currently on the mocked lcdClient and we are going to need to implement the deposit feature

fedekunze:
Closes #1548 

_Description:_

❤️ Thank you!

---

### DIFF
--- a/app/src/renderer/connectors/lcdClientMock.js
+++ b/app/src/renderer/connectors/lcdClientMock.js
@@ -1,4 +1,6 @@
 "use strict"
+
+import moment from "moment"
 const b32 = require(`../scripts/b32.js`)
 const { getHeight } = require(`./rpcWrapperMock.js`)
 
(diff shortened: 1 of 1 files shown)


app/src/renderer/connectors/lcdClientMock.js
```diff
@@ -353,13 +355,13 @@ let state = {
       description: `custom text proposal description`,
       initial_deposit: [
         {
-          denom: `stake`,
+          denom: `steak`,
```
faboweb: didn't we want to move towards `stake`?

app/src/renderer/connectors/lcdClientMock.js
```diff
@@ -891,6 +916,142 @@ module.exports = {
...
+      } else if (!coinBalance || coinBalance.amount < depositCoinAmt) {
+        results.push(txResult(1, `Not enough ${coin.denom}s in your account`))
+        return results
+      }
+
+      coinBalance.amount -= depositCoinAmt
+
+      // ============= TOTAL PROPOSAL's DEPOSIT =============
+      // Increment total deposit of the proposal
+      let depositIndex = proposal.total_deposit.findIndex(
```
faboweb: I think you would reduce the complexity to read the code a little by using `.find` instead of `.findIndex` as you can edit the found deposit and do not need to query it again via the index.
What was your thought process to use `.findIndex`?

app/src/renderer/connectors/lcdClientMock.js
fedekunze: yeah but it's not going to be used until GoS and the denom for governance is still steak

app/src/renderer/connectors/lcdClientMock.js
```diff
@@ -891,6 +916,142 @@ module.exports = {
...
+      // ============= USER'S DEPOSITS =============
+      // check if there's an existing deposit by the depositer
+      let prevDepositIndex = state.deposits[proposal_id].findIndex(
+        deposit => deposit.depositer === depositer
+      )
+
+      if (prevDepositIndex === -1) {
+        // if no previous deposit by the depositer, we add it to the existing deposits
+        state.deposits[proposal_id].push(submittedDeposit)
+        break // break since no need to iterate over other coins
```
faboweb: As this is the only break, you could also loop once through all deposits checking for an existing deposit. This would reduce the if then elses in the second "update deposits" loop.

app/src/renderer/connectors/lcdClientMock.js
```diff
@@ -891,6 +916,142 @@ module.exports = {
...
+    // TODO: double check if we need to update it inside the loop
+    incrementSequence(fromAccount)
+
+    // check if the propoposal is now active
+    if (proposal.proposal_status === `Pending`) {
+      // TODO: get min deposit coin from gov params instead of stake params
+      let depositCoinAmt = proposal.total_deposit.find(coin => {
+        return coin.denom === `steak`
+      }).amount
+      if (parseInt(depositCoinAmt) >= 10) {
```
faboweb: >= 10 is the min deposit?

test/unit/specs/lcdClientMock.spec.js
```diff
@@ -884,6 +884,269 @@ describe(`LCD Client Mock`, () => {
...
+
+          // should have added the deposit from the depositer
+          let userDepositAfter = await client.getProposalDeposit(
+            `5`,
+            lcdClientMock.addresses[0]
+          )
+          expect(userDepositAfter).toEqual(deposit)
+        })
+
+        it(`when user has previous deposit`, async () => {
```
faboweb: I really like the very understandable test description! Great work.

app/src/renderer/connectors/lcdClientMock.js
fedekunze: yup, but maybe because is an `sdk.Dec` I need to multiply it ? thoughts ?

test/unit/specs/lcdClientMock.spec.js
fedekunze: haha otherwise I couldn't understand it either

app/src/renderer/connectors/lcdClientMock.js
```diff
@@ -891,6 +916,130 @@ module.exports = {
...
+          prevDeposit.amount.push(coin)
+        } else {
+          // there's a previous deposit from the depositer with the same coin
+          let newAmt = parseInt(prevDepCoin.amount) + parseInt(coin.amount)
+
+          prevDepCoin.amount = String(newAmt)
+        }
+      }
+    }
+    // TODO: double check if we need to update it inside the loop
```
fedekunze: @faboweb is this ok ?

app/src/renderer/connectors/lcdClientMock.js
fedekunze: yeah I started using `.find()` but then I changed Idk, why though... updated

app/src/renderer/connectors/lcdClientMock.js
fedekunze: yeah I started using .find() but then I changed Idk why though... updated

app/src/renderer/connectors/lcdClientMock.js
fedekunze: yeah but it's an `sdk.Dec`, do I need to multiply the amount as well ?

app/src/renderer/connectors/lcdClientMock.js
faboweb: sure, but this is just the mock implementation and not the real network. does it conflict with settings for the real network?

app/src/renderer/connectors/lcdClientMock.js
faboweb: I was only implying, that this could use a comment or taking the value of the min deposit from a global constant.

app/src/renderer/connectors/lcdClientMock.js
faboweb: I think you are good there.

app/src/renderer/connectors/lcdClientMock.js
fedekunze: > taking the value of the min deposit from a global constant.

Indeed, it's actually part of the governance params, which aren't yet merged on the SDK. I'll add a comment

fedekunze: #1564

app/src/renderer/connectors/lcdClientMock.js
fedekunze: not really, but I'd have to add another proposal or refactor it to change that, bc it'll probably break other unit tests that use proposals. Can I do that on the scope of #1378 ?

faboweb: Sure. I just did that PR. Let's see afterwards what breaks. 💥

PR #1558, opened 2018-11-13
Spike separate window for account creation.

### LINKED ISSUE #1519: Spike for Create separate process to display/capture seed words.
Spike for https://github.com/cosmos/voyager/issues/1359.

NodeGuy:
Closes #1519

_Description:_

This PR is not intended to be merged and it doesn't matter if the tests pass or not.

Things to note:

* A separate BrowserWindow is created just for displaying the `Create Account` form (with the seed words).
* The seed words are requested from Gaia Lite via HTTP and displayed.
* No modules are imported.
* Vue is not used.
* Custom HTML is used with a section for CSS (but I didn't bother to fill it in).
* It will look fairly seamless for the user—just like the current `Create Account` page.
* Type in a name under `Account Name` and hit `Next`.  The window will close and the name will be sent via IPC back to the main process.  You can see it displayed in the terminal.

Conclusion:  All of the mechanisms required to make this work have been demonstrated.
---

### DIFF
--- a/app/src/main/index.js
+++ b/app/src/main/index.js
@@ -20,6 +20,7 @@ global.config = config // to make the config accessable from renderer
 require(`electron-debug`)()
 
 let shuttingDown = false
+let createAccountWindow
 let mainWindow
 let lcdProcess
 let streams = []
@@ -122,6 +123,19 @@ async function shutdown() {
   })
 }
 
+const showCreateAccountWindow = parent => {
+  createAccountWindow = new BrowserWindow({
+    center: true,
+    modal: true,
+    parent,
+    webPreferences: {
+      devTools: false
+    }
+  })
+
+  createAccountWindow.loadURL(`${winURL}/createAccount.html`)
+}
+
 function createWindow() {
   mainWindow = new BrowserWindow({
     show: false,
@@ -139,6 +153,8 @@ function createWindow() {
   mainWindow.once(`ready-to-show`, () => {
     setTimeout(() => {
       mainWindow.show()
+      showCreateAccountWindow(mainWindow)
+
       if (DEV || JSON.parse(process.env.COSMOS_DEVTOOLS || `false`)) {
         mainWindow.webContents.openDevTools()
         // we need to reload at this point to make sure sourcemaps are loaded correctly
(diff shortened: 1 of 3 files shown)


faboweb:
> The seed words are requested from Gaia Lite via HTTP and displayed.

The could be got directly from the CLI.

app/src/main/index.js
```diff
@@ -428,6 +444,11 @@ const eventHandlers = {
 
   "successful-launch": () => {
     console.log(`[START SUCCESS] Vue app successfuly started`)
+  },
+
+  createAccount: (event, name) => {
```
faboweb: would probably need name + seed?

app/src/main/index.js
```diff
@@ -139,6 +153,8 @@ function createWindow() {
   mainWindow.once(`ready-to-show`, () => {
     setTimeout(() => {
       mainWindow.show()
+      showCreateAccountWindow(mainWindow)
```
faboweb: should probably be created as a result of the main UI starting an account creation. but I guess this was not intended here in terms of it being a spike.

faboweb:
This should then probably also be used for account imports, am I correct?

NodeGuy:
> Why is it not intended to be merged ?

Because it's a spike.  The purpose of a spike is to reduce risk, not to create usable code.

app/src/main/index.js
NodeGuy: Yes.  The word "spike" refers to something that pierces as many layers of the software as possible, narrow instead of broad, so I deliberately sent only one piece of information that will ultimately be passed through.

app/src/main/index.js
NodeGuy: Correct.

faboweb:
> The could be got directly from the CLI.

Well then we would reduce the attack surface more as we communicate the seed only directly between two processes.

NodeGuy:
> Well then we would reduce the attack surface more as we communicate the seed only directly between two processes.

Oh right, because of the HTTPS issue.  It turns out we can do HTTPS from a renderer process but there's the extra complication of the certificate so you raise a good point.  I'll try it using the CLI.

faboweb:
Closing the spec as not needed anymore

PR #1569, opened 2018-11-13
Jordan/1556 misc fixes

### LINKED ISSUE #1556: misc. fixes
- `LiProposal` status indicator is not in the right place
- `TmDataError` missing from `PageProposal` and `PageTransactions`
- `PageValidator` back button goes to wrong tab
- proper messaging for `TabMyDelegations`
- proper messaging for `PageTransactions`

jbibla:
Closes #1556 

_Description:_

- `LiProposal` status indicator is not in the right place
- `TmDataError` missing from `PageProposal` and `PageTransactions`
- `PageValidator` back button goes to wrong tab
- proper messaging for `TabMyDelegations`
- proper messaging for `PageTransactions`
❤️ Thank you!

---

* [x] Added entries in `CHANGELOG.md` with issue # and GitHub username

### DIFF
--- a/app/src/renderer/vuex/modules/transactions.js
+++ b/app/src/renderer/vuex/modules/transactions.js
@@ -3,6 +3,7 @@ import { uniqBy } from "lodash"
 export default ({ node }) => {
   let emptyState = {
     loading: false,
+    error: false,
     wallet: [], // {height, result: { gas, tags }, tx: { type, value: { fee: { amount: [{denom, amount}], gas}, msg: {type, inputs, outputs}}, signatures} }}
     staking: [],
     governance: []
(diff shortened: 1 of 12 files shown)


app/src/renderer/vuex/modules/transactions.js
```diff
@@ -48,19 +52,26 @@ export default ({ node }) => {
...
-        transactions: allTxs
-      })
+        const walletTxs = await dispatch(`getTx`, `wallet`)
+        commit(`setWalletTxs`, walletTxs)
+
+        const allTxs = stakingTxs.concat(governanceTxs.concat(walletTxs))
+        await dispatch(`enrichTransactions`, {
+          transactions: allTxs
+        })
+      } catch (error) {
```
jbibla: i think we need to be catching errors in all the vuex modules - @faboweb @NodeGuy @fedekunze

faboweb: We should make sure, that the resulting errors appear in some log and are sent via Raven. So that we can act on odd behavior with remote clients.

PR #1605, opened 2018-11-20
Fabo/remove prepush test

faboweb:
Removes running of all unit tests on push.


fedekunze:
need to fix the changelog conflicts

Issue #1627, opened 2018-11-22
Display atoms instead of sub-atoms


faboweb:
Follow up of #1594

We are currently interpreting values (tokens) from the SDK wrong. Tokens will always be send in the form of "sub-atoms" where 1 atom = 10^18 sub-atoms. 
We should keep sub-atoms (as BigNumber) in the state. But we will show tokens in the form of 1 atom. This means we need to:
- [ ] increase the amount of default tokens on a local node to reflect this increased amount
- [ ] create a displaying function to show  atoms from sub-atoms
- [ ] create a input conversion function to convert user inputs to sub-atoms

faboweb:
Let's discuss this Chris.

jbibla:
> Tokens will always be send in the form of "sub-atoms" where 1 atom = 10^18 sub-atoms.

we discussed addressing this in the REST API instead of on the client. this will be important for all clients and wallet providers.

jbibla:
can we link this in the SDK repo and close this issue?

faboweb:
https://github.com/cosmos/cosmos-sdk/issues/3261

jbibla:
danke!

faboweb:
Will not be done in the SDK

PR #1643, opened 2018-11-26
fedekunze/1594 disable float amounts on transactions

### LINKED ISSUE #1594: Disable all transfer buttons if amount is not int
Description:

On the newest version I'm getting the following error if I try to delegate/deposit/send an amount with decimals

`math/big: cannot unmarshal "1.2" into a *big.Int`

fedekunze:
Closes #1594 

_Description:_

❤️ Thank you!

---


faboweb:
This is wrong :/
We allow floats. Just the sending needs to be converted to "sub atoms". See #1627

PR #1684, opened 2018-12-03
Fabo/1533 show waiting for connection

### LINKED ISSUE #1533: Disable any action buttons and loading when disconnected
Follow up on: https://github.com/cosmos/voyager/issues/1532

In the situation of a missing connection we need to show the user, he is not able to do any interaction with the app. So we need to
- disable all action buttons
- show "waiting for connection" messages when we show "loading" messages

faboweb:
Closes #1533 

_Description:_

Will need a follow up:
- [ ] Check all modal interactions for being disabled if offline
- [ ] Add unit test for reconnection if node not reachable

❤️ Thank you!

---


app/src/renderer/vuex/modules/delegates.js
```diff
@@ -10,6 +10,7 @@ export default ({ node }) => {
     delegates: [],
     globalPower: null,
     loading: false,
+    loaded: false,
```
jbibla: why do we need `loaded` and `loading`?

app/src/renderer/vuex/modules/governance/proposals.js
```diff
@@ -52,12 +58,13 @@ export default ({ node }) => {
...
         state.error = error
       }
-      state.loading = false
     },
     async getProposal({ state, commit }, proposal_id) {
       state.loading = true
       try {
         state.error = null
+        state.loading = false
+        state.loaded = true // TODO make state for only proposal
```
jbibla: can you clarify this `TODO`?

app/src/renderer/vuex/modules/send.js
```diff
@@ -14,6 +14,10 @@ export default ({ node }) => {
   }
 
   async function doSend({ state, dispatch, commit, rootState }, args) {
+    if (!rootState.connection.connected) {
+      throw Error(`Currently not connected to the send, try again later.`)
```
jbibla:
```suggestion
      throw Error(`Currently not connected to a secure node. Please try again when Voyager has secured a connection.`)
```

test/unit/specs/components/governance/PageGovernance.spec.js
```diff
@@ -4,7 +4,6 @@ import Vuelidate from "vuelidate"
 import PageGovernance from "renderer/components/governance/PageGovernance"
 import ModalPropose from "renderer/components/governance/ModalPropose"
 import lcdClientMock from "renderer/connectors/lcdClientMock.js"
-import { mount } from "@vue/test-utils"
```
jbibla: shouldn't we start using `test-utils` instead of `vuex-setup` magic? 🎩🐰

app/src/main/index.js
```diff
@@ -435,13 +426,7 @@ Object.entries(eventHandlers).forEach(([event, handler]) => {
...
 // test an actual node version against the expected one and flag the node if incompatible
 async function testNodeVersion(client, expectedGaiaVersion) {
-  let result
-  try {
-    result = await client.nodeVersion()
-  } catch (error) {
-    debugger
-    return { compatible: false, nodeVersion: null }
-  }
+  let result = await client.nodeVersion()
```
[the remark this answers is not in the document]
faboweb: Catching would make a version incompatibility and disconnectedness the same, where say need to be handled differently. As I removed the catch, the error now bubbles to the pickAndConnect function and can be handled there.

app/src/renderer/vuex/getters.js
```diff
@@ -67,7 +67,7 @@ export const pool = state => state.pool
 export const parameters = state => state.parameters
 
 // governance
-export const proposals = state => state.proposals.proposals
```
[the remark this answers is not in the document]
faboweb: proposals is the module containing also the connection states `loading`, `loaded` and `error` which we want to also use in some components. If `proposals` is  `state.proposals.proposals` we can't access i.e. `state.proposals.loading' .

app/src/renderer/vuex/modules/delegates.js
faboweb: `loading`: the module has requested data,
`loaded`: the module has loaded data once so we can show "cached" data,

if we haven't `loaded` data yet and are currently `loading` we show an according message

app/src/renderer/vuex/modules/governance/proposals.js
faboweb: As we are doing a request here, we should probably have `loading`, `loaded` states for it. I wasn't sure if we want another module or a substate `proposals.proposal.[loading, loaded, error]` here. Ideas?

app/src/main/index.js
```diff
@@ -533,6 +513,14 @@ async function connect() {
       rpcURL: config.node_rpc
     })
   })
+}
+
+async function connect() {
+  log(`starting gaia rest server with nodeURL ${config.node_lcd}`)
+
+  const gaiaLite = await startLCD(lcdHome, config.node_rpc)
```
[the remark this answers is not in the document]
faboweb: `connect()` bubbles errors to upstream functions so those functions know when connection was successful. Does that feel weird?

app/src/main/index.js
```diff
@@ -533,6 +513,14 @@ async function connect() {
...
+  lcdProcess = gaiaLite.process
```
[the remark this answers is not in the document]
faboweb: Done

app/src/renderer/components/governance/PageProposal.vue
```diff
@@ -158,9 +161,9 @@ export default {
   }),
   computed: {
     // TODO: get denom from governance params
-    ...mapGetters([`bondingDenom`, `proposals`]),
+    ...mapGetters([`bondingDenom`, `proposals`, `connected`]),
     proposal() {
-      let proposal = this.proposals[this.proposalId]
+      let proposal = this.proposals.proposals[this.proposalId]
```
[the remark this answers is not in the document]
faboweb: See comment above.

app/src/renderer/components/staking/TabMyDelegations.vue
```diff
@@ -1,9 +1,11 @@
 <template>
   <div>
-    <div v-if="yourValidators.length > 0">
+    <div v-if="delegation.loaded && yourValidators.length > 0">
       <table-validators :validators="yourValidators" />
     </div>
-    <tm-data-msg v-if="yourValidators.length < 1" icon="info_outline">
+    <tm-data-connecting v-if="!delegation.loaded && !connected" />
```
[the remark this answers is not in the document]
faboweb: I agree.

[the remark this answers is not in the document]
faboweb: https://github.com/cosmos/voyager/issues/1688

test/unit/specs/components/governance/PageGovernance.spec.js
faboweb: Using `test-utils` and specifying each moving part was more magic than `vuex-setup` for testing the templates. I.e. the router need to be specified. I think using the `vuex-setup` is the better choice in this case. For modules I agree with just testing the functions directly.

app/src/main/index.js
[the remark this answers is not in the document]
jbibla: good question. i'm not sure. but i'm ok with it now.

app/src/renderer/components/governance/PageProposal.vue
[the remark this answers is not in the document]
jbibla: not sure which one you are referring to?

app/src/renderer/vuex/getters.js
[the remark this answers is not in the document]
jbibla: hmm right ok. perhaps we can change the name of it then. `governance.proposals` or something. (not in scope for this PR - but worth considering as a refactor ticket)

app/src/renderer/vuex/modules/delegates.js
jbibla: awesome.

app/src/renderer/vuex/modules/governance/proposals.js
jbibla: ohhhh i see. good question. not sure what the vue-recommended way of doing this is.

test/unit/specs/components/governance/PageGovernance.spec.js
jbibla: what would @NodeGuy do 🤣

app/src/renderer/components/staking/TabMyDelegations.vue
```diff
@@ -12,12 +14,15 @@
...
-    <div v-if="yourValidators.length > 0" class="check-out-message">
+    <div
+      v-if="delegation.loaded && yourValidators.length > 0"
+      class="check-out-message"
+    >
       Check out
       <router-link :to="{ name: 'Validators' }">the validator list</router-link>
       to find other validators to delegate to.
     </div>
-    <div v-if="undelegatedValidators.length">
```
fedekunze: lol

app/src/renderer/vuex/modules/send.js
```diff
@@ -14,6 +14,12 @@ export default ({ node }) => {
   }
 
   async function doSend({ state, dispatch, commit, rootState }, args) {
+    if (!rootState.connection.connected) {
+      throw Error(
+        `Currently not connected to a secure node. Please try again when Voyager has secured a connection.`
```
fedekunze: what are the steps to secure a connection ? if it's just waiting we should display  `Please wait until Voyager has secured a connection`

test/unit/specs/components/governance/PageGovernance.spec.js
fedekunze: > what would @NodeGuy do 🤣
😢

app/src/renderer/vuex/modules/governance/deposits.js
```diff
@@ -27,7 +33,6 @@ export default ({ node }) => {
         Raven.captureException(error)
         state.error = error
       }
```
sabau: Just asking: In case of errors the state will remain { ..., loading: true}? Shouldn't we keep the state.loading = false in a finally block?

sabau: Got it! Alternatively I was thinking of a single property ~ enum with the possible states, but probably won't be easy to manage without types.

fedekunze:
need to fix conflicts and then we should be ok

app/src/renderer/vuex/modules/governance/deposits.js
faboweb: Also possible. Enum:
```
const Enum = {
  Loading: 1,
  Loaded: 2,
  Failed: 3
}

let state = Enum.Loading
```

But we could have an error and still retry loading in the background, so using multiple states at the same time. But I see you will have the better solution here in the end. ;)

app/src/renderer/vuex/modules/send.js
faboweb: This should not happen. All interaction should be disabled if not connected. This is just in case we screwed up. And in that case we currently don't send a transaction again automatically.

PR #1717, opened 2018-12-07
e2e test command [Proposal]

### LINKED ISSUE #1673: add instructions to run individual e2e tests
Description:

run all e2e tests: 
```
$ yarn test:e2e
``` 

run single a e2e test: 
```
$ yarn run pack
$ npx tape test/e2e/wallet.js
```

sabau:
_Description:_

- grouped README setions
- added default value if no TEST variable has been set
- *Before merging we should check and eventually cover Windows machine*

Closes #1673 

❤️ Thank you!

---


faboweb:
So using env variables on Windows is not that straight forward (no one liner). I added an alternative for Windows. Question is, if the extra command has a benefit if we need to declare two.

sabau:
Thanks for testing it and updating the README!

Is it working even with a clean environment? without having never ran the first part of the command: `cross-env ANALYTICS=false yarn run pack`

I would like to keep the command, it feels more standard than running from our node_modules  + hide an implementation detail from the user, so he does not have to know that we are using tape

Without the prefix to remove the tests from google analytics and the yarn pack, so the command takes care of this.

PR #1727, opened 2018-12-10
Fabo/1670 update balance after gov interactions

### LINKED ISSUE #1670: balance not updated on fee-only transactions
Description:

 
I don’t recall seeing my balance being updated after a redelegation or voting on a proposal. This transactions should update the balance as the user is still paying gas on them.

faboweb:
Closes #1670

_Description:_

Voyager is now querying the balance after any tx with a "sender", "depositer", "voter", "proposer" or "recipient" tag with the account of the currently logged in user.

❤️ Thank you!

---

### DIFF
--- a/app/src/renderer/vuex/modules/wallet.js
+++ b/app/src/renderer/vuex/modules/wallet.js
@@ -159,7 +159,8 @@ export default ({ node }) => {
         `tm.event = 'Tx' AND sender = '${state.address}'`,
         `tm.event = 'Tx' AND recipient = '${state.address}'`,
         `tm.event = 'Tx' AND proposer = '${state.address}'`,
-        `tm.event = 'Tx' AND depositer = '${state.address}'`
+        `tm.event = 'Tx' AND depositer = '${state.address}'`,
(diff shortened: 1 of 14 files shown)


app/src/renderer/vuex/modules/wallet.js
```diff
@@ -155,19 +155,22 @@ export default ({ node }) => {
...
-          query: `tm.event = 'Tx' AND sender = '${state.address}'`
-        },
-        onTx
-      )
+      const queries = [
+        `tm.event = 'Tx' AND sender = '${state.address}'`,
+        `tm.event = 'Tx' AND recipient = '${state.address}'`,
+        `tm.event = 'Tx' AND proposer = '${state.address}'`,
+        `tm.event = 'Tx' AND depositer = '${state.address}'`,
+        `tm.event = 'Tx' AND voter = '${state.address}'`
```
fedekunze: can we add also:
`tm.event = 'Tx' AND delegator = '${state.address}'`

fedekunze: (based on Billy's feedback)

PR #1728, opened 2018-12-10
Avoid double voting with same option

### LINKED ISSUE #1683: Governance Proposal voting
UI Version: 0.X.X

Description:

## Problem
After double voting on a proposal I get an error message saying that my action is already in the cache and I'm prevented to act on this proposal and on others too.

## Steps
- Create a proposal
- Deposit (I did it 1 + 2 + 40) 43 Steak then the Vote button is enabled
- Vote for it, let's say YES
- Vote again with the same value
- Every other action gets blocked: *Invalid Sequence. Got 9 expected 10*

## Possible solution
As far as I understood the root cause is the double voting with identical value, it will try to create two identical blocks, we can prevent this by storing in the state the last vote and disable just that value so you can't vote twice with the same value

Related: https://github.com/cosmos/cosmos-sdk/issues/3002

sabau:
_Description:_
- getVotes has been used instead of getVote as the latter throws error if the specified address did't vote yet
- disable the already voted option
- fixed setProposalVotes action parameters

Closes #1683

❤️ Thank you!

---

### DIFF
--- a/app/src/renderer/components/governance/PageProposal.vue
+++ b/app/src/renderer/components/governance/PageProposal.vue
@@ -121,6 +121,7 @@
         :show-modal-vote.sync="showModalVote"
         :proposal-id="proposalId"
         :proposal-title="proposal.title"
+        :last-vote-option="lastVote && lastVote.option"
         @castVote="castVote"
       />
     </template>
@@ -157,11 +158,18 @@ export default {
   },
   data: () => ({
     showModalDeposit: false,
-    showModalVote: false
+    showModalVote: false,
+    lastVote: undefined
   }),
   computed: {
     // TODO: get denom from governance params
-    ...mapGetters([`bondingDenom`, `proposals`, `connected`]),
+    ...mapGetters([
+      `bondingDenom`,
+      `proposals`,
+      `connected`,
+      `wallet`,
+      `votes`
+    ]),
     proposal() {
       let proposal = this.proposals.proposals[this.proposalId]
       if (proposal) {
@@ -244,8 +252,20 @@ export default {
     }
   },
   methods: {
-    onVote() {
+    async onVote() {
       this.showModalVote = true
+
+      try {
+        await this.$store.dispatch(`getProposalVotes`, this.proposalId)
+        this.lastVote =
+          this.votes[this.proposalId] &&
+          this.votes[this.proposalId].find(e => e.voter === this.wallet.address)
+      } catch ({ message }) {
+        this.$store.commit(`notifyError`, {
+          title: `Error while getting votes for proposal #${this.proposalId}`,
+          body: message
+        })
+      }
     },
     onDeposit() {
       this.showModalDeposit = true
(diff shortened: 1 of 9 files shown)


app/src/renderer/connectors/lcdClientMock.js
```diff
@@ -1184,8 +1184,11 @@ module.exports = {
     results.push(txResult(0))
     return results
   },
-  async getProposalVotes(proposalId) {
-    return state.votes[proposalId] || []
+  async queryProposalVotes(proposalId) {
+    return (
+      state.votes[proposalId] ||
+      Promise.reject(`Invalid proposalId ${proposalId}`)
```
fedekunze: we should use the error structure format from the SDK. I'd propose to leave it now as it is and modify this everywhere on another PR

app/src/renderer/components/governance/PageProposal.vue
```diff
@@ -244,8 +252,14 @@ export default {
     }
   },
   methods: {
-    onVote() {
+    async onVote() {
       this.showModalVote = true
+
+      // The error is already handled with notifyError in votes.js
+      await this.$store.dispatch(`getProposalVotes`, this.proposalId)
```
fedekunze: let's move this logic to `mounted()` instead, will need it afterwards for the `TabVotes` so better to make it available globally on the component, not just for the modal

test/unit/specs/store/governance/votes.spec.js
```diff
@@ -21,8 +21,11 @@ describe(`Module: Votes`, () => {
 
   it(`adds votes to state`, () => {
     let { mutations, state } = module
-    mutations.setProposalVotes(state, `1`, votes)
-    expect(state.votes[`1`]).toEqual(votes)
+    mutations.setProposalVotes(state, {
+      proposalId: proposals[`1`].proposal_id,
+      votes
+    })
+    expect(state.votes[proposals[`1`].proposal_id]).toEqual(votes)
```
fedekunze:
```suggestion
    expect(state.votes[`1`]).toEqual(votes)
```

app/src/renderer/components/governance/PageProposal.vue
sabau: Currently the data should be available globally in the lastVote property.
I think it's more effective to keep the dispatching bounded with the action, if we move the call away from the action it won't be refreshed, let's say we move it to mounted:
- I vote Yes and the modal closes
- click again to vote without leaving the proposal page
- Any button is disabled because the component is already mounted and the lastVote has not been updated
- I can vote Yes again

app/src/renderer/connectors/lcdClientMock.js
sabau: Agree! I just tried locally and currently:

https://localhost:9070/gov/proposals/3/votes => null
https://localhost:9070/gov/proposals/1/votes => [{proper array of votes}]

So instead of that message I should place the same value I've got from the sdk ok?

sabau: Ok we are currently missing a proper error message there, I will leave this message but wrapped in the proper object:

https://github.com/cosmos/cosmos-sdk/issues/3078

app/src/renderer/components/governance/PageProposal.vue
fedekunze: good argument, makes sense to me

PR #1738, opened 2018-12-11
1724 tab navigation for ModalProposals

### LINKED ISSUE #1724: Input and button order when using tab navigation
Description:

 If you try to navigate using tabs, the order not always reflect the rendered boxes, we should keep them aligned.

ModalsProposal i.e. is mostly OK, but amount comes after the readonly denom

enyan94:
Part of #1724 

Tab navigation now skips readonly denom in ModalProposals. This seems to be the only readonly element that was causing "improper" tab navigation.


faboweb:
Thank you very much for your contribution. :)

PR #1743, opened 2018-12-12
Added undelegation transactions to TabMyDelegation

### LINKED ISSUE #1387: display the unbonding tx in the "Unstaked Validators" list instead of the validator Li
Split from https://github.com/cosmos/voyager/issues/1364#issuecomment-425074861.

46092551-64d93d00-c1b5-11e8-8323-ea73b0505ba5

sabau:
### Description
- Added the list of transactions related to unbound.
- Changed the title of the unbounding validartor list 
- added title to the Active Validators list
- refactored a the modifiers in the delegations store
- updated e2e test

Graphical suggestions are really welcome :) I personally don't like the solution I pourpose here, but was a starting point to have the data ready by recycling the view.

An idea I had was to make the transacitons part of the table, so every transacitons will appear as subrow of the unbounding validator (eventually collapsable)

Closes #1387 

### Note
- as this PR grew enough I think the redelegation will be a separate one

We should create a separate PR for that
❤️ Thank you!

---

screenshot 2018-12-13 at 13 26 11


fedekunze:
See my comment on https://github.com/cosmos/voyager/issues/1387#issuecomment-446951028. Let's rename the tab to `My Validators` and change the the `Delegated steak` header from the second section to `Unbonded steak` so we can get rid of the unbonding tx. 

@faboweb what do you think ?

fedekunze:
@sabau had a cool idea to on make the `LiValidator` component a dropdown so when you click it, it displays the unbonding txs from that validator. Maybe we can implement that on another PR ?

faboweb:
> @sabau had a cool idea about making a dropdown on the LiValidator component so when you click it, it displays the unbonding txs from that validator.

sounds cool. but let's maybe sketch it first.

app/src/renderer/components/staking/TabMyDelegations.vue
```diff
@@ -24,30 +30,58 @@
     </div>
     <div v-if="delegation.loaded && undelegatedValidators.length > 0">
```
faboweb: Should we get rid of this section now?

test/e2e/delegation.js
```diff
@@ -62,11 +62,13 @@ test(`delegation`, async function(t) {
   t.test(`Stake`, async t => {
     let totalAtoms = (await app.client
       .$(`.header-balance .total-atoms h2`)
-      .getText()).split(`.`)[0] // 130.000...
+      .getText()).split(`.`)[0] // 30.000...
```
faboweb: huh? why is it suddenly 100 less?

faboweb: I think bonded atoms are not calculated in to the total atoms any longer. Needs to be fixed.

jbibla:
amazing work thank you @sabau! 

@fedekunze we don't use the B (bond, unbond, bonding, unbonding) word, bud. 😉 

> making a dropdown on the LiValidator component so when you click it, it displays the unbonding txs from that validator.

very cool idea. i think i would prefer if we showed a link to the transaction on the transaction page as opposed to the actual tx component - but i can mock this up. either way, agreed - should be in a different PR.

app/src/renderer/components/staking/TabMyDelegations.vue
```diff
@@ -1,6 +1,12 @@
 <template>
   <div>
     <div v-if="delegation.loaded && yourValidators.length > 0">
+      <h3 class="tab-header">
+        My active validators
```
jbibla:
```suggestion
        My Validators
```

test/unit/specs/components/staking/__snapshots__/TabMyDelegations.spec.js.snap
```diff
@@ -51,12 +51,29 @@ exports[`Component: TabMyDelegations should show a message if not staked yet to
...
 `;
 
 exports[`Component: TabMyDelegations should show committed validators 1`] = `
 <div>
   <div>
+    <h3
+      class="tab-header"
+    >
+      
+      My active validators
```
jbibla:
```suggestion
      My Validators
```

test/e2e/delegation.js
sabau: Exaclty this point I was a uncertain on how this changed!

app/src/renderer/components/staking/TabMyDelegations.vue
sabau: Yeah the initial idea was to just drop this and replace with the transactions, then I left to keep the discussion open on whether we want or not a kind of grouping per validator node for the u-transactions

app/src/renderer/components/staking/TabMyDelegations.vue
```diff
@@ -24,30 +30,58 @@
...
+    TmDataMsg,
+    TmDataConnecting,
+    TmDataLoading,
+    TmLiStakeTransaction
+  },
   data: () => ({
     bondInfo: `Validators you are currently bonded to`,
-    unbondInfo: `Your bonded validators in unbonding process`
+    unbondInfo: `Your bonded validators in unbonding process`,
+    unbondTransactions: `The transactions currently in unbouding period`
```
jbibla:
```suggestion
    unbondTransactions: `The transactions currently in unbonding period`
```

faboweb: #no-bonding ;)

faboweb:
My proposal:
- remove unbounding validators list
- rename "unbounding transactions" to "undelegation transactions"
- align transaction better with validator list
(- maybe have a better list component that fits better with the table layout of the validator list)

app/src/renderer/components/staking/TabMyDelegations.vue
fedekunze: that's how the term is called, unbonding period. We agreed not to use bonding instead of delegation

fedekunze:
My proposal:

- remove unbonding delegations transactions
- update LiValidator to show amount undelegated instead of delegated steak
- rename `My delegations` to `My validators`, bc the first is inaccurate

fedekunze:
I vote NoWithVeto on showing the undelegation txs

sabau:
I think we should discuss this synchronously. I would like to answer and all together agree on those two questions:
- What do we want to offer in this section exactly?
- What the user would need from this section and why?
  - I think he needs to see a reaction from his delegation and undelegation actions (so the lists were OK)
  - Probably he wants to know where he put his atoms
  - I would like to know when I will have my atoms back
  - How the list should look depends on what he can do, as for now he can do a single undelegation per validator = we have a single validator with a single amount a single period to show
     - but we are working in the direction of allowing several, so probably I would choose to work already in that direction, and show several amounts with different periods per validator or just not grouped at all but ordered from the most recent to the oldest

sabau:
Show just 2 list with minimal redisign until Staking 2
- Get rid of liValidator 2 (unbounding validator)
- make it nice

In Staking 2 group per validator the txs and we will decide the details there
- working on design in staking 2!

app/src/renderer/components/staking/TabMyDelegations.vue
```diff
@@ -23,31 +29,52 @@
       to find other validators to delegate to.
     </div>
     <div v-if="delegation.loaded && undelegatedValidators.length > 0">
```
faboweb: do we need the distinction into `undelegatedValidators`?

test/e2e/delegation.js
faboweb: Still an issue?

app/src/renderer/components/staking/TabMyDelegations.vue
sabau: Not necessarily, Thanks! `yourValidators` will always be a superset of it, I'm gonna remove `undelegatedValidators`

app/src/renderer/components/staking/TabMyDelegations.vue
```diff
@@ -23,31 +29,52 @@
...
         >
+          info_outline
+        </i>
       </h3>
-      <table-validators :validators="undelegatedValidators" />
+      <div class="transactions">
+        <template v-for="transaction in unbondingTransactions">
+          <tm-li-stake-transaction
+            :transaction="transaction"
+            :validators="undelegatedValidators"
```
sabau: and use here `yourValidators`

app/src/renderer/vuex/modules/delegation.js
```diff
@@ -33,32 +33,28 @@ export default ({ node }) => {
       state.delegates = state.delegates.filter(c => c.id !== delegate)
     },
     setCommittedDelegation(state, { candidateId, value }) {
-      let committedDelegates = Object.assign({}, state.committedDelegates)
+      state.committedDelegates[candidateId] = value
```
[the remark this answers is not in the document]
sabau: Oh no back then I missed this comment! I just fixed this

app/src/renderer/components/staking/TabMyDelegations.vue
```diff
@@ -22,52 +28,85 @@
...
-      )
-    },
     yourValidators({ committedDelegations, delegates: { delegates } } = this) {
       return delegates.filter(
         ({ operator_address }) => operator_address in committedDelegations
       )
-    }
+    },
+    unbondingTransactions: ({ allTransactions, delegation } = this) =>
+      allTransactions
```
faboweb: will this show only active unbonding transactions or all unbonding transactions?

sabau: Only unbonding, the filter underneath will remove all the others

faboweb: yes, but I have the feeling this will show unbonding transactions that are already passed the unbonding period. In this case they shouldn't show up anymore as the funds already have returned to the user. Am I mistaken?

sabau: correct, I just added a helper and used it here

test/e2e/delegation.js
sabau: fixed! 130.000 again

sabau: by replacing `state.stuff = x` with `Vue.set(state, 'stuff', x)`

app/src/renderer/components/staking/TabMyDelegations.vue
faboweb: NIce

PR #1751, opened 2018-12-14
Simplify local boot and enable multi node

### LINKED ISSUE #1661: add instructions to start a local testnets with X validators
Description:

Scope:

- Create a variable `N_VALIDATORS` that receives a positive integer to start a local testnet
- Update `README`

sabau:

Closes #1661 

_Description:_

- reduce side effects: 
  - at configuration level: I don't have to know that informations are being placed in my home folder, everything is local to the project and the network name I've chosen
  - at code level helpers are functional and don't rely on global status (first of a series of this kind)
  - You can import helpers without side effect (and test them lately 🎉)
  - if I clone several times the project I can expect them to run indipendently without iteracting on their mutual configurations (of course not at the same time :) )
- taking out the functions to work with multiple nodes
- on the road to reduce the code amount in the boot process for local node
- enabling build + running with one command for multiple nodes
- reduce hardcoded parameters, instead pass them through functions

❤️ Thank you!

---

### DIFF
--- a/tasks/build/local/helper.js
+++ b/tasks/build/local/helper.js
@@ -2,56 +2,38 @@
 
 const fs = require(`fs-extra`)
 const path = require(`path`)
-const startLocalNode = require(`../../gaia`).startLocalNode
-const getNodeId = require(`../../gaia`).getNodeId
-const makeValidator = require(`../../gaia`).makeValidator
-
-let {
+const {
+  startLocalNode,
+  getNodeId,
+  makeValidator,
   initNode,
   createKey,
   initGenesis,
   makeExec,
   nodeBinary
-} = require(`../../gaia.js`)
+} = require(`../../gaia`)
 
-const buildLocalNode = buildTestnetPath => async options => {
+const buildLocalNode = buildTestnetPath => async ({
+  numberNodes,
+  ...options
+}) => {
   try {
-    const chainId = `local-testnet`
-    const moniker = `local`
-    const localTestnetPath = `${buildTestnetPath}/${chainId}`
-    // remove existing config
-    if (options.overwrite) {
-      fs.removeSync(localTestnetPath)
-    }
-
-    const clientHome = `./${localTestnetPath}/main-node-cli/lcd`
-    const nodeHome = `${localTestnetPath}/main-node-home`
-    const defaultAccountInfo = {
-      keyName: `local`,
-      password: options.password,
-      clientHomeDir: clientHome
-    }
-    await initNode(
-      chainId,
-      moniker,
-      nodeHome,
-      options.password,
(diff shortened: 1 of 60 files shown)


tasks/testnet.js
```diff
@@ -18,6 +20,13 @@ async function main() {
     }
     const TESTNET_NODE_FOLDER = path.join(userHome, `.gaiad-testnet`)
     startLocalNode(TESTNET_NODE_FOLDER)
+    if (process.env[`VALIDATORS`] && process.env[`VALIDATORS`] > 1) {
+      // let's re-t
```
faboweb: ?

faboweb:
Awesome :)

tasks/testnet.js
sabau: Ah this should have been stashed sorry.
Basically I planned to use the cli() feature to specify how many nodes you want + done some experiments to add accounts.

I moved here the idea about extra parameters that I was playing experimenting on:
https://github.com/cosmos/voyager/issues/1752

faboweb: :D no problem. I like the proposal.

tasks/testnet.js
```diff
@@ -1,27 +1,53 @@
...
+        chainId: network,
+        password: `1234567890`,
+        overwrite: false,
+        moniker: `local`,
+        keyName: `main-account`
+      },
+      numberNodes
+    )
+    await startNodes(nodes, mainAccountSignInfo, network)
+    fs.copySync(join(nodes[1].home, `config`), cliHomePrefix)
```
sabau: This mimic the behaviour of electron start, I tried hard not to put those three lines here and rely on that hook but I couldn't find a nice way of doing it. Any idea?

sabau: From copySync to WriteFileSync should happen at boot in my mind, instead I'm forcing it to happen at network creation time (I didn't wanted to touch also that process right now, to keep this PR small)

faboweb: boot process will die anyway ;)

tasks/build/local/helper.js
```diff
@@ -0,0 +1,223 @@
...
+const saveVersion = nodeHome => {
+  const versionPath = join(nodeHome, `config`)
+  let versionFilePath = join(versionPath, `gaiaversion.txt`) // nodeHome/config is used to copy created config files from
+  return makeExec(
+    `mkdir -p ${versionPath} && ${nodeBinary} version > ${versionFilePath}`
+  )
+}
+
+// nodes[0] is a placeholder just to be aligned with the enumeration used in gaia
+// TODO: next PR refactor also gaia to simplify numeration
```
faboweb: 👍

PR #1753, opened 2018-12-17
Jordan/1674 created PageProfile for shared stylez

### LINKED ISSUE #1674: wrong class on PageProposal elements
Description:

`PageProposal` is using `validator-profile...` classes, for eg `class="validator-profile__header validator-profile__section proposal"` .

We should change them to have `proposal-page...` classes

jbibla:
Closes #1674 

_Description:_

- created PageProfile for shared stylez
- fixed problematic naming of validator classes on proposal elements

❤️ Thank you!

---

### DIFF
--- a/app/src/renderer/components/staking/PageValidator.vue
+++ b/app/src/renderer/components/staking/PageValidator.vue
@@ -1,5 +1,5 @@
 <template>
-  <tm-page data-title="Validator"
+  <page-profile data-title="Validator"
     ><template slot="menu-body">
       <tm-balance />
     </template>
@@ -11,7 +11,7 @@
       </tm-tool-bar>
     </div>
     <tm-data-error v-if="!validator" /><template v-else>
-      <div class="validator-profile__header validator-profile__section">
+      <div class="page-profile__header page-profile__section">
         <div class="column">
           <img
             v-if="validator.keybase"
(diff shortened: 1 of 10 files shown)


app/src/renderer/components/staking/PageValidator.vue
```diff
@@ -70,16 +70,14 @@
...
               <dt>Voting Power</dt>
-              <dd id="validator-profile__power">
-                {{ pretty(powerRatio * 100) }} %
-              </dd>
+              <dd id="page-profile__power">{{ pretty(powerRatio * 100) }} %</dd>
             </dl>
             <dl v-if="config.devMode" class="colored_dl">
               <dt>Uptime</dt>
-              <dd id="validator-profile__uptime">
+              <dd id="page-profile__uptime">
```
faboweb: are you sure about name spacing these under `page-profile`?

jbibla: i think it's an ok name, intended to describe our pages with a "profile-like" layout. open to suggestions if you have them.

sabau: If we recycle all this I'm happy! 
The `uptime` case seems a bit contextualized in validators although, everything else seems general enough to be recycled in other pages.

Ideally I would keep separate (declare in validator page) only those three classes:
- self-bond 
- commissions
- uptime

jbibla: yeah, i agree with you @sabau - i didn't actually write or refactor any css in this PR - just renamed things. i would be happy to improve the css here to make it more generalized or i could do this in another PR ... whatever you guys think is best.

faboweb: I am fine with doing it in another PR. Going to merge this. Can you open a follow up?

PR #1770, opened 2018-12-21
WIP: fedekunze/1769 rm hardcoded params

### LINKED ISSUE #1769: use parameters instead of hardcoded values
Description:

Eg: `min_deposit`, `bonding_denom`, etc

fedekunze:
Closes #1769 

_Description:_

❤️ Thank you!

---


app/src/renderer/components/governance/PageGovernance.vue
```diff
@@ -67,8 +67,15 @@ export default {
...
-    // TODO: get min deposit denom from gov params
-    ...mapGetters([`proposals`, `filters`, `bondingDenom`, `connected`])
+    ...mapGetters([
+      `proposals`,
+      `filters`,
+      `connected`,
+      `governanceParameters`
+    ]),
+    depositDenom() {
+      return this.governanceParameters.parameters.deposit.min_deposit[0].denom
```
faboweb: min_deposit can be several denoms?

fedekunze: LOL yeah, idk why tho

faboweb: Can we put that into a getter as it is used several times?

fedekunze: hmm ideally we'd like to support multiple deposit denoms for proposals in the future, so we'd need to have something like an array of them

fedekunze: maybe leave it for another issue when we support multiple coins everywhere?

faboweb: I mean put `depositDenom` in a getter as you use it in too places. Can return an array when we have multiple denom min fees.

test/unit/specs/components/wallet/__snapshots__/PageSend.spec.js.snap
```diff
@@ -1356,7 +1356,7 @@ exports[`PageSend should show bech32 error when address length is too long 1`] =
                 </h3>
                  
                 <h2>
-                  1,000.0000…
+                  0.0000…
```
faboweb: This looks wrong.

test/unit/specs/components/wallet/__snapshots__/PageSend.spec.js.snap
```diff
@@ -46,7 +46,7 @@ exports[`PageSend disables sending if not connected 1`] = `
                 <h2
                   class="total-atoms__value"
                 >
-                  1,000.0000…
+                  0.0000…
```
sabau: Wow all to 0. Is this expected?

test/unit/specs/components/common/PagePreferences.spec.js
```diff
@@ -1,24 +1,29 @@
...
 import PagePreferences from "renderer/components/common/PagePreferences"
+import lcdClientMock from "renderer/connectors/lcdClientMock.js"
+
 jest.mock(`renderer/google-analytics.js`, () => () => {})
 
 describe(`PagePreferences`, () => {
   let wrapper, store
-  let instance = setup()
+  let { stakingParameters } = lcdClientMock.state
+  let { mount } = setup()
```
sabau: You know I'm a kind of fan of this :)

Minor: `let` to `const` where possible, there are some instances of those possible changes along the PR

test/unit/specs/components/wallet/__snapshots__/PageSend.spec.js.snap
fedekunze: Page send uses `fermions` and `mycoins`, so (i think ?) it's expected

fedekunze: (the balance header uses `stake`)

test/unit/specs/components/common/PagePreferences.spec.js
fedekunze: I'd prefer to do that refactor on all the test files afterwards along with the `({ wrapper, store })`

app/src/renderer/components/common/TmBalance.vue
```diff
@@ -51,7 +51,7 @@ export default {
     }
   },
   computed: {
-    ...mapGetters([`bondingDenom`, `user`, `totalAtoms`]),
```
jbibla: what is the thinking here @fedekunze ? isn't it nicer to have `bondingDenom` as a getter than having to use `stakingParameters.parameters.bond_denom` every time?

fedekunze: hmmm just deleted it bc is was calling config.denom which was a mocked value. Tbh I didn't thought much about it. Do you think we need a `getter` for it ?

test/unit/specs/components/wallet/__snapshots__/PageSend.spec.js.snap
fedekunze: see reply above

test/unit/specs/components/wallet/__snapshots__/PageSend.spec.js.snap
fedekunze: updated

app/src/renderer/components/common/TmBalance.vue
fedekunze: updated

test/unit/specs/components/staking/__snapshots__/PageValidator.spec.js.snap
```diff
@@ -3592,7 +3592,7 @@ Array [
   Array [
     "notify",
     Object {
-      "body": "You have successfully undelegated 10 STAKEs.",
+      "body": "You have successfully undelegated 10 undefineds.",
```
faboweb: I think this should show a denom

fedekunze: damn true

fedekunze: uhh ? On my local file it shows the correct STAKE. but when I try to push it says `Everything up-to-date`

fedekunze: updated

app/src/renderer/components/governance/PageGovernance.vue
fedekunze: updated

PR #1777, opened 2018-12-24
[WIP] Initialize status component [After UI merge]

### LINKED ISSUE #1688: Create component for connecting, loading, loaded states
We currently use tm-data-connecting, tm-data-loading, tm-data-msg in almost all views to handle connection states. We should refactor those 3 into one component.

sabau:
Closes #1688 

_Description:_

- fix some css property in variables and app.css
- create managed-body component
- use it wallet and page transactions

❤️ Thank you!

---


app/src/renderer/components/common/ManagedBody.vue
```diff
@@ -0,0 +1,65 @@
...
+</template>
+
+<script>
+import { TmDataMsg, TmDataLoading, TmDataEmpty } from "@tendermint/ui"
+import DataEmptySearch from "common/TmDataEmptySearch"
+import TmDataError from "common/TmDataError"
+import ModalSearch from "common/TmModalSearch"
+import TmDataConnecting from "common/TmDataConnecting"
+export default {
+  name: `managed-body`,
```
jbibla: not sure i understand the naming choice here @sabau

sabau: I wasn't sure neither honestly.
I was thinking of the goal of this component, basically it try to manage the statuses of data you pass to it, with a mandatory body slot, so I tried to express it somehow but the naming is not really clear.

Do you have some suggestion on how to make it change it?

It manages all the statuses in one place so I thought of manage something

sabau: Oh wait, I just realized that this component will fit as sub component of tm-page, once the merge is complete, will simply define a new slot

sabau:
I would like to re-open this in a cleaner manner (locally this grew too much)

Issue #1742, opened 2018-12-12
Deprecate lcdClientMock


fedekunze:
- the mocked LCDClient supports the demo mode, which is not being used even by Voyager. 
- it takes a lot of development time to copy the same behaviour from Gaia into the lcdClientMock (specially POST requests)
   - use contract tests for that

Scope:

- [ ] replace mocked values used by unit tests for swagger file values
- [x] Deprecate demo mode
- [ ] Delete `lcdClientMock.js`

jbibla:
please include a description of why

fedekunze:
the main a reason was bc the mocked LCDClient supports the demo mode, which is not being used even by us. The second reason was bc it takes a lot of development time to copy the same behaviour from Gaia into the lcdClientMock (specially POST requests) and for that we can just use contract tests.

We talked about this on the retro yesterday but please feel free to add a `discuss` label it you think it needs more sync discussion

jbibla:
thanks for the explanation! fabo mentioned to me in our call yesterday - makes sense!

PR #1792, opened 2019-01-03
WIP: fedekunze/remove demo mode

fedekunze:
Addresses second checkbox from #1742

_Description:_

❤️ Thank you!

---


test/unit/specs/components/common/TmConnectedNetwork.spec.js
```diff
@@ -2,19 +2,24 @@ import TmConnectedNetwork from "common/TmConnectedNetwork"
 import setup from "../../../helpers/vuex-setup"
 
 describe(`TmConnectedNetwork`, () => {
-  let wrapper, store, instance
+  let wrapper, store
```
sabau: That's a nice way  👍I would only switch the 
`let instance` to `const instance`

test/unit/specs/components/common/TmConnectedNetwork.spec.js
```diff
@@ -2,19 +2,24 @@ import TmConnectedNetwork from "common/TmConnectedNetwork"
...
+    })
     store = instance.store
     wrapper = instance.wrapper
-    wrapper.setData({ lastHeader: { chain_id: `Test Net`, height: 42 } })
-    await store.dispatch(`setMockedConnector`, true)
   })
 
   it(`has the expected html structure`, () => {
     expect(wrapper.vm.$el).toMatchSnapshot()
+    console.log(wrapper.vm.$el.outerHTML)
```
sabau: leftover

app/src/renderer/components/common/PagePreferences.vue
```diff
@@ -5,16 +5,6 @@
       <tool-bar />
     </template>
     <tm-part title="Settings">
-      <tm-list-item type="field" title="Select network to connect to">
```
sabau: Although this has to be killed here, I liked the idea of being able to switch easily between networks. When we will be on the web maybe we could recycle something from here

fedekunze: I like the idea, specially if we support multiple testnets/chains

test/unit/specs/components/common/TmConnectedNetwork.spec.js
```diff
@@ -87,7 +77,7 @@ describe(`TmConnectedNetwork`, () => {
   })
 
   it(`shows the connected node`, async () => {
-    instance = mount(TmConnectedNetwork)
+    let instance = mount(TmConnectedNetwork)
```
sabau: Ah sorry I wasn't explicit, I meant all the instance of `let instance`

PR #1808, opened 2019-01-08
fedekunze/1707 fix amount

### LINKED ISSUE #1707: ProposalModal Amount bug
### Description:
- I can't find a way to choose from which coin I want to Deposit
- Once I have 0 Steaks available the UI let me create a proposal with 1 steak anyway, then an error is thrown (which is OK i think)

Here the response:
[proposalCreation.txt](https://github.com/cosmos/voyager/files/2649983/proposalCreation.txt)

and the error: 
screen shot 2018-12-05 at 19 03 43

fedekunze:
Closes #1707

_Description:_

❤️ Thank you!

---


app/src/renderer/components/governance/ModalPropose.vue
```diff
@@ -59,17 +59,11 @@
...
-        :min="0"
-        v-model="amount"
-        type="number"
-      />
+      <tm-field id="amount" :min="0" v-model="amount" type="number" />
       <tm-form-msg
         v-if="!$v.amount.between && amount > 0"
         :max="$v.amount.$params.between.max"
-        :min="$v.amount.$params.between.min"
+        :min="1"
```
sabau: but in this case on line 62 we are saying that the minimum is 0 and in the error message the minimum is 1

sabau: or is the `1` referring to the first parameter in line 186? 

In both cases I think is safer to use the actual parameters of the validation function in the error message, to avoid having to remember to update both places if we change this in the future.

app/src/renderer/components/governance/ModalDeposit.vue
```diff
@@ -25,18 +25,11 @@
...
-        :min="0"
-        v-model="amount"
-        type="number"
-      />
+      <tm-field v-focus id="amount" :min="0" v-model="amount" type="number" />
       <tm-form-msg
         v-if="!$v.amount.between && amount > 0"
         :max="$v.amount.$params.between.max"
-        :min="$v.amount.$params.between.min"
+        :min="1"
```
sabau: As `ModalPropose`

app/src/renderer/components/governance/ModalPropose.vue
fedekunze: you mean updating the validation minLength to be 1 ?

app/src/renderer/components/governance/ModalDeposit.vue
```diff
@@ -25,18 +25,11 @@
...
       />
-      <tm-field
-        v-focus
-        id="amount"
-        :max="balance"
-        :min="0"
-        v-model="amount"
-        type="number"
-      />
+      <tm-field v-focus id="amount" :min="0" v-model="amount" type="number" />
```
sabau: Removing the :max here I think will remove the limit on the input field, then the user won't know why the button is disabled (although it should be pretty obvious)

app/src/renderer/components/governance/ModalPropose.vue
fedekunze: > This should be just the value to put in the string we will see in the error message

So 1 ?

app/src/renderer/components/governance/ModalDeposit.vue
fedekunze: but then it shows the error message

app/src/renderer/components/governance/ModalPropose.vue
faboweb: I am on @sabau side here. What was your intent while changing this?

app/src/renderer/components/governance/ModalDeposit.vue
sabau: True, I haven't thought of that

faboweb:
What is the progress here?

fedekunze:
updated @sabau @faboweb

PR #1819, opened 2019-01-11
Refactor: TmPage manage status

### LINKED ISSUE #1688: Create component for connecting, loading, loaded states

sabau:
Closes #1688 

_Description:_
- remove duplicated code for loading, error and searching functionalities
- use props and slots through TmPage and TmPageHeader

❤️ Thank you!

---


app/src/renderer/components/governance/PageGovernance.vue
```diff
@@ -67,14 +57,14 @@ export default {
     showModalPropose: false
   }),
   computed: {
-    ...mapGetters([`proposals`, `filters`, `depositDenom`, `connected`])
+    ...mapGetters([`proposals`, `filters`, `depositDenom`, `connected`]),
+    proposalList() {
```
fedekunze:
```suggestion
    proposalIds() {
```

app/src/renderer/components/governance/PageGovernance.vue
```diff
@@ -1,22 +1,19 @@
...
-      <tool-bar
-        ><a v-tooltip.bottom="'Search'" @click="setSearch()"
-          ><i class="search material-icons">search</i></a
-        ></tool-bar
-      >
-    </template>
-    <modal-search type="proposals" />
+  <tm-page
+    :tabs="tabs"
+    :dataset="proposalList"
```
fedekunze:
```suggestion
    :dataset="proposalIds"
```

app/src/renderer/components/governance/PageGovernance.vue
```diff
@@ -67,14 +57,14 @@ export default {
     showModalPropose: false
   }),
   computed: {
-    ...mapGetters([`proposals`, `filters`, `depositDenom`, `connected`])
+    ...mapGetters([`proposals`, `filters`, `depositDenom`, `connected`]),
+    proposalList() {
+      return Object.keys(this.proposals.proposals)
```
fedekunze: hmm, it'd be better to change the getter to just use `this.proposals` (separate issue probably)

app/src/renderer/components/wallet/PageWallet.vue
```diff
@@ -125,6 +83,9 @@ export default {
       }
       return balances
     },
+    allBalances() {
```
fedekunze: why don't we use just `this.wallet.balances` instead of this func ?

app/src/renderer/vuex/modules/connection.js
```diff
@@ -22,16 +23,16 @@ export default function({ node }) {
 
   const mutations = {
     stopConnecting(state, stop) {
-      state.stopConnecting = stop
+      Vue.set(state, `stopConnecting`, stop)
```
fedekunze: +++, nice !

test/unit/specs/components/governance/PageGovernance.spec.js
```diff
@@ -43,8 +44,8 @@ describe(`PageGovernance`, () => {
   })
 
   it(`should show the search on click`, () => {
-    wrapper.find(`.tm-tool-bar i.search`).trigger(`click`)
-    expect(wrapper.contains(`.tm-modal-search`)).toBe(true)
+    wrapper.vm.$el.querySelector(`.tm-tool-bar a.search-button i`).click()
```
fedekunze: find didn't work ?

test/unit/specs/components/governance/__snapshots__/PageProposal.spec.js.snap
```diff
@@ -94,78 +94,77 @@ exports[`PageProposal has the expected html structure 1`] = `
...
-               
-              <router-link-stub
-                id="settings"
-                to="/preferences"
+                help_outline
+              </i>
+            </a>
+             
+            <router-link-stub
+              id="settings"
```
fedekunze: it weird we're using settings and preferences instead of sticking to one or the other (other issue)

test/unit/specs/components/staking/PageStaking.spec.js
```diff
@@ -25,10 +25,10 @@ describe(`PageStaking`, () => {
     expect(wrapper.vm.$el).toMatchSnapshot()
   })
 
-  it(`should show the search on click`, () => {
-    wrapper.find(`.tm-tool-bar i.search`).trigger(`click`)
-    expect(wrapper.contains(`.tm-modal-search`)).toBe(true)
-  })
+  // it(`should show the search on click`, () => {
```
fedekunze: leftover comment

test/unit/specs/components/wallet/PageTransactions.spec.js
```diff
@@ -72,31 +68,19 @@ describe(`PageTransactions`, () => {
...
-  })
-
-  it(`should not show search when there is nothing to search`, () => {
-    store.commit(`setWalletTxs`, [])
-    store.commit(`setStakingTxs`, [])
-    store.commit(`setGovernanceTxs`, [])
-    wrapper.find(`.search-button`).trigger(`click`)
-    expect(wrapper.contains(`modal-search-stub`)).toBe(false)
-  })
+  // it(`should show an error if there are no transactions`, () => {
```
fedekunze: ditto

app/src/renderer/components/common/ToolBar.vue
```diff
@@ -26,8 +48,27 @@ import { mapGetters, mapMutations } from "vuex"
 export default {
   // the name needs to be different from TmToolBar (tm-tool-bar) or else recursive rendering takes place
   name: `tool-bar`,
+  props: {
+    refresh: {
+      type: Object,
+      default: undefined
```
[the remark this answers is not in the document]
faboweb: `required: false` doesn't work. You have to set a default value.

app/src/renderer/components/common/PagePreferences.vue
```diff
@@ -1,6 +1,6 @@
 <template>
-  <tm-page data-title="Preferences"
-    ><template slot="menu-body">
+  <tm-page data-title="Preferences">
```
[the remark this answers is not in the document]
sabau: yeah I can't read it straight, sometimes I miss completely blocks

app/src/renderer/components/common/TmPageHeader.vue
```diff
@@ -3,23 +3,35 @@
     <div class="tm-page-header-container">
       <div class="tm-page-header-text">
         <div v-if="$slots['title']" class="tm-page-header-title">
-          <slot name="title"></slot>
+          <slot name="title" />
```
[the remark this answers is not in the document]
sabau: yeah I think the order in our eslintrc is overriding:
`extends: ["plugin:vue/recommended", "prettier", "prettier/vue"],` 0 errors

Instead:
`extends: ["prettier", "prettier/vue", "plugin:vue/recommended"],` 540 errors

Maybe we can fix it along with the other linting PR

test/unit/specs/components/governance/PageGovernance.spec.js
sabau: I'll try again with find + trigger.
The `trigger` vs `click` convinced me to switch when I was doing this for some reason trigger wasn't working

app/src/renderer/components/wallet/PageWallet.vue
sabau: Mainly for testing, so I don't have to mock wallet if I need just this.

app/src/renderer/components/governance/PageGovernance.vue
sabau: I like it more, and it makes me think of the `dataset` property name

app/src/renderer/components/common/PagePreferences.vue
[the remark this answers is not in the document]
sabau: Just made a PR (broken for now, let's decide where to apply those rules `web` VS `develop`)

CHANGELOG.md
```diff
@@ -97,6 +97,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - [\#1379](https://github.com/cosmos/voyager/issues/1379) Changed some tests so they don't display errors on the console to better identify real errors @faboweb
 - [\#1792](https://github.com/cosmos/voyager/pull/1792) removed mocked demo mode @fedekunze
 - [\#1720](https://github.com/cosmos/voyager/issues/1720) Time format from 12 to 24h @sabau
+- [\#1688](https://github.com/cosmos/voyager/issues/1688) TmPage handles recurrent statuses management and search modal @sabau
```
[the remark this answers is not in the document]
sabau: I'll try to rephrase it, basically the concept is that some redundant code has moved from many pages to a single place

app/src/renderer/components/common/ToolBar.vue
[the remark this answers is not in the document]
sabau: Yeah unfortunately it won't work without, even if logically it makes sense

app/src/renderer/components/governance/PageGovernance.vue
sabau: Ok I've seen in every other places I was passing the whole dataset, here I was wrong in getting only the keys, so I was thinking of keeping the name but instead of keys I'll pass the Object.values, so it behaves equally along the pages

app/src/renderer/components/governance/PageGovernance.vue
sabau: Yeah this was the renaming issue we were tackling some weeks ago, but I think we are reading tallies, loading and some other property

test/unit/specs/components/governance/PageGovernance.spec.js
sabau: Yep it's working, Thanks! I'll revert to trigger('click')

test/unit/specs/components/governance/__snapshots__/PageProposal.spec.js.snap
sabau: Agree! The id should be `preferences` as we are going to that route, the only thing that leave us related to the word `settings` will be the material icon name

sabau:
The only thing I left as it was is 
```
proposalList() {
--
proposalIds() {
```

Instead I changed what the method is doing, now it returns really the list instead of just the keys, so it's consistent with the other pages

app/src/renderer/components/governance/PageGovernance.vue
faboweb: Is `dataset` a dataset or a set of ids? can't follow ^^'

app/src/renderer/components/wallet/PageWallet.vue
faboweb: you currently still need to provide `wallet.balances` as `allDenomBalances` requires it. either use `this.allBalances` in there or use `wallet.balances` imo.

app/src/renderer/vuex/modules/connection.js
```diff
@@ -22,16 +23,16 @@ export default function({ node }) {
...
     },
     setConnected(state, connected) {
-      state.connected = connected
+      Vue.set(state, `connected`, connected)
```
faboweb: This is only necessary if manipulating the property of a property of state.

sabau: So here I'm abusing the Vue.set function, connected is always a boolean right?

faboweb: I think so. But it's not wrong. ;)

Issue #1221, opened 2018-08-31
Make linting a seperate check in CircleCI


faboweb:
Description:

Currently linting is part of the testing step. It would be nice to have this as a separate check to quickly access the problem.

NodeGuy:
Please add the `discuss` label to new issues.

PR #1841, opened 2019-01-16
Enable prefer-const eslint rule

### LINKED ISSUE #1802: Set const linter on
let's just turn this on https://eslint.org/docs/rules/prefer-const

_Originally posted by @jbibla in https://github.com/cosmos/voyager/pull/1782_

sabau:
Closes #1802 

_Description:_

- Enforce the prefer-const eslint rule

- making green this PR takes few minutes, but it will change most of vue files. To minimize conflits we decided to apply this after web merges into develop

❤️ Thank you!

---


sabau:
Ah you mean split liting from unitTesting. That can be nice, we can put it as prerequiste, the fast failing test that will not even make the others start, what do you think guys? It can be also just another step, but I liked the idea of saving CCI job if the PR won't succeed anyway.

Let's create an issue :)

fedekunze:
There's one already #1221, the thing is that if we merge this as it is then develop will always fail until we fix all the linter errors

jbibla:
this makes me very happy 

`const` FTW

shout out to @NodeGuy who would love this too.

NodeGuy:
LOL, it's nice to be reminded of you guys.  Welcome to the team, @sabau!

PR #1851, opened 2019-01-17
Web deploy prerequisites

### LINKED ISSUE #1833: Voyager CI online release
- Deploy to website (voyager-staging.cosmos.network)
- Start a compatible testnet
- Start a faucet for that testnet

sabau:
Closes #1833

_Description:_

- reduce bundle size by optimizing few imports
- reduce lodash/fp dependency (a single usage for a whole library)
- change build command as yarn pack is reserved and does not call webpack

[demo](https://s3.eu-central-1.amazonaws.com/cosmos-voyager/index.html#/)

Deployment of nodes followup (I won't do all this in the first PR, but we know it's possible and we can specify those details to the SRE team next call)

My plan is to have two node types:
- gaia in a Docker exposing RPC port
- gaiacli in another Docker, exposing the rest server, but pointing to the first machine of type `gaia`
- Eventually an nginx or APIGateway, the chapest solution will win :tada: 

- Is possible, BUT the genesis file should be on both node types

In this way we will be able to spin up gaia nodes as hell and just one gaiacli node pointing to (let's say) the first gaia node, so the gaiacli node can scaleup indipendently if he receives burst of requests, but the real nodes are always the same

`gaiacli [command] --node=IP:PORT`

❤️ Thank you!

---


package.json
```diff
@@ -134,6 +134,7 @@
     "vue-directive-tooltip": "1.4.5",
     "vue-router": "3.0.1",
     "vuelidate": "0.6.2",
-    "vuex": "3.0.1"
+    "vuex": "3.0.1",
+    "webpack-bundle-analyzer": "3.0.3"
```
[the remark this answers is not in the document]
sabau: totally

faboweb:
This branch doesn't seem to trigger CI?!

faboweb:
Code coverage is missing

.aws/.ebextensions/01_nginx_custom.config
```diff
@@ -0,0 +1,60 @@
+files:
```
faboweb: probably easy for nginxers but I think the config could use some comments

.aws/first.sh
```diff
@@ -0,0 +1,52 @@
+#!/bin/bash
```
faboweb: could also use some comments

.aws/first.sh
```diff
@@ -0,0 +1,52 @@
...
+
+PASSWORD=1234567890
+ACCOUNT=main_account
+PORT=26657
+API_PORT=8080
+NETWORK=testnet
+AMOUNT=100stake
+
+./gaiad init --home . --moniker ${ACCOUNT} --chain-id ${NETWORK}
```
faboweb: Thos steps could need some comments

.aws/first.sh
```diff
@@ -0,0 +1,52 @@
...
+screen -dmS gaia ./gaiad start --home .
+screen -dmS rest ./gaiacli rest-server --laddr tcp://0.0.0.0:${API_PORT} --home . --node http://localhost:${PORT} --chain-id ${NETWORK} --trust-node true
+
+# get my address to use it as source of richness for others
+ADDRESS=$(./gaiacli keys show ${ACCOUNT} --home . --address)
+
+while true
+do
+    # Is anyone asking for money
+    LIST=$(aws s3 ls s3://cosmos-gaia/addresses/)
```
faboweb: very hacky solution ^^

.aws/first.sh
```diff
@@ -0,0 +1,52 @@
...
+    if [[ -n "${LIST}" ]];
+    then
+        PID=$(lsof -ti tcp:${API_PORT})
+        if [[ -n "${PID}" ]];
+        then
+            # we cannot use gaiacli while the rest server is running, it locks the resources
```
faboweb: in theory we could have a second gaiacli instance where we import the main key which performs the action

.circleci/config.yml
```diff
@@ -1,104 +1,148 @@
-version: 2
+version: 2.1
+
+aliases:
```
faboweb: cool

.circleci/config.yml
```diff
@@ -1,104 +1,148 @@
...
+  - &docker-browsers circleci/node:10.15-browsers
+  - &docker-go circleci/golang:1.11
+
+commands:
+  yarn-update:
+    description: "[YARN] fetch new version"
+    steps:
+      - run:
+          name: Update Yarn
+          command: sudo npm -g uninstall yarn ; sudo npm -g install yarn@1.13.0
```
faboweb: why did you fix the version to 1.13.0?

.circleci/config.yml
```diff
@@ -234,22 +261,39 @@ workflows:
...
+      - build-gaia
+      - deploy2s3:
+          requires:
+            - testUnit
+            - build
+          filters:
+            branches:
+              only: develop
+
+      #       - deploy-gaia
```
faboweb: nice idea, but you did remove the step right. maybe remove the comment

app/index.ejs
```diff
@@ -4,6 +4,7 @@
 <head>
   <meta name=viewport content="width=device-width, user-scalable=no">
   <meta charset="utf-8">
+  <link rel="shortcut icon" href="icons/png/24x24.png" />
```
faboweb: nice :)

app/src/config.js
```diff
@@ -0,0 +1,17 @@
+const dev = process.env.NODE_ENV === `develop`
```
faboweb: how about just having a `config.json` on the server with specifications for the deployed paths instead of hardcoding all of these values?

app/src/renderer/main.js
```diff
@@ -4,130 +4,7 @@
...
-  return new Vue({
-    router,
-    ...App,
-    store
-  }).$mount(`#app`)
-}
+import { main } from "scripts/boot"
 
 // run
 main()
```
faboweb: this file now only does this one command. is that useful?

app/src/renderer/vuex/modules/transactions.js
```diff
@@ -1,4 +1,3 @@
-import fp from "lodash/fp"
```
faboweb: nice

app/src/renderer/vuex/store.js
```diff
@@ -6,7 +6,7 @@
 
 import Vue from "vue"
 import Vuex from "vuex"
-import _ from "lodash"
+import { merge } from "lodash"
```
faboweb: I think you could just import the one function with (didn't test it):
```suggestion
import merge from "lodash/merge"
```

circle.ebs_deploy.sh
```diff
@@ -0,0 +1,63 @@
...
+    # add all the extensions
+    cp -r .aws/.ebextensions .
+
+    # zip and push it to S3 ALWAYS to the same file, no jokes, just a static veriable in CircleCi env variable
+    zip -r ${DOCKER_DEPLOY_FILE} .ebextensions Dockerrun.aws.json
+    aws s3 cp ${DOCKER_DEPLOY_FILE} $S3_PATH --region=${AWS_REGION}
+}
+
+deploy_ebs() {
+    # Create a version of the APPLIOCATION_NAME with the current biuild number as version. The source is the S3 file we know
```
faboweb:
```suggestion
    # Create a version of the APPLICATION_NAME with the current build number as version. The source is the S3 file we know
```

webpack.renderer.config.js
```diff
@@ -8,16 +8,22 @@ const fs = require(`fs-extra`)
...
+const devPlugins = process.env.CIRCLECI
+  ? []
+  : [new CleanWebpackPlugin([buildPath]), new BundleAnalyzerPlugin()]
+
 const rendererConfig = {
-  devtool:
-    process.env.NODE_ENV === `production`
-      ? `#source-map`
-      : `#inline-source-map`,
+  devtool: process.env.NODE_ENV === `production` ? false : `#inline-source-map`,
```
faboweb: why? we need source-maps, are they produces somewhere else?

webpack.renderer.config.js
```diff
@@ -102,22 +113,42 @@ const rendererConfig = {
       govern: resolve(`app/src/renderer/components/govern`),
       staking: resolve(`app/src/renderer/components/staking`),
       wallet: resolve(`app/src/renderer/components/wallet`)
+      // moment: `moment/src/moment`, // this will remove locales, we rely on them
```
faboweb: do we need these comments?

webpack.renderer.config.js
```diff
@@ -126,9 +157,6 @@ const rendererConfig = {
  */
 if (process.env.NODE_ENV === `production`) {
   rendererConfig.plugins.push(
-    new webpack.DefinePlugin({
-      "process.env.NODE_ENV": `"production"`
```
faboweb: this is never used?

webpack.renderer.config.js
sabau: Was the alias system for reduce the size of moment and lodash, but we are using their inner dependencies so we can't optimize for now..I'll remove them

app/src/renderer/main.js
sabau: In this way we can test the other functions without running for real the `main` function once we import the file. We could move this call somewhere else, I tried to minimize the impact on touched files

faboweb:
But we need them to have proper stack traces with Sentry, don't we?

Karoly Albert Szabo <notifications@github.com> schrieb am Fr., 1. Feb.
2019, 16:27:

app/src/config.js
sabau: I'm not 100% I got the server, you mean CircleCi? On S3 we will have only bundled files.

The conversion from json to js was to simplify the process, in this way I bundle and push without need of setup on the destination. Maybe I could use some ENV variables on CircleCi to override localhost, but I would do this anyway with a conf.js file.

.circleci/config.yml
sabau: The security is available only since 1.13.0 (fixed) they delivered it in 1.12.x (where x was a bit bigger than our version) so once I had to update I decided to apply the fixed version

.aws/first.sh
sabau: Yeah..I feel a bit ashamed 👅this was the quickest win I found.

I'll introduce a rate and address limiter on the API Gateway so at least greediness will be slowed down

.aws/first.sh
sabau: correct! @mircea-c told me something like this will become true in their infrastructure + stargate will be on a separate machine, so for now I left the hacky solution, but I'll add a TODO to remember that when we'll adopt the real network deployment this PID stuff has to be removed, sounds ok?

.circleci/config.yml
sabau: Yeee all the new commands 🎉

app/src/renderer/vuex/store.js
sabau: I'll try now

app/src/renderer/main.js
faboweb: 👍

app/src/config.js
faboweb: My idea was to no bundle the `config.js` file but to have a `config.json` that get's loaded from the server when Voyager starts so it is easy to manipulate the properties. Voyager will probably be deployed in different places and different scenarios so hardcoding our values into the config might not be the best idea. But that would need additional work so maybe we leave this as is.

.circleci/config.yml
faboweb: But the security feature is still there in a later version so I was asking why we need to fix to that specific version? we might loose security updates this way.

webpack.renderer.config.js
faboweb: But we need them to have proper stack traces with Sentry, don't we?

.aws/.ebextensions/01_nginx_custom.config
sabau: Even better: all this .ebextension folder will be gone. 
We are not using it automatically, and SRE is doing this job, better reduce code conflicts

webpack.renderer.config.js
sabau: Sentry! Thanks for reminding me! Ok I'll put them back as separate file ok?
cheap-source-map should be a bit lighter but provide the stack trace and module references. Should be readable code although not indented

app/src/config.js
sabau: Hum sorry I still don't get who is the server. In this moment Voyager will be served as static content (the e2e test could run from circleci artifacts directly once we have a proper network boot system i.e. https://15157-99653950-gh.circle-artifacts.com/0/tmp/voyager/app/dist/index.html#/)

I'll try the ENV variables for RPC / REST and keep only a fallback, sounds good?

app/src/renderer/vuex/store.js
sabau: It works! but I think I like more the `{ }` version, they both import only the merge function thanks to webpack tree shaking

.circleci/config.yml
sabau: Ah got it! mostly the same reason why we are setting the node version in the docker files, to be sure they don't insert some change we haven't tested.

Initially I tried with @latest, should we go that way? yarn seems stable in his future releases, no breaking changes until 2.0, and even there seems ok for our needs

app/src/renderer/vuex/modules/transactions.js
sabau: I felt a bit sorry to remove the only example of pure functional stuff XD but that was a too big dependency

.aws/first.sh
sabau: Added some line by line / every two lines for coupled commands

sabau:
The coverage is a bit funny, I think it says that is not covered anymore since when I removed he side effect (main() was called for real once the file was imported and it was touching this function but was not tested for real)

.aws/first.sh
```diff
@@ -0,0 +1,67 @@
...
+    # Is anyone asking for money
+    LIST=$(aws s3 ls s3://cosmos-gaia/addresses/)
+    if [[ -n "${LIST}" ]];
+    then
+        # TODO: SRE team is working on a clean solution.
+        # This hacky version is temporary and will be replaced by a stargate running on a separate machine
+        # Remove this logic as soon as we put back the deploy-gaia automation
+
+        # Check if a stargate is locally running
+        PID=$(lsof -ti tcp:${API_PORT})
```
sabau: no more needed since https://github.com/cosmos/cosmos-sdk/pull/3514

.aws/first.sh
```diff
@@ -0,0 +1,67 @@
+#!/bin/bash
```
jbibla: do we need to include `.aws` files? @cosmos/sre ?

.circleci/config.yml
jbibla: is this question resolved?

app/src/config.js
jbibla: seems like we can move forward with this for now. but fabo's idea is a good one - so we'll have to find a solution later.

app/src/renderer/scripts/boot.js
```diff
@@ -0,0 +1,126 @@
...
+  })
+
+  // add error handlers in production
+  if (env.NODE_ENV === `production`) {
+    // Sentry is used for automatic error reporting. It is turned off by default.
+    Sentry.init({})
+
+    // this will pass the state to Sentry when errors are sent.
+    // this would also sent passwords...
+    // Sentry.configureScope(scope => {
```
jbibla: commented out?

webpack.renderer.config.js
```diff
@@ -8,15 +8,24 @@ const fs = require(`fs-extra`)
...
+
+const devPlugins = process.env.CIRCLECI
+  ? []
+  : [new CleanWebpackPlugin([buildPath]), new BundleAnalyzerPlugin()]
+
 const rendererConfig = {
   devtool:
     process.env.NODE_ENV === `production`
-      ? `#source-map`
+      ? `#cheap-source-map`
```
jbibla: what does this mean?

.circleci/config.yml
faboweb: Ok good argument. I would anyway go with at least allowing patch versions, so `1.13.x`. But I think we would be fine with `1.x`.

app/src/config.js
faboweb: I will open an issue explaining my idea. We can move on with this PR without this change. (https://github.com/cosmos/voyager/issues/1951)

.aws/first.sh
faboweb: sweet

app/src/renderer/scripts/boot.js
faboweb: left in as a warning to not do this. see the comment

webpack.renderer.config.js
faboweb: `transformed code (lines only)`
I am not sure why we shouldn't just go with full source-maps?!

faboweb:
Re coverage: I think the router setup could be escaped via `/* istanbul ignore next */` and the on route changed handler should be tested.

.circleci/config.yml
sabau: Ok, now this logic will be moved away from here, it has been set into tendermint/devops repo

webpack.renderer.config.js
sabau: the bundle size will increase, that's basically a version that grants similar results with half space.
- source maps is the whole codebase, so the slowest option both to create and to download, usually is advised not to put it on production albeit is one of the options, in our case since we are open source this won't impact security
- cheap source maps are a representation both tree-shacked and transpiled, build time decrease a lot and to download them is a bit quicker too

- nosources-source-map would be perfect (but is slow) for Sentry only, to report lines and files, but the code is not on the map at all, but is as slow as source-map and the size doesn't improve much too

.aws/first.sh
sabau: those are used by the current (temporary) node setup on EC2, until the real network boot on merge is ready

webpack.renderer.config.js
faboweb: :+1: let's it that way then. I wasn't aware that the source-maps increase the bundle size :X

.circleci/config.yml
```diff
@@ -1,104 +1,151 @@
...
+aliases:
+  # Define paths and never think about them again
+  - &GAIA /tmp/gaia
+  - &WORKSPACE /tmp/voyager
+  - &DIST /tmp/voyager/app/dist
+  # Pick docker versions here only, then use the aliases in the executors definition
+  - &docker-node circleci/node:10.15
+  - &docker-browsers circleci/node:10.15-browsers
+  - &docker-go circleci/golang:1.11
+  - &docker-aws cibuilds/aws:1.15
```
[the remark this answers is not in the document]
sabau: My idea was to define at the top of the file every hardcoded string I will use everywhere else, so if I change things in the future will be easier to catch errors  if all the definitions lives together, I'll remove the aws executor and docker-image as it seems they are no longer needed

.circleci/config.yml
```diff
@@ -115,69 +162,52 @@ jobs:
...
-      - run: yarn install
-      - save_cache:
-          paths:
-            - yarn.lock
-            - node_modules
-          key: v2-dependencies-root-{{ checksum "package.json" }}
-
+          at: *GAIA
+      - yarn-update
+      - yarn-install
```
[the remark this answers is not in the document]
sabau: You're right, the E2E tests are currently disables and we probably have to rewrite a portion of this, I'll comment it out for now

.circleci/config.yml
```diff
@@ -115,69 +162,52 @@ jobs:
...
-
-      - restore_cache:
-          keys:
-            - v2-dependencies-root-{{ checksum "package.json" }}
-            - v2-dependencies-root-
-
-      - run: sudo npm -g uninstall yarn ; sudo npm -g install yarn@1.13.0
-      - run: yarn install
+      - yarn-update
+      - yarn-install
```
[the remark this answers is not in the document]
sabau: In this case we are using the node_modules only but it's mostly a series of git operations

.aws/others.sh
```diff
@@ -0,0 +1,50 @@
...
+# get the key to make my node validator
+PUBKEY=$(./gaiad tendermint show-validator --home .)
+echo ${PASSWORD} | ./gaiacli keys add ${ACCOUNT} --home . > secondary_node.log
+ADDRESS=$(./gaiacli keys show ${ACCOUNT} --home . --address)
+echo ${PUBKEY} > ./${ADDRESS}
+
+# ask money by declaring my node
+aws s3 cp ${ADDRESS} s3://cosmos-gaia/addresses/${ADDRESS}
+
+poor=true
```
fedekunze: lol

cicle.run_deploy.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+set -xeuo pipefail
+
+# TODO: region should be ENV determined
```
fedekunze: create issue here

package.json
```diff
@@ -18,14 +18,14 @@
     "build:testnets": "cd tasks/build/testnets && ./localBuild.sh",
     "build:gaia": "cd tasks/build/Gaia && ./localBuild.sh",
     "build:local": "node tasks/build/local/build",
+    "build:ui": "cross-env NODE_ENV=production webpack --colors --config webpack.renderer.config.js --mode=production --profile --json > app/dist/build.stats.json",
```
fedekunze: document this on `README`

cicle.run_deploy.sh
sabau: I'll drop the whole file, last version does not use it anymore

package.json
sabau: thanks! I missed it completely

test/unit/specs/App.spec.js
```diff
@@ -72,4 +71,95 @@ describe(`App Start`, () => {
     expect(startApp).toHaveBeenCalled()
     expect(startApp.mock.calls[0][0]).toHaveProperty(`x`, 1)
   })
+
+  it(`Check the calls on VUE`, async () => {
+    // const mockVue = jest.mock(`Vue`)
```
faboweb:
```suggestion
```

test/unit/specs/App.spec.js
```diff
@@ -72,4 +71,95 @@ describe(`App Start`, () => {
     expect(startApp).toHaveBeenCalled()
     expect(startApp.mock.calls[0][0]).toHaveProperty(`x`, 1)
   })
+
+  it(`Check the calls on VUE`, async () => {
```
faboweb: I would have maybe just called the function and checked if it doesn't throw.

sabau: yeah I tried to test the real call but was taking a weird path to mock them, so I just counted the number of calls

PR #1861, opened 2019-01-22
Fabo/1818 fix validators message

### LINKED ISSUE #1818: No Results msg on no validators
Description:

When there are no validators, or the response hasn't been received (in my case because of `Unchecked runtime.lastError: The message port closed before a response was received.`), the Validators tab shows a No Results message. It should display a message that there are no validators on the network.

screen shot 2019-01-11 at 11 39 03 am

faboweb:
Closes #1818

_Description:_

On error the module would set `loading` to false. But on error `loading` should stay `true` to indicate the pending loading.

❤️ Thank you!

---


fedekunze:
can you attach a snapshot on the description or add a testcase w/Snapshot when the validators are empty ?

faboweb:
> missing some coverage tho

not true! ^^

PR #1923, opened 2019-02-01
Jordan/1894 favicons

### LINKED ISSUE #1894: voyager needs a favicon
UI Version: 0.X.X

Description:

jbibla:
Closes #1894 

_Description:_

- added favicons for all browsers and devices

❤️ Thank you!

---


app/index.ejs
```diff
@@ -4,13 +4,22 @@
...
   <% if (htmlWebpackPlugin.options.appModules) { %>
-    <% } %>
+  <% } %>
 
-      <body style="background: #15182d">
-        <script async src='https://www.google-analytics.com/analytics.js'></script>
-        <div id="app" style="background: #15182d"></div>
-      </body>
+<body style="background: #15182d">
+  <script async src='https://www.google-analytics.com/analytics.js'></script>
```
jbibla: unrelated to this PR: we should probably bundle this instead of making this request here? what do you guys think?

@sabau @faboweb @fedekunze

webpack.renderer.config.js
```diff
@@ -46,7 +46,7 @@ const rendererConfig = {
             loader: `url-loader`,
             query: {
               limit: 10000,
-              name: `imgs/[name].[ext]`
+              name: `images/[name].[ext]`
```
jbibla: i don't know how / if this was working before - but `imgs` wasn't an actual path

app/index.ejs
sabau: I think it's ok to keep it in the main template, those resources won't change often

Thanks for fixing the indentation here 👍

webpack.renderer.config.js
sabau: webpack was moving the files in the imgs folder, but this renaming is confusing and does not add value at all 👍 good cleanup

sabau: screenshot 2019-02-01 at 19 42 35

jbibla:
this was simpler because i didn't need to tell webpack to move the assets (there is an icon folder, separate from the assets folder) and because the paths didn't seem to work in `index.ejs` ... if you can, would you mind taking a look to see if this is easy to accomplish?

sabau:
I used the `static` vue folder, this morning I was curious why they suggested 2 assets in their structure suggestions, now I found out, usually I was using the webpack plugin to copy static stuff, but Vue apparently is doing this out of the box

jbibla:
thanks for the help @sabau!!

PR #1925, opened 2019-02-02
Jordan/1869 network page

### LINKED ISSUE #1869: Explorer: page Network
Description:
- import and improve style from old Explorer

Subticket of: #1744 

- [ ] skeleton page
- [ ] network info header

jbibla:
Closes #1869 

_Description:_

- implemented PageNetwork

❤️ Thank you!

---


app/src/renderer/components/network/PageNetwork.vue
```diff
@@ -0,0 +1,112 @@
...
+    ]),
+    status() {
+      const color = this.connected ? `green` : `red`
+      const message = this.connected
+        ? `Network is up and running`
+        : `Network is down`
+      return { color, message }
+    },
+    lastBlock() {
+      // moment.relativeTimeThreshold(`ss`, 1)
```
jbibla: @sabau i'd like to uncomment this line, but it breaks the tests and i can't figure out how to mock it correctly.

app/src/renderer/routes.js
```diff
@@ -80,7 +80,11 @@ export default [
     name: `transactions`,
     component: require(`./components/wallet/PageTransactions`).default
   },
-
+  {
+    path: `/network`,
```
fedekunze: This should be the default route

app/src/renderer/components/network/PageNetwork.vue
sabau: Just restored, basically the workaround worked but was an unreadable spaghetti code of side effects.

Then I realised..why are we globabally mocking moment? 
So I tried removing that mock with the idea of fixing the tests that needed it => None 🎉 
Another global gone 🌮

app/src/renderer/routes.js
jbibla: interesting idea @fedekunze! i don't want to make this change right now, but we should talk about it at the next team meeting. i've added it to the agenda. 

remember, eventually we will have a dashboard as the default route - but this might be a good interim solution.

faboweb:
A screenshot of the page would have been nice.

PR #1926, opened 2019-02-04
[WIP] Vuex for explorer

### LINKED ISSUE #1865: Explorer: Blocks Vuex module
Description:
- We should create the vuex module for blocks

Subticket of: #1744

sabau:

Closes #1865 

_Description:_

- Add blocks in append only
- action to reset blocks
- add peers requested via RPC

❤️ Thank you!

---

### DIFF
--- a/app/src/renderer/vuex/modules/blockchain.js
+++ b/app/src/renderer/vuex/modules/blockchain.js
@@ -1,4 +1,7 @@
 import * as Sentry from "@sentry/browser"
+import Vue from "vue"
+
+const maxBlocks = 100
 
 export default ({ node }) => {
   const state = {
@@ -9,10 +12,19 @@ export default ({ node }) => {
     blockMetas: {},
     subscribedRPC: null,
     loading: false,
-    error: null
+    error: null,
+    peers: [],
+    blocks: []
   }
 
   const mutations = {
+    setLoading(state, loading) {
+      state.loading = loading
+    },
+    setError(state, error) {
+      console.log(`EEROOOOOOO`)
+      state.error = error
+    },
     setBlockHeight(state, height) {
       state.blockHeight = height
     },
@@ -21,6 +33,19 @@ export default ({ node }) => {
     },
     setBlockMetas(state, blockMetas) {
       state.blockMetas = blockMetas
+    },
+    setPeers(state, peers) {
+      Vue.set(state, `peers`, peers)
+    },
+    setBlocks(state, blocks) {
+      Vue.set(state, `blocks`, blocks)
+    },
+    addBlock(state, block) {
+      const blocks = state.blocks
+      if (blocks.length > maxBlocks) blocks.splice(-1, 1)
+      blocks.unshift(block)
+      // immutable version [block, ...blocks] gets slower if maxBlocks is over 100K
+      Vue.set(state, `blocks`, blocks)
     },
     setSubscribedRPC(state, subscribedRPC) {
       state.subscribedRPC = subscribedRPC
@@ -42,13 +67,13 @@ export default ({ node }) => {
         if (blockMetaInfo) {
           return blockMetaInfo
         }
-        state.loading = true
+        commit(`setLoading`, true)
         const { block_metas } = await node.rpc.blockchain({
           minHeight: String(height),
           maxHeight: String(height)
         })
         blockMetaInfo = block_metas ? block_metas[0] : undefined
-        state.loading = false
+        commit(`setLoading`, false)
 
         commit(`setBlockMetas`, {
           ...state.blockMetas,
@@ -61,8 +86,8 @@ export default ({ node }) => {
           body: error.message
         })
         Sentry.captureException(error)
-        state.loading = false
-        state.error = error
+        commit(`setLoading`, false)
+        commit(`setError`, error)
         return null
       }
     },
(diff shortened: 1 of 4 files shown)


app/src/renderer/vuex/modules/blockchain.js
```diff
@@ -9,10 +12,18 @@ export default ({ node }) => {
     blockMetas: {},
     subscribedRPC: null,
     loading: false,
-    error: null
+    error: null,
+    peers: [],
```
jbibla: what's `peers` for?

jbibla:
can we call the module `blocks.js` instead of `blockchain.js`? i think this will be more self evident.

app/src/renderer/vuex/modules/blockchain.js
```diff
@@ -85,11 +109,21 @@ export default ({ node }) => {
       commit(`setSyncing`, false)
 
       // only subscribe if the node is not catching up anymore
```
jbibla: is this comment true? we used to run into an issue where voyager would connect to nodes that weren't caught up yet. we don't have this problem right now because we're using hosted nodes... so we should implement this now to accommodate for syncing nodes or delete the comment and wait until it's an issue later.

app/src/renderer/vuex/store.js
```diff
@@ -83,6 +83,7 @@ export function storeUpdateHandler(mutation, state, pending) {
  */
 function persistState(state) {
   const cachedState = JSON.stringify({
+    //blocks: state.blockchain.blocks,
```
jbibla: why don't we want to cache blocks?

app/src/renderer/vuex/modules/blockchain.js
sabau: I was thinking the same, but if our stargate went off for a while is probably wise to let him catch up before reading blocks. But once we remove completely the RPC then we have to move this logic to stargate itself (if we want to keep it)

app/src/renderer/vuex/modules/blockchain.js
sabau: If you think we won't show them I can remove it, I was thinking would be nice to have that list to show extra informations about the network, even just the versions involved (or a pie-like)

app/src/renderer/vuex/modules/blocks.js
```diff
@@ -9,25 +16,25 @@ export default ({ node }) => {
...
-    setBlockMetas(state, blockMetas) {
-      state.blockMetas = blockMetas
-    },
-    setSubscribedRPC(state, subscribedRPC) {
-      state.subscribedRPC = subscribedRPC
-    },
-    setSubscription(state, subscription) {
-      state.subscription = subscription
-    }
+    setLoading: (state, loading) => (state.loading = loading),
```
jbibla: woahhh compression!

app/src/renderer/vuex/store.js
jbibla: hmm yeah not sure. we don't want there to be no blocks, but we also don't want to show old blocks. so probably we don't need the cache.

app/src/renderer/vuex/modules/blockchain.js
jbibla: interesting. `node_info` could be valuable and is used in a couple places - but how could peer `node_info` be useful?

app/src/renderer/vuex/modules/blockchain.js
jbibla: sure ok. that makes sense. but what i was trying to say is that we're not checking if the node is syncing or not... are we?

jbibla:
i think if you remove the cache and the peers then we can merge your blocks PR! unless you think we should keep both, then that’s ok too - just remove the commented out code.

app/src/renderer/vuex/modules/blockchain.js
sabau: peers give back the node_info of every peer in the network

jbibla: ok actually i think this will be cool. let's leave it in and i'll add something to PageNetwork to make use of this data.

app/src/renderer/vuex/store.js
jbibla: this is the comment i was talking about

app/src/renderer/vuex/modules/blocks.js
```diff
@@ -1,4 +1,11 @@
 import * as Sentry from "@sentry/browser"
+import Vue from "vue"
+
+export const cache = (list, element, maxSize = 1000) => {
```
faboweb: This could use a comment

app/src/renderer/vuex/modules/blocks.js
```diff
@@ -83,13 +90,25 @@ export default ({ node }) => {
       }
 
       commit(`setSyncing`, false)
+      // New RPC endpoint in sync, reset UI block list
+      commit(`setBlocks`, [])
```
faboweb: If we reconnect to the same node, we now clear all cached blocks. Does that make sense?

test/unit/helpers/node_mock.js
```diff
@@ -50,6 +50,33 @@ module.exports = {
         },
         node_info: { network: `test-net` }
       }
+    },
+    async net_info() {
```
faboweb: going forward we probably shouldn't work with the `node_mock.js` but use the new more unity approach to testing and add mocked values to the tests individually.

test/unit/specs/store/blocks.spec.js
```diff
@@ -0,0 +1,329 @@
...
+        status: () =>
+          Promise.resolve({
+            sync_info: {
+              latest_block_height: 0,
+              catching_up: false
+            }
+          }),
+        subscribe: (query, cb) => {
+          cb({ block: `subscribe here` })
+          expect(commit).toBeCalledWith(`setSubscription`, true)
```
faboweb: dupe test of line 169

app/src/renderer/vuex/modules/blocks.js
jbibla: good catch - probably not.

PR #1933, opened 2019-02-06
Jordan/1866 blocks

### LINKED ISSUE #1744: Explorer in Voyager
Many users are looking to have a read only block explorer to see the state of the network and verify txs.

Proposed structure:
New page `Network`: Holds information about the current network (similar to first page in old explorer). Can maybe be enriched with data from https://hubble.figment.network/chains/gaia-9002.
image

New page `Blocks`: Holds information about blocks. Similar to our old block explorer in Voyager with blocks flying in. You can search for a block and open it in a new page (able to link to it).

New page `Block`(no link): Holds block meta info (see below). Holds link to transactions page (pre-filtered for block).
image

(old) Page `Transactions`: Add ability to filter for: `hash`, `block`, `address`. Ability to set those as an URL-query so we can link to it. 
- https://github.com/cosmos/voyager/issues/1719
- https://github.com/cosmos/voyager/issues/1701
- https://github.com/cosmos/voyager/issues/965

### LINKED ISSUE #1784: Remove links to Explorer for now
The links to the explorer do not work as the explorer is offline.

### LINKED ISSUE #1866: Explorer: Blocks page
Description:
- Creation of the Blocks page for the Explorer

Subticket of: #1744

### LINKED ISSUE #1867: Explorer: Block Detail
Description:
- Define how it should look: Page or Modal
- Should be accessible directly via URL (i.e. open on incognito tab to check)
- Implement and test

Subticket of: #1744

jbibla:
closes #1744 - bringing explorer back to voyager
closes #1866 - added blocks to page network
closes #1867 - added block page
closes #1784 - removed links to external explorer

❤️ Thank you!

screen shot 2019-02-11 at 3 32 03 pm
screen shot 2019-02-11 at 3 32 17 pm

---


sabau:
oops

closes #1744 - bringing explorer back to voyager
closes #1866 - added blocks to page network
closes #1867 - added block page with back and forward buttons
closes #1784 - removed links to external explorer

app/src/renderer/components/network/PageBlock.vue
```diff
@@ -0,0 +1,154 @@
...
+  watch: {
+    $route: `getBlock`
+  },
+  async mounted() {
+    await this.getBlock()
+  },
+  methods: {
+    async getBlock() {
+      const url =
+        `http://localhost:8080/` + `blocks/` + this.$route.params.height
```
sabau: `http://localhost:8080` should come from config

sabau:
```suggestion
        `${config.stargate}/blocks/${this.$route.params.height}`
```

app/src/renderer/components/network/PageBlock.vue
```diff
@@ -0,0 +1,154 @@
...
+      const block = await axios.get(url)
+      this.header = block.data.block.header
+      this.block = block.data
```
sabau: I think we may want to move this logic away of the UX, it should stay into an action in `blocks.js`, that could be dispached form here and it returns the block (optionally we could add it to the state but I think it's fine to fetch it again OR we have to build up a dict of already fetched blocks)

sabau: let's check if `queryBlockInfo` suits your needs

jbibla: yep. great idea. i've refactored this.

app/src/renderer/components/common/TmConnectedNetwork.vue
```diff
@@ -38,33 +38,26 @@
...
 export default {
   name: `tm-connected-network`,
   data: () => ({
-    num: num
+    num
   }),
   computed: {
     ...mapGetters([`lastHeader`, `nodeUrl`, `connected`]),
-    networkTooltip({ connected, nodeUrl } = this) {
+    networkTooltip({ connected, nodeUrl, lastHeader } = this) {
```
sabau: So happy to see this `this` gone

app/src/renderer/components/common/TmConnectedNetwork.vue
```diff
@@ -38,33 +38,26 @@
...
-      return `#` + num.prettyInt(this.lastHeader.height)
-    },
-    explorerLink() {
-      return `https://explorecosmos.network/blocks/` + this.lastHeader.height
-    }
-  },
-  methods: {
-    closeMenu() {
-      this.$store.commit(`setActiveMenu`, ``)
+      return `#` + this.num.prettyInt(this.lastHeader.height)
```
sabau: that I was wondering why adding one more here

app/src/renderer/components/common/TmPage.vue
```diff
@@ -216,6 +213,7 @@ export default {
   justify-content: space-between;
   align-items: center;
   width: 100%;
+  padding: 1rem;
```
sabau: Good eye 👍

app/src/renderer/routes.js
```diff
@@ -85,6 +85,11 @@ export default [
     name: `network`,
     component: require(`./components/network/PageNetwork`).default
   },
+  {
+    path: `/block/:height`,
```
fedekunze: curious about why isn't this `/blocks/:height` ?

app/src/renderer/vuex/modules/blocks.js
```diff
@@ -18,7 +18,10 @@ export default ({ node }) => {
     loading: false,
     error: null,
     peers: [],
-    blocks: []
+    blocks: [],
+    block: {
```
fedekunze: what is `block` ? is it the latest one ? In any case, this should go with a comment or variable renaming

app/src/renderer/components/transactions/TmLiTransaction.vue
```diff
@@ -47,10 +47,7 @@ export default {
   },
   computed: {
     date() {
-      const time = moment(this.time)
-      return time.format(
-        `${moment().isSame(time, `day`) ? `` : `YYYY/MM/DD `}HH:mm`
-      )
+      return moment(this.time).format(`MMM Do YYYY, HH:mm:ss`)
```
sabau: this logic was in place to show only the hour if the transaction happened today, and the whole date if it happened earlier, are we sure we want to drop this?

app/src/renderer/components/network/PageBlock.vue
```diff
@@ -0,0 +1,151 @@
...
+            </div>
+          </div>
+        </div>
+
+        <div class="row">
+          <dl class="info_dl colored_dl">
+            <dt>Time</dt>
+            <dd>
+              {{
+                moment(block.block.header.time).format("MMM Do YYYY, HH:mm:ss")
```
sabau: would be nicer to move this function downstairs in the computed ones, so it can be tested OR even nicer to move this in the time helper and have it defined in a single place

app/src/renderer/components/network/PageBlock.vue
```diff
@@ -0,0 +1,151 @@
...
+                  <tr
+                    v-for="(precommit, index) in block.block.last_commit
+                      .precommits"
+                    :key="index"
+                    class="block data-table__row"
+                  >
+                    <td>{{ precommit.validator_address }}</td>
+                    <td>
+                      {{
+                        moment(precommit.timestamp).format(
```
sabau: same as up, would be nicer to have the formatting logic coded only once

app/src/renderer/components/network/PageBlock.vue
```diff
@@ -0,0 +1,151 @@
...
+        {
+          title: `Round`
+        }
+      ]
+    }
+  },
+  watch: {
+    $route: `getBlock`
+  },
+  mounted() {
```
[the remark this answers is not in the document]
faboweb: `getBlock` can be executed while other stuff is happening in parallel. We don't really need to wait for it. Therefor we don't need make this `async` and `await`.

app/src/renderer/components/transactions/TmLiTransaction.vue
jbibla: ahhhhhh i see. no, not my intention - will revert.

app/src/renderer/routes.js
jbibla: mistake. good catch.

app/src/renderer/vuex/modules/blocks.js
jbibla: added comment - done!

app/src/renderer/components/transactions/TmLiTransaction.vue
```diff
@@ -47,7 +47,10 @@ export default {
   },
   computed: {
     date() {
-      return moment(this.time).format(`MMM Do YYYY, HH:mm:ss`)
+      const time = moment(this.time)
+      return time.format(
+        `${moment().isSame(time, `day`) ? `` : `MMM Do YYYY `}HH:mm:ss`
```
faboweb: is there a test for showing the result when it's not the same day?

app/src/renderer/components/transactions/TmLiTransaction.vue
```diff
@@ -47,7 +47,10 @@ export default {
   },
   computed: {
     date() {
-      return moment(this.time).format(`MMM Do YYYY, HH:mm:ss`)
+      const time = moment(this.time)
```
faboweb: If you move everything to `this` destructuring should you here also use `date({ time } = this) {`.

sabau: + makes the test on this sweetier

app/src/renderer/components/transactions/TmLiTransaction.vue
sabau: As for now nothing else than a snapshot that look only at the same day version, and not the other one.
We should add a tiny test on this, I'll push it here ok?

what if we move this into app/src/renderer/scripts/time.js

sabau: Just added the test (I had to merge develop to have Date working)

PR #1934, opened 2019-02-06
Fix errors at boot time

sabau:
_Description:_
After running a yarn build:gaia from develop I encountred some errors
- --moniker seems not to be a named parameter anymore
- tls seems to be disabled by default instead of what the --help states

❤️ Thank you!

---

### DIFF
--- a/tasks/gaia.js
+++ b/tasks/gaia.js
@@ -27,7 +27,7 @@ async function initNode(
   password = `1234567890`,
   overwrite = false
 ) {
-  let command = `${nodeBinary} init --home ${homeDir} --moniker ${moniker} --chain-id ${chainId}`
+  let command = `${nodeBinary} init ${moniker} --home ${homeDir} --chain-id ${chainId}`
   if (overwrite) {
     command += ` -o`
   }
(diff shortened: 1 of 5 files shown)


tasks/gaia.js
```diff
@@ -53,7 +53,7 @@ async function initGenesis(
   nodeHomeDir
 ) {
   await makeExec(
-    `${nodeBinary} add-genesis-account ${address} 150stake,1000photino  --home ${nodeHomeDir}`
+    `${nodeBinary} add-genesis-account ${address} 200000000stake,1000photino  --home ${nodeHomeDir}`
```
jbibla: 🤑

jbibla:
from our develop or from SDK develop? can we confirm these are intended changes, not just bugs or mistakes?

jbibla:
also, changelog 😄

tasks/gaia.js
sabau: richness is gooood, but this change happened because of an error in the creation under 100000000stake

sabau:
From our develop, I removed the local-testnet folder and tried to build and start, then some errors occurred

sabau:
The seed phrase was stripped recently as was printed in non standard way, I just created an issue to restore it https://github.com/cosmos/cosmos-sdk/issues/3518

sabau:
> from our develop or from SDK develop? can we confirm these are intended changes, not just bugs or mistakes?

Yep, except the huge amount of stakes needed to boot. I shuould report also that one.

PR #1935, opened 2019-02-06
Gaia build: Keep old versions if available

### LINKED ISSUE #1855: Keep local gaia build for each commit
We currently delete gaia builds when we rebuild for a new commit. I propose we:
- keep gaia builds per commit
- have a command that checks if the gaia build exists and builds if not

sabau:

Closes #1855 

_Description:_

- Save compiled version into separate folder
- link that folder to the Gaia folder for retrocompatibility
- if the hash we want to build exists, just change the link, otherwise create it and link it

❤️ Thank you!

---


tasks/build/Gaia/localBuild.sh
```diff
@@ -1,14 +1,23 @@
...
 
 . ./COMMIT.sh
+version="$(pwd)/../../../builds/GaiaVersions/$COMMIT"
 source="$(pwd)/../../../builds/Gaia"
-mkdir -p "$source"
-export TARGET=/mnt
+mkdir -p "$version"
+rm -rf "$source"
+if [[ -f "$version/linux/gaiad" ]]; then
+    echo "Already built"
```
jbibla: do we need to be able to rebuild existing versions?

faboweb: you could just remove the existing version from your local file system. should be an edge case

faboweb:
A thought: Maybe we also default to just building the Gaia version for our current OS not all of them (this was necessary for building the Electron version we don't longer have).

tasks/build/Gaia/localBuild.sh
sabau: if we need it we can add a hook

sabau:
Nice idea, just done it, someone that is not me can double check on their local machine if everything works as expected, I cloned the clean repo and seems ok, but just to be sure

sabau:
I kept the empty linux folder because basically it does nothing else than trying to copy the nothing built in /go/bin/gaia* inside it

faboweb:
Working but getting error:
image

Plus: `yarn start` is broken because of reordering of commands in `package.json`

faboweb:
> Plus: `yarn start` is broken because of reordering of commands in `package.json`

See https://github.com/cosmos/voyager/pull/1956

sabau:
But how was it broken before #1956? 
what errors did you get? I'm using the other version and basically the only thing that changes is the command the is not detached is proxy instead of nodes 2 right?

faboweb:
`yarn start` calls `yarn backend skip-rebuild` which calls `yarn nodes 2 & yarn stargate & yarn proxy`. But `skip-rebuild` is a parameter for `yarn nodes {node count} {skip rebuilding genesis}`.

sabau:
ah that's it! so the parametrer was always ignored

sabau:
cp warning now does not happen anymore + I tried to improve a bit the readibility of localBuild.sh

PR #1952, opened 2019-02-11
Fabo/eslint pretty

faboweb:
Fixed the compatibility between eslint and prettier

Missing:
- [x] fix linting issues from updated packages
- [x] add Karolys rules
- [x] add `eslint --fix` to `yarn lint:fix`


.eslintrc.js
```diff
@@ -2,7 +2,12 @@
 
 module.exports = {
   root: true,
-  extends: ["plugin:vue/recommended", "prettier", "prettier/vue"],
+  extends: [
+    "plugin:vue/recommended",
```
sabau: in this way we override again the rules, basically this won't change prettier behaviour much

faboweb: I thought the wished outcome is to fix the html in the vue templates. This is now the case.

app/src/renderer/components/common/AnchorCopy.vue
```diff
@@ -1,8 +1,8 @@
 <template>
-  <a class="anchor-copy" @click="click"
```
faboweb: @sabau see here

app/src/renderer/components/common/TmBtn.vue
```diff
@@ -1,36 +1,32 @@
...
+        {{ icon }}
+      </i>
+      <img
         v-else-if="img"
         :src="img"
         :class="'tm-btn__img'"
         aria-hidden="true"
-      /><span v-if="value" class="tm-btn__value">{{ value }}</span></span
-    >
+      /><span v-if="value" class="tm-btn__value">{{ value }}</span>
```
sabau: It improves but is still closing/opening sibilings

faboweb: What would you expect here?

fedekunze:
unit tests failing

PR #2047, opened 2019-02-21
fee distribution module

fedekunze:
Closes #2035

**Description:**

Thank you! 🚀

---

For contributor:

- [x] Added entries in `CHANGELOG.md` with issue # and GitHub username
- [ ] Reviewed `Files changed` in the github PR explorer
- [ ] Attach screenshots of the UI components on the PR description (if applicable)
- [ ] Scope of work approved for big PRs

For reviewer:

- [ ] Manually tested the changes on the UI

Issue #2044, opened 2019-02-21
Big numbers means big responsibilities


sabau:
**Describe the bug**
Big numbers are not displayed nicely on the header with a 15" screen

**To Reproduce**
1. Login when you are rich AF
2. Look at your money and dislike the way they are printed, they make you look poorer

**Expected behavior**
See 19M or some other cool expandable shorthand that on hover or on click show the whole number

**Screenshots**
<img width="1023" alt="screenshot 2019-02-21 at 11 32 43" src="https://user-images.githubusercontent.com/1656471/53162536-84da2480-35cc-11e9-939d-a48d080dd4b7.png">

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser Chrome
 - Version 72

We have to decide if the fix I suggest make sense for everyone and/or what else we can do to show massive numbers, the chain will grow so also block numbers will hit big numbers

jbibla:
the SDK is going to send us Atoms, not uAtoms in the future... so this will be resolved when that is completed.

fedekunze:
Lol this description is golden

jbibla:
still @sabau is correct - we may have some millionaire-atom-hodlers who this will be ugly for.

faboweb:
Other UIs just shorten the numbers at some point: `1000000` becomes `1m`

jbibla:
ah true! 

1k, 1M, 1B

jbibla:
still, we agreed to do the math until the REST API returns Atoms instead of uAtoms

faboweb:
Let'S make big numbers font size smaller

jbibla:
- [ ] add a class if long numbers
- [ ] do calculation for uAtoms

jbibla:
i'd like to leave the font size as is for now - if this is an issue in the future we can add more dynamic font sizes. is that ok?

PR #2049, opened 2019-02-21
fedekunze/show rewards

### LINKED ISSUE #135: Show commission earned
I as a user
Want to see the amount of commission I earned
So that I stay engaged in holding atoms.

AC:
- [ ] a user should be able to see how many rewards they have been issued from each validator they have delegated to
- [ ] a user should be able to see a sum of the total rewards he has in all validators
- [ ] rewards should update in the UI as they update

See following event for updates on rewards (taken from https://github.com/cosmos/cosmos-sdk/blob/develop/docs/spec/distribution/state.md):

dependencies: #209

fedekunze:
Closes #135

**Description:**

Show delegation rewards on validators pages and outstanding rewards on network page

screen shot 2019-02-28 at 3 57 47 pm
screen shot 2019-02-28 at 3 57 54 pm
screen shot 2019-02-28 at 3 58 01 pm

Thank you! 🚀

---

For contributor:

For reviewer:


app/src/renderer/components/network/PageNetwork.vue
```diff
@@ -132,7 +132,10 @@ export default {
         }
       ]
     }
-  }
+  },
+  async mounted() {
```
jbibla: nice!

this will only get the data once... right? should we update it every block / every 10 blocks? is this too many requests?

app/src/renderer/components/staking/LiValidator.vue
```diff
@@ -43,7 +43,7 @@
       }}
     </td>
     <td class="li-validator__rewards data-table__row__cell__separator">
-      n/a
+      {{ rewards }}
```
jbibla: 🎉

app/src/renderer/components/staking/LiValidator.vue
```diff
@@ -97,6 +97,13 @@ export default {
       }
       return `n/a`
     },
+    rewards() {
+      const validatorRewards = this.distribution.rewards[this.validator.operator_address]
+      if(validatorRewards) {
+        return this.num.prettyInt(validatorRewards[`stake`])
```
jbibla: should this use bondingDenom instead of `stake`? or some other param?

app/src/renderer/components/staking/LiValidator.vue
```diff
@@ -97,6 +97,13 @@ export default {
...
+      }
+      return `n/a`
```
jbibla:
```suggestion
      return `--`
```

app/src/renderer/scripts/common.js
```diff
@@ -27,18 +37,6 @@ module.exports.ratToBigNumber = function(rat) {
   }
   return new BN(rat)
 }
-// TODO uncomment when validator comission is done
-// module.exports.parseValidatorShares = function(validator) {
```
jbibla: are these functions not usable now?

app/src/renderer/vuex/modules/distribution.js
```diff
@@ -0,0 +1,178 @@
+import * as Sentry from "@sentry/browser"
+import Vue from "vue"
+import { coinsToObject } from "scripts/common.js"
+
+export default ({ node }) => {
+    const emptyState = {
+        loading: false,
+        loaded: false,
+        error: null,
+        /* totalRewards use the following format:
```
jbibla: i don't know if i like having these comments like this.

app/src/renderer/components/staking/TabMyDelegations.vue
```diff
@@ -99,6 +99,12 @@ export default {
               transaction.tx.value.msg[0].value.validator_addr
             ]
         }))
+  },
+  async mounted() {
+    this.$store.dispatch(`getTotalRewards`)
+    this.yourValidators.forEach(validator => {
+      this.$store.dispatch(`getRewardsFromValidator`, validator.operator_address)
```
jbibla: this is good for now. might be an issue if a user has too many validators... maybe a good candidate for a new special endpoint on the REST API. 

@alexanderbez 

maybe the endpoint `/distribution/delegators/cosmos1address/rewards` could return the data per validator instead of as a single total value? just an idea.

jbibla: not sure about making multiple requests for this. should we request a new endpoint for this?

app/src/renderer/vuex/modules/distribution.js
```diff
@@ -0,0 +1,178 @@
...
+            { state, rootState: { session }, commit }
+        ) {
+            state.loading = true
+            try {
+                const rewardsArray = await node.getDelegatorRewards(session.address)
+                const rewards = coinsToObject(rewardsArray)
+                commit(`setTotalRewards`, rewards)
+                state.error = null
+            } catch (error) {
+                commit(`notifyError`, {
```
jbibla: no more notify

app/src/renderer/vuex/modules/distribution.js
```diff
@@ -0,0 +1,178 @@
...
+        },
+        async getRewardsFromValidator({ state, rootState: { session }, commit }, validatorAddr) {
+            state.loading = true
+            try {
+                const rewardsArray = await node.getDelegatorRewardsFromValidator(session.address, validatorAddr)
+                const rewards = coinsToObject(rewardsArray)
+                commit(`setDelegationRewards`, { validatorAddr, rewards })
+                state.error = null
+            } catch (error) {
+                commit(`notifyError`, {
```
jbibla: no more notify

app/src/renderer/vuex/modules/distribution.js
```diff
@@ -0,0 +1,178 @@
...
+                state.error = error
+            }
+            state.loading = false
+            state.loaded = true
+        },
+        async withdrawRewardsFromValidator(
+            { rootState: { wallet }, commit, dispatch },
+            { validatorAddr, password, submitType }
+        ) {
+            await dispatch(`sendTx`, {
```
jbibla: not for this issue - but this seems kind of dangerous from an implementation perspective. can we make `sendTx` a more generalized function that takes `type` and other relevant `...args`? having to add `wallet.address` and `password` like this is maybe problematic?

@faboweb @sabau ?

app/src/renderer/vuex/modules/distribution.js
```diff
@@ -0,0 +1,178 @@
...
+                type: `postWithdrawDelegatorRewardsFromValidator`,
+                to: wallet.address,
+                pathParameter: validatorAddr,
+                password,
+                submitType
+            })
+            // optimistic update
```
jbibla: 😄

app/src/renderer/vuex/modules/distribution.js
```diff
@@ -0,0 +1,178 @@
...
+            commit(`setDelegationRewards`, { validatorAddr, rewards: {} })
+            // update rewards
+            await dispatch(`getTotalRewards`)
+            await dispatch(`getRewardsFromValidator`, validatorAddr)
+        },
+        // TODO: move to a common parameters module
```
jbibla: what is your thinking here? what params are there?

app/src/renderer/vuex/modules/distribution.js
```diff
@@ -0,0 +1,178 @@
...
+        async getDistributionParameters({ commit }) {
+            state.loading = true
+            try {
+                const parameters = await node.getDistributionParameters()
```
jbibla: **not for this PR** - i would love to rename `node` to something a lot more descriptive. 

@faboweb @sabau @fedekunze any ideas?

app/src/renderer/vuex/modules/distribution.js
```diff
@@ -0,0 +1,178 @@
...
+                commit(`setDistributionParameters`, parameters)
+                state.error = null
+            } catch (error) {
+                commit(`notifyError`, {
```
jbibla: no mo' notify

jbibla: use https://github.com/cosmos/voyager/pull/2049/files#diff-b37c59a991d76e68f3b9a84431a39ab5R9 instead

app/src/renderer/vuex/modules/distribution.js
jbibla: use https://github.com/cosmos/voyager/pull/2049/files#diff-b37c59a991d76e68f3b9a84431a39ab5R9 instead

app/src/renderer/vuex/modules/distribution.js
jbibla: use https://github.com/cosmos/voyager/pull/2049/files#diff-b37c59a991d76e68f3b9a84431a39ab5R9 instead

app/src/renderer/vuex/modules/distribution.js
```diff
@@ -0,0 +1,178 @@
...
+            state.loaded = true
+        },
+        async getOutstandingRewards({ commit }) {
+            state.loading = true
+            try {
+                const oustandingRewardsArray = await node.getDistributionOutstandingRewards()
+                const oustandingRewards = coinsToObject(oustandingRewardsArray)
+                commit(`setOutstandingRewards`, oustandingRewards)
+            } catch (error) {
+                commit(`notifyError`, {
```
jbibla: no mo` notify 

use https://github.com/cosmos/voyager/pull/2049/files#diff-b37c59a991d76e68f3b9a84431a39ab5R9 instead

app/src/renderer/vuex/modules/send.js
```diff
@@ -89,13 +89,23 @@ export default ({ node }) => {
       const submitType = args.submitType || `local`
       delete args.submitType
 
-      // extract "to" address
+      // TODO: refactor to get path request parameters at once
+      // extract path parameters
       const to = args.to
```
jbibla: would `toAddress` be more descriptive?

app/src/renderer/vuex/modules/validators.js
```diff
@@ -54,6 +59,29 @@ export default ({ node }) => {
       if (validatorHash === state.validatorHash) return
       commit(`setValidatorHash`, validatorHash)
       await dispatch(`getValidators`)
+    },
+    async getValidatorDistributionInfoAndRewards({ commit }, validatorAddr) {
```
jbibla: can this be split into two functions? 

`getValidatorDistributionInfo()`
`getValidatorRewards()`

app/src/renderer/vuex/modules/validators.js
```diff
@@ -54,6 +59,29 @@ export default ({ node }) => {
...
+        const rewardsArray = await node.getValidatorRewards(validatorAddr)
+
+        const rewards = coinsToObject(rewardsArray)
+        self_bond_rewards = coinsToObject(self_bond_rewards)
+        val_commission = coinsToObject(val_commission)
+
+        commit(`setValidatorDistributionInfo`, { self_bond_rewards, val_commission, rewards })
+        state.error = null
+      } catch (error) {
+        commit(`notifyError`, {
```
jbibla: no mo' notify

https://github.com/cosmos/voyager/pull/2049/files#diff-b834fe9e3ec5cca6b2d0a8b6526c1208R9

jbibla:
please add rewards to the balance header too!

app/src/renderer/vuex/modules/distribution.js
faboweb: `api`?

jbibla: that's what i was thinking too

app/src/renderer/components/network/PageNetwork.vue
faboweb: issue states: `* rewards should update in the UI as they update`

app/src/renderer/scripts/common.js
faboweb: they are decimals now and not `1000/2500000`

app/src/renderer/vuex/modules/distribution.js
faboweb: if we don't have types this is a good documentation on the formats. what is bugging you with them?

app/src/renderer/components/staking/TabMyDelegations.vue
```diff
@@ -99,6 +99,12 @@ export default {
               transaction.tx.value.msg[0].value.validator_addr
             ]
         }))
+  },
+  async mounted() {
```
faboweb: This is also just querying the rewards once.

app/src/renderer/scripts/common.js
```diff
@@ -18,6 +18,16 @@ module.exports.shortAddress = function(address, length = 4) {
   }
 }
 
+module.exports.coinsToObject = function(coinArray) {
+  const coins = {}
```
faboweb: proposal for a reduce:
```
const coins = coinArray.reduce((dictionary,{ denom, amount }) => { dictionary[denom] = amount; return dictionary; }, {})
return coins
```

app/src/renderer/vuex/modules/distribution.js
```diff
@@ -0,0 +1,178 @@
...
+        resetSessionData({ rootState }) {
+            rootState.delegation = JSON.parse(JSON.stringify(emptyState))
+        },
+        async getTotalRewards(
+            { state, rootState: { session }, commit }
+        ) {
+            state.loading = true
+            try {
+                const rewardsArray = await node.getDelegatorRewards(session.address)
+                const rewards = coinsToObject(rewardsArray)
```
faboweb: I like the `coinsToObject` refactor

fedekunze:
thanks for your comments ! I will address the distr module comments on #2047

app/src/renderer/vuex/modules/distribution.js
```diff
@@ -0,0 +1,178 @@
...
+            state.loading = false
+            state.loaded = true
+        },
+        async withdrawAllRewards(
+            { rootState: { wallet }, dispatch },
+            { password, submitType }
+        ) {
+            await dispatch(`sendTx`, {
+                type: `postWithdrawDelegatorRewards`,
+                to: wallet.address,
```
faboweb: please always use `session.address` so we don't store the address in 2 places.

app/src/renderer/vuex/modules/distribution.js
```diff
@@ -0,0 +1,178 @@
...
+            state.loading = false
+            state.loaded = true
+        },
+        async withdrawRewardsFromValidator(
+            { rootState: { wallet }, commit, dispatch },
+            { validatorAddr, password, submitType }
+        ) {
+            await dispatch(`sendTx`, {
+                type: `postWithdrawDelegatorRewardsFromValidator`,
+                to: wallet.address,
```
faboweb:
```suggestion
                to: session.address,
```

app/src/renderer/vuex/modules/distribution.js
fedekunze: Agree, but I'd also propose to leave notifications for `devMode` only. I like to see if something fails when I'm coding. Thoughts @faboweb @sabau ?

app/src/renderer/vuex/modules/distribution.js
faboweb: we need to refactor this. but we will always need to send the password along like this if we don't store it.

faboweb: and `address` is a parameter of `postWithdrawDelegatorRewardsFromValidator`. but open for ideas.

app/src/renderer/vuex/modules/distribution.js
fedekunze: I'll delete them here but maybe we can update all of them in a later refactor

app/src/renderer/vuex/modules/distribution.js
```diff
@@ -0,0 +1,178 @@
...
+                    body: error.message
+                })
+                Sentry.captureException(error)
+                state.error = error
+            }
+            state.loading = false
+            state.loaded = true
+        },
+        async getOutstandingRewards({ commit }) {
+            state.loading = true
```
faboweb: note: we are using the same variable `loading` for rewards and parameters. so they are influencing each other. this might result into inconveniences.

app/src/renderer/vuex/modules/send.js
```diff
@@ -89,13 +89,23 @@ export default ({ node }) => {
...
+      const pathParameter = args.pathParameter
       delete args.to
+      delete args.pathParameter
 
       // get the generated tx by querying it from the backend
-      const req = to ? node[type](to, args) : node[type](args)
-      const generationRes = await req.catch(handleSDKError)
+      let request
```
faboweb: this looks very complex and screams refactor :)

app/src/renderer/vuex/modules/validators.js
faboweb: I agree

app/src/renderer/vuex/modules/distribution.js
jbibla: can we remove `.address` from the wallet module?

app/src/renderer/vuex/modules/distribution.js
jbibla: ah! multiple loading properties are required then!

app/src/renderer/vuex/modules/send.js
jbibla: 😱

fedekunze:
ledger uses wallet.address too

app/src/renderer/vuex/modules/distribution.js
faboweb: Yes please

app/src/renderer/vuex/modules/distribution.js
jbibla: they take up many lines and we don't do this anywhere else.

app/src/renderer/vuex/modules/distribution.js
jbibla: > I like to see if something fails when I'm coding.

we need to display the error message and corresponding component in the UI, just not as a notify.

app/src/renderer/vuex/modules/distribution.js
jbibla: i understand. i'm just proposing that we send a tx with type - but data that exists in getters can be populated elsewhere - instead of in the component.

faboweb: > data that exists in getters can be populated elsewhere

sure

app/src/renderer/components/staking/TabMyDelegations.vue
fedekunze: imo we should eventually use something like graphQL

app/src/renderer/vuex/modules/distribution.js
fedekunze: move all parameters into the parameters module and call all of them using for example `parameters.staking`, `parameters.distribution`, etc

app/src/renderer/vuex/modules/send.js
fedekunze: will address that on the refactor PR for this, I'd like to get the path params for the endpoints at once

app/src/renderer/vuex/modules/validators.js
fedekunze: done

app/src/renderer/vuex/modules/distribution.js
fedekunze: #2114

app/src/renderer/components/network/PageNetwork.vue
```diff
@@ -40,15 +40,23 @@
             <dl class="info_dl">
               <dt>Total Liquid {{ bondDenom }}</dt>
               <dd id="loose_tokens">
-                {{ pool.pool.loose_tokens ? pool.pool.loose_tokens : `n/a` }}
+                {{ pool.pool.not_bonded_tokens ? num.pretty(pool.pool.not_bonded_tokens) : `n/a` }}
```
sabau: Line 43 and 51 share the same logic (and 59 is partially doing that via `outstandingRewards`), can be moved to a function that takes as argument an integer

app/src/renderer/components/staking/LiValidator.vue
```diff
@@ -98,10 +105,10 @@ export default {
...
     },
     yourVotes() {
-      return this.committedDelegations[this.validator.id]
+      return this.committedDelegations[this.validator.operator_address]
         ? calculateTokens(
           this.validator,
-          this.committedDelegations[this.validator.id]
+          this.committedDelegations[this.validator.operator_address]
         )
         : BigNumber(0)
```
sabau: Why BigNumber to represent a 0? is not part of this PR, just curious

test/unit/specs/store/distribution.spec.js
```diff
@@ -106,6 +137,35 @@ describe(`Module: Fee Distribution`, () => {
...
+          validators[1].operator_address
+        )
+      })
+
+      it(`fails`, async () => {
+        const validators = [
+          { operator_address: undefined },
+          { operator_address: `cosmosvaloper1address2` },
+        ]
+        dispatch = jest.fn(async () => Promise.reject(Error(`invalid address`)))
```
sabau: dispatch does not need the jest.fn wrapping, the simplified version should work
```suggestion
        dispatch = () => Promise.reject(Error(`invalid address`))
```

app/src/renderer/vuex/modules/distribution.js
sabau: I think here the problem we have is `we don't do this anywhere else`, it's true, but we should

app/src/renderer/components/common/TmBalance.vue
```diff
@@ -16,7 +16,7 @@
       </div>
       <div v-if="rewards" class="top-section">
         <h3>Total Rewards</h3>
-        <h2>{{ rewards }}</h2>
+        <h2>{{ num.short(distribution.totalRewards[bondDenom] || 0) }}</h2>
```
faboweb: we agreed on not using `short` and wait for #2044 
also I think putting this formatting in the computed as it was before is better for testability

app/src/renderer/vuex/modules/wallet.js
```diff
@@ -51,6 +51,7 @@ export default ({ node }) => {
       commit(`setWalletAddress`, address)
       await dispatch(`queryWalletBalances`)
       dispatch(`loadDenoms`)
+      dispatch(`getTotalRewards`)
```
faboweb: you can instead use a `initializeWallet` action in the `distribution` module which will in parallel be triggered. this way the code is more modular and domain specific.

app/src/renderer/vuex/modules/distribution.js
```diff
@@ -66,9 +66,6 @@ export default ({ node }) => {
         await dispatch(`getTotalRewards`)
       }
     },
-    async initializeWallet({ dispatch }) {
-      await dispatch(`getTotalRewards`)
```
faboweb: lol this was already here. let's do it this way, see other comment.

app/src/renderer/components/staking/LiValidator.vue
faboweb: `calculateTokens` returns `BigNumber`. this way the return type stays predictable.

app/src/renderer/vuex/modules/wallet.js
fedekunze: `initializeWallet` it's not distribution module specific

fedekunze: I'd prefer to call everything on the original `initializeWallet` action

faboweb: let's rename `initializeWallet` to `initializeAccount` or `initializeSession` then

app/src/renderer/components/staking/LiValidator.vue
sabau: nice, got it thanks!

app/src/renderer/components/common/TmBalance.vue
sabau: yeah I missed this one, same as `PageNetwork`

test/unit/specs/store/distribution.spec.js
fedekunze: I like to test if the dispatches were called with the right parameters, happy to discuss it afterwards tho

app/src/renderer/components/network/PageNetwork.vue
fedekunze: > can be moved to a function that takes as argument an integer

I like this idea, I'd add also a string parameter for the default response too (e.g: `n/a`). We could move this into a helper function in `scripts` since we use it in more pages as well

app/src/renderer/components/network/PageNetwork.vue
```diff
@@ -40,23 +40,19 @@
             <dl class="info_dl">
               <dt>Total Liquid {{ bondDenom }}</dt>
               <dd id="loose_tokens">
-                {{ pool.pool.not_bonded_tokens ? num.pretty(pool.pool.not_bonded_tokens) : `n/a` }}
+                {{ pool.pool.not_bonded_tokens ? num.pretty(
```
faboweb: defensive programming

PR #2066, opened 2019-02-22
Fix Proposal page

### LINKED ISSUE #1908: Refresh proposal status after deposit
It currently doesn't update itself
Same for voting

sabau:
Closes #1908 

**Description:**
- fix accessing the page directly
- number of votes temporary fixed, when we will receive atoms instead of micro atoms this multiplier should be removed

Thank you! 🚀

---

For contributor:

For reviewer:


app/src/renderer/components/governance/PageProposal.vue
```diff
@@ -183,64 +180,55 @@ export default {
...
     },
-    totalVotes() {
-      return (
-        this.tally.yes +
-        this.tally.no +
-        this.tally.no_with_veto +
-        this.tally.abstain
-      )
+    totalVotes({tally: {yes, no, no_with_veto, abstain}} = this) {
+      return (yes + no + no_with_veto + abstain)
```
faboweb: Are these guaranteed to be defined? I.e.: `undefined + 1 + undefined = NaN`

app/src/renderer/vuex/modules/governance/proposals.js
```diff
@@ -25,34 +37,18 @@ export default ({ node }) => {
...
-                final_tally_result = JSON.parse(
-                  JSON.stringify(proposal.value.final_tally_result)
-                )
-              }
-              commit(`setProposalTally`, {
-                proposal_id: proposal.value.proposal_id,
-                final_tally_result
-              })
-            })
+            proposals.map(setProposalTally(commit, node))
```
faboweb: nice refactor

app/src/renderer/vuex/modules/governance/proposals.js
```diff
@@ -9,7 +21,7 @@ export default ({ node }) => {
     proposals: {},
     tallies: {}
   }
-  const state = JSON.parse(JSON.stringify(emptyState))
+  const state = {...emptyState}
```
faboweb: 👍

faboweb:
unit tests fail

app/src/renderer/components/governance/PageProposal.vue
sabau: In theory yes, I tried the counter check and I can't find a way to have this object only partially defined

app/src/renderer/vuex/modules/governance/proposals.js
faboweb: Note: I think if I edit `state.proposals` I also edit `emptyState.proposals`

faboweb:
conflicts

app/src/renderer/vuex/modules/governance/proposals.js
```diff
@@ -94,6 +80,7 @@ export default ({ node }) => {
         Sentry.captureException(error)
         state.error = error
       }
+      return undefined
```
jbibla: why do we want to return `undefined`?

faboweb: we did as well before, it is just more explicit this way

jbibla: i think a number of our modules do not have any return statements ... ?

app/src/renderer/vuex/modules/governance/proposals.js
```diff
@@ -70,22 +66,12 @@ export default ({ node }) => {
...
-        } else {
-          final_tally_result = JSON.parse(
-            JSON.stringify(proposal.value.final_tally_result)
-          )
-        }
-        commit(`setProposalTally`, { proposal_id, final_tally_result })
+        setProposalTally(commit, node)(proposal)
+        return proposal
       } catch (error) {
         commit(`notifyError`, {
```
jbibla: please remove the notify and add loading and loaded properties in the catch block

app/src/renderer/vuex/modules/governance/proposals.js
sabau: yeah it's what was already happening under the hood, I was thinking is more direct for readers in this way

app/src/renderer/vuex/modules/governance/proposals.js
sabau: Are you sure? I can't find any other place where we do that

sabau: I will move the flag switch after the real action, as it is now it doesn't make sense, I haven't seen this earlier

sabau: I think I will never see the error, we had a little discussion with @faboweb about this, I'll open an issue where we can develop

Here the loaded status is shared across `getProposals` and `getProposal`, when you access a proposal that `proposalId` is going to the `proposals` dictionary, but loading, loaded and error are promiscuous, same happen for transactions etc.. 

We were thinking of changing the priority in TmPage so error rule, and error in this case should not come from `proposals.error` but from either:
- `proposals.statuses[proposalId]` if we want a structure with `{loading, loaded, error}`
- `proposals.errors[proposalId]` if we want to store just errors

All this apply only to proposals that got accessed directly from URL, the others won't mind, they are already in the store thanks to `getProposals` and adding a bunch of `true false null` for every proposal scanned is not really useful I think, so we should expect those flags to be also undefined in case of succesfully fetched proposals

This shuold be generic so we can replace the word `proposal` with transaction etc..

PR #2071, opened 2019-02-24
automatic release created for v1.0.0-beta.7

faboweb:
### Added

- Enable dev mode from url param @faboweb

### Changed

- Refactored tests: PageStaking, PanelSort, TabStakingParameters @faboweb


faboweb:
@sabau version bump is created even without changes

PR #2080, opened 2019-02-25
Fabo/1337 page validator

faboweb:
Closes #2080

**Description:**

Thank you! 🚀

---

For contributor:

For reviewer:


sabau:
Seems like the `oldBondedAtoms ` function is running only with empty objects, so the coverage of the inner loop is missing

test/unit/specs/components/staking/PageValidator.spec.js
```diff
@@ -1,15 +1,29 @@
...
-import setup from "../../../helpers/vuex-setup"
+import {shallowMount, createLocalVue} from "@vue/test-utils"
 import PageValidator from "renderer/components/staking/PageValidator"
-import lcdClientMock from "renderer/connectors/lcdClientMock.js"
+import BigNumber from "bignumber.js";
 
-const { stakingParameters } = lcdClientMock.state
-
-// TODO refactor comming up, not doing too much here
+const stakingParameters = {
```
sabau: I would move all the objects that references to payloads outside the spec file, so if the payload changes we have to update only in 1 place + all tests will fail in different suites, instead of just in one, every tests that need them imports it

faboweb: I agree, and they should be tested against the swagger spec probably. but can we do this in a follow up (based on your first changes).

faboweb:
> Seems like the `oldBondedAtoms ` function is running only with empty objects, so the coverage of the inner loop is missing

nice catch

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Changed
 
+- [\#1337](https://github.com/cosmos/voyager/issues/1337) Refactor tests: TmModalHelp, PageValidator @jbibla @faboweb
```
fedekunze: this should go on `Unreleased`

app/src/renderer/vuex/getters.js
```diff
@@ -42,7 +42,7 @@ export const oldBondedAtoms = (state, getters) => {
     delegatorAddress => {
       const shares = getters.delegation.committedDelegates[delegatorAddress]
       const delegator = getters.delegates.delegates.find(
-        d => d.id === delegatorAddress
+        d => d.operator_address === delegatorAddress
```
fedekunze: 👍

test/unit/specs/components/staking/PageValidator.spec.js
```diff
@@ -22,269 +36,154 @@ const validator = Object.assign({}, lcdClientMock.state.candidates[0], {
...
       }
     })
-    wrapper = instance.wrapper
-    store = instance.store
-
-    wrapper.vm.$refs.undelegationModal = { open: () => {} }
-    wrapper.vm.$refs.delegationModal = { open: () => {} }
   })
 
   describe(`has the expected html structure`, () => {
```
fedekunze: update description

test/unit/specs/components/staking/__snapshots__/PageValidator.spec.js.snap
```diff
@@ -1,3772 +1,659 @@
...
-      maximum="100"
-      to="cosmos15ky9du8a2wlstz6fpx3p4mqpjyrm5ctpesxxn9"
-      validator="[object Object]"
-    />
-     
-    <!---->
-  </main>
-</div>
-`;
+exports[`PageValidator has the expected html structure errors signing info is missing 1`] = `"This validator is actively validating"`;
```
fedekunze: is this an expected error ?

PR #2085, opened 2019-02-25
refactored pool module tests

jbibla:
---

For contributor:

For reviewer:


app/src/renderer/vuex/modules/pool.js
```diff
@@ -7,7 +7,7 @@ export default ({ node }) => {
     loaded: false,
     error: null
   }
-  const state = JSON.parse(JSON.stringify(emptyState))
+  const state = emptyState
```
faboweb: if I edit now `state.pool.x` I change `emptyState.pool` as well. Cloning the objects is a safety pattern.

jbibla:
```suggestion
  const state = JSON.parse(JSON.stringify(emptyState))
```

test/unit/specs/store/pool.spec.js
```diff
@@ -1,64 +1,83 @@
...
+    await actions.getPool({ state, commit, rootState })
+    expect(commit).not.toHaveBeenCalled()
   })
 
   it(`should query pool`, async () => {
-    await store.dispatch(`getPool`)
-    expect(store.state.pool.pool).toEqual(mockPool)
-  })
+    const commit = jest.fn()
+    const rootState = {
```
fedekunze: why not define `rootState` above with `true` as default value and only change to `false` when needed? to avoid dup code.

faboweb: good idea. This way the `rootState` could be changed in one place instead of in many.

test/unit/specs/store/pool.spec.js
```diff
@@ -1,64 +1,83 @@
...
-const mockPool = lcdClientMock.state.pool
-const mockRootState = {
-  connection: {
-    connected: true
-  }
-}
-
 describe(`Module: Pool`, () => {
-  let store
+  let module, state, actions, mutations
```
sabau: We should avoid use reserved js words `module`

fedekunze: agree, let's change it

CHANGELOG.md
```diff
@@ -19,6 +19,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Fixed
```
fedekunze: why are there 2 fixed entries ?

app/src/renderer/vuex/modules/pool.js
```diff
@@ -29,16 +29,16 @@ export default ({ node }) => {
       try {
         const pool = await node.getPool()
         commit(`setPool`, pool)
+
         state.error = null
         state.loading = false
         state.loaded = true
       } catch (error) {
-        commit(`notifyError`, {
```
fedekunze: 👍

PR #2093, opened 2019-02-26
Fabo/2089 changed devmode flag to insecure

### LINKED ISSUE #2089: Rename devMode url flag to insecure

faboweb:
Closes #2089 

**Description:**

Thank you! 🚀

---

For contributor:

For reviewer:


jbibla:
change the html too!

faboweb:
You mean the "Development Mode" warning?

fedekunze:
don't we need to change also the features that are on `devMode` ?

fedekunze:
I think we should keep the development mode, I don't see why for eg using ledger on dev mode can be insecure

jbibla:
> I don't see why for eg using ledger on dev mode can be insecure

it wouldn't be. are you concerned it will be misleading?

fedekunze:
@jbibla yes indeed

fedekunze:
I'd kill this PR, devMode assumes it's insecure

jbibla:
> devMode assumes it's insecure

i really don't think this is true. devMode could imply all kinds of things, like we're in a semi-stable-beta, or we're displaying experimental features that have nothing to do with security.

IMO we need a feature flag (which we can call devMode) and we need an insecure flag (for using seeds / passwords on the web).

> I don't see why for eg using ledger on dev mode can be insecure

this could be misleading and i agree with you there. can we only display the insecure mode banner when the user has signed in with a seed or a password?

faboweb:
So you want 2 flags? `experimental` and `insecure`?

jbibla:
> So you want 2 flags? `experimental` and `insecure`?

i think this would be ideal but we don't need this now. i don't think we have features that need to be flagged that aren't insecure.

faboweb:
Switched to `experimental` as the flag name and updated the position of the warning on smaller screens a little

image

jbibla:
@faboweb i'm sorry! i think i was misunderstood. i think it is really important that it says insecure. if we want a dev / experimental flag we could add that too at a later date.

jbibla:
> devMode is a place where you can play with code, literally development, so in a bundled environment 
I will expect not to find a flag with that name/meaning.

@sabau does this mean you don't think we should use `devMode` at all?

---

sorry to be strong about this - but i think it's important that we show "insecure mode" for when the password or seeds are going to be in the browser.

faboweb:
Updated

PR #2100, opened 2019-02-26
Fabo/dispatch reconnected event

faboweb:
The "reconnected" event was not dispatched on a reconnect


sabau:
let's add another line also in connection.spec 🎉

PR #2125, opened 2019-02-28
Jordan/1897 little action modals

### LINKED ISSUE #1897: ActionModal on smaller screens
The `ActionModal` is pretty big. In the case of the `ModalPropose` is spanning more than 100% of the screen on my Mac. Can we improve the styling so it always fits? Google Inboxs modal is about half the size.

ModalPropose
image

ModalSend
image

jbibla:
Closes #1897 

Thank you! 🚀

---

For contributor:

For reviewer:


app/src/renderer/components/governance/PageGovernance.vue
```diff
@@ -7,6 +7,7 @@
     data-title="Governance"
   >
     <tm-btn
+      v-if="session.signedIn"
```
faboweb: Should this not be there when not signed in or should an action modal say "Please sign in"?

faboweb:
Screenshots would be nice

jbibla:
> Screenshots would be nice

sorry. always forget. @sabau can we automate this one day with browserstack? 😝

sabau:
We have the artifacts, I guess adding screenshot automatically would be too much, we should know which page / modal / state trigger the interesting part that changed.
But once a E2E test run on Browserstack the whole screencast of it will be available, we could use that one.

https://19419-99653950-gh.circle-artifacts.com/0/tmp/voyager/app/dist/index.html

app/src/renderer/components/governance/PageGovernance.vue
jbibla: ahh good point. the button should be present, the action modal should contain the sign in message.

jbibla: i don't know how to solve this now. i'd rather have it hidden then have the double button stack. so if it's ok with you, i would leave this as is for now.

faboweb: This would be inconsistent thoughout the app as on some routes the sign in is there and on others it isn't. Imo we should fix the header design so this "stacking" is not an issue anymore.

jbibla: i believe the sign in is present everywhere. what's unique is the placement of this create button actually - which i need to find a different spot for.

faboweb: Ah right this is the create button. Sorry.

faboweb: https://github.com/cosmos/voyager/issues/2157

fedekunze:
screenshots ?

jbibla:
here are some screenshots. these are still _too big_ IMO, but the solution is having a multi-step action modal. this is the best we can do with tweaking css.

screen shot 2019-03-04 at 1 06 09 pm
screen shot 2019-03-04 at 1 06 21 pm
screen shot 2019-03-04 at 1 05 59 pm

faboweb:
> The feedback thing looks really annoying on mobile. Any chance we can take it out on the modals ? cc: @faboweb

I wouldn't. We can try to position it differently.

PR #2148, opened 2019-03-04
Fabo/readd insecure mode

faboweb:
The insecure mode PR was missing the insecure URL parameter.


test/unit/specs/App.spec.js
```diff
@@ -1,5 +1,50 @@
 import { startApp, main, routeGuard } from "renderer/scripts/boot"
 
+async function start(urlParams, environment) {
```
sabau: nice!

PR #2151, opened 2019-03-04
Fabo/sign in msg

faboweb:
Added a sign in required message
Readded the primary color to buttons.

image

### DIFF
--- a/app/src/renderer/components/common/TmPage.vue
+++ b/app/src/renderer/components/common/TmPage.vue
@@ -14,9 +14,9 @@
       <slot slot="header-buttons" name="header-buttons" />
     </tm-page-header>
     <main class="tm-page-main">
-      <card-sign-in-required v-if="signinRequired && !session.signedIn" />
-      <template v-else-if="this.$slots['managed-body']">
-        <tm-data-connecting v-if="!loaded && !connected" />
+      <template v-if="this.$slots['managed-body']">
+        <card-sign-in-required v-if="signinRequired && !session.signedIn" />
+        <tm-data-connecting v-else-if="!loaded && !connected" />
         <tm-data-loading v-else-if="!loaded && loading" />
         <tm-data-error v-else-if="error" />
         <slot
--- a/app/src/renderer/components/common/TmPage.vue
+++ b/app/src/renderer/components/common/TmPage.vue
@@ -14,7 +14,8 @@
       <slot slot="header-buttons" name="header-buttons" />
     </tm-page-header>
     <main class="tm-page-main">
-      <template v-if="this.$slots['managed-body']">
+      <card-sign-in-required v-if="signinRequired && !session.signedIn" />
+      <template v-else-if="this.$slots['managed-body']">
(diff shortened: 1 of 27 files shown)


app/src/renderer/components/common/TmPage.vue
```diff
@@ -90,6 +93,10 @@ export default {
     refresh: {
       type: Function,
       default: undefined
+    },
+    signinRequired: {
```
jbibla:
```suggestion
    signInRequired: {
```

faboweb: done

test/unit/specs/components/staking/TabMyDelegations.spec.js
```diff
@@ -17,7 +17,8 @@ const getters = {
   committedDelegations: {
   },
   connected: true,
-  bondDenom: `stake`
+  bondDenom: `stake`,
+  session: { signedIn: true }
```
fedekunze: we should add a snapshot when not signed in

app/src/renderer/components/common/TmPage.vue
```diff
@@ -14,7 +14,8 @@
       <slot slot="header-buttons" name="header-buttons" />
     </tm-page-header>
     <main class="tm-page-main">
-      <template v-if="this.$slots['managed-body']">
+      <card-sign-in-required v-if="signinRequired && !session.signedIn" />
```
jbibla: should this go inside of the template below with connecting, loading, and error?

app/src/renderer/components/common/TmPage.vue
```diff
@@ -90,6 +93,10 @@ export default {
     refresh: {
       type: Function,
       default: undefined
+    },
+    signInRequired: {
```
jbibla: thinking ahead a little bit (maybe too far ahead) maybe this should be called `addressRequired` because soon you will not need to "sign in" or "be signed in" - you will just have to provide an address.

fedekunze: although the address will still be provided on the sign in session

app/src/renderer/components/common/TmPage.vue
faboweb: either way. I can do it.

app/src/renderer/components/common/TmPage.vue
faboweb: I could go either way

test/unit/specs/components/staking/TabMyDelegations.spec.js
faboweb: Done. tests should move into TmPage (which is barely tested) https://github.com/cosmos/voyager/issues/2156

app/src/renderer/components/common/TmPage.vue
jbibla: let's merge it for now.

faboweb:
Some things were missing. Will add them to your PR

PR #2185, opened 2019-03-06
display transactions on page block

### LINKED ISSUE #1959: Network page: display LiTransactions in PageBlock
parse transaction data from blocks endpoint

fedekunze:
Closes #1959 

**Description:**

Screen Shot 2019-03-08 at 4 51 27 PM

Thank you! 🚀

---

For contributor:

For reviewer:


app/src/renderer/components/network/PageBlock.vue
```diff
@@ -30,9 +32,12 @@
         <div class="row">
```
jbibla: should we `v-if="!block.block.evidence"`?

jbibla: should we `v-if="!block.block.evidence"`?

app/src/renderer/components/network/PageBlock.vue
```diff
@@ -113,10 +141,11 @@ export default {
         `queryBlockInfo`,
         $route.params.height
       )
+      await $store.dispatch(`getBlockTxs`, $route.params.height)
 
       if (
         !blockInfo &&
-        Number($route.params.height) > Number(lastHeader.height)
+				Number($route.params.height) > Number(lastHeader.height)
```
jbibla: not sure about this spacing

app/src/renderer/components/network/PageNetwork.vue
```diff
@@ -69,10 +69,10 @@
         <tbody>
           <tr v-if="blocks.length === 0" class="block data-table__row">
             <td>
-              <img class="loading-icon" src="~assets/images/loader.svg" />
```
jbibla: @sabau flagging this for you as i believe we have some linting conflict about this rule

app/src/renderer/components/transactions/LiBankTransaction.vue
```diff
@@ -4,7 +4,17 @@
     :time="transaction.time"
     :block="transaction.height"
   >
-    <template v-if="sent">
+    <template v-if="address === ''">
+      <div slot="caption">
+        Transfer&nbsp;<b>{{ full(atoms(coins.amount)) }}</b><span>&nbsp;{{ coins.denom.toUpperCase() }}</span>
```
jbibla:
```suggestion
        Sent&nbsp;<b>{{ full(atoms(coins.amount)) }}</b><span>&nbsp;{{ coins.denom.toUpperCase() }}</span>
```

app/src/renderer/components/transactions/LiStakeTransaction.vue
```diff
@@ -156,6 +156,7 @@ export default {
     timeDiff() {
       // only show time diff if still waiting to be terminated
       if (this.state !== `locked`) return ``
+      console.log(this.time)
```
jbibla: remove

app/src/renderer/components/transactions/LiTransaction.vue
```diff
@@ -39,7 +39,7 @@ export default {
       default: null
     },
     time: {
-      type: Number,
+      type: [String, Number],
```
jbibla: we are not supposed to allow this

app/src/renderer/vuex/modules/blocks.js
```diff
@@ -14,26 +14,32 @@ export default ({ node }) => {
...
     loading: false,
     error: null,
     peers: [],
     blocks: [],
     // one block, specified by height
     block: {
-      block_meta: {}
+      block: {},
+      block_meta: {},
+      trasactions: {}
```
jbibla:
```suggestion
      transactions: {}
```

app/src/renderer/vuex/modules/blocks.js
```diff
@@ -14,26 +14,32 @@ export default ({ node }) => {
...
     }
   }
 
   const mutations = {
-    setLoading: (state, loading) => (state.loading = loading),
+    setBlocksLoading: (state, loading) => (state.loading = loading),
```
jbibla: would be good to add the `loaded` property as well

test/unit/specs/connectors/api.spec.js
```diff
@@ -98,6 +98,20 @@ describe(`API`, () => {
       })
     })
 
+    it(`queries a transaction by height`, async () => {
+      axios.mockReturnValue({})
+      await client.getTxsByHeight(`2`)
+      expect(axios.mock.calls).toEqual([
+        [
+          {
+            data: undefined,
```
jbibla: is this really expected?

app/src/renderer/components/transactions/LiBankTransaction.vue
fedekunze: transfer or sent ? I feel used _Transfer_ bc i feel _Sent_ is more like you're sending something, whereas here is a transfer from a random address to another

app/src/renderer/components/transactions/LiTransaction.vue
fedekunze: in this case the time is not calculated, because the block is already available, so it's returned directly instead of getting the ISO time (number).

app/src/renderer/vuex/modules/blocks.js
fedekunze: good catch !

test/unit/specs/connectors/api.spec.js
fedekunze: we're just testing that the right endpoint is called, not its returned value

app/src/renderer/components/network/PageBlock.vue
fedekunze: what is the evidence used for in the first place ?

app/src/renderer/components/transactions/LiBankTransaction.vue
jbibla: right. i see. but it's still _sent_ _from_ and _to_

app/src/renderer/components/network/PageBlock.vue
fedekunze: deleted evidence

app/src/renderer/components/transactions/LiBankTransaction.vue
fedekunze: updated

app/src/renderer/components/transactions/LiTransaction.vue
faboweb: then we should calculate the time string in the other cases so the types are more predictable

fedekunze: I'd leave this as it is because removing it breaks the transaction dates

app/src/renderer/components/network/PageBlock.vue
sabau: yeah in theory the && should be in the newline, although I don't like it too much

app/src/renderer/components/network/PageNetwork.vue
sabau: Ok, I'll try to fix in the lint lint lint PR :) which rules are conflicting?

app/src/renderer/components/network/PageBlock.vue
faboweb: also tabspace is off. 1 tab = 2 spaces.

app/src/renderer/components/transactions/LiTransaction.vue
faboweb: no, let's please change it in this PR. we agreed that we don't have several types per prop.

fedekunze: updated to be `ISOString`

app/src/renderer/components/network/PageBlock.vue
fedekunze: > 1 tab = 2 spaces

I have that on my settings

faboweb:
sweet! much cleaner now.

app/src/renderer/components/network/PageBlock.vue
faboweb: but it is 4 spaces here.

sabau: got it, I have to add the rule to reject any tab

test/unit/specs/components/transactions/LiTransaction.spec.js
```diff
@@ -25,14 +25,14 @@ describe(`LiTransaction`, () => {
...
+    expect(LiTransaction.computed.date({ time: new Date(420000).toISOString() })).toEqual(
+      `00:07:00`
     )
   })
 
   it(`Should print the datetime if we are in a different day`, () => {
     expect(
-      LiTransaction.computed.date({ time: Date.now() + 999999999 })
-    ).toEqual(`Jan 12th 1970 13:47:21`)
+      LiTransaction.computed.date({ time: new Date(4200000 + 999999999).toISOString() })
```
sabau: let's user Date.new() if we change the date.now(function to another time all this hardcoded 4200000) will break

app/src/renderer/components/network/PageNetwork.vue
jbibla: probably how we handle self closing elements (with a slash or without, with a space or without)

test/unit/specs/components/transactions/LiTransaction.spec.js
fedekunze: `Date.new is not a function`

fedekunze: and `new Date()` defaults to `Date.now()`

faboweb: this is about not hardcoding `4200000` in the test but using `Date.now()` so the tests don't break if we change the value `4200000` in the `Date.now` mock.

sabau: yeah if you look into `test/unit/helpers/fixed_time.js` there is the mock

sabau: only `Date.now()` is mocked

sabau: My  bad, that's a typo: `Date.new()`

app/src/renderer/components/transactions/LiTransaction.vue
```diff
@@ -39,8 +39,8 @@ export default {
       default: null
     },
     time: {
-      type: Number,
-      default: null
+      type: String,
+      default: null // TODO: fails with required: true
```
jbibla: why?

fedekunze: idk

jbibla: that is not really a satisfactory answer... 

should it be required?

jbibla:
oy fede. i'm sorry this PR is taking so long to get in. the value of poquito PRs amigo!

app/src/renderer/components/network/PageBlock.vue
```diff
@@ -34,10 +34,17 @@
             class="column"
           >
             <h3>Transactions</h3>
-            <data-empty-block
+            <tm-data-msg
```
faboweb: nice

app/src/renderer/components/transactions/LiTransaction.vue
```diff
@@ -22,7 +22,7 @@
           :to="{ name: `block`, params: { height: block } }"
         >
           Block #{{ block }}&nbsp;
-        </router-link>@ {{ date }}
+        </router-link>{{ date !== `Invalid date` ? `@ ${date}` : `` }}
```
sabau: nice to have: moving into a computed

fedekunze: #2250

test/unit/specs/components/transactions/LiTransaction.spec.js
```diff
@@ -25,14 +25,20 @@ describe(`LiTransaction`, () => {
...
+      `00:07:00`
     )
   })
 
   it(`Should print the datetime if we are in a different day`, () => {
     expect(
-      LiTransaction.computed.date({ time: Date.now() + 999999999 })
-    ).toEqual(`Jan 12th 1970 13:47:21`)
+      LiTransaction.computed.date(
+        { time: new Date(4200000 + 999999999).toISOString() }
```
sabau: I still think is more readable with new Date(Date.now() + 999..) it's basically for the reader to know the rendering changes when you change day
Even better would be to change the 9999....99 with a meaningful multiplication.

that's just becasuse is hard to notice that 
42000 < 420000 < 4200000

sabau: the `day * 2` is in the lcdClientMock if you want to recycle from there (before it's dropped)

fedekunze: updated

PR #2214, opened 2019-03-07
Jordan/2188 page network bug

### LINKED ISSUE #2188: Network page error message is incorrect
**Describe the bug**

Network page error message is incorrect while cosmos wallet is trying to connect to a node.

**To Reproduce**

- open cosmos wallet without being connected to the API
- go to network page

**Expected behavior**

Should show the waiting for connection component.

**Screenshots**

screen shot 2019-03-06 at 7 48 02 am

jbibla:
Closes #2188 

**Description:**

- using TmPage loading, loaded, dataset props
- updated Connecting Msg copy
- added `blocks.loaded` to module

Thank you! 🚀

---

For contributor:

For reviewer:


app/src/renderer/components/common/TmPage.vue
```diff
@@ -87,7 +87,7 @@ export default {
       default: undefined
     },
     dataset: {
-      type: Array,
+      type: [Array, Object],
```
jbibla: i know we agreed to not permit this. however, this might be time for an exception _or_ please share a suggestion for how to avoid this.

app/src/renderer/components/network/PageNetwork.vue
```diff
@@ -1,7 +1,12 @@
 <template>
-  <tm-page data-title="Network">
-    <tm-data-error v-if="!connected" />
-    <template v-else>
+  <tm-page
+    :loading="blocks.loading"
+    :loaded="blocks.loaded"
+    :error="blocks.error"
+    :dataset="lastHeader"
```
jbibla: page network is a bit strange because it doesn't need the "items" (dataset) to load. it just needs a `connection` to render the top part of the page. the `blocks` are the second "dataset". `lastHeader` and `connection` don't have loading, loaded, or error states...

reveals room for a refactor - but this might be sufficient for now.

app/src/renderer/components/common/TmPage.vue
faboweb: this needs to be an array or more then this needs to be refactored. see line 28

faboweb: proposal: change to "dataAvailable" and make it a boolean. checks are then performed on each page on the individual dataset

sabau: Also cheaper to pass

app/src/renderer/components/network/PageNetwork.vue
sabau: I think we have in the store the list of blocks, but I like more @faboweb idea of booleanize

app/src/renderer/components/common/TmPage.vue
jbibla: ok makes sense. do you guys think this should be done in this PR and not separately?

app/src/renderer/components/network/PageNetwork.vue
jbibla: me too

app/src/renderer/components/common/TmPage.vue
faboweb: let's do it in here. the change needs to happen in only 3 places and we don't pile up more dept.

jbibla: > in only 3 places

yeah. one line 😝

sabau:
I think this should be merged right after TmPage and PageNetwork stuff is done, all good and and is a bug fix

PENDING.md
```diff
@@ -1,4 +1,17 @@
+### Added
+
+- [\#2104](https://github.com/cosmos/voyager/issues/2104) fix transaction times @fedekunze
```
fedekunze: why so many `PENDING` entries ?

app/src/renderer/components/governance/PageGovernance.vue
```diff
@@ -1,11 +1,5 @@
 <template>
-  <tm-page
-    :tabs="tabs"
-    :dataset="proposalList"
-    search="proposals"
-    class="governance"
-    data-title="Governance"
-  >
+  <tm-page :tabs="tabs" class="governance" data-title="Governance">
```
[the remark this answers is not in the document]
jbibla: yeah, it should be covered in the proposal tab though, not up here.

PR #2287, opened 2019-03-15
Switch from hash to tag

sabau:
**Description:**
- Use Tags

Thank you! 🚀

---

For contributor:

For reviewer:


tasks/build/Gaia/COMMIT.sh
```diff
@@ -2,4 +2,4 @@
 
 # This is the commit of the SDK version to use for building Gaia.  We use an
```
fedekunze: fix commits 
```suggestion
# This is the SDK release version to use for building Gaia.
```

tasks/build/Gaia/COMMIT.sh
```diff
@@ -2,4 +2,4 @@
...
 # explicit hash instead of a tag so we don't have to trust GitHub.
```
fedekunze:
```suggestion
```

jbibla: .... why are we switching to tags if this means trusting github?

faboweb: Because if the SDK team decides to switch the version tag to another commit, we are covered. it is easier to identify and talk about a version tag then about a cryptic commit hash.

jbibla: good reasoning.

tasks/build/Gaia/COMMIT.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/sh
 
-# This is the commit of the SDK version to use for building Gaia.  We use an
+# This is the SDK release version to use for building Gaia.
 # explicit hash instead of a tag so we don't have to trust GitHub.
-export COMMIT=4c5038018163fbb51ad882c0505df4958dec2ea8
+export COMMIT=v0.33.0
```
faboweb: can we move this someplace more obvious while we are on it? Like on in `config.js`?

fedekunze: + renaming the file from `COMMIT` to `VERSION`

jbibla:
@sabau looks like a small change is required before we merge

tasks/build/Gaia/COMMIT.sh
sabau: the problem is that this is used only in bash / CI, I would prefer to have it either sh or just an file with the string I will put in the TAG variable

tasks/build/Gaia/COMMIT.sh
sabau: Plus we have the tagged docker images :) as soon as we have time we should remove all the building here and just run that container

tasks/build/Gaia/COMMIT.sh
fedekunze: kk let's create an issue for that ?

faboweb:
e2e breaking

sabau:
Fixed it! sorry I stopped working on this in favour of Browserstack, but was just two export missing

.circleci/config.yml
```diff
@@ -124,31 +124,28 @@ jobs:
...
           paths:
             - *GAIA
       - persist_to_workspace:
           root: *GAIA
           paths:
             - "*"
-      # - store_artifacts:
-      #     path: builds/Gaia/linux_amd64/*
+      - store_artifacts:
+          path: builds/Gaia/linux_amd64/*
```
fedekunze: why linux?

.circleci/config.yml
```diff
@@ -124,31 +124,28 @@ jobs:
...
-          key: v8-Gaia-{{ checksum "tasks/build/Gaia/COMMIT.sh" }}
-
-      # If Gaia isn't in the cache then build it.
+          key: v8-gaia-{{ checksum "tasks/build/Gaia/VERSION" }}
       - run: |
           if [ ! -d /tmp/gaia ]; then
             cd tasks/build/Gaia
-            . ./COMMIT.sh
-            TARGET=/tmp/gaia PLATFORM=linux ./build.sh
+            TAG=`cat ./VERSION` TARGET=/tmp/gaia PLATFORM=linux ./build.sh
```
fedekunze: why linux only?

.circleci/config.yml
faboweb: the circle ci environment docker we run in is linux I suppose

.circleci/config.yml
sabau: It runs on top of alpine so it's the only supported OS

fedekunze: is that documented somewhere ?

sabau: It's maybe not that straightforward now that you ask.
we are using the official `circleci/golang:1.11` 
that is built on top of `golang:1.11.1` 
that is built on top of `alpine:stable`

PR #2310, opened 2019-03-18
Fabo/browserstack

faboweb:
Anyway blocked on: https://github.com/browserstack/browserstack-integration-nodejs/issues/11


.circleci/config.yml
```diff
@@ -199,6 +199,8 @@ jobs:
     environment:
       - BINARY_PATH: "/tmp/voyager/linux_amd64/gaiacli"
       - NODE_BINARY_PATH: "/tmp/voyager/linux_amd64/gaiad"
+      - RUN_ON_BSTACK: true
```
jbibla:
```suggestion
      - RUN_ON_BROWSERSTACK: true
```

.circleci/config.yml
```diff
@@ -199,6 +199,8 @@ jobs:
...
+      - BSTACK_LOCAL: true
```
jbibla:
```suggestion
      - BROWSERSTACK_LOCAL: true
```

package.json
```diff
@@ -85,7 +86,7 @@
     "lodash": "4.17.11",
     "markdown-it": "8.4.1",
     "moment-timezone": "0.5.16",
-    "nightwatch": "1.0.19",
+    "nightwatch": "^0.9",
```
jbibla: ?

.circleci/config.yml
faboweb: Those are environment variables used by browserstack I didn't name them myself

package.json
faboweb: We are blocked as the browserstack tunnel integration doesn't work with nightwatch `^1.x`

faboweb:
We need to switch to this: https://github.com/browserstack/nightwatch-browserstack

package.json
sabau: damn this is so annoying I'm trying to make it work anyway XD

test/e2e/tests/test.spec.js
```diff
@@ -3,14 +3,18 @@ module.exports = {
   'Demo test' : function (browser) {
     browser
       .url(browser.launch_url)
+    browser
```
sabau: Are they all atomic due to version `0.9`?

faboweb: I think I just played around because Browserstack didn't work as expected. It would execute a lot of the commands before the page was loaded.

faboweb:
Closing as anyway outdated

PR #2321, opened 2019-03-20
Fabo/2301 keybase request throttle

### LINKED ISSUE #2301: Keybase requests are inefficient and failing
**Describe the bug**

- [ ] keybase requests are failing sporadically
- [ ] we can request ONLY the avatar - instead of receiving the whole user object and parsing it out locally
    - https://keybase.io/docs/api/1.0/call/user/lookup
    - https://keybase.io/docs/api/1.0/user_objects
- [ ] failed keybase requests don't need a notify

**To Reproduce**
- visit the validators table

faboweb:
Closes #2301 

**Description:**

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/app/src/renderer/vuex/modules/keybase.js
+++ b/app/src/renderer/vuex/modules/keybase.js
@@ -45,29 +45,24 @@ export default () => {
       return lookupId(state, keybaseId)
     },
     async getKeybaseIdentities({ dispatch, commit, state }, validators) {
-      try {
-        state.loading = true
-        const identities = await Promise.all(
-          validators.map(async validator => {
-            if (validator.description.identity) {
-              return dispatch(
-                `getKeybaseIdentity`,
-                validator.description.identity
-              )
-            }
-          })
-        )
-        state.error = null
-        state.loading = false
-        state.loaded = true
-        commit(`setKeybaseIdentities`, identities.filter(x => !!x))
+      state.loading = true
+      const identities = await Promise.all(
+        validators.map(async validator => {
+          if (validator.description.identity) {
+            return dispatch(
+              `getKeybaseIdentity`,
+              validator.description.identity
+            )
+          }
+        })
+      )
+      state.error = null
+      state.loading = false
+      state.loaded = true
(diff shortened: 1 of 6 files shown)


app/src/renderer/vuex/modules/keybase.js
```diff
@@ -75,3 +69,35 @@ export default () => {
     mutations
   }
 }
+
+async function lookupId(state, keybaseId) {
```
jbibla: why do we ever need the ID?

app/src/renderer/vuex/modules/keybase.js
```diff
@@ -75,3 +69,35 @@ export default () => {
...
+        avatarUrl: user.pictures.primary.url,
+        userName: user.basics.username,
+        profileUrl: `https://keybase.io/` + user.basics.username
+      }
+    }
+  }
+}
+
+async function lookupUsername(state, username) {
+  const urlPrefix = `https://keybase.io/_/api/1.0/user/lookup.json?usernames=`
```
jbibla: i believe you're still requesting the full user object...

```
User objects are massive

...

To avoid getting more data than you need, you can (should!) specify a comma-separated list of fields you desire. Acceptable values are:

pictures
basics
public_keys
profile
proofs_summary
remote_key_proofs
sigs
cryptocurrency_addresses NEW
```

from https://keybase.io/docs/api/1.0/call/user/lookup

app/src/renderer/vuex/modules/keybase.js
faboweb: we get the ID from chain, the validators only enter the ID

app/src/renderer/vuex/modules/keybase.js
faboweb: Nice tip

app/src/renderer/vuex/modules/keybase.js
```diff
@@ -1,4 +1,5 @@
 import axios from "axios"
+import moment from "moment"
```
sabau: I think moment is being used through the externals

PR #2384, opened 2019-03-29
Revert "Fabo/changelog tool"

fedekunze:
Reverts cosmos/lunie#2381

Command doesn't work and doesn't add any content to `changes/`


sabau:
Agree, I tested just that does not returns errors

PR #2385, opened 2019-03-29
Fabo/changelog tool

faboweb:
Same but with fixed changed folder path


tasks/changelog.js
```diff
@@ -0,0 +1,93 @@
...
+        validate: function(value) {
+          if (value) return true
+
+          return `You need to specify the change.`
+        }
+      },
+      {
+        type: `input`,
+        name: `issue`,
+        message: `(optional) Which GitHub issue is discussing this change?`
```
fedekunze: can we make this mandatory ? adding the PR number also works

faboweb: Sure we can. Can you tell what the benefits are of adding the PR number to the changelog?

fedekunze: so people can follow the link to the PR that introduced a specific entry (it can be useful for reference, ease of debugging, etc)

sabau: Nice idea, it gives context in the long run if someone wants to come to the repo and read what happened / why stuff was done

fedekunze: Or we can put it on a blockchain :D

faboweb: okidoki, doing it

faboweb: done

PR #2391, opened 2019-04-01
Fabo/ catching potential error when sending

faboweb:
Closes #ISSUE

**Description:**

Thank you! 🚀

---

For contributor:

For reviewer:


app/src/renderer/vuex/modules/send.js
```diff
@@ -200,6 +200,11 @@ export default ({ node }) => {
       // check response code
       assertOk(res)
 
+      // Sometimes we get back failed transactions, which shows only by them having a `code` property
+      if (res.code) {
+        throw new Error(`Something went wrong while sending the transaction.`)
```
jbibla: any way we can be more specific?

fedekunze: I agree we should display a better error, but we'd have to hardcode the base errors from the SDK. See https://github.com/cosmos/cosmos-sdk/issues/4013

app/src/renderer/vuex/modules/send.js
```diff
@@ -200,6 +200,12 @@ export default ({ node }) => {
       // check response code
       assertOk(res)
 
+      // Sometimes we get back failed transactions, which shows only by them having a `code` property
+      if (res.code) {
```
sabau: nice catch!

app/src/renderer/vuex/modules/send.js
```diff
@@ -265,3 +271,44 @@ function handleSDKError(err) {
   }
   throw new Error(message)
 }
+
+function getErrorMessage(code) {
```
sabau: I think would be nice to have this out of any particular module, to have it usable by others

PR #2402, opened 2019-04-02
Fabo/ update dependencies

faboweb:
Closes #ISSUE

**Description:**

Thank you! 🚀

---

For contributor:

For reviewer:


app/src/renderer/connectors/tendermint-ws.js
```diff
@@ -0,0 +1,149 @@
+'use strict'
+
+/*
+* Extracted from https://github.com/nomic-io/js-tendermint
```
jbibla: awesome!

app/src/renderer/scripts/boot.js
```diff
@@ -51,7 +51,12 @@ export const startApp = async (
   enableGoogleAnalytics = _enableGoogleAnalytics
 ) => {
   Vue.use(Router)
-  Vue.use(Tooltip, { delay: 1 })
+  Vue.use(Tooltip, {
+    delay: 500,
+    placement: `bottom`,
```
jbibla: i think if you don't set this then it uses some kind of "auto" placement feature... so why always bottom?

faboweb: I will change it back to auto. There were some bugs when updating this library. I need to revert the update. Might be interesting to migrate to another more maintained lib.

jbibla: feel free to merge after changing this back to auto

PR #2485, opened 2019-04-25
Fabo/2265 memo

### LINKED ISSUE #2265: Add memo to transactions via UI
There isn't any way to set custom memo message to transaction sent via Voyager.
Can we have an option to include it when sending coins?
Would be nice to see the memos in the Transactions listing view.

faboweb:
Closes #2265

**Description:**

Adds memo input and displaying. Strips (new) Lunie suffix when displaying the memo.

image
image
image

Thank you! 🚀

---

For contributor:

For reviewer:


app/src/renderer/vuex/modules/send.js
```diff
@@ -74,7 +74,7 @@ export default ({ node }) => {
         gas: args.gas,
         gas_prices: args.gas_prices,
         simulate: args.simulate,
-        memo: `Sent via Lunie`
+        memo: `${args.memo ? args.memo + ` ` : ``}(Sent via Lunie)`
```
fedekunze: imho this goes against what the user inputs on the memo input field. You should either prefill the input field with `Sent via Lunie` as a default. You don't know if the transaction that's required on exchanges will be valid for them if you add the extra `Sent via Lunie`.

app/src/renderer/components/wallet/SendModal.vue
```diff
@@ -115,7 +127,8 @@ export default {
     address: ``,
     amount: null,
     denom: ``,
-    num
+    num,
+    memo: null
```
fedekunze: See comment, imo it's more transparent if you prefill this with `Sent via Lunie`
```suggestion
    memo: `Sent via Lunie`
```

faboweb:
@fedekunze see new proposal. I want to keep the suffix in though.

jbibla:
> have another mean on the SDK

if you build it, we will use it. but i think this is good for now.

@faboweb can we move this to the fee screen instead of adding it to the first page?

app/src/renderer/components/wallet/SendModal.vue
```diff
@@ -90,6 +90,21 @@
         type="between"
       />
     </tm-form-group>
+    <tm-form-group
+      class="action-modal-group"
+      field-id="memo"
+      field-label="Memo"
+    >
+      <span class="input-suffix">
+        {{ (memo ? " - " : "") + "(Sent via Lunie)" }}
```
jbibla:
```suggestion
        {{ (memo ? " + " : "") + "(Sent via Lunie)" }}
```

jbibla: there is also a max memo size specified in the genesis file which should be included as part of the form validation (has to include Sent via Lunie too)

faboweb:
> @faboweb can we move this to the fee screen instead of adding it to the first page?

I had it there first but didn't want to pollute the other transaction types for jow.
I still could put it there, but then the screen is inconsistent.

app/src/renderer/components/wallet/SendModal.vue
faboweb: Are you sure about the plus?

Compare:
`Memo xyz + (Sent via Lunie)`
`Memo xyz - (Sent via Lunie)`

jbibla: no. you are right. i was trying to show that it was appended. we can leave it as a dash.

jbibla:
> I had it there first but didn't want to pollute the other transaction types for jow. I still could put it there, but then the screen is inconsistent.

if there is a memo field on every tx type we should include this option on every tx. but i would recommend hiding the input initially and having a link-button that says "customize memo" or something like that. not everyone will want to use this and many won't know what it is for.

thanks for adding the validation.

jjangg96:
Some exchange use memo as an identification of user.
If you add some sentence to memo, It will make error of deposit.

dlguddus:
I represent strong disagreement on default memo value by the behalf of Korean delegator community.
They really don't like it, related to exchange related problems.

fedekunze:
> Some exchange use memo as an identification of user.
If you add some sentence to memo, It will make error of deposit.

That's what I meant with my original comment as well

faboweb:
Ok, we will remove the forced suffix.

faboweb:
> @faboweb can we move this to the fee screen instead of adding it to the first page?

The memo is also included in the simulation. But we could just not do that and be a little unprecise.

faboweb:
Updated

mdyring:
There seems to be very big demand for this feature as some of the larger exchanges require a memo for deposits. Same goes for Binance listing ATOMs tomorrow.

PR #2493, opened 2019-04-26
Fabo/ledger version check

faboweb:
Throws an error if the Ledger Cosmos App version is below 1.0.1 to prevent certain errors.


app/src/renderer/vuex/modules/ledger.js
```diff
@@ -98,16 +98,13 @@ export default () => {
...
-      } catch (error) {
-        Sentry.captureException(error)
-        commit(`setLedgerError`, error)
+      const response = await state.cosmosApp.get_version()
+      actions.checkLedgerErrors(response)
+      const { major, minor, patch, test_mode } = response
+      const version = { major, minor, patch, test_mode }
+      commit(`setCosmosAppVersion`, version)
+      if (minor === 0 && patch === 0) {
+        throw new Error(`Comos Ledger App is outdated. Please update to at least version 1.0.1`)
```
fedekunze:
```suggestion
        throw new Error(`Comos Ledger App is outdated. Please update it to the latest version`)
```

faboweb:
Note, this doesn't check for the Ledger firmware update, which caused problems. And which we should also check.

app/src/renderer/vuex/modules/ledger.js
```diff
@@ -98,16 +98,13 @@ export default () => {
...
-        commit(`setCosmosAppVersion`, version)
-      } catch (error) {
-        Sentry.captureException(error)
-        commit(`setLedgerError`, error)
+      const response = await state.cosmosApp.get_version()
+      actions.checkLedgerErrors(response)
+      const { major, minor, patch, test_mode } = response
+      const version = { major, minor, patch, test_mode }
+      commit(`setCosmosAppVersion`, version)
+      if (minor === 0 && patch === 0) {
```
fedekunze: this will fail when `v1.5.0` (i.e the next upcoming version) gets released
```suggestion
      if (major < 1 || (major === 1 && minor === 0 && patch === 0)) {
```

jbibla:
> Note, this doesn't check for the Ledger firmware update, which caused problems. And which we should also check.

do you want to include this in this PR or separately?

faboweb:
We can merge this without

PR #2537, opened 2019-05-03
Miguel/1995 show warning use ledger mobile safari firefox

### LINKED ISSUE #1995: Show warning if trying to use Ledger in Safari or Firefox
Description:

Safari throws an error if you click on the icon to connect ledger on mobile: [https://voyager-staging.interblock.io](https://voyager-staging.interblock.io)

unadjustednonraw_mini_2cb2

migueog:
Closes #1995 

**Description:**

- Prevent users from signing in or creating a new account if their user agent is not Chrome

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/src/components/common/ActionModal.vue
+++ b/src/components/common/ActionModal.vue
@@ -92,7 +92,7 @@
             }}
           </div>
           <div v-if="!browserWithLedgerSupport">
-            `Please use Chrome, Brave, or Opera.`
+            Please use Chrome, Brave, or Opera.
           </div>
         </HardwareState>
         <TmFormGroup
--- a/src/components/common/ActionModal.vue
+++ b/src/components/common/ActionModal.vue
@@ -83,7 +83,7 @@
           icon="usb"
           :loading="!!sending"
         >
-          <div v-if="checkBrowser">
+          <div v-if="browserWithLedgerSupport">
             {{
               sending
                 ? `Please verify and sign the transaction on your Ledger`
@@ -91,8 +91,8 @@
             the Cosmos app`
             }}
           </div>
-          <div v-if="!checkBrowser">
-            {{ `Please use Chrome, Brave, or Opera.` }}
+          <div v-if="!browserWithLedgerSupport">
+            `Please use Chrome, Brave, or Opera.`
           </div>
         </HardwareState>
         <TmFormGroup
@@ -151,7 +151,7 @@
                 v-else
                 color="primary"
                 value="Submit"
-                :disabled="checkBrowser"
+                :disabled="browserWithLedgerSupport"
                 @click.native="validateChangeStep"
               />
             </div>
@@ -268,7 +268,7 @@ export default {
       }
       return signWithLocalKeystore
     },
-    checkBrowser() {
+    browserWithLedgerSupport() {
       if (
         navigator.userAgent.includes(`Chrome`) ||
         navigator.userAgent.includes(`Opera`)
--- a/src/components/common/ActionModal.vue
+++ b/src/components/common/ActionModal.vue
@@ -151,7 +151,7 @@
                 v-else
                 color="primary"
                 value="Submit"
-                :disabled="browserWithLedgerSupport"
+                :disabled="!browserWithLedgerSupport"
                 @click.native="validateChangeStep"
               />
             </div>
--- a/src/components/common/ActionModal.vue
+++ b/src/components/common/ActionModal.vue
@@ -1,14 +1,6 @@
 <template>
-  <transition
-    v-if="show"
-    name="slide-fade"
-  >
-    <div
-      v-focus-last
-      class="action-modal"
-      tabindex="0"
-      @keyup.esc="close"
-    >
+  <transition v-if="show" name="slide-fade">
+    <div v-focus-last class="action-modal" tabindex="0" @keyup.esc="close">
       <div class="action-modal-header">
         <img
           class="icon action-modal-atom"
@@ -25,22 +17,13 @@
           <i class="material-icons">close</i>
         </div>
       </div>
-      <div
-        v-if="requiresSignIn"
-        class="action-modal-form"
-      >
+      <div v-if="requiresSignIn" class="action-modal-form">
         <p>You need to sign in to submit a transaction.</p>
       </div>
-      <div
-        v-else-if="step === `txDetails`"
-        class="action-modal-form"
-      >
+      <div v-else-if="step === `txDetails`" class="action-modal-form">
         <slot />
       </div>
-      <div
-        v-else-if="step === `fees`"
-        class="action-modal-form"
-      >
+      <div v-else-if="step === `fees`" class="action-modal-form">
         <TmFormGroup
           v-if="session.experimentalMode"
           :error="$v.gasPrice.$error && $v.gasPrice.$invalid"
@@ -81,10 +64,7 @@
           :gas-price="Number(gasPrice)"
         />
       </div>
-      <div
-        v-else-if="step === `sign`"
-        class="action-modal-form"
-      >
+      <div v-else-if="step === `sign`" class="action-modal-form">
         <TmFormGroup
           v-if="signMethods.length > 1"
           class="action-modal-form-group"
@@ -105,9 +85,9 @@
         >
           <div v-if="session.browserWithLedgerSupport">
             {{
-            sending
-            ? `Please verify and sign the transaction on your Ledger`
-            : `Please plug in your Ledger&nbsp;Nano&nbsp;S and open
+              sending
+                ? `Please verify and sign the transaction on your Ledger`
+                : `Please plug in your Ledger&nbsp;Nano&nbsp;S and open
             the Cosmos app`
             }}
           </div>
--- a/src/components/common/ActionModal.vue
+++ b/src/components/common/ActionModal.vue
@@ -374,12 +374,15 @@ export default {
       }
     },
     async connectLedger() {
-      if (navigator.userAgent.includes(`Chrome`) || navigator.userAgent.includes(`Opera`)) {
+      if (
+        navigator.userAgent.includes(`Chrome`) ||
+        navigator.userAgent.includes(`Opera`)
+      ) {
         try {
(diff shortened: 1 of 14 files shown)


src/components/common/TmSessionWelcome.vue
```diff
@@ -79,7 +79,14 @@ export default {
   },
   methods: {
     setState(value) {
-      this.$store.commit(`setSessionModalView`, value)
+      if (navigator.userAgent.includes(`Chrome`)) {
+        this.$store.commit(`setSessionModalView`, value)
+      } else {
+        this.$store.commit(`notifyError`, {
```
faboweb: notifyError is a better we avoid these days. Can you better show a message inline?
also the problem is only signing in with Ledger, here you throw an error whenever the user clicks any one of the sign in options. I would recommend to show the warning only on the sign in screen with Ledger. And when the user tryes to sign a transaction in the ActionModal.

src/components/common/TmSessionWelcome.vue
```diff
@@ -79,7 +79,14 @@ export default {
...
+          title: ``,
+          body: `Please use Chrome or Brave.`
```
faboweb: Opera is also a working browser

migueog: @faboweb Thanks! I've added that to the message.

src/components/common/TmSessionWelcome.vue
migueog: @faboweb Thanks for the feedback. 

I've fixed this so that the error message only occurs when an user has selected the Ledger option. I've also applied this to the ActionModal component.

src/components/common/TmSessionHardware.vue
```diff
@@ -58,12 +58,19 @@ export default {
...
+      ) {
+        this.setStatus(`detect`)
+        try {
+          await this.$store.dispatch(`connectLedgerApp`)
+        } catch (error) {
+          this.setStatus(`connect`)
+          this.setConnectionError(error.message)
+        }
+      } else {
+        this.setConnectionError(`Please use Chrome, Brave, or Opera`)
```
faboweb: Please add the warning right away. So when the user using Firefox visits the page, he sees the message. Then also disable the "Submit"/"Sign In" button in that case.
The thinking: right now the user expects everything to work. then he presses a button expecting to get signed in and then is sad as it doesn't work. where if we show the message and disable the button, the user never creates an expectation.

src/components/common/TmSessionWelcome.vue
```diff
@@ -12,6 +12,13 @@
 
       <div class="tm-session-main">
         <LiSession
+          v-if="!checkBrowser"
```
migueog: @faboweb Just wondering if this approach is correct. I feel like it's kinda clunky but it gets the job done. 

I was thinking about modifying the `LiSession` component to achieve the same effect but it seemed like I could spend my time on other more pressing issues and leave this to be refactored at a later time.

src/components/common/ActionModal.vue
```diff
@@ -262,6 +268,16 @@ export default {
       }
       return signWithLocalKeystore
     },
+    checkBrowser() {
```
jbibla: i might call this `browserWithLedgerSupport` or something a little bit more descriptive like that.

src/components/common/ActionModal.vue
```diff
@@ -83,12 +83,17 @@
...
+          <div v-if="checkBrowser">
+            {{
+              sending
+                ? `Please verify and sign the transaction on your Ledger`
+                : `Please plug in your Ledger&nbsp;Nano&nbsp;S and open
+            the Cosmos app`
+            }}
+          </div>
+          <div v-if="!checkBrowser">
+            {{ `Please use Chrome, Brave, or Opera.` }}
```
jbibla:
```suggestion
            Please use Chrome, Brave, or Opera.
```

src/components/common/TmSessionWelcome.vue
```diff
@@ -12,6 +12,13 @@
 
       <div class="tm-session-main">
         <LiSession
+          v-if="!browserWithLedgerSupport"
+          icon="usb"
+          title="Sign in with Ledger Nano S"
```
colw: This is not important, but perhaps a non-breaking space between `Nano` and `S`?

Screenshot 2019-05-07 at 17 30 46
vs.

Screenshot 2019-05-07 at 17 30 58

It's just the first thing I saw, and it bothered me, but the nbsp doesn't look that great either. Am I being too nit picky @jbibla ? :)

*Or perhaps change the total width.

src/components/common/TmSessionWelcome.vue
```diff
@@ -12,6 +12,13 @@
...
+          subtitle="Please use Chrome, Opera, or Brave."
```
colw: If the LiSession doesn't have a click event  we should remove the `:hover` pointer and background changes. Could confuse the user.

Ideally the 'Please use…` message is highlighted, but perhaps not a deal breaker.

src/components/common/ActionModal.vue
```diff
@@ -262,6 +268,16 @@ export default {
       }
       return signWithLocalKeystore
     },
+    browserWithLedgerSupport() {
```
colw: I would extract this function, and use it here and in `TmSessionWelcome`.  Best not to repeat it.

src/components/common/TmSessionWelcome.vue
migueog: @colw Yeah, I don't like how this looks. I'll change it.

jbibla: no such thing as _too_ nit picky 😉 

`nbsp;` is a great idea.

src/components/common/ActionModal.vue
```diff
@@ -146,6 +151,7 @@
                 v-else
                 color="primary"
                 value="Submit"
+                :disabled="browserWithLedgerSupport"
```
jbibla: why is this disabled if browser has ledger support?

src/components/common/TmSessionWelcome.vue
colw: 😆 

Also, I should have been clear this was taken on the iPhone 8 (simulator).

src/components/common/ActionModal.vue
jbibla: good call! @migueog you can move it to `session.js`

src/components/common/TmSessionWelcome.vue
colw: If you do go with a `nbsp;` it would be a a good idea to make it obvious it's there, either through a unicode character, or even a small comment.

src/components/common/ActionModal.vue
migueog: @colw Ah great thanks!

@jbibla It has been moved! Thanks for the help

src/components/common/ActionModal.vue
migueog: Oops. This is a typo. This button should be disabled if the browser does not have ledger support

src/components/common/TmSessionWelcome.vue
migueog: @colw I agree with you. I think I'll need to modify `LiSession` to do this but I feel like this should be done as a separate issue.

@jbibla Thoughts?

src/components/common/TmSessionWelcome.vue
```diff
@@ -14,7 +14,7 @@
         <LiSession
           v-if="!browserWithLedgerSupport"
           icon="usb"
-          title="Sign in with Ledger Nano S"
+          title="Sign in with Ledger Nano nbsp; S"
```
colw: This is rendered directly in Vue. You'll need to find another way to encode it.

faboweb:
I reduced the complexity a little by not going a action+mutation way but setting the value directly in the store on startup

src/components/common/TmSessionWelcome.vue
migueog: @jbibla I think I'll open a new issue for this since the scope of this issue has expanded.

test/unit/specs/components/common/ActionModal.spec.js
```diff
@@ -132,6 +133,21 @@ describe(`ActionModal`, () => {
     expect($store.dispatch).toHaveBeenCalledWith(`connectLedgerApp`)
   })
 
+  it(`should check for browser support if browser is Chrome`, () => {
```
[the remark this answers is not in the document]
migueog: @jbibla OK thanks for catching that. I've removed this test.

test/unit/specs/components/common/TmSessionWelcome.spec.js
```diff
@@ -64,6 +69,21 @@ describe(`TmSessionWelcome`, () => {
     it(`has the expected html structure`, () => {
       expect(wrapper.vm.$el).toMatchSnapshot()
     })
+
+    it(`should check for browser support if browser is Chrome`, () => {
```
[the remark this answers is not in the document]
migueog: Ditto for this one.

src/components/common/TmSessionWelcome.vue
migueog: Ah OK my bad. Misunderstood. 

Were you thinking something more like this -> ` `

src/components/common/TmSessionWelcome.vue
migueog: @colw Thanks for the help! Looking better for me now.

src/components/common/TmSessionWelcome.vue
```diff
@@ -14,13 +14,13 @@
         <LiSession
           v-if="!browserWithLedgerSupport"
           icon="usb"
-          title="Sign in with Ledger Nano nbsp; S"
+          title="Sign in with Ledger Nano&#160;S"
```
colw: Nice 👍

migueog:
@jbibla I think this PR is good to go!

jbibla:
awesome. i'll review tonight (and hopefully merge!)

faboweb:
Updated styling:
image
image

PR #2552, opened 2019-05-07
colw/Refactor Proposal tally display code

### LINKED ISSUE #2430: Proposal tally doesn't sum to 100%
**Describe the bug**

Screen Shot 2019-04-05 at 7 58 31 PM

colw:
Closes #2430 

**Description:**

Proposal integer tallies do not sum to 100%, so we need to mask this by either rounding up/down numbers, or displaying more decimal places so the original problem is less obvious.

The solution here will display percentages rounded to 2 decimals places:

Screenshot 2019-05-07 at 21 03 54

Code changes:
- Add convenience method for checking DepositPeriod status
- Add filter method for percentages
- Replace individual tally methods with filtering
- Replace if structure with switch
- Replace `percentInt` with `percent` (default 2 decimal places)

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/src/utils/index.js
+++ b/src/utils/index.js
@@ -1,0 +1,39 @@
+// This will take an object and for each (k,v) will return
+// v rounded such that the sum of all v is 100.
+export const roundObjectPercentages = o => {
+  // This algorithm workson integers and we want 2 decimal places.
+  // So round up first.
+  let asArray = Object.entries(o).map(v => {
+    v[1] = v[1] * 100
+    return v
+  })
+
+  // Calculate difference from 100%
+  var margin =
+    10000 -
+    asArray.reduce(function(acc, x) {
+      return acc + Math.round(x[1])
+    }, 0)
+
+  const cmp = (a, b) => {
+    return b[1] - Math.round(a[1]) - a[1]
+  }
+
+  // Sort and distribute difference amongst values
+  asArray.sort(cmp)
+  const result = asArray.map(function(x, i) {
+    const rounded = [
+      x[0],
+      (Math.round(x[1]) + (margin > i) - (i >= asArray.length + margin)) / 100.0
+    ]
+    return rounded
+  })
+
+  // Turn the array back into the original data structure layout
+  const resultObject = {}
+  result.forEach(x => {
+    resultObject[x[0]] = x[1]
+  })
+
+  return resultObject
+}
--- a/src/utils/index.js
+++ b/src/utils/index.js
@@ -1,11 +1,11 @@
 // This will take an object and for each (k,v) will return
 // v rounded such that the sum of all v is 100.
-export const roundObjectPercentages = o => {
+export const roundObjectPercentages = dataMap => {
   // This algorithm workson integers and we want 2 decimal places.
   // So round up first.
-  let asArray = Object.entries(o).map(v => {
-    v[1] = v[1] * 100
-    return v
+  let asArray = Object.entries(dataMap).map(keyValue => {
+    keyValue[1] = keyValue[1] * 100
+    return keyValue
   })
 
   // Calculate difference from 100%
--- a/src/utils/index.js
+++ b/src/utils/index.js
@@ -1,31 +1,40 @@
 // This will take an object and for each (k,v) will return
 // v rounded such that the sum of all v is 100.
+
+// Used the following as a reference:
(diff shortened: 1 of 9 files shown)


colw:
How does this look @jbibla ?

_Note that I still need to fix the case where it adds to 99.99%_

colw:
@jbibla Let me know if it's important the numbers add to exactly 100%, and I'll put it in.

faboweb:
I think it's ok. I was just thinking that the added decimals create more noise. But your call.

jbibla:
this is how we show decimals on the validator table, so for the sake of consistency i think 2 decimal places makes sense.

@colw how much work would it be to have them always add up to 100?

colw:
@jbibla I could do it today. Potential it can be reused in other parts, so there's value.

colw:
Sum to 1 or 2 decimal places:

Screenshot 2019-05-09 at 21 47 03

src/utils/index.js
```diff
@@ -0,0 +1,39 @@
+// This will take an object and for each (k,v) will return
+// v rounded such that the sum of all v is 100.
+export const roundObjectPercentages = dataMap => {
+  // This algorithm workson integers and we want 2 decimal places.
+  // So round up first.
+  let asArray = Object.entries(dataMap).map(keyValue => {
```
faboweb: Maybe something like this would be more readable?

```suggestion
  let asArray = Object.entries(dataMap).map(([key, value])=> {
```

src/utils/index.js
```diff
@@ -0,0 +1,39 @@
...
+  })
+
+  // Calculate difference from 100%
+  var margin =
+    10000 -
+    asArray.reduce(function(acc, x) {
+      return acc + Math.round(x[1])
+    }, 0)
+
+  const cmp = (a, b) => {
```
faboweb: what's compared here?

src/utils/index.js
```diff
@@ -0,0 +1,39 @@
...
+  // This algorithm workson integers and we want 2 decimal places.
+  // So round up first.
+  let asArray = Object.entries(dataMap).map(keyValue => {
+    keyValue[1] = keyValue[1] * 100
+    return keyValue
+  })
+
+  // Calculate difference from 100%
+  var margin =
+    10000 -
```
faboweb: why 10000?

src/utils/index.js
```diff
@@ -0,0 +1,39 @@
...
+  const cmp = (a, b) => {
+    return b[1] - Math.round(a[1]) - a[1]
+  }
+
+  // Sort and distribute difference amongst values
+  asArray.sort(cmp)
+  const result = asArray.map(function(x, i) {
+    const rounded = [
+      x[0],
+      (Math.round(x[1]) + (margin > i) - (i >= asArray.length + margin)) / 100.0
```
faboweb: I think this line would love a small comment

src/utils/index.js
```diff
@@ -0,0 +1,39 @@
...
+    asArray.reduce(function(acc, x) {
+      return acc + Math.round(x[1])
+    }, 0)
+
+  const cmp = (a, b) => {
+    return b[1] - Math.round(a[1]) - a[1]
+  }
+
+  // Sort and distribute difference amongst values
+  asArray.sort(cmp)
```
colw: // const test = 1

src/utils/index.js
colw:
```suggestion
      // A small comment
      (Math.round(x[1]) + (margin > i) - (i >= asArray.length + margin)) / 100.0
```

src/utils/index.js
colw: Good one!

src/utils/index.js
colw: True. I'll rename it

src/utils/index.js
colw: I'll make this clearer. Scaling for 100%, accounting for 2 extra decimal places.

src/utils/index.js
colw: It's good you say so actually. I did base this of an algorithm I found, and I thought I understood it completely until now.

colw: I've added a basic description and a link to a stack overflow article on the matter. I hope hat's sufficient. It can do a better job of describing what's happening, without cluttering up the code base with a larger explanation.

Issue #2560, opened 2019-05-09
add a careers page


jbibla:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

PR #2571, opened 2019-05-11
automatic release created for v1.0.0-beta.64

faboweb:
### Added

- [#2560](https://github.com/cosmos/lunie/issues/2560) Added careers page @jbibla

### Changed

- [#2568](https://github.com/cosmos/lunie/pull/2568) Show amount of source stake available to delegate or redelgate @colw

### Fixed

- [#2568](https://github.com/cosmos/lunie/pull/2568) Do not add stake to total when redelegating @colw
- [#2570](https://github.com/cosmos/lunie/pull/2570) Load validators when linking to a validator directly @faboweb
- [#1995](https://github.com/cosmos/lunie/issues/1995) Prevent users from signing in or creating a new account if their user agent is not Chrome @migueog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,22 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+## [1.0.0-beta.64] - 2019-05-11
+
+### Added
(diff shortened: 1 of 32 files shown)


faboweb:
Signing doesn't work:
image

PR #2585, opened 2019-05-13
Miguel/1735 implement multi step action modals

### LINKED ISSUE #1735: Implement multi step action modals
We get some bad feedback about the action modal. It seems to not be well designed for the expectation of customers. Let's come up with 2 proposals on how the action modal should look like and collect feedback from our colleagues.

migueog:
Closes #1735 

**Description:**

Creating a step component that is added to the Action modal to describe which step the user is currently on.

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/src/components/common/ActionModal.vue
+++ b/src/components/common/ActionModal.vue
@@ -27,9 +27,9 @@
           step-number="2"
         ></StepComponent>
         <StepComponent
-          step-title="Sign"
+          step-title="Submit"
           :step="step"
-          step-name="sign"
+          step-name="submit"
           step-number="3"
           :include-line="false"
         ></StepComponent>
--- a/src/components/common/ActionModal.vue
+++ b/src/components/common/ActionModal.vue
@@ -24,21 +24,21 @@
           :step="step"
           step-name="txDetails"
           step-number="1"
-          :includeLine=true
+          :include-line="true"
         ></StepComponent>
         <StepComponent
           step-title="Fees"
           :step="step"
           step-name="fees"
           step-number="2"
-          :includeLine=true
+          :include-line="true"
         ></StepComponent>
         <StepComponent
           step-title="Sign"
           :step="step"
           step-name="sign"
           step-number="3"
-          :includeLine=false
+          :include-line="false"
         ></StepComponent>
       </div>
       <div v-if="requiresSignIn" class="action-modal-form">
--- a/src/components/common/ActionModal.vue
+++ b/src/components/common/ActionModal.vue
@@ -17,21 +17,28 @@
           <i class="material-icons">close</i>
         </div>
       </div>
+      <!-- This step stuff should be refactored into something smaller -->
       <div class="step--container">
         <StepComponent
           step-title="Details"
           :step="step"
           step-name="txDetails"
+          step-number="1"
+          :includeLine=true
         ></StepComponent>
         <StepComponent
-          step-title="Details"
+          step-title="Fees"
           :step="step"
-          step-name="txDetails"
+          step-name="fees"
+          step-number="2"
+          :includeLine=true
         ></StepComponent>
         <StepComponent
-          step-title="Details"
+          step-title="Sign"
           :step="step"
-          step-name="txDetails"
+          step-name="sign"
+          step-number="3"
+          :includeLine=false
         ></StepComponent>
       </div>
       <div v-if="requiresSignIn" class="action-modal-form">
--- a/src/components/common/ActionModal.vue
+++ b/src/components/common/ActionModal.vue
@@ -482,7 +482,9 @@ export default {
   display: flex;
   flex-direction: row;
   text-align: center;
-  justify-content: center;
+  justify-content: space-between;
+  width: 13rem;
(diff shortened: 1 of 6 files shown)


src/components/common/ActionModal.vue
```diff
@@ -17,6 +13,30 @@
...
+        ></StepComponent>
+        <StepComponent
+          step-title="Fees"
+          :step="step"
+          step-name="fees"
+          step-number="2"
+          :include-line="true"
+        ></StepComponent>
+        <StepComponent
+          step-title="Sign"
```
jbibla: 3rd step should be `Submit` not `Sign`

src/components/common/ActionModal.vue
```diff
@@ -17,6 +13,30 @@
           <i class="material-icons">close</i>
         </div>
       </div>
+      <!-- This step stuff should be refactored into something smaller -->
```
jbibla: this is a good start but will restrict all action modals to the same 3 steps (which might be ok for now). it would be worth thinking about how to generalize this for any number of steps.

src/components/common/ActionModal.vue
```diff
@@ -17,6 +13,30 @@
...
+      <div class="step--container">
+        <StepComponent
+          step-title="Details"
+          :step="step"
+          step-name="txDetails"
+          step-number="1"
+          :include-line="true"
+        ></StepComponent>
+        <StepComponent
+          step-title="Fees"
```
jbibla: maybe 2nd step can be called `Review` instead of fees? We'll add the `TmAnyTx` component to the second step - and then the user will be reviewing their transaction and hitting next if they want to "Submit". 

What do you think @colw @faboweb?

src/components/common/StepComponent.vue
```diff
@@ -0,0 +1,90 @@
+<template>
+  <div class="stepItem">
+    <svg width="30" height="30">
```
jbibla: what was the reason you went with inline svg as opposed to css?

src/components/common/StepComponent.vue
```diff
@@ -0,0 +1,90 @@
...
+
+.text--default {
+  color: var(--grey);
+  font-size: var(--xs);
+  font-weight: 500;
+  line-height: normal;
+  letter-spacing: normal;
+  text-align: center;
+
+  /* This needs to be improved */
```
jbibla: yeah. the words aren't centered and will be different in size. i might consider grouping the number, circle, and step name in a div so they can be center aligned together.

src/components/common/StepComponent.vue
```diff
@@ -0,0 +1,90 @@
...
+  fill: rgba(252, 49, 167, 0.25);
+}
+
+svg {
+  display: inline-block;
+}
+
+.text--default {
+  color: var(--grey);
+  font-size: var(--xs);
```
jbibla: step numbers should be this size too

src/components/common/StepComponent.vue
migueog: I think it was a bit of a brain fart when I started this issue. It's way simpler if make the circles using CSS. Thanks for the feedback!

src/components/common/ActionModal.vue
migueog: Yeah I was thinking about how to create a `stepGroup` component that takes in an array of strings. I guess that could work but it feels like passing an array of strings is pretty clunky.

src/components/common/StepComponent.vue
migueog: Yeah that's a better way of doing that.

src/components/common/ActionModal.vue
faboweb: How would this be called in classic e-commerce? "Overview", "Checkout"?

jbibla: I think the whole flow would be called a checkout flow. I think `Overview` could be good. Review is also okwme because it is the action the user is supposed to take.

I feel pretty good about `Details` `Review` `Submit` but certainly open to better options!

colw: Summary also works.

colw: `Details / Review / Submit`, but perhaps the `Details` could be set by the caller modal.

faboweb: Hahahaha I think we should lean into the "shopping" domain. People understand this.

faboweb: How about doing 4 steps? <Action>/Review/Sign/Submit ?

jbibla: how / why would sign and submit be separate?

faboweb: sign is the step where the user signs the transactions / picks a method and waits for the signature to be available
submit is the step where the user waits for the tx so be submitted and finally included in a block

jbibla: ok let's explore this more but not in this PR

faboweb:
I have the feeling the steps-container and the steps should be combined and take an array of steps plus the current step number. But probably fine for now.

PR #2590, opened 2019-05-15
colw/2378 disable withdrawl on zero delegations

### LINKED ISSUE #2378: Disable withdrawal when user doesn't have any delegations
**Describe the bug**

Attempting to do a withdrawal without delegations renders the following error on the modal:

`Withdrawal failed: JSON Missing msgs.`

**To Reproduce**

1. Sign in with an account with no delegations
2. Click withdraw and try to submit the transaction

**Expected behavior**

Show a message that user needs to delegate to be able to withdraw.

colw:
Closes #2378

**Description:**

Disable option to withdraw when the user has no rewards.

withdraw-zero

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/src/components/common/TmBalance.vue
+++ b/src/components/common/TmBalance.vue
@@ -16,6 +16,7 @@
         <h3>Rewards</h3>
         <h2>{{ rewards }}</h2>
         <TmBtn
+          v-show="rewards > 0"
           id="withdraw-btn"
           class="withdraw-rewards"
           :value="connected ? 'Withdraw' : 'Connecting...'"
--- a/changes/colw_2378-disable-withdrawl-on-zero-delegations
+++ b/changes/colw_2378-disable-withdrawl-on-zero-delegations
@@ -1,0 +1,1 @@
+[Fixed] [#2378](https://github.com/cosmos/lunie/issues/2378) Do not show withdrawal button when user has no rewards avilable @colw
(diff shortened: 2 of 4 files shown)


src/components/common/TmBalance.vue
```diff
@@ -16,6 +16,7 @@
         <h3>Rewards</h3>
         <h2>{{ rewards }}</h2>
         <TmBtn
+          v-show="rewards > 0"
```
faboweb: Why not disabling it? Hiding it does also hide the functionality.

colw: Good question

It's a link, not a button, and perhaps 'dead text' could be confusing? I could do that and add a tool tip perhaps? Explaining why it's not 

Then again hiding it initially is maybe not a bad thing? As a user gets rewards it becomes intuitive why it's not there at the beginning (or again not present in the future)

I'll experiment, and post a gif.

colw: @jbibla Do you have thoughts on this? I'm hesitant to 'disable' an anchor text. Not quite sure what it should look like. A more dull colour with no cursor pointer?

faboweb: Ah! Sorry I was mistaken. I thought this is the button on the Actionmodal. I think this is fine.

jbibla:
i think a nice improvement on this would be letting users click withdraw, but displaying a message on the withdraw modal that says "make a delegation to start earning rewards!" or something like that.

PR #2646, opened 2019-05-21
Jordan/add ga vue router

jbibla:
Closes #ISSUE

**Description:**

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/src/scripts/boot.js
+++ b/src/scripts/boot.js
@@ -100,6 +100,7 @@ export const startApp = async (
   setGoogleAnalyticsPage(router.currentRoute.path)
   router.beforeEach(routeGuard(store))
   router.afterEach(to => {
+    /* istanbul ignore next */
     setGoogleAnalyticsPage(to.path)
   })
 
--- a/src/scripts/boot.js
+++ b/src/scripts/boot.js
@@ -17,7 +17,10 @@ import _Node from "../connectors/node"
 import _Store from "../vuex/store"
 import * as urlHelpers from "scripts/url.js"
 import _config from "src/config"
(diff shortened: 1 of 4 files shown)


src/scripts/google-analytics.js
```diff
@@ -37,3 +37,9 @@ module.exports.track = function track(...args) {
     window.ga(`send`, ...args)
   }
 }
+
+module.exports.setGoogleAnalyticsPage = function track(...args) {
+  if (window.ga) {
+    window.ga(`set`, `page`, ...args)
```
faboweb: will this `send` the event or just set the current page?

src/scripts/boot.js
```diff
@@ -92,7 +97,11 @@ export const startApp = async (
     routes
   })
 
+  setGoogleAnalyticsPage(router.currentRoute.path)
   router.beforeEach(routeGuard(store))
+  router.afterEach(to => {
+    setGoogleAnalyticsPage(to.path)
```
faboweb: if you don't want to test this, please add a `/* istanbul ignore next: <REASON> */`

jbibla: it would be good to test it - but i couldn't figure out how to mock the router in in app.spec - odd setup there. do you know how?

src/scripts/google-analytics.js
jbibla: good call. upon further investigation, `send` is required.

src/scripts/boot.js
faboweb: there are actually tests for the routeGuard in `App.spec` which could be used as a reference

PR #2647, opened 2019-05-21
colw/2637 withdraw from 1-5 validators at a time

### LINKED ISSUE #2637: Ledger Nano S is not able to process big transactions (i.e. many withdrawels)
**Describe the bug**

Ledger Nano S is not able to process big transactions. When I delegate to a bunch of validators and try to withdraw from all of them at once (default) the Ledger reports an error, that it received too many "tokens" which stands for too big of a transaction to fit into the memory of the Ledger.

**To Reproduce**

Delegate to many validators.
Withdraw

**Expected behavior**

I am able to withdraw rewards. An easy fix is to add a withdraw button to each validators page. We probably need to come up with a better solution later.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: 
 - Browser: 
 - Version: 

**Smartphone (please complete the following information):**
 - Device: 
 - OS: 
 - Browser 
 - Version 

**Additional context**

colw:
Closes #2637 

**Description:**

Withdraw in one of two ways:

- Global withdraw: Will withdraw from the 5 validators with the highest rewards.
- Validator page withdraw: Withdraw from single validator only

Screenshot 2019-05-23 at 11 50 15

If there are no rewards from a given validator, the button is disabled:
Screenshot 2019-05-23 at 11 51 46

The modal supports both variations and displays a message when withdrawing globally.
Screenshot 2019-05-23 at 11 49 17

Message: `Note: Lunie withdraws only the top 5 rewards at a time.`

Thank you! 🚀

---

For contributor:

For reviewer:


src/components/staking/ModalWithdrawRewards.vue
```diff
@@ -0,0 +1,82 @@
...
+    atoms,
+    viewDenom,
+    fullDecimals
+  },
+  props: {
+    validatorAddress: {
+      type: String,
+      required: true
+    },
+    rewards: {
```
faboweb: Thinking about the ActionModal webcomponent, does it make sense to make the rewards an input? Or would we wrap this in a controller?

colw: Yep, I thought about this. I believe wrapping it in a controller is better.

IMHO, the Modals should be 'dumb' views, and at most receive data and forward user actions, but that's it. Separation of concerns.

colw: (Note: This should apply to most views)

src/components/staking/ModalWithdrawRewards.vue
```diff
@@ -0,0 +1,82 @@
...
+  >
+    <TmFormGroup
+      class="action-modal-form-group"
+      field-id="amount"
+      field-label="Amount"
+    >
+      <span class="input-suffix">{{ denom | viewDenom }}</span>
+      <TmField id="amount" :value="rewards | atoms | fullDecimals" readonly />
+    </TmFormGroup>
+    <span v-if="!validatorAddress" class="form-message">
```
faboweb: should the message show always or only when the user actually has more then 5 delegations?

src/components/staking/ModalWithdrawRewards.vue
```diff
@@ -0,0 +1,82 @@
...
+      Note: Lunie withdraws only the top 5 rewards at a time.
```
faboweb:
```suggestion
      Note: Lunie withdraws only the top 5 rewards at a time. The Ledger Nano S can't process bigger transactions.
```

src/utils/index.js
```diff
@@ -47,3 +47,41 @@ export const roundObjectPercentages = dataMap => {
 
   return resultObject
 }
+
+const cmpSecondElementDesc = (a, b) => {
+  return b[1] - a[1]
+}
+
+// Takes an object:
```
faboweb: 👍  Nice comment!

src/vuex/modules/distribution.js
```diff
@@ -95,21 +97,28 @@ export default ({ node }) => {
...
-      {
-        rootState: { session },
-        getters: { committedDelegations },
-        dispatch
-      },
-      { gas, gas_prices, password, submitType }
+      { rootState, getters, dispatch },
+      { gas, gasPrice, denom, validatorAddress, password, submitType }
     ) {
+      const top5Delegations = getTop5Delegations(getters.committedDelegations)
```
faboweb: should only limit to 5 delegations if the user is using the Ledger Nano S, but we can figure this out once we have more signers

src/components/staking/ModalWithdrawRewards.vue
colw: 🤦‍♂ You're right!

src/components/staking/ModalWithdrawRewards.vue
colw:
```suggestion
      Note: Lunie withdraws only the top 5 rewards at a time, due to a limitation in the Ledger Nano S.
```

jbibla: thanks colw - i prefer that language!

src/components/staking/ModalWithdrawRewards.vue
colw: Actually, is it necessary? We could leave it there to inform the user regardless?

src/utils/index.js
colw: Fair call. Improved it 👍

src/components/staking/ModalWithdrawRewards.vue
faboweb: I personally like to reduce noise that is not important to users. @jbibla preferences?

src/vuex/modules/distribution.js
colw: Indeed. There's some time to spare before the Nano X is in wide distribution, and we'll need to detect the difference. Is that right?

src/components/staking/ModalWithdrawRewards.vue
colw: * It would have a minor impact on the message wording. The current phrasing is generic.

src/vuex/modules/distribution.js
faboweb: yep

PR #2652, opened 2019-05-22
Fabo/check out of gas after broadcasting

faboweb:
This checks if a tx fails after is has been broadcasted like this tx https://stargate.lunie.io/txs/10ec0001a581f7666a6e2efae3ae3e4b0e964c3c189ae52a7f12210eb760529f

See change here: https://github.com/luniehq/cosmos-js/blob/master/src/send.js#L58

**Description:**

Thank you! 🚀

---

For contributor:

For reviewer:


src/components/common/ActionModal.vue
```diff
@@ -365,10 +389,6 @@ export default {
         this.step = feeStep
       } catch ({ message }) {
         this.submissionError = `${this.submissionErrorPrefix}: ${message}.`
-
-        setTimeout(() => {
```
faboweb: Is that ok @jbibla? The hides pretty quick. I don't think hiding it is actually necessary

Issue #2231, opened 2019-03-11
Add e2e tests for all actions


faboweb:
**Is your feature request related to a problem? Please describe.**

As a user
I want the developers to test the main actions in e2e tests to prevent them from producing breaking bugs
So that I have a working UI at all time.

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

AC:
@faboweb 
- [x] have e2e test for sending coins
- [x]  have e2e test for delegating
- [x]  have e2e test for undelegating
- [x]  have e2e test for redelegating
@colw
- [ ]  have e2e test for creating a proposal
- [ ]  have e2e test for depositing on a proposal
- [ ]  have e2e test for voting on a proposal
- [ ] have e2e tests for withdrawing rewards from single and multiple validators

PR #2665, opened 2019-05-28
Fabo/2231 delegation e2e

faboweb:
Part of #2231 

Note: The update in the balance header has currently a not predictable behavior which we need to fix.

**Description:**

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/src/vuex/modules/wallet.js
+++ b/src/vuex/modules/wallet.js
@@ -113,7 +113,6 @@ export default ({ node }) => {
 
 function subscribeToTxs(rpcClient, address, dispatch) {
   function onTx(data) {
-    console.log(data)
     dispatch(`queryWalletStateAfterHeight`, data.TxResult.height + 1)
   }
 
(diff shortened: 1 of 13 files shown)


src/vuex/modules/wallet.js
```diff
@@ -113,6 +113,7 @@ export default ({ node }) => {
 
 function subscribeToTxs(rpcClient, address, dispatch) {
   function onTx(data) {
+    console.log(data)
```
jbibla: 😱

faboweb: Upsi

faboweb:
Could happen if your testnet is already running for a while. Should not happen if it's freshly started like on the CI as there are no rewards piled up.

colw:
It was a fresh build running for a minute or so. We'll tweak these things out as we go.

faboweb:
I could set one validator to charge 100% commission so the rewards would be 0. Would that be a better solution?

colw:
That could work. Different validators for different test purposes.

PR #2667, opened 2019-05-28
Jordan/misc style improvements

jbibla:
**Description:**

- added "Your Public Address" to wallet page (user feedback)
- added a min-height to action modal so it doesn't jump in size too much
- widened action-modal steps and improved spacing
- step count starts from 1, not 0
- removed subtotal if 0 because this confusing (user feedback)
- removed hover colour from unclickable things (@colw feedback 💪)

Thank you! 🚀

---

For contributor:

For reviewer:


jbibla:
SCREENSHOTS COMING SOON :D

jbibla:
Screen Shot 2019-05-30 at 3 48 32 PM

colw:
@faboweb Happy with the bech32 adjustment?

colw:
@jbibla I removed the tool tip for the long form address. What do you think?

test/unit/specs/components/common/ShortBech32.spec.js
```diff
@@ -38,6 +38,26 @@ describe(`ShortBech32`, () => {
...
+    wrapper.setProps({ address: `cosmosaddress1asdfasdfasdf`, longForm: true })
+    expect(wrapper.find(".address").text()).toBe(`cosmosaddress1asdfasdfasdf`)
+  })
+
+  it(`should calculate a long address`, () => {
+    wrapper.setProps({ address: `cosmosaddress1asdfasdfasdf`, longForm: true })
+    expect(wrapper.vm.bech32).toBe(`cosmosaddress1asdfasdfasdf`)
+  })
+
+  it(`should return 'address not found when invalid in long form'`, () => {
```
colw:
```suggestion
  it(`should return 'address not found' when invalid in long form`, () => {
```

src/components/common/ShortBech32.vue
```diff
@@ -9,6 +10,15 @@
     >
       {{ shortBech32 }}
```
faboweb: suggestion: switch longForm only here instead of copying the whole element. this would reduce code

src/components/common/ShortBech32.vue
```diff
@@ -1,6 +1,7 @@
 <template>
   <div class="short-bech32">
```
faboweb: suggestion: rename to "Address"

jbibla: there is bech32 specific logic here though - so maybe `bech32address` or something?

not sure if this is necessary for this PR

src/components/common/ShortBech32.vue
jbibla: i don't know what you mean.

faboweb: You are copying the whole element
```
    <div
      v-else
      id="address"
      v-clipboard:copy="address"
      v-clipboard:success="() => onCopy()"
      class="address"
    >
      {{ bech32 }}
    </div>
```

This is producing code repetition which should be avoided. Better just switch on the things that actually change, which is only the actual value shown.

src/components/common/ShortBech32.vue
faboweb: I meant renaming the whole component as it is not only "Short" anymore.

jbibla: yep, agreed. but i think it should remain more specific than `address` because of the bech32 logic

src/components/common/ShortBech32.vue
jbibla: ah ok! nice suggestion. 

CC @colw

colw: @faboweb Ah wait, the reason I copy the whole element is to remove the tool tip on the long form address (it's the same content). It might be a necessary evil.

@jbibla If the tool tip and the copy to clipboard are considered 'tied' together, I could just leave the tool tip in?

src/components/common/ShortBech32.vue
colw: Both fair points. Keeping it as Bech32 works when looking at the templates code. It reads like:

```
<Bech32 :address="value" />
```


Though perhaps

`<Address :bech32="…" />` or `<Address :value="…" />` is better?

jbibla: yeah good illustration. whatever you think is best.

src/components/common/ShortBech32.vue
jbibla: ah i see. maybe if we leave a tooltip for the long one it should say "Click to Copy". what do you think about that?

@colw i trust you to make the right call - i think it's good either way.

colw:
I don't think that Code Climate error should matter so much. Similar code (perhaps they should be merged at some point), but part of larger functions with different intentions..

jbibla:
going to test the ledger withdrawal today before we merge this.

PR #2679, opened 2019-06-05
Fabo/cosmos keys

faboweb:
This extracts our signer script into an own repository so we can isolate security vulnerable code


colw:
This could use a small description 👍 

Also, I had a quick test, and it did not succeed.

faboweb:
:( the error is bad. I try to look into it later if noone else wants to pick it up

faboweb:
@colw fixed! was just a typo

PR #2682, opened 2019-06-05
Benji/2234 Allows enter key to submit ActionModal when input field is highlighted

### LINKED ISSUE #2234: Pressing enter on action modals should submit forms
**Describe the bug**

Pressing enter on action modals should submit the form. Currently, it does not.

**To Reproduce**

- go to wallet page
- open send modal
- fill in empty form fields
- press `Enter`
- form is not submitted

**Expected behavior**

I would expect the form to be submitted.

**Additional context**

I believe this would be more semantic and accessible HTML.

thebkr7:
Closes #2234 

**Description:**

Adds functionality so that when the password input field is selected the enter button can be used to send the transaction rather than having to click Submit

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/src/components/common/ActionModal.vue
+++ b/src/components/common/ActionModal.vue
@@ -112,6 +112,7 @@
             v-model="password"
             type="password"
             placeholder="Password"
+            @keyup.enter.native="validateChangeStep"
           />
           <TmFormMsg
             v-if="$v.password.$error && !$v.password.required"


jbibla:
this looks like a good fix but i think that we'd prefer to have a closer look at the forms to see why they aren't already behaving this way. html forms are supposed to work this way by default (implicit submission).

very likely, our forms are not conforming to the rules of semantic HTML. can you see if this is the case?

https://www.tjvantoll.com/2013/01/01/enter-should-submit-forms-stop-messing-with-that/

faboweb:
This is also not working on all the steps. I.e.:
Try to delegate
Add amount
Hit enter

colw:
@jbibla There are no form elements in the ActionModal, only inputs.

jbibla:
> There are no form elements in the ActionModal, only inputs.

@thebkr7 adding form elements is the way to resolve this issue.

colw:
> > There are no form elements in the ActionModal, only inputs.
> 
> @thebkr7 adding form elements is the way to resolve this issue.

Yep! But also, we need to add the type=submit to the next button(s). And make sure an item is focused in the modal.

faboweb:
@thebkr7 and @colw feel free to hack this out together today/tomorrow. Could be a good first bonding. :)

colw:
@thebkr7 I scaled back the refactor so you should be able to tackle this easier, See #2692

thebkr7:
Will begin from the start with a new PR due to taking a different approach as discussed above.

PR #2690, opened 2019-06-06
Benji/2165 show end voting period

### LINKED ISSUE #2165: Show end of voting period

A proposal passes after a certain period. Currently this is not shown.

Show the remaining voting period.

**Additional context**

thebkr7:
Closes #2165 

Adds a div showing when the voting for a proposal in Governance ends and re-organizes time-specific elements.

Fixing that the end time for proposals was not showing.

Before:
image

After:
image

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/src/components/governance/PageProposal.vue
+++ b/src/components/governance/PageProposal.vue
@@ -204,18 +204,20 @@ export default {
         return false
       }
     },
-    votingStartedAgo({ proposal} = this) {
+    votingStartedAgo({ proposal } = this) {
       return moment(new Date(proposal.voting_start_time)).fromNow()
     },
     proposalStatus({ proposal, depositEndsIn, votingStartedAgo } = this) {
-      if (proposal.proposal_status === 'DepositPeriod') {
+      if (proposal.proposal_status === "DepositPeriod") {
         return `Deposit period ends ${depositEndsIn}`
-      } else if (proposal.proposal_status === 'VotingPeriod') {
+      } else if (proposal.proposal_status === "VotingPeriod") {
         return `Voting started ${votingStartedAgo}`
-      } else if (proposal.proposal_status === 'Rejected') {
-        return 'Rejected'
-      } else if (proposal.proposal_status === 'Passed') {
-        return 'Passed'
+      } else if (proposal.proposal_status === "Rejected") {
+        return "Rejected"
+      } else if (proposal.proposal_status === "Passed") {
+        return "Passed"
+      } else {
+        return false
       }
     },
     depositEndsIn({ proposal } = this) {
--- a/src/components/governance/PageProposal.vue
+++ b/src/components/governance/PageProposal.vue
@@ -58,17 +58,8 @@
 
           <dl class="info_dl colored_dl">
             <dt>Proposal Status</dt>
-            <dd v-if="proposal.proposal_status === 'DepositPeriod'">
-              {{ `Deposit period ends ${depositEndsIn}.` }}
-            </dd>
-            <dd v-if="proposal.proposal_status === 'VotingPeriod'">
-              {{ `Voting started ${votingStartedAgo}.` }}
-            </dd>
-            <dd v-if="proposal.proposal_status === 'Rejected'">
-              Rejected
-            </dd>
-            <dd v-if="proposal.proposal_status === 'Passed'">
-              Passed
+            <dd>
+              {{ proposalStatus }}
             </dd>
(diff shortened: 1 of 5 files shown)


faboweb:
Missing changelog. Please run `yarn changelog`.

src/components/governance/PageProposal.vue
```diff
@@ -66,6 +67,12 @@
             </dd>
           </dl>
 
+          <dl class="info_dl colored_dl">
+            <dt>End Date</dt>
```
jbibla: let's make sure we specify if it's a deposit end date or voting end date

jbibla:
looks good. what will be displayed if during the deposit period?

thebkr7:
Good point. I did not understand the end date isn't always set once proposed but rather only when the required atom is deposited. I will adjust

jbibla:
> I did not understand the end date isn't always set once proposed but rather only when the required atom is deposited

no worries. there is probably an end deposit date (or we can calculate it based on the deposit period param). we might want to show _a_ date but have it be a bit smart about what state the proposal is in.

thebkr7:
How about only conditionally displaying the end date cell based on if it is passed the DepositPeriod? Proposal Status already shows when the deposit end date is if the proposal is in the DepositPeriod.

thebkr7:
The last commit should have been a separate PR. It shows Rejected and Passed for when votes are complete instead of always saying when voting started a month ago.

faboweb:
> How about only conditionally displaying the end date cell based on if it is passed the DepositPeriod? Proposal Status already shows when the deposit end date is if the proposal is in the DepositPeriod.

How about fusing everything into one field:
- when depositing: show deposit end date
- when depositing failed: show failed
- when voting: show voting end date
- when voting over: show success/failure

thebkr7:
This is how we have it set up with the exception of during the voting period we show two fields: 1. Voting End Date, 2. Voting Start Date.

faboweb:
> This is how we have it set up with the exception of during the voting period we show two fields: 1. Voting End Date, 2. Voting Start Date.

does knowing the start date provide additional value for the user?

thebkr7:
While I agree that it isn't critical information I do imagine it could provide some value. Example: They would like to see how long this proposal has been getting voted on without them. 1hr vs 1 day could give a lot more information as to why there are so few votes? But you know what the end user would like best. Should I remove it?

faboweb:
> But you know what the end user would like best

Hahahaha I don't! But one of our design principles is to remove not necessary information to reduce noise.

But let's leave this as is. We also aim to not cramp too many changes into one PR.

src/components/governance/PageProposal.vue
```diff
@@ -55,17 +55,31 @@
...
           <dl class="info_dl colored_dl">
             <dt>Proposal Status</dt>
-            <dd>
-              {{
-                proposal.proposal_status === `DepositPeriod`
-                  ? `Deposit period ends ${depositEndsIn}.`
-                  : `Voting started ${votingStartedAgo}.`
-              }}
+            <dd v-if="proposal.proposal_status === 'DepositPeriod'">
+              {{ `Deposit period ends ${depositEndsIn}.` }}
```
faboweb: The text content should probably be a computed value instead of copying the element n times.

src/components/governance/PageProposal.vue
```diff
@@ -189,6 +200,19 @@ export default {
     submittedAgo({ proposal } = this) {
       return moment(new Date(proposal.submit_time)).fromNow()
     },
+    endDate({ proposal } = this) {
+      return moment(proposal.voting_end_time).format("MMMM Do YYYY, HH:mm")
+    },
+    displayEndDate({ proposal } = this) {
+      if (
+        proposal.proposal_status !== "DepositPeriod" &&
+        proposal.total_deposit[0].amount >= 521000000
```
faboweb: We should use the proper parameter from the `governanceParameters` module instead of hardcoding this value here.

thebkr7: Could you please point me in the right direction to find the dynamic value for the minimum deposit amount before a proposal enters the voting period? I explored the getters but was unsuccessful.

thebkr7: I do see the correct Getter. I initially ruled it out due to it giving the wrong value but it is the test value for min. deposits in development.

faboweb:
Nice!

PR #2700, opened 2019-06-11
Benji/2699 validator status bug

### LINKED ISSUE #2699: Validator status icon is only accurate when viewing all validators and inaccurate for single validators.
**Describe the bug**
Validator status icon is only accurate when viewing all validators and inaccurate for single validators.

Accurate icon/popup:
image

Then when you click into the validator no matter what it shows it as "Actively Validating" which is not the case:
image

**To Reproduce**
View all validators
Find one that doesn't have a green status
Click into it
Now you will see in that validator it shows it as if it has a green status and is running as normal

thebkr7:
Closes #2699 

**Description:**

Updated PageValidator and it's corresponding tests to follow the same pattern as LiValidator. Now validators accurately reflect their status when expanded and not only in the list view.

@faboweb I proceeded with updating this code as if "revoked" and "voting_power"  had been used accidentally instead of "jailed" and "status". If this assumption was incorrect please point me in the right direction.

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/changes/Benji_2699-Validator-Status-Bug
+++ b/changes/Benji_2699-Validator-Status-Bug
@@ -1,0 +1,1 @@
+[Fixed] [#2699](https://github.com/cosmos/lunie/issues/2699) Fixed the bug that was always displaying validators as active when expanded even if they were not @thebkr7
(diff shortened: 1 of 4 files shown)


thebkr7:
I've updated the test but am very open to suggestions on how to word it better.

PR #2761, opened 2019-06-24
Fabo/remove old tests

### LINKED ISSUE #1742: Deprecate lcdClientMock

faboweb:
Closes #1742

**Description:**

Thank you! 🚀

---

For contributor:

For reviewer:


jbibla:
nice fabo! thanks for doing this. do you think we should be getting away from `mockValues` altogether? or this pattern is ok?

faboweb:
> nice fabo! thanks for doing this. do you think we should be getting away from mockValues altogether? or this pattern is ok?

hmm I think having a central place to store validated data structures which match the APIs we use could still prevent us from diverging from the API by accident.
E2E test should help on top.

PR #2763, opened 2019-06-24
Fabo/added extension signer method

faboweb:
image

Any preferences on what to show?

### DIFF
--- a/src/ActionModal/components/ActionModal.vue
+++ b/src/ActionModal/components/ActionModal.vue
@@ -1,6 +1,14 @@
 <template>
-  <transition v-if="show" name="slide-fade">
-    <div v-focus-last class="action-modal" tabindex="0" @keyup.esc="close">
+  <transition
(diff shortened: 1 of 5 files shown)


src/ActionModal/components/ActionModal.vue
```diff
@@ -300,12 +330,32 @@ export default {
...
+      }
       return [
         {
           key: `(Unsafe) Local Account`,
           value: signWithLocalKeystore
         }
       ]
+    },
+    submitButtonCaption() {
+      return (() => {
```
colw: I need to know your thinking behind this `function x{return () => {}()}` call?

:)

faboweb: brainfreez ;D need to remove it

jbibla:
> Any preferences on what to show?

i think a graphic would be great here. i can mock up a browser bar with the chrome extension icon in it?

PR #2784, opened 2019-06-27
Benji/extension SessionAccounts displays real accounts and has Add Account button

thebkr7:
Extension SessionAccounts displays real accounts and has Add Account button.

image

Thank you! 🚀

---

For contributor:

For reviewer:


thebkr7:
Sign Out button needs css work

src/components/common/SessionAccounts.vue
```diff
@@ -14,40 +14,45 @@
...
-    setState(value) {
-      this.$emit(`route-change`, value)
-    },
-    close() {
-      this.$emit(`close`)
+  computed: {
+    accounts() {
+      let accounts = this.$store.state.accounts
+      return accounts.map(({ name, address }) => ({
+        value: address,
```
jbibla: any reason we're changing this from `address` to value: address`?

thebkr7:
No Sign Out button

src/components/common/SessionAccounts.vue
thebkr7: Just keeping the code uniform as it was in TmSessionSignIn https://github.com/luniehq/lunie/blob/26a9e93b3012ce1c4cb94d24585a5787049cb9e9/src/components/common/TmSessionSignIn.vue#L88

thebkr7:
I took the functionality of the "Add Account" button away as this will use Jordan's new routes in https://github.com/luniehq/lunie/pull/2798

src/components/common/SessionApprove.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <div class="approve-tran" hide-header>
+  <div class="approve-tran">
```
jbibla:
```suggestion
  <div class="session-approve">
```

src/components/common/SessionAccounts.vue
```diff
@@ -1,6 +1,5 @@
 <template>
-  <div class="accounts" :hide-header="true">
-    <ToolBar :display-text="true" />
+  <div class="accounts">
```
jbibla:
```suggestion
  <div class="session-accounts">
```

jbibla: just a suggestion to keep with conventions

src/components/common/SessionApprove.vue
jbibla: to keep with convention - but also avoid the shortform `tran`

test/unit/specs/components/common/SessionAccounts.spec.js
```diff
@@ -6,32 +6,37 @@ describe(`SessionAccounts`, () => {
...
   })
 
-  it(`should close`, () => {
-    const self = {
-      $emit: jest.fn()
-    }
-    SessionAccounts.methods.close.call(self)
-    expect(self.$emit).toHaveBeenCalledWith(`close`)
+  it(`renders the correct account`, () => {
+    expect(wrapper.html()).toContain("<h3>accountName</h3>")
```
faboweb:
```suggestion
    expect(wrapper.html()).toContain("accountName")
```

src/components/common/SessionAccounts.vue
```diff
@@ -17,37 +16,30 @@
           <h3>{{ account.name }}</h3>
           <Bech32 :address="account.address" short-form />
         </div>
-        <TmBtn value="Go to Lunie" color="primary" />
       </div>
     </div>
+    <div class="button-add-account">
```
faboweb: This has no functionality yet, is that intended?

test/unit/specs/components/common/SessionAccounts.spec.js
```diff
@@ -6,32 +6,37 @@ describe(`SessionAccounts`, () => {
...
 
-  let wrapper
+  let wrapper, $store
 
   beforeEach(() => {
+    $store = {
+      state: {
+        accounts: [
+          {
+            name: "accountName",
```
colw: I would even change this to "Benji", or "Foobar's Account Name". Currently it looks like a variable, and this makes the test below harder to understand what is expected.

test/unit/specs/components/common/SessionAccounts.spec.js
colw: @thebkr7 Also, see above.

test/unit/specs/components/common/SessionAccounts.spec.js
```diff
@@ -6,32 +6,37 @@ describe(`SessionAccounts`, () => {
...
   })
 
-  it("moves to other session pages", () => {
-    const self = {
-      $emit: jest.fn()
-    }
-    SessionAccounts.methods.setState.call(self, "welcome")
-    expect(self.$emit).toHaveBeenCalledWith("route-change", "welcome")
+  it(`should show an account`, () => {
+    expect(wrapper.find(`.card-content`)).toBeDefined()
```
colw: This class doesn't exist I believe. Should it be `card.content`? In fact you can put anything in this `find()` and the test still passes.

What you need is:

`expect(wrapper.find(".card.content").exists()).toBe(true)`

p.s. In fact, I don't think you need this test, since the above `accountName` check should be sufficient to prove an account was listed.

src/components/common/SessionAccounts.vue
thebkr7: It was intentionally left without functionality because it will use the functionality introduced in Jordans PR2798. I have commented it out for now.

PR #2785, opened 2019-06-27
Refactors so TableInvoice takes the total fee

thebkr7:
Closes #28 for the Extension Repo

**Description:**

Refactors so TableInvoice takes the total fee rather than taking two numbers as props and calculating it there.

Thank you! 🚀

---

For contributor:

- [x] Added changes entries. Run `yarn changelog` for a guided process.
- [x] Reviewed `Files changed` in the github PR explorer
- [ ] Attach screenshots of the UI components on the PR description (if applicable)
- [ ] Scope of work approved for big PRs

For reviewer:

- [ ] Manually tested the changes on the UI

PR #2787, opened 2019-06-28
Benji/ext issue 29 calculate subtotal in SessionApprove TableInvoice

### LINKED ISSUE #29: Network Monitoring

thebkr7:
Closes #29 Extension

**Description:**

Refactors SessionApprove to pass TableInvoice the subtotal fee and amount of the transaction. TableInvoice was refactored in PR #2785

Thank you! 🚀

---

For contributor:

For reviewer:


src/components/common/SessionApprove.vue
```diff
@@ -114,6 +112,10 @@ export default {
     },
     senderAddress() {
       return this.signRequest ? this.signRequest.senderAddress : null
+    },
+    amount() {
+      //Unable to test wit real transaction, may be: this.tx.tx.value.msg.value
```
thebkr7: Can I get confirmation on this?

thebkr7: I can test this now that I can send transactions to the extension

src/components/common/SessionApprove.vue
```diff
@@ -89,6 +87,11 @@ function parseTx(signMessage) {
   }
 }
 
+function parseFee(message) {
```
thebkr7: Is there a specific script file that parseTx and parseFee could be kept in? @faboweb

faboweb: Not yet, feel free to create one. Something like 'scripts/parsers.je:?

thebkr7:
@colw Any chance you could look into the failing tests. I imagine it is something simple that I am missing.

faboweb:
fixed the JSON.parse issue. but there are more

test/unit/specs/components/common/SessionApprove.spec.js
```diff
@@ -5,16 +5,12 @@ import SessionApprove from "common/SessionApprove"
...
 
   let wrapper, $store
 
   beforeEach(() => {
     const getters = {
-      signRequest: {
-        senderAddress: "cosmos1234",
-        signMessage: "{}"
-      }
+      signRequest: `{signMessage:{"account_number":"1","chain_id":"testnet","fee":{"amount":[{"amount":"40","denom":"stake"}],"gas":"39953"},"memo":"(Sent via Lunie)","msgs":[{"type":"cosmos-sdk/MsgSend","value":{"amount":[{"amount":"12000000","denom":"stake"}],"from_address":"cosmos1ek9cd8ewgxg9w5xllq9um0uf4aaxaruvcw4v9e","to_address":"cosmos1324vt5j3wzx0xsc32mjhkrvy5gn5ef2hrwcg29"}}],"sequence":"0"}}`
```
colw: This is why your tests are failing. Your computed properties are trying to parse `this.signRequest.signMessage`, which is incorrect. At this point `this.signRequest` is a string, and there is no `signMessage` property.

(The giveaway is in the error here: `SyntaxError: Unexpected token u in JSON at position 0`. The `u` it refers to is `'undefined'`)

My suggestion is to have one computed property which runs a `JSON.parse()` on the request, and then have other computed properties use this result.

I also couldn't find a getter for `signRequest`, but perhaps have that parse the JSON for you?

src/scripts/parsers.js
```diff
@@ -0,0 +1,30 @@
+"use strict"
+
+export const parseTx = signMessage => {
+  const { msgs, fee, memo } = JSON.parse(signMessage)
```
colw: As per my other comment, I would avoid `JSON.parse()` being called 3 times.

src/components/common/SessionApprove.vue
```diff
@@ -111,11 +97,21 @@ export default {
     tx() {
       return this.signRequest ? parseTx(this.signRequest.signMessage) : null
     },
-    // fees() {
-    //   return this.tx ? this.tx.tx.value.fee.amount : null
-    // },
+    fees() {
+      return this.signRequest ? parseFee(this.signRequest.signMessage) : null
```
colw: You may even be able to do away with the `parsers.js` file if you parse it once and then have each of these computed properties just access the object as normal.

src/components/common/SessionApprove.vue
```diff
@@ -20,9 +20,11 @@
         From
         <Bech32 :address="senderAddress" />
       </div>
-
-      <!-- <TableInvoice :fees="fees" /> -->
-
+      <TableInvoice
+        :amount="amount"
```
colw: This won't (yet) pass the tests since the `TableInvoice` code is in the other PR #2802

test/unit/specs/components/common/SessionApprove.spec.js
faboweb: `signRequest` should be parsed when adding it to the store in the extension I think. The getter is only in the extension.

src/scripts/parsers.js
faboweb: I reduced the amounts of `JSON.parse`

PR #2790, opened 2019-06-28
Benji/ext 30 validators monikers fix

thebkr7:
Closes #ISSUE

Passes delegates to chrome extension components so that SessionApprove can access the Moniker. There is a more efficient way of only passing the Moniker rather than passing all the delegates and filtering them in LiStakeTransaction but this would take reworking how that side of Lunie.io works.

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/src/ActionModal/components/ActionModal.vue
+++ b/src/ActionModal/components/ActionModal.vue
@@ -457,7 +457,7 @@ export default {
       }
 
       try {
-        await this.actionManager.send(memo, feeProperties)
+        await this.actionManager.send(memo, feeProperties, this.transactionData.valdiators)        
         this.trackEvent(
           `event`,
           `successful-submit`,
--- a/src/ActionModal/components/ActionModal.vue
+++ b/src/ActionModal/components/ActionModal.vue
@@ -457,7 +457,11 @@ export default {
       }
 
       try {
-        await this.actionManager.send(memo, feeProperties, this.transactionData.valdiators)        
+        await this.actionManager.send(
+          memo,
+          feeProperties,
+          this.transactionData.valdiators
+        )
         this.trackEvent(
           `event`,
           `successful-submit`,
--- a/src/ActionModal/components/ActionModal.vue
+++ b/src/ActionModal/components/ActionModal.vue
@@ -460,7 +460,7 @@ export default {
         await this.actionManager.send(
           memo,
           feeProperties,
-          this.transactionData.valdiators
+          this.transactionData.validators
         )
         this.trackEvent(
           `event`,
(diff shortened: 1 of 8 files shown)


src/ActionModal/components/ActionModal.vue
```diff
@@ -457,7 +457,11 @@ export default {
       }
 
       try {
-        await this.actionManager.send(memo, feeProperties)
+        await this.actionManager.send(
+          memo,
+          feeProperties,
+          this.transactionData.valdiators
```
jbibla:
```suggestion
          this.transactionData.validators
```

src/components/common/SessionApprove.vue
```diff
@@ -114,6 +114,9 @@ export default {
     },
     senderAddress() {
       return this.signRequest ? this.signRequest.senderAddress : null
+    },
+    deligates() {
```
jbibla:
```suggestion
    delegates() {
```

src/components/common/SessionApprove.vue
```diff
@@ -75,15 +75,16 @@ import { required } from "vuelidate/lib/validators"
...
 
   return {
     tx: {
       type: "auth/StdTx",
       value: {
         msg: msgs,
         fee,
-        memo
+        memo,
+        delegates
```
faboweb: this doesn't look correct. the delegates which serve as meta data shouldn't be part of the actual transaction

src/components/common/SessionApprove.vue
```diff
@@ -114,6 +114,9 @@ export default {
     },
     senderAddress() {
       return this.signRequest ? this.signRequest.senderAddress : null
+    },
+    delegates() {
+      return this.tx ? this.tx.tx.value.delegates : null
```
faboweb: shouldn't this be `this.signRequest.delegates`?

src/ActionModal/components/ActionModal.vue
```diff
@@ -457,7 +457,11 @@ export default {
       }
 
       try {
-        await this.actionManager.send(memo, feeProperties)
+        await this.actionManager.send(
+          memo,
+          feeProperties,
+          this.transactionData.validators
```
colw: My spidey sense tells me this should be wrapped in some kind of generic context object. There might be other properties, apart from delegates, that need to get passed into the ActionManager, and we don't want to overcrowd the parameter list.

thebkr7: I'd agree. Fabo has suggested that we make a call from the extension to get the validators information to avoid a potential man in the middle attack. Doing so will eliminate this change.

colw:
I'll jump into this now to help @thebkr7 out.

PR #2798, opened 2019-07-03
Jordan/session refactor

jbibla:
This is the start of a refactor to use `vue-router` for session screens. This PR accomplishes the following:

- dramatic reduction in duplicated code (back buttons, close buttons. `setState` methods)
- reduce code required in session.js (`toggleSessionModal`, `setSessionModalView`)
- nice URLs for session screens
- delete some components (`SessionRouter`, `TmModal`)
- more semantic HTML (using A tags to change routes and URLs)
- nicer, proper looking session screens instead of modals 😸 

This seems to be a more correct approach for this implementation. With a few more hours I could clean this up, remove a lot of code, and have everything working as expected again.

Thank you! 🚀

---

For contributor:

For reviewer:


src/components/common/SessionFrame.vue
```diff
@@ -0,0 +1,76 @@
...
+  computed: {
+    ...mapGetters([`session`])
+  }
+}
+</script>
+
+<style>
+@import "../../styles/session.css";
+
+.session-modal {
```
faboweb: should we split the styles in `session.css` and here or should we merge them?

src/routes.js
```diff
@@ -82,6 +82,55 @@ export default [
     name: `block`,
     component: require(`./components/network/PageBlock`).default
   },
+  {
```
colw: Can we tell Code Climate to ignore this file?

Screenshot 2019-07-03 at 15 08 40

faboweb: It's not a required check

colw: Sure, but I think it's a good habit to treat warnings and errors as much the same. In this case there is nothing wrong with this code.

faboweb: I marked the errors at `wontfix`

colw:
@jbibla Just to confirm, we don't have back/close buttons on the Lunie sign-in form anymore?

faboweb:
Not in the extension but in Lunie this should stay the same I think.

colw:
> In the extension there should be no close button. But in Lunie everything should stay the same.

Screenshot 2019-07-04 at 12 42 54

colw:
> In the extension there should be no close button. But in Lunie everything should stay the same.

Screenshot 2019-07-04 at 12 43 44

jbibla:
it’s true! i removed the back and close buttons :dizzy_face:

modals create a lot of dark ux patterns.

browsers have back buttons and closing a screen with an `x` does not really make sense on proper routes.

faboweb:
> browsers have back buttons and closing a screen with an x does not really make sense on proper routes.

ios apps don't

colw:
> it’s true! i removed the back and close buttons 😵
> 
> modals create a lot of dark ux patterns.
> 
> browsers have back buttons and closing a screen with an `x` does not really make sense on proper routes.

Ah, right 👍 

> closing a screen with an `x` does not really make sense on proper routes.

This still _looks_ like a modal and routes are mostly an implementation detail, you can't even see them easily on mobile for example. 

Also the back button looks weird in mobile.

Screenshot 2019-07-04 at 13 27 09

jbibla:
> ios apps don't

that's true. we will resolve this when we have mobile apps.

> This still looks like a modal and routes are mostly an implementation detail, you can't even see them easily on mobile for example.

over time, they should look less like modals. take a look at some web app sign in / sign up pages. they look more or less how ours do now. 

i don't think routes are an implementation detail. try using the back button while on the session modals or sending a link to "sign in with ledger".

> Also the 'Back to Lunie' button looks weird on mobile.

definitely. i'll fix this.

faboweb:
Hmm I feel stuck on the session flow. Like here:
image

faboweb:
Live: https://9792-99653950-gh.circle-artifacts.com/0/tmp/voyager/dist/index.html#/

src/components/common/TmSessionExtension.vue
```diff
@@ -1,28 +1,26 @@
...
-      <div class="session-header">
-        <a @click="goBack">
-          <i class="material-icons session-back">arrow_back</i>
-        </a>
+  <SessionFrame>
+    <div class="session">
+      <div class="session-container">
         <h2 class="session-title">
-          Use Lunie Chrome extension
+          Use Lunie Browser extension
```
thebkr7: Is `Chrome` not more specific for other browser users that may think their browser is supported?

src/components/common/TmSessionExtension.vue
```diff
@@ -48,29 +46,37 @@
...
-          <p class="ledger-install">
-            You can find the Lunie Chrome extension on the
-            <a href="">Google Play Store</a>.
-          </p>
+
+          <div class="session-footer">
+            <p class="ledger-install">
+              You can find the Lunie Browser Extension in the
+              <a
+                href="https://chrome.google.com/webstore/category/extensions"
```
thebkr7: If we are going to link to the extension in the chrome store we may want to have the link as a config value to update all locations at once. It would avoid 404's down the road if we miss updating a link.

src/components/common/TmSessionExtension.vue
jbibla: Brave 🙃

src/components/common/TmSessionExtension.vue
jbibla: nice idea. i would propose making a separate ticket for this task. would apply to many affiliate links - we could already do this for ledger too.

PR #2802, opened 2019-07-03
Benji/2801 Refactors TableInvoice to receive bondDenom as props and updates tests

### LINKED ISSUE #2801: Refactor TableInvoice to get all data from props rather than store getters

thebkr7:
Closes #2801

**Description:**

Refactors TableInvoice to receive bondDenom as props, updates ActionModal to pass props to TableInvoice, and updates tests.
Does not update SesionApprove to pass bondDenom props to Table invoice as that is completed in https://github.com/luniehq/lunie/pull/2785

Thank you! 🚀

---

For contributor:

For reviewer:


colw:
Looks great. Do you mind if I add some tests as examples for the TableInvoice?

colw:
I took out the`// already in atoms` comments with the intention of change the property names.

Something like `subTotalAtoms`? Is that a good idea?

thebkr7:
The comment was in there from before but looked outdated. Good call, looks good.

test/unit/specs/components/ActionModal/components/__snapshots__/ActionModal.spec.js.snap
```diff
@@ -98,6 +98,7 @@ exports[`ActionModal should show the action modal when user has logged in with e
      
     <tableinvoice-stub
       amount="0"
+      bonddenom="uatom"
```
jbibla: why is this `uatom` if `bondDenom: `STAKE``

test/unit/specs/components/ActionModal/components/TableInvoice.spec.js
```diff
@@ -6,25 +6,31 @@ describe(`TableInvoice`, () => {
...
       }
     })
   })
 
-  it(`should show the transaction invoice summary`, async () => {
+  it(`should render correctly`, async () => {
     expect(wrapper.vm.$el).toMatchSnapshot()
   })
+
+  it(`should display the correct subtotal`, async () => {
```
jbibla: you need to test the computed properties, not the rendering of the component in a specific scenario. 

eg. `networkFee` with different gas prices and gas estimates.

test/unit/specs/components/ActionModal/components/TableInvoice.spec.js
```diff
@@ -6,25 +6,31 @@ describe(`TableInvoice`, () => {
...
+    expect(wrapper.text()).toMatch(/17.2[0]+ STAKE/)
```
jbibla: try to test `wrapper.vm.subTotal` instead of `wrapper.text`

jbibla: same for the two tests below

test/unit/specs/components/ActionModal/components/__snapshots__/ActionModal.spec.js.snap
thebkr7: Would this not be determined by the test? https://github.com/luniehq/lunie/blob/f929ee5d1f8d831588fdd7fb5287bbbb638077e5/src/ActionModal/components/ActionModal.vue#L62

test/unit/specs/components/ActionModal/components/TableInvoice.spec.js
thebkr7: @colw Do you want to update them our should I?

test/unit/specs/components/ActionModal/components/__snapshots__/ActionModal.spec.js.snap
jbibla: yep yep. my mistake.

test/unit/specs/components/ActionModal/components/TableInvoice.spec.js
jbibla: ah ok - fabo filled me in on the debate.

not worth bike shedding over - but my feeling is that we should test the way things are rendered with snapshots and we should use unit tests to test computed properties.

you guys sort it out so we can get it merged!

test/unit/specs/components/ActionModal/components/TableInvoice.spec.js
```diff
@@ -6,25 +6,31 @@ describe(`TableInvoice`, () => {
...
-  }
-
   beforeEach(() => {
     wrapper = shallowMount(TableInvoice, {
       localVue,
-      mocks: { $store },
       propsData: {
-        amount: 1,
+        amount: 17.2,
         gasEstimate: 1234567,
```
faboweb: can we get rid of gasEstimate and gasPrice right away? Doesn't make sense to keep this right?

test/unit/specs/components/ActionModal/components/TableInvoice.spec.js
colw: Snapshots are only useful to let you know something changed. It has not bearing on wether it is correct or not.

If you test the internal implementation, it also means you need to change the tests, which kind of defeats the purpose of the tests.

This way, you can change the internal methods as much as you want. Change a computed property name, no worries. Change a tag order, no worries. Accidentally change that `*` into a `+`? Test fails. It's job done.

thebkr7: Just to clarify my comment now that I read it with a different perspective. By "them" I meant the tests. @jbibla Sorry for any confusion

test/unit/specs/components/ActionModal/components/TableInvoice.spec.js
thebkr7: Can't get rid of gasEstimate because it is a required prop in TableInvoice but gasPrice we can.

PR #2812, opened 2019-07-05
Jordan/accounts fixes

jbibla:
- refactored the AccountList into its own component to reduce duplicate code and ensure styles are the same when we render account lists.
- also added session frame to SessionAccounts because it was missing and inconsistent

Thank you! 🚀

---

For contributor:

For reviewer:


src/components/common/AccountList.vue
```diff
@@ -0,0 +1,67 @@
+<template>
+  <ul class="account-list">
+    <li v-for="account in accounts" :key="account.name" class="account">
+      <div class="account-info">
+        <h3>{{ account.name }}</h3>
+        <Bech32 :address="account.address || ''" />
+      </div>
+      <TmBtn
```
faboweb: This won't hide the Use Account button in the extension right?

test/unit/specs/components/common/SessionAccounts.spec.js
```diff
@@ -19,7 +19,7 @@ describe(`SessionAccounts`, () => {
         ]
       }
     }
-    wrapper = shallowMount(SessionAccounts, {
+    wrapper = mount(SessionAccounts, {
```
faboweb: Why `mount` > `shallowMount`?

colw: I'm guessing because the `AccountList` isn't rendered, which is what is tested. Maybe it's better to ShallowMount and test that the accounts are present and passed to the `AccountList`?

@jbibla I can't comment on the line because it isn't diffed, but does the computed property `accounts` do anything? It seems to pass the same array?

colw: I'm guessing because the `AccountList` isn't rendered, which is what is tested. Maybe it's better to ShallowMount and test that the accounts are present and passed to the `AccountList`?

src/components/common/SessionAccounts.vue
```diff
@@ -1,37 +1,38 @@
...
   name: `session-accounts`,
   components: {
-    Bech32
-    // TmBtn
+    AccountList,
+    SessionFrame,
+    TmBtn
   },
   computed: {
     accounts() {
```
colw: Does this pass the same array straight through?

src/components/common/SessionAccounts.vue
```diff
@@ -1,37 +1,38 @@
...
+            <router-link to="/wallet">
+              Lunie
+            </router-link>
+            and to approve transactions.
+          </p>
+
+          <AccountList :accounts="accounts" />
+
+          <div class="button-add-account">
+            <TmBtn value="Add Another Account" to="/accounts" color="primary" />
```
colw: Do you need a `type="link"` on this button? Otherwise it won't become a `router-link`?

src/components/common/AccountList.vue
```diff
@@ -0,0 +1,67 @@
...
+        class="account-button"
+        value="Use Account"
+        color="primary"
+        @click.native="signIn(account.address)"
```
colw: This doesn't apply to the extension?

src/components/common/AccountList.vue
```diff
@@ -0,0 +1,67 @@
+<template>
+  <ul class="account-list">
+    <li v-for="account in accounts" :key="account.name" class="account">
+      <div class="account-info">
+        <h3>{{ account.name }}</h3>
+        <Bech32 :address="account.address || ''" />
```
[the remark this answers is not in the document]
jbibla: good ideas!

src/components/common/AccountList.vue
jbibla: nope

src/components/common/AccountList.vue
jbibla: good call - i'll fix this now.

src/components/common/SessionAccounts.vue
jbibla: yeah that's correct - but this will display as a button that links to `/accounts`

colw: Ok, but I believe the `TmBtn` component only handles the `to` prop when it is also a `type === `'ink'`.

src/components/common/AccountList.vue
faboweb: But shouldn't it?!

faboweb:
A bunch of comments were not resolved. Was there a need to fast track this?

PR #2814, opened 2019-07-08
Colw/ext get accounts

colw:
**Description:**

Display accounts from extension in sign on screen and extension correctly

Add tests

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/src/scripts/extension-utils.js
+++ b/src/scripts/extension-utils.js
@@ -15,9 +15,6 @@ const processMessage = (store, type, payload) => {
       store.commit("setExtensionAvailable")
       store.dispatch("getAddressesFromExtension")
       break
-    case "GET_WALLETS_RESPONSE":
-      store.commit("setExtensionAccounts", payload)
-      break
   }
 }
 
@@ -62,7 +59,8 @@ function waitForResponse(type) {
 }
 
 const sendAsyncMessageToContentScript = async payload => {
-  sendMessageToContentScript(payload, true)
+  // I think we can deal with async console errors problems by returning true
+  sendMessageToContentScript(payload, false)
 
   // await async response
   const response = await waitForResponse(`${payload.type}_RESPONSE`)
@@ -72,10 +70,11 @@ const sendAsyncMessageToContentScript = async payload => {
   return response
 }
 
-export const getAccounts = () => {
-  sendMessageToContentScript({
+export const getWalletsFromExtension = async () => {
+  const wallets = await sendAsyncMessageToContentScript({
     type: "GET_WALLETS"
   })
+  return wallets
 }
 
 export const signWithExtension = async (signMessage, senderAddress) => {
--- a/src/scripts/extension-utils.js
+++ b/src/scripts/extension-utils.js
@@ -15,6 +15,11 @@ const processMessage = (store, type, payload) => {
       store.commit("setExtensionAvailable")
       store.dispatch("getAddressesFromExtension")
       break
+    case "GET_WALLETS_RESPONSE":
+      store.commit("setExtensionAccounts", payload)
+      break
+    default:
+      return
   }
 }
 
@@ -70,11 +75,8 @@ const sendAsyncMessageToContentScript = async payload => {
   return response
 }
 
-export const getWalletsFromExtension = async () => {
-  const wallets = await sendAsyncMessageToContentScript({
-    type: "GET_WALLETS"
-  })
-  return wallets
+export const getAccountsFromExtension = () => {
+  sendMessageToContentScript({ type: "GET_WALLETS" })
 }
 
 export const signWithExtension = async (signMessage, senderAddress) => {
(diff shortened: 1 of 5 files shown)


src/scripts/extension-utils.js
```diff
@@ -72,10 +70,11 @@ const sendAsyncMessageToContentScript = async payload => {
   return response
 }
 
-export const getAccounts = () => {
-  sendMessageToContentScript({
+export const getWalletsFromExtension = async () => {
```
faboweb: We agreed to name `wallets` `accounts` in this context as we only send the name and the address not the actual keys.

src/scripts/extension-utils.js
```diff
@@ -15,9 +15,6 @@ const processMessage = (store, type, payload) => {
       store.commit("setExtensionAvailable")
       store.dispatch("getAddressesFromExtension")
       break
-    case "GET_WALLETS_RESPONSE":
```
faboweb: We might still want this to allow the extension to push new accounts. I.e. if the user creates an account while Lunie is open.

src/scripts/extension-utils.js
```diff
@@ -72,10 +70,11 @@ const sendAsyncMessageToContentScript = async payload => {
   return response
 }
 
-export const getAccounts = () => {
-  sendMessageToContentScript({
+export const getWalletsFromExtension = async () => {
+  const wallets = await sendAsyncMessageToContentScript({
```
faboweb: Do we need my `sendAsyncMessageToContentScript` with the new async handling in the extension?

src/scripts/extension-utils.js
colw: OK. In that case, it does it make sense to just remove the async call below?

src/scripts/extension-utils.js
colw: Good thinking.

src/scripts/extension-utils.js
faboweb: Sounds reasonable. What do you think?

colw: Less code! 👍  :)

src/scripts/extension-utils.js
colw: We can still use it. We just don't need to 'skipResponse'.

Ideally we could handle every response in `processMessage((store, type, payload)`. It means all responses go in the store. Would that work for every request? Sign requests?

faboweb: Maybe. But it would mean that we need to handle the pending signRequest state in the store. Might be reasonable.

src/vuex/modules/extension.js
```diff
@@ -22,10 +22,12 @@ export default () => {
...
     async getAddressesFromExtension({
       state: {
-        externals: { getAccounts }
-      }
+        externals: { getAccountsFromExtension }
+      },
+      commit
     }) {
-      getAccounts() // response will be handled by extension message listener
+      const wallets = await getAccountsFromExtension()
```
faboweb: This is now not working anymore I think

src/scripts/extension-utils.js
colw: Technically we should store all state in the store. But, it's hard to tell if this (the actionmodal) is an exception)

…thinking. 🤔

src/vuex/modules/extension.js
colw: Woops, it should still set the accounts, but it's not necessary code here.

src/scripts/extension-utils.js
```diff
@@ -18,6 +18,8 @@ const processMessage = (store, type, payload) => {
     case "GET_WALLETS_RESPONSE":
       store.commit("setExtensionAccounts", payload)
       break
+    default:
+      return
```
jbibla: missing test ;)

colw: 🤦‍♂

colw: I'm still seeing the red line above, but I have a test for it? Does CodeClimate take a little while… or?

test/unit/specs/scripts/extension-utils.spec.js
```diff
@@ -75,6 +75,20 @@ describe(`Extension Utils`, () => {
         }
       ])
     })
+
+    it("should do nothing with no message type", () => {
+      const result = processLunieExtensionMessages(store)({
+        source: global,
+        data: {
+          type: "NOT_FROM_LUNIE_EXTENSION",
```
faboweb:
```suggestion
          type: "FROM_LUNIE_EXTENSION",
```
This would get filtered out by `filterExtensionMessage`

colw: 🤦‍♂  Derp. Thanks!

src/scripts/extension-utils.js
jbibla: green line now!

PR #2826, opened 2019-07-11
Benji/ext30 query delegates

thebkr7:
Closes # extension issue # 30

**Description:**

Adds monikers to the delegation transaction data to be passed to the extension.

Thank you! 🚀

---

For contributor:

For reviewer:


src/ActionModal/components/ActionModal.vue
```diff
@@ -317,6 +317,10 @@ export default {
       type: Function,
       default: undefined
     },
+    validatorMoniker: {
```
faboweb: This makes the actionmodal more dependent on the wrapping application. I would recommend querying this from the actionmanager. What do you think @colw ?

src/ActionModal/utils/ActionManager.js
```diff
@@ -67,13 +67,14 @@ export default class ActionManager {
...
-  async send(memo, txMetaData) {
+  async send(memo, txMetaData, subData) {
     this.readyCheck()
 
     const { gasEstimate, gasPrice, submitType, password } = txMetaData
     const signer = getSigner(config, submitType, {
       address: this.context.userAddress,
-      password
+      password,
+      subData
```
faboweb: What is 'subdata'?

test/unit/specs/components/ActionModal/utils/actions.js
```diff
@@ -8,6 +8,23 @@ export const sendTx = {
       }
     ]
   },
+  txDelProps: {
```
faboweb: We avoid abbreviations.

test/unit/specs/components/ActionModal/utils/actions.js
```diff
@@ -8,6 +8,23 @@ export const sendTx = {
...
+    amounts: [
+      {
+        amount: "20000",
+        denom: "uatom"
+      }
+    ],
+    validator_address: "cosmosvaloper1pnmn05d9prestn2vxujr8h88fpm9c0skeg72p5",
+    message: {
+      value: {
+        validatorMoniker: "operator_account_2"
```
faboweb: This data structure looks weird. "Message.value.validatorMoniker" is that intended?

thebkr7: Yes, it was intended due to the way the action managers message was structured. But It does not matter because the way that it is structured now this test will be removed as it is the same functionality for any message.

src/ActionModal/utils/ActionManager.js
thebkr7: txMetaData was taken so it is an alternative. I tried avoiding a name that would suggest it is only for validator monikers in case we want to pass more data through in the future. Would just `metaData` be better?

src/ActionModal/components/ActionModal.vue
```diff
@@ -318,7 +318,7 @@ export default {
       default: undefined
     },
     validatorMoniker: {
-      type: String,
+      type: [String, Object],
```
faboweb: We try to avoid two types ;)

thebkr7: Woa quickest feedback ever!  🚀

src/ActionModal/utils/ActionManager.js
faboweb: would it make sense to merge it with txMetaData? Or do you think splitting the 2 data blobs makes more sense?

src/ActionModal/components/ActionModal.vue
colw: Shouldn't this info be in the `fromOptions`?

src/ActionModal/components/DelegationModal.vue
```diff
@@ -136,6 +137,21 @@ export default {
 
       return this.fromOptions[this.selectedIndex].address
     },
+    validatorMoniker() {
```
colw: Something about this function (and I've seen the code elsewhere), is we seem to be calculating something we should already know?

src/ActionModal/components/ActionModal.vue
colw: The problem is we are treating the validator and the moniker as separate things.

src/ActionModal/components/DelegationModal.vue
```diff
@@ -136,6 +137,21 @@ export default {
...
+      if (this.from === this.modalContext.userAddress) {
```
faboweb: There is the computed property `this.isRedelegation` already.

PR #2828, opened 2019-07-12
Fabo/add disconnected bar

faboweb:
Closes #ISSUE

image

**Description:**

Thank you! 🚀

---

For contributor:

For reviewer:


colw:
This is probably related to the underlying connection module, but what of the use case that you are connected, and then lose the connection?

I can disconnect in chrome, and get notification warnings (failed to fetch etc.) but the `DisconnectedBar` would not appear.

src/components/common/DisconnectedBar.vue
```diff
@@ -0,0 +1,73 @@
+<template>
+  <div v-if="connection.stopConnecting" class="disconnected-bar">
+    <i></i>
+    <p>
+      You are currently not connected to a node. Retry?
```
colw: I think the word 'Retry?' should initiate `reconnect` too.

src/vuex/getters.js
```diff
@@ -110,7 +110,7 @@ export const depositDenom = getters =>
     : `uatom`
 
 // connection
-export const approvalRequired = state => state.connection.approvalRequired
+export const connection = state => state.connection
```
colw: Thinking about this. I believe we might be making too much use of getters, where we could just use `mapState` in the component.

faboweb: That's a great idea. Hadn't considered this.

faboweb:
> This is probably related to the underlying connection module, but what of the use case that you are connected, and then lose the connection?

That should be handled by this. After 5 attempts to connect to the websocket of the fullnode the bar appears.

jbibla:
> I can disconnect in chrome, and get notification warnings (failed to fetch etc.) but the DisconnectedBar would not appear.

has this been resolved?

src/components/common/DisconnectedBar.vue
```diff
@@ -2,8 +2,8 @@
   <div v-if="stopConnecting" class="disconnected-bar">
     <i></i>
     <p>
-      You are currently not connected to a node.
-      <a class="link" @click="reconnect">Retry</a>?
+      You are currently not connected.
+      <a class="link" @click="reconnect">Try Reconnecting</a>?
```
thebkr7: Nice 👍

PR #2832, opened 2019-07-12
Benji/Ext58 Adds back buttons to TmSessionSignUp and TmSessionImport for Ext

thebkr7:
Closes na

**Description:**

Adds back buttons to TmSessionSignUp and TmSessionImport that are only shown in the extension

Thank you! 🚀

---

For contributor:

For reviewer:


test/unit/specs/components/common/__snapshots__/TmSessionImport.spec.js.snap
```diff
@@ -21,6 +21,480 @@ exports[`TmSessionImport has the expected html structure 1`] = `
...
+    class="tm-btn session-close"
+  >
+    <span
+      class="tm-btn__container tm-btn--secondary"
+    >
+      <!---->
+      <span
+        class="tm-btn__value"
+      >
+        Back to Lunie
```
faboweb: Back button is showing

test/unit/specs/components/common/TmSessionSignUp.spec.js
```diff
@@ -34,6 +35,32 @@ describe(`TmSessionSignUp`, () => {
     expect(wrapper.vm.$el).toMatchSnapshot()
   })
 
+  it(`should not show back button if in website`, () => {
+    expect(wrapper.vm.$el).toMatchSnapshot()
```
faboweb: the test should check for the back button existence

test/unit/specs/components/common/__snapshots__/TmSessionImport.spec.js.snap
thebkr7: @faboweb That is not the back button I added but rather the close button from SessionFrame. It is not visible in the extension and has been there since sessionFrame was introduced.

thebkr7:
Note: the branch name is misleading as that is another PR. The PR name is accurate. This was an error on my part.

src/components/common/TmSessionImport.vue
```diff
@@ -165,9 +168,19 @@ export default {
     }
   }),
   computed: {
-    ...mapGetters([`connected`])
+    ...mapGetters([`connected`]),
```
faboweb:
```suggestion
    ...mapGetters([`connected`, `extension`]),
```

that replaces all of `inExtension` I think

test/unit/specs/components/common/TmSessionImport.spec.js
```diff
@@ -16,7 +16,8 @@ describe(`TmSessionImport`, () => {
   beforeEach(() => {
     store = new Vuex.Store({
```
faboweb: We usually just mock the store without using the actual Vuex store. This might be still old in here. Checkout https://github.com/luniehq/lunie/blob/develop/test/unit/specs/components/ActionModal/components/SendModal.spec.js#L33

src/components/common/TmSessionImport.vue
```diff
@@ -1,6 +1,9 @@
 <template>
   <SessionFrame>
     <div class="session">
+      <a v-if="inExtension" @click="goBack">
```
jbibla: this is a good implementation! 

instead of duplicating the back button in both session screens, should we make this 1 component that can be tested on its own?

thebkr7: Thanks :)
Do you think just adding it to the SessionFrame with conditionals to display based on the route would make sense? A separate component may be the answer but the back buttons are only in two places. @jbibla

src/components/common/TmSessionImport.vue
thebkr7: This will cause an error when the getter `extension` is not defined in the extension. It will work but an error will be in the console. Think it's worth it?

test/unit/specs/components/common/TmSessionImport.spec.js
thebkr7: Yeah I like the new way better too. That was just there already. I'll update it 👍

src/components/common/TmSessionImport.vue
faboweb: How about defining the getter `extension` in the extension?

src/components/common/TmSessionImport.vue
faboweb: Do you guys think that it makes sense to have the back button only in the extension? If there is a need for it in the extension, wouldn't it make for a better UX in the browser as well?

src/components/common/TmSessionImport.vue
```diff
@@ -165,9 +168,19 @@ export default {
     }
   }),
   computed: {
-    ...mapGetters([`connected`])
+    ...mapGetters([`connected`]),
+    inExtension() {
+      if (this.$store.getters.extension) {
```
colw: I believe this would always return true.

In addition to what Fabian said above, you need to check `extension.enabled`.

thebkr7: It isn't always true and it does work in my testing. It comes back as undefined when the getter `extension` is not defined in the extension and as defined when in the website. An alternative would be checking the chrome object to see what type of window you are in.

src/components/common/TmSessionImport.vue
thebkr7: This works well

src/components/common/TmSessionImport.vue
thebkr7: Good idea, this works well.

src/components/common/TmSessionSignUp.vue
```diff
@@ -1,179 +1,175 @@
...
+      </h2>
+      <div v-if="session.insecureMode" class="session-main">
+        <div class="danger-zone">
+          <h2>DANGER ZONE</h2>
           <p>
-            Creating an address in the browser is unsafe. To offer you a secure
-            alternative we will be releasing a browser extension and a mobile
-            app soon.
+            Creating an address in the browser is not advised. This feature is
+            only enabled in insecure mode for testing purposes and should not be
```
jbibla: nice improvement.

PR #2869, opened 2019-08-08
colw/graphql keybase info

colw:
**Description:**

Loads key base info using GraphQL.

For now, create a `.env.development` file locally with the following:
```
VUE_APP_GRAPHQL_URL=<BACKEND_URL>
VUE_APP_HASURA_SECRET=<HASURA_KEY>
```

Rather than load the list, I have a query by validator keybase ID. So it runs for each item in the validator list, but we can also reuse it for the `PageValidator` page. Thoughts?

Also, it is possible to use `<ApolloXYZ query="...">` components that use slots to provide data to children. Could be a useful alternative.

Thank you! 🚀

---

For contributor:

For reviewer:


src/components/staking/PageValidator.vue
```diff
@@ -145,25 +139,23 @@
...
             <dl class="info_dl">
               <dt>Last Commission Change</dt>
               <dd>{{ lastCommissionChange }}</dd>
             </dl>
             <dl class="info_dl">
               <dt>Self Delegation</dt>
-              <dd id="page-profile__self-bond">
-                {{ selfBond }}
-              </dd>
+              <dd id="page-profile__self-bond">{{ selfBond }}</dd>
```
faboweb: Note: Somehow formatting flip flops. Let's watch this and fix it if it stays this way.

src/gql/index.js
```diff
@@ -0,0 +1,14 @@
+import gql from "graphql-tag"
+
+export const ValidatorKeybaseInfo = gql`
```
faboweb: How about naming this `ValidatorProfileInfo` so we can enrich the data with other information

colw: Perfect, good idea.

src/components/staking/PageValidator.vue
colw: I know, it's gone nuts. I use the cli to perform final linting/fixing now since it doesn't like my VSCode version. There are some conflicting rules somewhere.

colw:
Works fine, need to discuss backend keys etc.

PR #2891, opened 2019-08-20
[Simsala] automatic release created for v1.0.0-beta.110

faboweb:
Please manually test before merging this to master <a href='https://13243-99653950-gh.circle-artifacts.com/0/tmp/voyager/dist/index.html'>here</a>

### Repository

- this is needed for babel and jest to play nicely on my machine @jbibla


jest.config.js
```diff
@@ -4,7 +4,7 @@ module.exports = {
     "^.+\\.vue$": "vue-jest",
     ".+\\.(css|styl|less|sass|scss|svg|png|jpg|ttf|woff|woff2)$":
       "jest-transform-stub",
-    "^.+\\.jsx?$": "babel-jest"
+    "^.+\\.(js|jsx)?$": "babel-jest"
```
faboweb: @jbibla I needed to remove this again as it broke my develop.
This expression also means: Any file with the ending : `.js`, `.jsx` or `.`. I think that is not intended.

jbibla: saw your fix on the other branch. no problem.

jbibla: you don't want this included, right?

faboweb: better not. but will be overwritten by the mobile fixes anyway.

PR #2923, opened 2019-08-29
Fabo/2904 switch network features

### LINKED ISSUE #2904: network switch functionality
In Lunie we need a network switcher and following it, the functioniality to query data for the selected network.
To avoid incompatibility between different SDK versions we want to limit data shown to the defined data schema in our backend.
The dataschema has a feature schema. These are the features we are already support for that network. Lunie should now show a message `Lunie has no support for this feature yet` on pages that are flagged with false. Lunie should also show `Lunie has no support for this feature yet` in the ActionModal for disabled interactions.

faboweb:
Closes #2904 

image

**Description:**

Thank you! 🚀

---

For contributor:

For reviewer:


src/components/common/FeatureNotAvailable.vue
```diff
@@ -0,0 +1,24 @@
+<template>
+  <TmDataMsg icon="timeline">
+    <div slot="title">{{ feature }} not yet available</div>
+    <div slot="subtitle">
+      ¯\_(ツ)_/¯<br />
+      We are working hard to get this feature into Lunie. If you are super
```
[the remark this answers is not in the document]
faboweb: Feel free to change the content.

src/config.js
```diff
@@ -9,14 +9,13 @@ const rpc =
 export default {
   name: `Lunie`,
   development: dev,
-  network: process.env.NETWORK || `local-testnet`,
+  network: `cosmoshub`,
```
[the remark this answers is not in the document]
faboweb: on purpose, it's more the default network. it wasn't actually used before.

src/config.js
```diff
@@ -9,14 +9,13 @@ const rpc =
...
   stargate,
   rpc,
   google_analytics_uid: process.env.GOOGLE_ANALYTICS_UID || "",
   node_halted_timeout: 120000,
   block_timeout: 10000,
   default_gas_price: dev ? 1e-9 : 2.5e-8, // recommended from Cosmos Docs
-  version: process.env.RELEASE,
```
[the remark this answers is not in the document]
faboweb: was used for sentry which we don't use anymore

faboweb:
coverage uploading is shaky. tests actually pass

src/router.js
```diff
@@ -28,3 +33,28 @@ const router = new Router({
...
+        await waitForAvailable(selectorFn)
+        resolve()
+      }, 50)
+    )
+  }
+}
+
+// check if feature is allowed and redirect if not
+async function featureAvailable(store, to) {
+  await waitForAvailable(() => store.state.networks.network)
```
colw: I think there is a better way to do this.

The router guards will wait on promises automatically, and Apollo is reactive I believe, and will return some kind of promise based value I think? I haven't checked it, but I think we can just plug one into the other. The `waitForAvailable` doesn't feel right to me.

src/vuex/modules/networks.js
```diff
@@ -0,0 +1,88 @@
+import gql from "graphql-tag"
```
colw: I don't think we need this module. We can use just Apollo for storage and just store the `chain_id` in the session module. Each component that needs the Network info can request it.

src/router.js
```diff
@@ -5,14 +5,19 @@ import Vue from "vue"
...
+export const routeGuard = store => async (to, from, next) => {
   // Redirect if fullPath begins with a hash (fallback for old pre history mode urls)
   if (to.fullPath.includes("#")) {
     const path = to.fullPath.substr(to.fullPath.indexOf("#") + 1)
     next(path)
     return
   }
 
+  if (!(await featureAvailable(store, to))) {
+    next(`/feature-not-available/${to.meta.feature}`)
```
colw: Should this be lowerCase too for consistency?

faboweb: This was to pass the feature string that is actually displayed to the page. If we pass the lower case string, we would need to post process it in the page. Maybe could be done differently.

colw: We could use `this.$route.name` for for the title?

faboweb: `this.$route.name` actually should be lower case imo. It is used as an idea to describe routes. How would you pass `this.$route.name` to the feature page?

colw: That sounds reasonable.

I see now that `to.meta.feature` is used for more than one route, so we couldn't use the fields in reverse ways. So perhaps this is fine for now. We are optimising for the edge case, but I can see why.

(Perhaps each feature flag should have an associated title string, but it might be overkill for the moment.)

src/gql/index.js
```diff
@@ -60,5 +60,44 @@ export const SomeValidators = gql`
...
+    action_undelegate
+    action_withdraw
+    api_url
+    bech32_prefix
+    feature_activity
+    chain_id
+    feature_blocks
+    feature_portfolio
+    feature_proposals
+    feature_sessions
```
colw:
```suggestion
    feature_session
```

src/router.js
```diff
@@ -28,3 +38,21 @@ const router = new Router({
 })
 
 export default router
+
+// check if feature is allowed and redirect if not
+async function featureAvailable(apollo, networkId, to) {
+  const {
+    data: { networks }
+  } = await apollo.query({
+    query: NetworkCapabilities(networkId)
```
colw: It might be possible to query for the feature directly in the GraphQL. Effectively turning the result into a boolean. 

Just thought of this now, but I need to google the syntax.

PR #2926, opened 2019-08-29
Colw/graphql validators

colw:
Closes #ISSUE

**Description:**

Thank you! 🚀

---

For contributor:

For reviewer:


src/components/staking/PageValidator.vue
```diff
@@ -295,32 +271,18 @@ export default {
...
     },
-    statusColor() {
-      if (this.validator.jailed) return `red`
-
-      if (parseFloat(this.validator.status) === 0) return `orange`
-
-      return `green`
-    },
-    // empty descriptions have a strange '[do-not-modify]' value which we don't want to show
     website() {
```
faboweb: We don't want to show `[do-not-modify]` which is/was a default value and shows in some cases

colw: I caught this with the noBlanks filter (not the best name)

but… I don't think it works well for website. I'll fix this.

src/components/common/TmPage.vue
```diff
@@ -77,7 +77,7 @@ export default {
       default: false
     },
     error: {
-      type: Error,
+      type: [Error, Boolean],
```
faboweb: We have a rule about not allowing multiple types per property

src/components/staking/PageValidator.vue
```diff
@@ -503,17 +472,17 @@ span {
   border-radius: 0.25rem;
 }
 
-.validator-status.red {
+.validator-status.jailed {
```
faboweb: nice semantic classes

src/components/common/TmPage.vue
colw: Good call. `TMPage` already treated it like a Boolean, so I've removed the `Error` type. Other values will be coerced to Boolean.

Issue #1885, opened 2019-01-25
Create amount input component


faboweb:
In a lot of ActionModals we will check for the correct amount. @fedekunze recommended we refactor this component out and create a component for it.

We also introduced a "Max" button to maximize the input to the available amount. This should be included in this component.

AC:
- We have an input component that takes a max amount as a property
- The component checks if the amount entered is less or equal to the max amount
- The component allows to switch on a "Max" Button which would display on the right side of the input
- Hitting the "Max" button, the input would be set to the max amount defined by the property
- The component allows for easy binding to a value in a parent form

jbibla:
this is done #2956

PR #2956, opened 2019-09-05
Mario/2670 add max amount

### LINKED ISSUE #2670: add "max" button to amount fields

Often, users want to do make a tx with all the atoms. since there are 6 decimal places, this can be tedious to type out.

It is common in these situations to include a `max` button, which will automagically add the total amount left to the amount field for the user.

**Additional context**

Something to consider is the relationship with #2674.

mariopino:
Closes #2670

**Description:**

Add max button to amount field in transfer, delegate and undelegate modals.

Thank you! 🚀

---

For contributor:

For reviewer:


faboweb:
So this is missing the warning if there is no more funds for fees left as discussed in the issue

jbibla:
i might propose we add this button to `TmField` with a prop `maxAmount` indicating whether or not it should be shown and the amount. 

this would mean we wouldn't have to duplicate the HTML.

do you think this would be a good approach?

mariopino:
Added message:

Warning! you should probably leave some left over for fees

src/ActionModal/components/SendModal.vue
```diff
@@ -91,6 +98,9 @@
         name="Amount"
         type="between"
       />
+      <p v-if="isMaxAmount()" class="form-message notice">
+        Warning! you should probably leave some left over for fees
```
jbibla:
```suggestion
        You are about to use all your tokens for this transaction. Consider leaving a little bit left over to cover the network fees.
```

colw:
> i might propose we add this button to `TmField` with a prop `maxAmount` indicating whether or not it should be shown and the amount.
> 
> this would mean we wouldn't have to duplicate the HTML.
> 
> do you think this would be a good approach?

Good suggestion.

Looks great, I've no other comments apart from that.

faboweb:
> i might propose we add this button to TmField with a prop maxAmount indicating whether or not it should be shown and the amount.

TmField is already super cluttered. If we want to do this, let's create another component AmountInput or similar which has limited functionality. We thought about this here #1885. I propose we don't do the refactor here and I reopen the according issue.

PR #2959, opened 2019-09-06
Fabo/1771-tx grouping

### LINKED ISSUE #1771: Group txs by date on Tx History
Description:

I've seen in many mobile banking apps that they batch transactions according to the date they were submitted, I think it would be a nice to have for `transactions2`

Here's an image from [Revolut](https://www.revolut.com/):

screen shot 2018-12-22 at 12 02 51 pm

faboweb:
Closes #1771

**Description:**

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/src/components/transactions/TransactionList.vue
+++ b/src/components/transactions/TransactionList.vue
@@ -1,22 +1,46 @@
 <template>
   <div>
-    <TransactionItem
-      v-for="msg in transactions"
-      :key="msg.key"
-      :transaction="msg"
-      :validators="validators"
-      :address="address"
-    />
+    <template v-for="dayTxs in days">
+      <div :key="getDate(dayTxs[0])">
+        <h1 v-if="getDate(dayTxs[0]) === getDate({ time: new Date() })">
+          Today
+        </h1>
+        <h1 v-else>{{ dayTxs[0] | date }}</h1>
+        <TransactionItem
+          v-for="msg in dayTxs"
+          :key="msg.key"
+          :transaction="msg"
+          :validators="validators"
+          :address="address"
+        />
+      </div>
+    </template>
   </div>
 </template>
 
 <script>
 import TransactionItem from "./TransactionItem"
+import groupBy from "lodash.groupby"
+import moment from "moment"
+
+function getDate(tx) {
+  var dateObj = tx.time
+  var month = dateObj.getUTCMonth() + 1 //months from 1-12
+  var day = dateObj.getUTCDate()
+  var year = dateObj.getUTCFullYear()
+
+  return year + "/" + month + "/" + day
+}
 
 export default {
   name: `transaction-list`,
   components: {
     TransactionItem
+  },
+  filters: {
+    date(tx) {
+      return moment(tx.time).format("MMM Do YYYY")
+    }
   },
   props: {
     transactions: {
@@ -31,6 +55,14 @@ export default {
       type: Object,
       required: true
     }
+  },
+  computed: {
+    days() {
+      return groupBy(this.transactions, getDate)
+    }
+  },
+  methods: {
+    getDate
   }
 }
 </script>
--- a/package.json
+++ b/package.json
@@ -43,6 +43,7 @@
     "crypto-js": "^3.1.9-1",
     "graphql": "14.4.2",
     "jdenticon": "2.2.0",
+    "lodash.groupby": "4.6.0",
     "lodash.isempty": "^4.4.0",
     "lodash.orderby": "^4.6.0",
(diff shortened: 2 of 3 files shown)


colw:
Perhaps we could group by more than just date.

Perhaps, Today, yesterday, last week, August etc. (cc @jbibla )

src/components/transactions/TransactionList.vue
```diff
@@ -1,17 +1,55 @@
...
-    <TransactionItem
-      v-for="msg in transactions"
-      :key="msg.key"
-      :transaction="msg"
-      :validators="validators"
-      :address="address"
-    />
+    <template v-for="group in groupedTransactions">
+      <div :key="group[0].title">
+        <h1>{{ group[0].title }}</h1>
```
jbibla:
```suggestion
        <h3>{{ group[0].title }}</h3>
```

src/components/transactions/TransactionList.vue
```diff
@@ -31,6 +69,48 @@ export default {
...
+            tx
+          }
+        }
+
+        // check if tx is in a month this year
+        const date = moment(tx.time)
+        const today = moment()
+        if (date.year() === today.year()) {
+          return {
+            title: date.format("MMMM"),
```
jbibla: any reason we're grouping by month instead of day?

tests/unit/specs/components/transactions/__snapshots__/TransactionList.spec.js.snap
```diff
@@ -2,11 +2,35 @@
...
 <div>
-  <transactionitem-stub
-    address="cosmos1"
-    showmetadata="true"
-    transaction="[object Object]"
-    validators="[object Object]"
-  />
+  <div>
+    <h1>
+      July 19
```
jbibla: why is `July 19` above `Today`?

src/components/transactions/TransactionList.vue
```diff
@@ -1,17 +1,55 @@
...
 </template>
 
 <script>
 import TransactionItem from "./TransactionItem"
+import groupBy from "lodash.groupby"
+import orderBy from "lodash.orderby"
+import moment from "moment"
+
+function stripTime(momentTime) {
+  return momentTime.startOf("day") // sets the time to 12AM
```
colw: Could we do without this?

Compare dates using `inBetween` and the granularity?

src/components/transactions/TransactionList.vue
faboweb: See Colins proposal in the issue.

tests/unit/specs/components/transactions/__snapshots__/TransactionList.spec.js.snap
faboweb: Because `Today` in the tests is 1970

src/components/transactions/TransactionList.vue
faboweb: Ah, ok I wasn't aware of that. Will change it.

tests/unit/specs/components/transactions/__snapshots__/TransactionList.spec.js.snap
jbibla: groovy

src/components/transactions/TransactionList.vue
jbibla: ah didn't see that.

@colw what's your thinking here?

i'd like to remove date from the tx itself which won't be possible if we group by something other than date (day).

colw: If there is a 1:1 between date-title and transaction, it can add to the clutter, and defeat the purpose of a heading/separator. Also, as time goes on, the exact date becomes less important, hence grouping by month. At the top, or present time, it matters what day transactions took place.

Also, perhaps easier to read when formatted relatively. For example, I might not remember the exact date off-hand, but I certainly know when 'yesterday' was.

I'm quite flexible, I think we have the same goal of reducing clutter without hiding anything too important.

jbibla: interesting line of thinking! thanks for sharing. i suppose this highlights the value of user stories (which we have not written down / agreed upon).

in my mind the point stands though - if we are not grouping by date, then we can't remove date from the transaction component. the new groupings struck me as 1) new and 2) additional abstractions.

colw: >  if we are not grouping by date, then we can't remove date from the transaction component.

Of course. I wasn't aware of the direction when making the above suggestion. Your call, I like your approach too.

PR #2960, opened 2019-09-06
Mario/2858 add warning field backend

### LINKED ISSUE #2858: Add warning field to backend

We want to inform users quick, if the backend shows bad behavior. We currently need to update Lunie to do so.

Add a field to our backend (Hasura) where we can set a warning message. Lunie queries this field and displays the warning bar.

Only use Twitter for warnings. Update Lunie manually with the warnings.

**Additional context**

mariopino:
Closes #2858

**Description:**

When a row is found in maintenance table the message is displayed to the user.

Maintenance table in Hasura backend have two columns: message and type. 

Type can be ok, warning, danger or empty (default) and changes the maintenance bar background color accordingly to each type.

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/src/components/common/MaintenanceBar.vue
+++ b/src/components/common/MaintenanceBar.vue
@@ -63,16 +63,6 @@ export default {
   color: var(--bright);
 }
 
- /* don't change - should be consistent across themes */
-  --input-bg: transparent;
-  --success: hsl(120, 58%, 50%);
-  --success-bc: hsl(120, 58%, 41%);
-  --warning: hsl(35, 100%, 50%);
-  --warning-bc: hsl(35, 75%, 38%);
-  --danger: hsl(0, 100%, 55%);
-  --danger-bc: hsl(0, 100%, 41%);
-  --tendermint: #5ead37;
-
 .maintenance-bar.success {
   background-color: var(--success);
 }
--- a/src/components/common/MaintenanceBar.vue
+++ b/src/components/common/MaintenanceBar.vue
@@ -1,6 +1,6 @@
 <template>
   <div v-if="maintenance.length > 0">
-    <div v-for="message in maintenance">
+    <div v-for="message in maintenance" v-bind:key="message">
       <div v-if="message.show" class="maintenance-bar" v-bind:class="message.type">
         <i></i>
         <p>
@@ -21,7 +21,7 @@ export default {
   data: () => ({
     maintenance: []
   }),
-  methods: { 
+  methods: {
     close(message) {
       message.show = false
     }
@@ -37,7 +37,7 @@ export default {
           }
         }
       `,
-      update: result => result.maintenance,
+      update: result => result.maintenance
     }
   }
 }
--- a/src/components/common/MaintenanceBar.vue
+++ b/src/components/common/MaintenanceBar.vue
@@ -25,7 +25,7 @@
 
 <script>
 import { mapState } from "vuex"
-import { Maintenance, MaintenanceResult } from "src/gql"
+import gql from "graphql-tag"
 export default {
   name: `maintenance-bar`,
   data: () => ({
@@ -41,9 +41,24 @@ export default {
     }
   },
   apollo: {
+    // They key is the name of the data property
+    // on the component that you intend to populate.
     maintenance: {
-      query: Maintenance,
-      update: MaintenanceResult
+      // Yes, this looks confusing.
+      // It's just normal GraphQL.
+      query: gql`
+        query Maintenance {
+          maintenance {
+            message
+            type
+          }
+        }
+      `,
+
+      // Apollo maps results to the name of the query, for caching.
+      // So to update the right property on the componet, you need to
+      // select the property of the result with the name of the query.
+      update: result => result.maintenance,
     }
   }
 }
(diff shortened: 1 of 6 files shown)


faboweb:
@mariopino pending changelog missing

mariopino:
> @mariopino pending changelog missing

Done!

src/components/common/MaintenanceBar.vue
```diff
@@ -49,6 +75,18 @@ export default {
   color: var(--bright);
 }
 
+.maintenance-bar.ok {
+  background-color: #61b360;
```
faboweb: please use colors from variables that already exist or add the colors to the variables. This way we can have an overview about used colors.

src/components/common/MaintenanceBar.vue
```diff
@@ -1,23 +1,36 @@
...
-    <p>
-      We've identified problems with our servers that are causing issues for
-      some of our users. We apologize for the disruption and are working on a
-      fix.
-    </p>
-    <a class="close">
-      <i class="material-icons" @click="close">close</i>
-    </a>
+  <div>
+    <div v-if="session.maintenanceBar && show" class="maintenance-bar">
```
faboweb: Do we still need this switch? This should now be obsolete as we show the maintenance bar only of there are records in the DB?

src/components/common/MaintenanceBar.vue
```diff
@@ -1,23 +1,36 @@
...
+      <p>
+        We've identified problems with our servers that are causing issues for
+        some of our users. We apologize for the disruption and are working on a
+        fix.
+      </p>
+      <a class="close">
+        <i class="material-icons" @click="close">close</i>
+      </a>
+    </div>
+    <div v-if="maintenance.length > 0 && show" class="maintenance-bar" v-bind:class="maintenance[0].type">
```
faboweb: Why did you limit this just one record? Would it be possible to show multiple messages?

src/components/common/MaintenanceBar.vue
faboweb: This is also replicating the same html structure twice. I propose we remove this section completely. Remember to also remove the `session.maintenanceBar` in the store

src/components/common/MaintenanceBar.vue
mariopino: > Why did you limit this just one record? Would it be possible to show multiple messages?

Sure, I'll check the needed css so we can stack multiple messages.

mariopino:
- Obsolete maintenance bar code removed.

- Removed session.maintenanceBar from store.

- Multiple messages supported.

- Now possible message types are:

  - **success**  background-color: var(--success)
  - **warning** background-color: var(--warning)
  - **danger** background-color: var(--danger)

- Also there is a new field in maintenance table in Hasura backend: show (boolean), which let us have predefined messages and enable them on demand.

mariopino:
Pending to write some tests to cover the new behavior of the component mocking apollo query data ...

faboweb:
> Also there is a new field in maintenance table in Hasura backend: show (boolean), which let us have predefined messages and enable them on demand.

Sweet idea.

src/components/common/MaintenanceBar.vue
```diff
@@ -100,11 +100,11 @@ export default {
     justify-content: space-around;
   }
 
-  .maintenance-bar:nth-child(2) {
+  div > div.maintenance-bar:nth-child(2) {
```
jbibla: hmm. why is this necessary @mariopino ?

the css is `scoped` - seems like we could accomplish this somehow with less specificity

mariopino: In screen sizes with max-width less than 1025px, maintenance-bar messagess are shown in the bottom of the page, but they are not stacked, so only the last message is displayed.

So that css code is needed to stack the messages at the bottom. I'm not happy with the solution thought, sure must be a better way to acomplish that.

In the case we want to keep the current behavior (only the last message is displayed in screens <= 1024px width) we can simply delete those css statements.

jbibla: ok cool. i understand. 

i think you are right in trying to solve this stacking problem. but i think we should find a solution that doesn't involve `div > div`. perhaps class names for each banner? or some kind of architectural change might be in order.

mariopino: @jbibla  css was updated in the lasts commits so now it's:

```
 .message-2 {
    bottom: 36px;
  }
  .message-3 {
    bottom: 72px;
  }
  .message-4 {
    bottom: 108px;
  }
  .message-5 {
    bottom: 144px;
  }
```

I would like to use scss/vars to increment the bottom offset automatically in each message but I don't know if it is possible.

jbibla: i don't think we have scss at the moment. i think what you're trying to resolve is a symptom of a bigger issue.

i'll take a look now and try a couple things out.

faboweb: I am probably wrong, but can't you put the messages in a container and then just position them relative to each other?

jbibla: no you're correct. mario and i talked about this yesterday. it will require some modifications. 

i opened https://github.com/luniehq/lunie/issues/2967 as a separate issue.

PR #2973, opened 2019-09-10
Fabo/own websocket

faboweb:
Replaces tendermint-js with normal websockets. Solves socket switching issue.
Enables network selector publically


src/components/network/PageNetworks.vue
```diff
@@ -28,7 +29,18 @@ export default {
       return this.networks.filter(network => !network.testnet)
     },
     testNetworks() {
-      return this.networks.filter(network => network.testnet)
+      const networks = this.networks.filter(network => network.testnet)
+      if (config.development) {
+        networks.push({
+          title: "Local Testnet",
```
colw: I guess this isn't technically a local testnet since we need to change the graphql url for that.

src/components/staking/PageValidators.vue
```diff
@@ -4,7 +4,7 @@
     :loading="$apollo.queries.validators.loading"
     :loaded="!$apollo.queries.validators.loading"
     :error="$apollo.queries.validators.error"
-    :data-empty="validators.length === 0"
+    :data-empty="validators && validators.length === 0"
```
colw: Was this causing a bug before? It defaults to an array, so it should be ok.

Really not a big deal, but it would mean a few less characters if we can do without it.

src/initializeApp.js
```diff
@@ -48,15 +48,26 @@ export default function init(urlParams, env = process.env) {
...
-  store
-    .dispatch(`connect`)
-    // wait for connected as the check for session will sign in directly and query account data
-    .then(() => {
-      store.dispatch(`checkForPersistedSession`)
-      store.dispatch("getDelegates")
-      store.dispatch(`getPool`)
-      store.dispatch(`getMintingParameters`)
+  if (config.development) {
+    store.dispatch(`setNetwork`, {
```
colw: If we put this in, we should try not to dupliciate the config settings.

PR #3004, opened 2019-09-30
Jordan/misc fixes

jbibla:
**Description:**
- maintenance bar and sidebar nav arrangement fix
- fixed max button styling for new form height
- mobile portfolio view was split incorrectly
- connected network bar wasn't displaying chain-id and tooltip was broken
- graphql env default fix

Thank you! 🚀

---

For contributor:

For reviewer:


src/components/common/TmConnectedNetwork.vue
```diff
@@ -24,23 +24,24 @@
           class="tm-connected-network__string"
         >
           <span v-tooltip.top="networkTooltip" class="chain-id">
-            {{ connection.network }}
+            {{ connection.lastHeader.chain_id }}
```
faboweb: What is the thinking here?

jbibla: the networks page shows `Cosmos Hub` (title) and `cosmoshub-2` (chain ID) but `connection.network` is a third way of displaying network info that I found confusing (`cosmoshub`). 

maybe it would be better to show the same title as from the networks page, but i felt the chain ID was more specific and in talking about the connection this felt more appropriate.

faboweb: Yeah I was getting at maybe we should show the title. We wanted to make the chain_id transparent for the user.

jbibla: we can do that. i'm happy with chain-id but title is ok too i think. this means we have to query networks from the connected component, right?

faboweb: yep, check `PageNetworks` on how to query the networks. But also feel free to merge this branch and change this in another PR.

PR #3030, opened 2019-10-18
Fabo/signup steps

faboweb:
Closes #ISSUE

**Description:**

Thank you! 🚀

---

For contributor:

For reviewer:


faboweb:
@mariopino @colw do you have comments on the approach?

jbibla:
very nice!

src/components/common/StepsForm.vue
```diff
@@ -0,0 +1,89 @@
+<template>
```
faboweb: This is the important component

src/components/common/TmSessionSignUp.vue
```diff
@@ -1,155 +1,142 @@
...
+      <div v-else>
+        <div v-if="!session.insecureMode" class="session-main">
+          <h2 class="session-title">
+            Create a new address
+          </h2>
+          <InsecureModeWarning />
+        </div>
+        <div v-else>
+          <div class="session-main">
+            <StepsForm :steps="steps" @done="success" ref="steps">
```
faboweb: Here is an example implementation

src/components/common/TmSessionSignUp.vue
```diff
@@ -1,155 +1,142 @@
...
+              <div slot="name">
```
faboweb: Every step put into a slot with the name of the step

src/components/common/TmSessionSignUp.vue
```diff
@@ -214,36 +210,48 @@ export default {
       }
     }
   },
+  created() {
+    this.steps = [
```
faboweb: This defines the steps order and their slot names
It also defines the validation per step to do.

faboweb: Nice thing is, this parent component still keeps the whole form and data in one place and can handle the whole data.

mariopino:
Straightforward, I like it! We should change then create and recover flows to use the new component ... arg! :-)

jbibla:
our new onboarding flow may not require so many steps so maybe we could close for now. does this still apply / would it be helpful for the action modals?

faboweb:
won't there always be a "create account" and "import account" with at least 3 steps each?

jbibla:
> won't there always be a "create account" and "import account" with at least 3 steps each?

probably. but since we're redesigning we might as well wait and see what we come up with.

jbibla:
fabo is waiting for time on an airplane ✈️

PR #3041, opened 2019-10-21
WIP Colw/api action modals

colw:
Closes #ISSUE

**Description:**

Thank you! 🚀

---

For contributor:

- [ ] Added changes entries. Run `yarn changelog` for a guided process.
- [ ] Reviewed `Files changed` in the github PR explorer
- [ ] Attach screenshots of the UI components on the PR description (if applicable)
- [ ] Scope of work approved for big PRs

For reviewer:

- [ ] Manually tested the changes on the UI

PR #3049, opened 2019-10-23
Case insensitive search

colw:
- Name for clarity
- Do not display loading indicators

Closes #ISSUE

**Description:**

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/src/components/staking/PageValidators.vue
+++ b/src/components/staking/PageValidators.vue
@@ -1,9 +1,6 @@
 <template>
   <TmPage
     :managed="true"
-    :loading="$apollo.queries.validators.loading"
-    :loaded="!$apollo.queries.validators.loading"
-    :error="$apollo.queries.validators.error"
     :data-empty="validators && validators.length === 0"
     hide-header
   >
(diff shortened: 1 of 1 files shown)


src/components/staking/PageValidators.vue
```diff
@@ -1,9 +1,6 @@
 <template>
   <TmPage
     :managed="true"
-    :loading="$apollo.queries.validators.loading"
-    :loaded="!$apollo.queries.validators.loading"
```
faboweb: if this property is set to true, the loader doesn't show anymore. I would recommend on first update to set a `loaded` property to true instead of removing the whole logic.

src/components/staking/PageValidators.vue
```diff
@@ -1,9 +1,6 @@
...
-    :error="$apollo.queries.validators.error"
```
faboweb: Note: `$apollo.queries.validators.error` actually doesn't exist. This needs to be extracted from the result.

```
result(data) {
  this.error = data.error
}

colw: You're right. Not sure where I got that from in the docs.

True, either that or an error method.

colw:
API was updated, hence easier to Include in #3041

PR #3057, opened 2019-10-25
Atharva/fixes #2966 `url for "activity" still points to "transactions"`

### LINKED ISSUE #2966: url for "activity" still points to "transactions"
**Describe the bug**

The URL for the activity page is accessed via `/transactions`.

**Expected behavior**

- redirect `/transactions` to `/activity`
- add default route / URL for Activity to be `/activity`

atharva3010:
Closes #2966 

**Description:**

- redirects `/transactions` to `/activity`  
- added default route / URL for Activity to be `/activity`

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/changes/route-change
+++ b/changes/route-change
@@ -1,0 +1,1 @@
+[Changed] [##2966](https://github.com/cosmos/lunie/issues/#2966) /transaction is changed to /activity @atharva3010
--- a/src/components/common/AppMenu.vue
+++ b/src/components/common/AppMenu.vue
@@ -57,7 +57,7 @@
 
       <router-link
         class="app-menu-item hide-xs"
-        to="/transactions"
+        to="/activity"
         exact="exact"
         title="Transactions"
         @click.native="close"
--- a/src/components/common/MobileMenu.vue
+++ b/src/components/common/MobileMenu.vue
@@ -25,7 +25,7 @@
     </router-link>
     <router-link
       class="mobile-menu-item"
-      to="/transactions"
+      to="/activity"
       exact="exact"
       title="Transactions"
     >
--- a/src/components/common/Page404.vue
+++ b/src/components/common/Page404.vue
@@ -18,7 +18,7 @@
           </router-link>
         </li>
         <li>
-          <router-link to="/transactions">
+          <router-link to="/activity">
             Transactions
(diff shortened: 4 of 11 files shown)


faboweb:
Instead of redirecting the `/transactions` route to `/activity` we should change all occurences of `/transactions` so the code stays clean.

faboweb:
And you need to run `yarn changelog` to document your changes ;)

atharva3010:
Sure! I’ll do that.

atharva3010:
I was a bit confused as the issue #2966 clearly specified for the redirect so, I assumed that modifying the occurrences wasn’t needed. No worries!

One question, will the redirect still be needed? Even after replacing all occurrences of `/transactions` with `/activity`? Are there any external links which may point to `/transactions`?

Do we also need to change the component name too? Currently it’s `PageTransactions`

atharva3010:
@faboweb please check now

jbibla:
a redirect from transactions is a good idea @faboweb - there may be bookmarks and links. i'd rather we don't break the links.

jbibla:
can we avoid the commits that are just spacing changes (from a linter)?

atharva3010:
@jbibla
> can we avoid the commits that are just spacing changes (from a linter)?

We can certainly avoid them, but I think we should keep code linted and formatted.

jbibla:
@atharva3010 - i'm in favour of linting! just want to be sure our lint config is working as expected and these changes won't be reverted by someone else who has a different lint config.

@faboweb shouldn't all the files be linted already? if not, can we lint 'em all?

atharva3010:
@jbibla oh okay! No worries. I don’t have a global lint config, I have linted the code with the lunie workspace config.

faboweb:
I could imagine we don't have an explicit rule for some spacings. Let's add the rule to avoid those irrelevant changes.

faboweb:
Did you consider the URLs in `test/e2e/`? (the redirect would still work I guess, but would be cleaner to change them as well)

atharva3010:
@faboweb I guess I haven’t changed anything in the tests, I will change them too :)

jbibla:
yep! let's get this merged!

PR #3066, opened 2019-10-30
[Simsala] automatic release created for v1.0.0-beta.133

faboweb:
Please manually test before merging this to master

### Repository

- always set the encryption disclaimer for submitting the app to the ios app store @faboweb


jbibla:
again? 

i don't think we should be incrementing the version without actual changes.

faboweb:
The changelog file wasn't removed in the last update 🤷‍♂

PR #3098, opened 2019-11-07
Fix bug in Proposals

### LINKED ISSUE #3094: Bug in Proposals page
**Describe the bug**

Proposals list doesn't work in feature/api-gateway branch caused by PR #3088 

**To Reproduce**

1. Go to Proposals.

**Expected behavior**

To see the proposals list.

**Additional context**

This is caused by the modified query for single proposal. We are actually using the same query fragment
to fectch proposals list and a single proposal.

We only need these fields in proposals list query:

- id
- type
- title
- description
- status

We need a specific query with that fields for proposals list, and include it in `ProposalsPage.vue` for easy reading.

mariopino:
Closes #3094 

**Description:**

Fix bug in Proposals.

Thank you! 🚀

---

For contributor:

For reviewer:


src/components/governance/PageProposals.vue
```diff
@@ -87,7 +88,17 @@ export default {
     proposals: {
       query() {
         /* istanbul ignore next */
-        return ProposalList(this.network)
+        return gql`
+          query proposals {
+            proposals(networkId: "${this.network}") {
```
colw: I believe this will not update if the networkId changes.

To fix, the query needs to be a function. You can see other examples in the PageValidators page for example.

It could be that string interpolation (` "${this.network}"`) is best/easier, but for consistency le's use a `variables` property in the `apollo` object.

Issue #3048, opened 2019-10-22
Add Sentry to Frontend


faboweb:

atharva3010:
I would like to pick this up @faboweb   
Can you give me a bit more information if you need anything other than the basic setup?

faboweb:
I guess the basic setup is ok (all global errors should be catched and logged). We need to deploy the source maps to Sentry for all `master` deploys. Take a shot and let's see how it goes. @iambeone is also working on Sentry this week. Though ask here if anything is missing.

jbibla:
@iambeone can you confirm this is working as expected? i think there is an issue blocking this.

PR #3110, opened 2019-11-09
[Simsala] automatic release created for v1.0.0-beta.136

faboweb:
Please manually test before merging this to master

### Repository

- [#3048](https://github.com/cosmos/lunie/issues/3048) Sentry integration was added @iambeone


jbibla:
no actual changes here... simsala is broken 😭 ?

faboweb:
It's not simsala. It's the flow that the mobile release branches merge things back to develop. I need to change this so that they also run on a release but only if we update the version of the app.

PR #3111, opened 2019-11-10
Fabo/force network loading for refetch queries

faboweb:
Fixes the issue that only stale data shows after a tx

### DIFF
--- a/src/components/common/TmBalance.vue
+++ b/src/components/common/TmBalance.vue
@@ -42,6 +42,7 @@
 </template>
 <script>
 import { shortDecimals } from "scripts/num"
+import refetch from "scripts/apollo-refetch"
 import { noBlanks } from "src/filters"
 import TmBtn from "common/TmBtn"
 import SendModal from "src/ActionModal/components/SendModal"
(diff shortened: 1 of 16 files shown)


faboweb:
@colw feel free to merge right away

faboweb:
assumes https://github.com/luniehq/lunie-api/pull/85

src/components/common/TmBalance.vue
```diff
@@ -150,7 +145,7 @@ export default {
         query: UserTransactionAdded,
         result() {
           // query if successful or not as even an unsuccessful tx costs fees
-          this.$apollo.queries.overview.refetch()
+          refetch(this.$apollo.query.overview)
```
colw: I would suggest using the Vue Apollo API to just set the options: https://apollo.vuejs.org/api/smart-query.html#setoptions

(If we keep this it would be better to rename it to `refetchNetworkOnly`)

faboweb: I went the same path before. But I think we want to query the cache and only want to force an update manually when we know that it changed, right?

faboweb: Feel free to play with just setting the option though. Wanted to assist you on the problem only.

colw: Sure, I guess in the case of balances we would almost always want to query the network though?

faboweb: I would guess almost never. Only if the user does a transaction, right?

faboweb: Rewards: Always.

PR #3120, opened 2019-11-12
Fix scrolling behaviour

### LINKED ISSUE #3119: Scroll should be at the top when you navigate to a new page.
**Describe the bug**

When open a large page that is taller than browser viewport, scroll down the page, open another large page throught menu, you will find that the scroll is at the bottom of the new page.

**To Reproduce**
1. Open Lunie, and resize the window so it's less than 1024 px width.
2. Open top menu and click on Security.
3. Scroll down the page
4. Open top menu and click on Careers.
5. The Careers page have the scroll at the bottom of the page.

**Expected behavior**

Scroll should be at the top when you navigate to a new page.

**Additional context**

We should reset scrolling when detecting a route change.

mariopino:
Closes #3119

**Description:**

Fixes scrolling behaviour. Maybe there is a cleaner way to do it, feedback welcome!

Thank you! 🚀

---

For contributor:

For reviewer:


src/router.js
```diff
@@ -3,6 +3,9 @@ import routes from "./routes"
 import { NetworkCapability, NetworkCapabilityResult } from "./gql"
 import Vue from "vue"
 
+// Reset scroll
+Vue.prototype.$scrollToTop = () => window.scrollTo(0, 0)
```
faboweb: Why do you add it to the prototype? You can just add it as a method to `AppMenu`

mariopino: Totally right!

jbibla:
i don't think this is an issue that should be fixed in the menu - it should have something to do with vue router.

mariopino:
> i don't think this is an issue that should be fixed in the menu - it should have something to do with vue router.

Yeah, that's the way we handle actually, I used the same code in other Vue apps and it works perfect.

But the problem is that current code doesn't work in Lunie! Seems to be an css related issue (some css property is preventing scrolling to top in route changes but I was not able to find it).

That's why I would choose to implement a working solution instead of spent more time debugging it.

src/components/common/AppMenu.vue
```diff
@@ -24,7 +24,10 @@
         to="/portfolio"
         exact="exact"
         title="Portfolio"
-        @click.native="close"
+        @click.native="
+          close()
+          scrollToTop()
```
faboweb: I think you are missing a ;

faboweb: Note: instead of calling both functions here you could also create a "handleMenuClick" function who executes both functions to avoid the duplications.
Or just add the logic to the close function

jbibla:
```
scrollBehavior (to, from, savedPosition) {
  return { y: 0 }
}
```
it does work actually - just not when the menu is open and you scroll behind it. this isn't working when the menu is open because the scroll is still happening underneath the menu.

what i would propose is using a vuex module to set `position: fixed` to the body element when the action modal or mobile menu is open. this will not allow scrolling in the background.

jbibla:
hmmm. it seems to work for main routes but not for supporting pages.

it does work actually - just not when the menu is open and you scroll behind it. this isn't working when the menu is open because the scroll is still happening underneath the menu.

we're using `noScroll` as well - this might have something to do with it...?

should we set `position: fixed` to the body element when the action modal or mobile menu is open? with a vuex module maybe?

mariopino:
> we're using `noScroll` as well - this might have something to do with it...? 

Nope, I tested without it with same results.

A curious thing about noScroll is that we are not enabling it in any place with `noScroll.on();` (https://www.npmjs.com/package/no-scroll), so I think that in the way we are using it doesn't have any effect.

> should we set `position: fixed` to the body element when the action modal or mobile menu is open? with a vuex module maybe?

Are there any problem with scrolling when you open the action mobile or the mobile menu?

PR #3121, opened 2019-11-12
Fabo/new app release flow

faboweb:
Bitrise should now only release apps for every push to master if certain files changed:

For iOS: `ios/App/App/Info.plist`
For Android: `android/app/build.gradle`

Both hold the version. So when the version is updated. We release an app.


jbibla:
how do we test builds then?

faboweb:
We could just test before merging to master so on every release

jbibla:
i don't know what you mean. the changes here look like they will not make it possible to trigger a build in bitrise to test new changes.

faboweb:
> i don't know what you mean. the changes here look like they will not make it possible to trigger a build in bitrise to test new changes.

Note: We should not just trigger a build. this would distribute the app live.

What do you want to achieve? Testing if a build of the app works on your phone?

jbibla:
> Testing if a build of the app works on your phone?

yes

---

my understanding is that if we merge this change then the only way to get builds into testflight will be to merge to master. this is great. but for development purposes, we should leave an option to push a version manually to testflight / the app stores. (i think)

faboweb:
> we should leave an option to push a version manually to testflight / the app stores. (i think)

ok we can do that out of bound. we need a different workflow in bitrise then anyway. we will switch the current workflow to publish to live automatically I guess? We can have a second workflow which publishes to testflight? Or we automatically publish to testflight always and we release manually to the live audience.

jbibla:
i think checking to make sure mobile changes actually happened is a great idea on any branch.

ideally, i think the following flow makes the most sense:
```
develop -> test version
master -> iOS & Android stores
```
apparently bitrise can handle this for us - https://www.bitrise.io/features/cross-platform-features

> Deploy to testers
If you are looking for an easy way to send your app to testers, you can use Bitrise deploy. This will send a link to specific email addresses so that your testers can install the .IPA.

> Send straight to the store
You can automatically send your apps to Google Play or iTunes Connect.

faboweb:
Does your proposal include the manual update of the version or do you propose releasing an app on every merge to master? This still would mean about 1 new app version per day

colw:
Just an FYI: I believe `agvtool` will auto bump numbers for iOS. I'm not sure about Android.

Also, does Bitrise do this automatically? https://devcenter.bitrise.io/builds/build-numbering-and-app-versioning/

jbibla:
> This still would mean about 1 new app version per day

if there are changes to mobile, we should increment the version and release with the nightly release. so this should not be every day. (if i understand you correctly)

faboweb:
let's have a call around this

PR #3123, opened 2019-11-12
colw/Loading state for single validator

colw:
Closes #ISSUE

**Description:**

Thank you! 🚀

---

For contributor:

For reviewer:


src/components/staking/PageValidator.vue
```diff
@@ -2,6 +2,8 @@
   <TmPage
     :managed="true"
     :data-empty="!validator.operatorAddress"
+    :loading="$apollo.loading"
+    :loaded="!$apollo.loading"
```
faboweb: that doesn't work. this needs to be a value set after the initial load. i.e. in the `result` hook of the `validator` query.

faboweb: Currently it defeats the purpose of `loaded` as it is set to `false` on ever new loading process.

colw: It appears to work fine for me. Also when the validator address is unknown.

faboweb: What happens if you delegate to the validator and there is a refetch of the rewards?

colw: I see, a new transaction in the subscription would cause it to show the loading component.

faboweb: 💯

src/components/staking/PageValidator.vue
```diff
@@ -309,6 +312,7 @@ export default {
         }
       },
       update: result => {
+
```
jbibla: this new line is not necessary

how is it that these types of lint things are getting through?

src/components/staking/PageValidator.vue
```diff
@@ -325,6 +329,10 @@ export default {
           ...result.validator,
           statusDetailed: getStatusText(result.validator.statusDetailed)
         }
+      },
+      result(queryResult) {
+        console.log(queryResult)
```
jbibla: console not needed 😄

jbibla:
unit tests are fixed by adding
```
      if (this.network.testnet) {
        return true
      }
```
to `checkFeatureAvailable()` in `ActionModal.vue`

PR #3125, opened 2019-11-12
Jordan/3077 mobile form improvements (session and action modal)

jbibla:
closes #3077
closes #3101

**Description:**
- fixed button to bottom on forms
- got rid of console error for status bar #3101
- fix validator warning color
- stop background scrolling in action modal
- added mobile config to session module
- cleaning up session and seed screens

Thank you! 🚀

---

For contributor:

- [ ] Added changes entries. Run `yarn changelog` for a guided process.
- [ ] Reviewed `Files changed` in the github PR explorer
- [ ] Attach screenshots of the UI components on the PR description (if applicable)
- [ ] Scope of work approved for big PRs

For reviewer:

- [ ] Manually tested the changes on the UI

PR #3147, opened 2019-11-15
Fix insecure mode

### LINKED ISSUE #3146: Insecure mode bug
**Describe the bug**
Setting environment `MOBILE_APP=true` or passing the url param `?insecure=true` should set session.insecureMode to true, but it doesn't.

**To Reproduce**

1. Execute `yarn serve:mobile`
2. Open Lunie in the browser
3. Go to Sign In
4. Go to Create a new address
5. Insecure warn is showed

**Expected behavior**
You should be able to create a new address without the warn.

mariopino:
Closes #3146 

**Description:**

Fix insecure mode handling.

Thank you! 🚀

---

For contributor:

For reviewer:


src/initializeApp.js
```diff
@@ -21,7 +21,8 @@ function setOptions(urlParams, store) {
     store.dispatch(`setNetwork`, { id: urlParams.network })
   }
 }
-
+`
```
faboweb: I think you added characters here

mariopino: > I think you added characters here

Yes, it was fixed 2 seconds later :-)

package.json
```diff
@@ -49,6 +49,7 @@
     "cross-spawn": "7.0.1",
     "crypto-js": "^3.1.9-1",
     "graphql": "14.4.2",
+    "https-proxy-agent": ">=3.0.0",
```
jbibla: what does this have to do with insecure mode? maybe remove this?

jbibla:
it seems like this is a good fix for insecure mode, but to be clear i don't think we want to use insecure mode on mobile. we should keep insecure mode (or is it called experimental now?) for insecure / experimental things and we should use the mobile env variable for mobile specific things. this is part of what i tried to change in #3125

mariopino:
> it seems like this is a good fix for insecure mode, but to be clear i don't think we want to use insecure mode on mobile. we should keep insecure mode (or is it called experimental now?) for insecure / experimental things and we should use the mobile env variable for mobile specific things. this is part of what i tried to change in #3125

Thanks for re-clarification Jordan :-), I thought the opposite idea because that was the behaviour in feature/api-gateway.

package.json
mariopino: Not sure why this has been end here :-/, but this module it's not currently used in develop so it's safe to remove it.

PR #3156, opened 2019-11-17
ana/2982 Silently sign user out when switching networks

### LINKED ISSUE #2982: switching networks should sign the user out
- if we don't sign users out there will be strange things happening with the data
- if users are signed in when switching we should confirm they want to sign out before it happens automatically

Bitcoinera:
Closes #2982

**Description:**

Every time the user switches networks they will get automatically signed out.
This happens without the user noticing.

Thank you! 🚀

---

For contributor:

For reviewer:


src/vuex/modules/session.js
```diff
@@ -160,6 +160,7 @@ export default ({ apollo }) => {
 
       dispatch(`resetSessionData`)
       commit(`setSignIn`, false)
+      console.info(`Signing user out...`)
```
faboweb: Hmm can we show a notification? This feels weird if it is silent.

faboweb: Or we do ask with a prompt:

```
By switching the network you will be signed out of your address `cosmos1234`.

Switch Network / Cancel
```

faboweb: Checkout `ActionModal.js` line 521 on how to do a prompt

jbibla: yep - this is a good solution! will alerts work on mobile?

colw: Possible alternative: A warning message above the network list.

faboweb: we need to test it, but I assume it works also on mobile.

faboweb: > Possible alternative: A warning message above the network list.

Could also be an option.

Bitcoinera: I have gone down the modal route. But I think it would be cool to use a custom modal to make it look fancier & also more Lunie-like

src/components/network/NetworkList.vue
```diff
@@ -32,9 +32,21 @@ export default {
...
     selectNetworkHandler(network) {
-      if (this.connection.network !== network.id) {
+      let confirm = this.confirmModalOpen()
+      if (this.connection.network !== network.id && confirm) {
         this.$store.dispatch(`setNetwork`, network)
       }
+    },
+    confirmModalOpen() {
+      let confirm = false
+      if (localStorage.getItem(`session`)) {
```
faboweb: how about checking the vuex getter `address` or `session.signedIn`? Check `DelegationModal.vue` for example

src/vuex/modules/session.js
Bitcoinera: Do you think doing a complete ActionModal for this modal is worth it? Something like this:

image

faboweb: hmmm sounds like a little too much effort for the first step, if I am allowed to say that. Especially as we want to have a slightly different account flow soon.

tests/unit/specs/components/network/NetworkList.spec.js
```diff
@@ -68,4 +71,14 @@ describe(`NetworkList`, () => {
     wrapper.find(".select-network-item.selected").trigger("click")
     expect(wrapper.vm.$store.dispatch).not.toHaveBeenCalledWith()
   })
+
+  it(`shows windows about switching networks`, async () => {
+    wrapper.setData({
```
faboweb: This test is not doing what it should. I guess you want to mock the window object and check if the confirmation got triggered.

tests/unit/specs/components/network/NetworkList.spec.js
```diff
@@ -76,9 +76,11 @@ describe(`NetworkList`, () => {
     wrapper.setData({
       session: {
         signedIn: true,
-        windowsDevice: true
       }
     })
-    expect(wrapper.element).toMatchSnapshot()
+    const spy = jest.spyOn(window, `confirm`).mockImplementationOnce(() => {})
```
faboweb: You could in theory also test the result:
```
.mockImplementationOnce(() => true)
.mockImplementationOnce(() => false)
```

but it is ok for now

PR #3158, opened 2019-11-17
Hide tooltips on mobile app

### LINKED ISSUE #3115: hide tooltips on mobile

- the tooltips don't display nicely on mobile. would be best to not use them in the native apps.

- disable them / don't allow them to render based on `MOBILE_APP` flag

- perhaps there is some way they won't even be loaded (not just hidden)

**Additional context**

mariopino:
Closes #3115 

**Description:**

Hide tooltips on mobile app.

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/changes/mario_3115-hide-tooltips-mobile-app
+++ b/changes/mario_3115-hide-tooltips-mobile-app
@@ -1,0 +1,1 @@
+[Changed] [#3115](https://github.com/cosmos/lunie/issues/3115) Hide tooltips on mobile app @mariopino
--- a/src/main.js
+++ b/src/main.js
@@ -17,7 +17,6 @@ import * as Sentry from "@sentry/browser"
 import * as Integrations from "@sentry/integrations"
 import "material-design-icons-iconfont/dist/material-design-icons.css"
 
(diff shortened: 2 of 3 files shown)


mariopino:
By now implemented using the easy way.

Most optimal solution could be to not  import `vue-directive-tooltip` (and then don't use it with `Vue.use`) based on MOBILE_APP environment variable value.

Buut I think it is difficult to implement that way since tooltips are inline in the html of components using `v-tooltip` directive.

That would require to move all the tooltips from the components (content and other attributes like placement ... etc) to custom data directives (like data-tooltip, data-tooltip-placement... etc) and then implement a custom tooltip library to use that data attributes to render the tooltips.

Given that then we can conditionally import our tooltip library based on  the value of MOBILE_APP environment variable.

faboweb:
Tip:

Maybe we can import a fake tooltip directive. This is how I do it in tests (but for directives):

```
Tooltip = config.mobileApp ? () => {} : Tooltip
Vue.use(Tooltip)
```

But good for now

PR #3165, opened 2019-11-18
Fabo/remove stargate

faboweb:
Removes mentions of stargate as it is not used anymore. Only gets received from graphql API


faboweb:
@colw on a hunch checked this, and this is already obsolete

jbibla:
woah! big step!

PR #3180, opened 2019-11-22
Colw/transaction service

faboweb:
**Description:**

~Requires updated Lunie-API that fetches accountNum and sequence.~ Updated.

Set `VUE_APP_ENABLE_TX_API` to `true` to enable.

Thank you! 🚀

---

For contributor:

For reviewer:


src/ActionModal/utils/ActionManager.js
```diff
@@ -12,17 +12,14 @@ const txFetchOptions = {
...
-
 async function transactionAPIRequest(payload) {
   // console.log(config)
   const options = {
     ...txFetchOptions,
     body: JSON.stringify({ payload })
   }
 
-  return fetch(`${config.graphqlHost}/transaction`, options).then(r => r.json())
+  return fetch(`${config.graphqlHost}/transaction/estimate`, options).then(r => r.json())
```
faboweb: we have axios as a dependency. would you rather prefer using fetch?

colw: I would personally. It's supported in all modern browsers.

The reason I used it here was just convenience though to get something running.

src/ActionModal/utils/MessageConstructor.js
```diff
@@ -46,3 +61,167 @@ export const getMultiMessage = async (context, messages) => {
     }
   }
 }
+
+// Bank
+export function MsgSend(
```
faboweb: Better export these functions from the library instead?

colw: Agreed, but they are not currently exported from the library. I can do this at a later time.

faboweb:
@colw maybe add some coverage to ActionManager?

src/ActionModal/components/ActionModal.vue
```diff
@@ -618,7 +619,16 @@ export default {
       const { type, memo, ...properties } = this.transactionData
       await this.actionManager.setMessage(type, properties)
       try {
-        this.gasEstimate = await this.actionManager.simulate(memo)
+        if (!config.enableTxAPI) {
+          this.gasEstimate = await this.actionManager.simulate(memo)
+        } else {
+          this.gasEstimate = await this.actionManager.simulateTxAPI(
+            this.createContext(),
```
jbibla: what does this do @colw?

jbibla:
action modal is so large!

src/ActionModal/components/ActionModal.vue
colw: The old functionality is still there, and you can opt-in to the transaction service by setting an environment variable when building the front end.

faboweb: he meant the context maybe. that is like a bunch of variables you need for the tx to be build

colw:
> action modal is so large!

What do you mean?

src/ActionModal/components/ActionModal.vue
jbibla: yep, that's what i meant. but i don't see the function anywhere.

jbibla:
> What do you mean? Or rather, in which way?

so many lines!

src/ActionModal/components/ActionModal.vue
colw: It's around, line ~500. Not part of this PR though.

PR #3181, opened 2019-11-22
Transaction styling

### LINKED ISSUE #3078: improve transaction styling

- the transaction list appears cluttered and at times hard to read (mobile especially, but desktop as well

- simplify the tx component for all tx types
- simple icon on left (new icons coming)
- type of transaction on left (Send, Withdrawal, Restaked)
- amount on right
- more details dropdown on click or tap

Screen Shot 2019-11-01 at 12 37 06 PM

**Additional context**

Screen Shot 2019-11-01 at 12 12 49 PM
Screen Shot 2019-11-01 at 12 12 28 PM

---
feedback welcome
@faboweb @mariopino @colw @fedekunze

mariopino:
Closes #3078 

**Description:**

Implement the new transaction page design:

image

Bonus: Add support for multisend transactions.

Thank you! 🚀

---

For contributor:

For reviewer:


jbibla:
really nice work mario! 

- let's add a slight animation / transition to the extra info box
- the background color of the tx should change on hover (like with li-validators)
- we have new icons we can use (but i don't think fabo likes them 🤭)
- there are a number of alignment and spacing issues that need to be addressed - and many places where there are extra spaces or not enough spaces

Screen Shot 2019-11-22 at 11 50 54 AM

mariopino:
Thanks! I saw that there is room to improve in the mobile view :-D The transition was on my plans, for sure and will follow the rest of your suggestions.

jbibla:
@fedekunze i know this PR is close to your heart

faboweb:
Improvement idea: Use icons for the meta information.
Currently the information looks very noisy.
Also: It necessary to hide the transaction details like the recipient validator. I think they are very important. @jbibla

mariopino:
> Also: It necessary to hide the transaction details like the recipient validator. I think they are very important. @jbibla

What do you mean?

src/components/transactions/TransactionIcon.vue
```diff
@@ -15,6 +11,47 @@ export default {
...
       required: true
+    },
+    transactionType: {
+      type: String,
+      required: true
+    }
+  },
+  computed: {
+    icon() {
+      if (this.transactionType === `Sent`) {
```
faboweb: do you know ob the object switch approach?

```
const dictionary = {
  keyA: value,
  keyB: value
}

const valueForKey = dictionary[key]
```

mariopino: Definitely more cleaner but as I need to make some comparisons for Withdrawal, Vote and Deposit types like: 

`} else if (this.transactionType.search(`Voted`) === 0) {`

It's possible to do that in a dictionary?

Another option is just to simplify the caption for that messages and include the info in the tx detail,  the we can use a dictionary. Maybe a better solution.

mariopino:
Hey @jbibla, we have some missing icons for:

- Create validator
- Edit validator
- Submitted proposal
- Unjail

mariopino:
> Hey @jbibla, we have some missing icons for:
> 
> * Create validator
> * Edit validator
> * Submitted proposal
> * Unjail

Also having a generic icon could be useful, just to have a fallback.

src/components/transactions/TransactionIcon.vue
faboweb: Looks funky to me. What are the options for `transactionType` that there are multiple with `Voted` in it?

faboweb: yeah I think you are on the right track. there should be an enum "transactionType" that should be different then the title.

mariopino: I finally implemented that way. The icon name should be the caption string, so that way we don't need a dictionary, only need to handle `Update withdraw address`, which is the only specific case. A bit hacky but simplier :-)

```javascript
computed: {
    icon() {
      if (this.transactionType.indexOf(` `) !== -1) {
        if (this.transactionType === `Update withdraw address`) {
          return `/img/icons/activity/Withdrawal.svg`
        } else {
          return `/img/icons/activity/${this.transactionType.substr(
            0,
            this.transactionType.indexOf(` `)
          )}.svg`
        }
      } else {
        return `/img/icons/activity/${this.transactionType}.svg`
      }
    }
  }
```

tests/unit/specs/components/transactions/TransactionList.spec.js
```diff
@@ -12,7 +12,7 @@ describe(`TransactionList`, () => {
   let height = 1086769
   let oneDay = 86400 // One day
   let offset = 86400 / 7 // One day / 7s block time
-  let timestamp = new Date().getTime() // Today
+  let timestamp = new Date('2019-11-27T04:41:20') // cannot do new Date().getTime() because of snapshots
```
faboweb: @Bitcoinera `new Date()` is fixed to a certain time in `fixed_time.js`

src/components/transactions/TransactionIcon.vue
faboweb: I am still worried, that the transactionType is the title of the transaction. If we change the title now, we need to change the name of the icon. feels like an unnecessary linkage. But we can improve later.

tests/unit/specs/components/transactions/TransactionList.spec.js
mariopino: BTW Would be nice to set a different date in fixed_time.js to allow susbtract at least 1 day, 1 month and 1 year to allow categorization by date as in transactions page.

mariopino:
TODO: Simplify TransactionMetadata component to accept one prop (tx), I think does not make sense to assign the tx properties individually, that way we can work with properties that maybe not defined, like `liquidDate`.

And of course fix our beloved e2e tests!

tests/unit/specs/components/transactions/TransactionList.spec.js
Bitcoinera: Oh, I see. Good to know :ok_hand:

Bitcoinera:
I noticed that there is one type of transaction we don't really need, since it is always happening at the same time than another, the `WithdrawDelegatorReward`.

I believe we should get rid of it both in the frontend and in the API, since it is not an independent type by itself and doesn't add any information. Here you can check it: https://stargate.cosmos.network/txs?action=withdraw_validator_rewards_all

image

jbibla:
`WithdrawValidatorCommission` is a tx that validators make to claim their commissions. it would appear separately in the block explorer but users would not be able to see this on their activity page. 

i wouldn't worry about it for now.

jbibla:
Screen Shot 2019-11-29 at 4 27 38 PM
Screen Shot 2019-11-29 at 4 27 54 PM

jbibla:
this is done IMO and should be merged as soon as possible. 

i think @mariopino and @Bitcoinera encountered a number of issues while working on this code — please make sure you open issues so we can fix these problems in the future!

mariopino:
Looks great! New format is missing in multisend though, you can check it signing with `cosmos1z8mzakma7vnaajysmtkwt4wgjqr2m84tzvyfkz`

faboweb:
Great seeing Jordan helping out the team here!!

src/components/transactions/TransactionIcon.vue
```diff
@@ -15,40 +11,45 @@ export default {
...
       required: true
+    },
+    transactionType: {
+      type: String,
+      required: true
+    }
+  },
+  computed: {
+    icon() {
+      if (this.transactionType.indexOf(` `) !== -1) {
```
faboweb: I still have concerns around these string searches in comparison to some enum. Also `transactionType` is used here as `title` which I think is a bad naming.

src/scripts/transaction-utils.js
```diff
@@ -58,8 +58,25 @@ export const isPendingUndelegation = tx =>
 
 export const getCoin = transaction => {
   if (Array.isArray(transaction.value.amount)) {
+    // console.log(`getCoin:`, transaction.value.amount[0])
```
faboweb:
```suggestion
```

src/scripts/transaction-utils.js
```diff
@@ -58,8 +58,25 @@ export const isPendingUndelegation = tx =>
...
     return transaction.value.amount[0]
   } else {
     return transaction.value.amount
   }
 }
+
+export const getMultiSendCoin = (transaction, sessionAddress) => {
```
faboweb: Misses a comment, why this is necessary

src/components/transactions/TransactionIcon.vue
mariopino: True! let's do an small refactor (no new features, I promise! ;-)).

jbibla: @faboweb i think it should be in a new PR!

src/scripts/transaction-utils.js
```diff
@@ -63,3 +63,20 @@ export const getCoin = transaction => {
     return transaction.value.amount
   }
 }
+
+// Required by cosmos-sdk/MsgMultiSend txs to fetch coins from transaction inputs[] or outputs[] arrays
```
faboweb: This states what it does not why.

```suggestion
// We currently don't support MultiCoin transactions in the design. For simplicity we display only the first outgoing or incomming denomination.
```

src/components/transactions/TransactionItem.vue
```diff
@@ -1,24 +1,63 @@
...
+        <TransactionMetadata v-if="showMetaData" :transaction="transaction" />
+      </div>
+    </transition>
   </div>
 </template>
 
 <script>
-import TransactionIcon from "./TransactionIcon"
-import TransactionDetails from "./TransactionDetails"
+import { messageType } from "./messageTypes.js"
```
faboweb: here is the messageType enum I would have liked to see in the icon handling as well. no need for a separate form of declaring the types.

faboweb: But we can do that as a follow up.

PR #3191, opened 2019-11-24
Jordan/allow for async loading of vue components

jbibla:
Closes #ISSUE

**Description:**

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/src/ActionModal/components/ActionModal.vue
+++ b/src/ActionModal/components/ActionModal.vue
@@ -301,7 +301,7 @@ import { track } from "scripts/google-analytics"
 import { UserTransactionAdded } from "src/gql"
 import config from "src/../config"
 
-const ActionManager = () => import("../utils/ActionManager")
+const ActionManager = () => import(/* webpackChunkName: "ActionManager" */ "../utils/ActionManager")
 
 const defaultStep = `details`
 const feeStep = `fees`
--- a/src/ActionModal/components/ActionModal.vue
+++ b/src/ActionModal/components/ActionModal.vue
@@ -301,7 +301,7 @@ import { track } from "scripts/google-analytics"
 import { UserTransactionAdded } from "src/gql"
 import config from "src/../config"
 
-import ActionManager from "../utils/ActionManager"
+const ActionManager = () => import("../utils/ActionManager")
 
 const defaultStep = `details`
 const feeStep = `fees`
--- a/src/ActionModal/components/ActionModal.vue
+++ b/src/ActionModal/components/ActionModal.vue
@@ -300,8 +300,7 @@ import { between, requiredIf } from "vuelidate/lib/validators"
 import { track } from "scripts/google-analytics"
 import { UserTransactionAdded } from "src/gql"
 import config from "src/../config"
-
-const ActionManager = () => import(/* webpackChunkName: "ActionManager" */ "../utils/ActionManager")
+import ActionManager from "../utils/ActionManager"
 
 const defaultStep = `details`
 const feeStep = `fees`
--- a/src/ActionModal/components/ActionModal.vue
+++ b/src/ActionModal/components/ActionModal.vue
@@ -284,15 +284,16 @@
 <script>
 import gql from "graphql-tag"
 import noScroll from "no-scroll"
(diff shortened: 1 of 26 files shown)


faboweb:
Interesting! Does this bring a performance improvement?

jbibla:
> Does this bring a performance improvement?

yes! that's the idea! instead of downloading the whole bundle on first visit the user only downloads the js files for the components that are needed (code-splitting). webpack will only do this if we load components this way.

faboweb:
I meant, did you measure the performance improvement?

jbibla:
> did you measure the performance improvement?

yes, i'm trying to find the optimal setup.

one major issue is how large cosmos-ledger and cosmos-keys are. trying to figure out why these are loading on initialization instead of as needed. but even so, can we reduce the size of these libraries?

src/ActionModal/components/ActionModal.vue
```diff
@@ -301,7 +301,7 @@ import { track } from "scripts/google-analytics"
 import { UserTransactionAdded } from "src/gql"
 import config from "src/../config"
 
-const ActionManager = () => import("../utils/ActionManager")
+const ActionManager = () => import(/* webpackChunkName: "ActionManager" */ "../utils/ActionManager")
```
faboweb: that looks dirty ^^' is this necessary?

faboweb:
> one major issue is how large cosmos-ledger and cosmos-keys are. trying to figure out why these are loading on initialization instead of as needed. but even so, can we reduce the size of these libraries?

probably we can, but I don't consider this anyway near of our prios. what do you think?

src/components/staking/PageValidators.vue
```diff
@@ -40,11 +40,11 @@
...
-import TmDataLoading from "common/TmDataLoading"
-import TableValidators from "staking/TableValidators"
-import TmPage from "common/TmPage"
-import TmField from "common/TmField"
-import TmBtn from "common/TmBtn"
+const TmDataLoading = () => import("common/TmDataLoading")
+const TableValidators = () => import("staking/TableValidators")
+const TmPage = () => import("common/TmPage")
+const TmField = () => import("common/TmField")
+const TmBtn = () => import("common/TmBtn")
```
colw: This is _probably_ overkill. But curious once we test it anyway.

src/ActionModal/components/ActionModal.vue
jbibla: no just using chunk names for testing and debugging

src/routes.js
```diff
@@ -17,7 +17,7 @@ export default [
     meta: {
       feature: "Proposals"
     },
-    component: require(`./components/governance/PageProposals`).default
+    component: () => import(`./components/governance/PageProposals`)
```
faboweb: This is already loading the pages and there content async

src/components/common/TmBalance.vue
```diff
@@ -41,14 +41,15 @@
...
 import refetchNetworkOnly from "scripts/refetch-network-only"
+import { shortDecimals } from "scripts/num"
 import { noBlanks } from "src/filters"
-import TmBtn from "common/TmBtn"
-import SendModal from "src/ActionModal/components/SendModal"
-import ModalWithdrawRewards from "src/ActionModal/components/ModalWithdrawRewards"
 import { UserTransactionAdded } from "src/gql"
 import { mapGetters } from "vuex"
+const TmBtn = () => import("common/TmBtn")
+const SendModal = () => import("src/ActionModal/components/SendModal")
```
faboweb: let's load modals async and make sure they are not rendered directly (v-if) but let's remove the noise for the other components?

src/routes.js
jbibla: i don't think so. why do you think that?

jbibla:
i'm going to close this in favour of a much cleaner PR #3207 that accomplishes the same goal with only a couple changes. i would propose that in the future someone investigates the following:

- reducing size of cosmos-ledger.js and cosmos-keys.js
- v-if -ing the action modals (which is not trivial at first glance)
- more vue / webpack optimization options

PR #3192, opened 2019-11-24
Jordan/deps

jbibla:
Closes #ISSUE

**Description:**

Thank you! 🚀

---

For contributor:

For reviewer:


vue.config.js
```diff
@@ -32,6 +32,7 @@ module.exports = {
         extensions: [`.js`, `.vue`, `.css`]
       },
       plugins: [
+        new webpack.IgnorePlugin(/^\.\/locale$/, /moment$/),
```
faboweb: why did you add this and what does it do?

jbibla: significantly reduces the size of moment.js 

https://github.com/jmblog/how-to-optimize-momentjs-with-webpack#how-to-optimize-momentjs-with-webpack

faboweb: I just checked and we do not localize dates. Shouldn't we do that? Not everyone

jbibla: yeah, maybe but this has not come up as an issue and we don't really have an internationalization strategy right now. when we do localization we can re-add the locales we need - but for now, I think we should just remove this unused code.

jbibla:
we needed to upgrade cosmos-keys and cosmos-ledger, right?

vue.config.js
colw: This is a good moment to point out that the perfect place to 

> significantly reduces the size of moment.js
> 
> https://github.com/jmblog/how-to-optimize-momentjs-with-webpack#how-to-optimize-momentjs-with-webpack

This is a great moment to point out that information like this belongs in the commit message :)

jbibla:
i haven't tested if the ledger actually works here yet.

PR #3195, opened 2019-11-24
Ana/3173 non existent network not handled correctly

### LINKED ISSUE #3173: Non-existent network not handled correctly by the client side

Whenever you switch working with the local testnet to the staging server, you get lost of errors from the API which are not very clear what they refer to.

The frontend should automatically switch to the first existent network on the list, every time it gets a request for a non-existent network.

A prompt asking you directly to select the network you would like to work with. But I don't think this is necessary.

**Additional context**

image

Bitcoinera:
Closes #3173 

**Description:**

Now when it is initialized the app will directly connect to one of the available networks (right now either `cosmos-hub-mainnet` or `local-cosmos-hub-testnet`.

I had to use a little hack to reorder the array of networks given by our API so `cosmos-hub-mainnet` would be the first element. I wonder if there is a better way to have this element as first.

Thank you! 🚀

---

For contributor:

For reviewer:


Bitcoinera:
It messed up with tests, but fixing it is tricky since it's got a call with axios to API...
I will have another look tomorrow

src/vuex/modules/connection.js
```diff
@@ -17,11 +20,27 @@ export default function() {
...
-        await commit(`setNetworkId`, JSON.parse(network))
-      }
+    async checkForPersistedNetwork({ dispatch, commit }) {
+      const network = JSON.parse(localStorage.getItem(`network`))
+      dispatch(`checkAvailableNetworks`).then( async result => {
+        let availNetworks = result.data.data.networks
+        let availNetworksID = Object.values(availNetworks).map(
+          network => network.id
+        )
+        // Little hack to reorder the array so 'cosmos-hub-mainnet' stays at the beginning
```
faboweb: Instead of a complicated filtering, how about just hardcoding the default network for now?

src/vuex/modules/connection.js
```diff
@@ -20,26 +18,23 @@ export default function() {
   }
 
   const actions = {
-    async checkForPersistedNetwork({ dispatch, commit }) {
+    async checkForPersistedNetwork({ dispatch, commit }, apollo) {
```
faboweb: in line 4 you can use `export default function({ apollo }) {` as in `session.js`. It is provided to all modules.

src/vuex/modules/connection.js
```diff
@@ -20,26 +18,23 @@ export default function() {
...
-        console.error('Error', error)
+      const { data } = await apollo.query({
+        query: Networks
       })
-
+      let availNetworks = Object.values(data.networks).map( network => network.id)
+      if (network && availNetworks.includes(network)) {
+        await commit(`setNetworkId`, network)
+      } else if (network && !availNetworks.includes(network)) {
+        if (data.networks.length > 1) {
```
faboweb: maybe use a safer way by doing a `network.find` for `cosmos-hub-mainet` and if that is not there choose `local-cosmos-hub-testnet`.

Bitcoinera: Sounds cool, yes

Bitcoinera:
I still don't know how to fix the unit test. I am looking into `session.spec.js` but I still don't see how apollo is being mocked there... In fact, searching for the term "apollo" returns nothing there

colw:
I believe Fabian implemented a `default=true` property to the network (`cosmos-hub-mainnet`) in the API.

*Right now, I think it is a mistake that both testnet and mainnet have this enabled, but we can update that.

faboweb:
I would just hardcode the default network for now for simplicity

src/vuex/modules/connection.js
```diff
@@ -17,11 +18,24 @@ export default function() {
   }
 
   const actions = {
-    async checkForPersistedNetwork({ commit }) {
-      const network = localStorage.getItem(`network`)
-      if (network) {
-        await commit(`setNetworkId`, JSON.parse(network))
+    async checkForPersistedNetwork({ dispatch, commit }) {
+      const network = JSON.parse(localStorage.getItem(`network`))
```
faboweb:
```suggestion
      const persistedNetwork = JSON.parse(localStorage.getItem(`network`))
```

src/vuex/modules/connection.js
```diff
@@ -17,11 +18,24 @@ export default function() {
...
+      const { data } = await apollo.query({
+        query: Networks
+      })
+      let availNetworks = Object.values(data.networks).map( network => network.id)
+      if (network && availNetworks.includes(network)) {
+        await commit(`setNetworkId`, network)
+      } else if (network && !availNetworks.includes(network)) {
```
faboweb:
```suggestion
      } else {
```

src/vuex/modules/connection.js
```diff
@@ -17,11 +18,24 @@ export default function() {
...
+        if (availNetworks.find( network => network === `cosmos-hub-mainnet`)) {
+          // we connect as default to cosmos-hub-mainnet
+          await dispatch(`setNetwork`, data.networks[1])
```
faboweb:
```suggestion
          await dispatch(`setNetwork`, data.networks.find(({id}) => id === `cosmos-hub-mainnet`))
```

src/vuex/modules/connection.js
```diff
@@ -17,11 +18,24 @@ export default function() {
...
+        } else {
+          // otherwise we connect to local-cosmos-hub-testnet
+          await dispatch(`setNetwork`, data.networks[0])
```
faboweb:
```suggestion
          await dispatch(`setNetwork`, data.networks.find(({id}) => id === `local-cosmos-hub-testnet`))
```

src/vuex/modules/connection.js
Bitcoinera: Makes sense now that you put it that way :stuck_out_tongue_winking_eye:

src/vuex/modules/connection.js
Bitcoinera: Oh yes, this is much safer. Thank you

faboweb:
Awesome!

faboweb:
unit tests are failing though :D

Bitcoinera:
I know! :sweat_smile: xD I need to fix this

Bitcoinera:
Thank you Fabo :star:

faboweb:
coverage is missing

Bitcoinera:
Done! 100% coverage in `connection.js` :muscle:

PR #3228, opened 2019-11-29
Fabo/custom gas for votes

faboweb:
Increases the gas estimate for voting as the estimate is too low.

### DIFF
--- a/src/ActionModal/utils/ActionManager.js
+++ b/src/ActionModal/utils/ActionManager.js
@@ -77,6 +77,8 @@ export default class ActionManager {
     }
 
     // temporary fix as the SDK doesn't return proper estimates for votes
+    // TODO move into transacton service
+    /* istanbul ignore next */
     if (this.messageType === transaction.VOTE) {
       gasEstimate = 30000
     }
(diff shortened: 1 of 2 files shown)


jbibla:
simulation only causes errors on votes?

faboweb:
Noone else complained so far

src/ActionModal/utils/ActionManager.js
```diff
@@ -76,6 +76,13 @@ export default class ActionManager {
       this.message = await this.createWithdrawTransaction()
     }
 
+    // temporary fix as the SDK doesn't return proper estimates for votes
+    // TODO move into transacton service
+    /* istanbul ignore next */
+    if (this.messageType === transaction.VOTE) {
+      gasEstimate = 30000
```
jbibla: how does this compare to what is usually returned?

jbibla: and what gas prices should this typically generate?

faboweb: I can't say for sure. On the local testnet it is 13k. I saw gas prices so far up to 25k. I guess would be cool to track used gas of all txs.
This effects fees. It makes the txs proportionally more expensive. Still txs in general are very cheap.

PR #3251, opened 2019-12-04
Jordan/portfolio loading

### LINKED ISSUE #3225: portfolio page should load more gracefully

the portfolio page in the web and on mobile loads too slowly. in addition to this, we load an empty portfolio first, instead of a loading screen.

- the user should never see the empty portfolio screen if the they have a balance, rewards, and a portfolio of validators
- the page should also load faster

- we have a placeholder graphic we could use here

**Additional context**

- the way this works now causes anxiety and confusion. it would be better for the user to be annoyed by a loading graphic than worried about their balance or portfolio if that needs to be the first step.

jbibla:
Closes #3225

**Description:**
- added graceful loading to portfolio page (super easy, should be on every page)
- new designs for bars with less code complexity
- fixed minor style bug with disabled inputs

Screen Shot 2019-12-04 at 12 44 36 PM
Screen Shot 2019-12-04 at 12 44 46 PM

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/changes/jordan_portfolio-loading
+++ b/changes/jordan_portfolio-loading
@@ -1,0 +1,2 @@
+[Changed] show loading images before causing anxiety for user about the state of their portfolio @jbibla
+[Code Improvements] cleaned up bar styles to match new design and reduce code complexity @jbibla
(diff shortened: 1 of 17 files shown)


faboweb:
Pretty cool!

faboweb:
Design doesn't match 100% so there is some jumping, but a leap improvement

faboweb:
And there is a delay even before the portfolio shows

faboweb:
https://www.loom.com/share/fd6d6f76a75149dea83ee9b75d6442f1

jbibla:
> And there is a delay even before the portfolio shows

yep. this is our bundle being enormous. i'm not sure how to render the placeholder loader earlier than in the component we want to "place-hold"

colw:
This is awesome. Nice work.

I wonder though, if using empty divs+css (without the svgs) would be quicker?

faboweb:
Note: we might switch later to just use css

jbibla:
> I wonder though, if using empty divs+css (without the svgs) would be quicker?

this is probably a better approach

PR #3260, opened 2019-12-05
Ana/livepeer integration

Bitcoinera:
Closes # 105 in the BE (Put Livepeer into (Lunie) production)

**Description:**

In this branch I am making a few changes to integrate the most properly the new Livepeer network, which has some particularities.

These changes are going to be kept **minimal**.

This is a work in progress. 

Thank you! 🚀

---

For contributor:

For reviewer:


src/components/staking/LiValidator.vue
```diff
@@ -36,7 +36,7 @@
       />
       <div class="validator-info">
         <h3 class="li-validator-name">
-          {{ validator.name }}
+          {{ validator.shortenName }}
```
colw: This is a good opportunity to use a custom Vue filter. It can be used like this for example:

`{{ validator.name | truncate(30) }}`

https://v1.vuejs.org/guide/custom-filter.html

Bitcoinera: Thank you! I should have supposed there was something like this at hand

Bitcoinera: I think this truncate will have to work responsively... It is going to be better

Bitcoinera:
This is also good to go in!

src/components/staking/LiValidator.vue
```diff
@@ -35,9 +35,12 @@
         :alt="`validator logo for ` + validator.name"
       />
       <div class="validator-info">
-        <h3 class="li-validator-name">
+        <h3 v-if="!validator.name.startsWith('0x')" class="li-validator-name">
```
jbibla: i don't think we should have special network-specific logic here. 

we could pass 0x addresses into bech32 where the truncation could happen. we can open a separate ticket to refactor the name of the component so it's more generic.

src/components/staking/LiValidator.vue
```diff
@@ -35,9 +35,12 @@
...
           {{ validator.name }}
         </h3>
+        <h3 v-else>
+          <Bech32 :address="validator.name" />
```
jbibla: this shouldn't be used for validator names

src/components/staking/PageValidator.vue
```diff
@@ -107,11 +107,11 @@
             {{ (validator.selfStake / validator.tokens) | percent }}
           </span>
         </li>
-        <li>
+        <li v-if="validator.startHeight">
```
jbibla: i'm assuming that livepeer doesn't have this data? 

is there no way to find out when a transcoder came online / became a transcoder? 

if no, let's say "unknown" or "n/a" or "--" instead of hiding the div.

jbibla: same for all the changes below in this file.

Bitcoinera: Ok :+1:

src/components/staking/LiValidator.vue
Bitcoinera: Ok, another misunderstanding :sweat_smile:

src/components/staking/LiValidator.vue
```diff
@@ -36,7 +36,7 @@
       />
       <div class="validator-info">
         <h3 class="li-validator-name">
-          {{ validator.shortenName }}
+          {{ validator.name | truncate(30) }}
```
faboweb: instead of truncate please use the address shortening format we already use everywhere else like in the Bech32 component

Bitcoinera: Oops I should have read this earlier xD

Bitcoinera: Alright, this in now implemented

src/components/staking/PageValidator.vue
```diff
@@ -109,29 +113,47 @@
...
         </li>
         <li>
           <h4>Uptime</h4>
-          <span id="page-profile__uptime">
-            {{ validator.uptimePercentage | percent }}
-          </span>
+          <span v-if="validator.uptimePercentage" id="page-profile__uptime">{{
+            validator.uptimePercentage | percent
+          }}</span>
+          <span v-else id="page-profile__uptime">{{
```
faboweb: there is a bunch of douplicated code in this PR. can this be handled differently?

Bitcoinera: Yes, you are right. I just changed it. The only issue was using filters inside a function. But SO saved me (again) :smile:

src/components/staking/PageValidator.vue
```diff
@@ -272,6 +250,11 @@ export default {
         `undelegations`,
         `transactions`
       ]) // TODO use more finegrained query string (network and address)
+    },
+    chooseFilter(field) {
+      return field
+        ? this.$options.filters.percent(field)
```
faboweb: In case of "Last Commission Change" this is the `fromNow` filter. How about passing in the filter as a second argument? Or can we chain the filters better @colw ?

src/components/staking/LiValidator.vue
```diff
@@ -38,6 +38,9 @@
         <h3 class="li-validator-name">
           {{ validator.name }}
         </h3>
+        <h3 class="li-validator-name li-validator-name-short">
```
jbibla: we should use CSS to achieve this, not extra HTML. also, this is not related to livepeer so could be addressed in a separate PR. 

livepeer validator names are temporarily addresses. this is not going to be the case for long and is not a precedent we want to set.

src/components/staking/LiValidator.vue
```diff
@@ -54,13 +57,17 @@
       }}
     </td>
     <td :class="{ 'hide-xs': showOnMobile !== 'voting-power' }">
-      {{ validator.votingPower | percent }}
+      <span v-if="validator.votingPower">
+        {{ validator.votingPower | percent }}
+      </span>
+      <span v-else>{{ validator.votingPower | noBlanks }}</span>
```
jbibla: this doesn't really make sense.

if `!validator.votingPower` why would we display `validator.votingPower` with a filter?

validators should _always_ have voting power.  even if it's 0 (which it is for cosmos validators who are not in the set, for example) then the percent filter should just display/return 0. 

i don't think we need extra HTML or conditional logic here at all.

src/filters/index.js
```diff
@@ -21,6 +21,8 @@ export const percentOrPending = function(value, totalValue, pending) {
 export const formatBech32 = (address, longForm = false, length = 4) => {
   if (!address) {
     return `Address Not Found`
+  } else if (address.startsWith(`0x`)) {
```
jbibla: nice!

jbibla:
i don't think there should be any changes to `LiValidator` or `PageValidator`. to me, this would indicate an issue with the way the API is returning data.

this PR should only need to include the network image, changelog entry, and the change to `formatBech32` to account for `0x`. possibly also some css for long validator names - but i would even say that could be a separate ticket and corresponding PR.

src/components/staking/LiValidator.vue
Bitcoinera: 0 means the validators has 0 voting power and '--' means "this feature is not available in this network. I personally think '--' is a better option, more explanatory.
Also, this is an outdated change. Now this is implemented much more beautifully with the chooseFilter method in just a single line of code.

src/components/staking/LiValidator.vue
Bitcoinera: Ok, you are right. I just deleted the whole logic for `validator.name` from `LiValidator`. It is better to invest time adding those names to the DB than implementing this for a short period of time only

Bitcoinera: Ok, you are right. I just deleted the whole logic for `validator.name` from LiValidator. It is better to invest time adding those names to the DB than implementing this for a short period of time only

src/components/staking/LiValidator.vue
faboweb: livepeer doesn't have something like a stake distribution??! sounds wrong.

jbibla: please push the latest changes so we can discuss only relevant things.

src/components/staking/PageValidator.vue
colw: I would think `abc | percent | noBlanks` should work. 

`noBlanks` should just display as is, or `--` if empty.

src/components/staking/LiValidator.vue
Bitcoinera: There is no voting and there are no proposals in Livepeer

src/components/staking/PageValidator.vue
Bitcoinera: No because `| percent` transforms `null` into 0, and `0` is not a value `noBlanks` considers as "blank".

src/components/staking/LiValidator.vue
```diff
@@ -35,12 +35,12 @@
         :alt="`validator logo for ` + validator.name"
       />
       <div class="validator-info">
-        <h3 class="li-validator-name">
-          {{ validator.name }}
-        </h3>
-        <h3 class="li-validator-name li-validator-name-short">
+        <h3 v-if="validator.name.startsWith('0x')" class="li-validator-name">
           {{ validator.name | formatBech32 }}
```
faboweb: why not do this on the server so the FE stays agnostik?

src/components/staking/LiValidator.vue
faboweb: `votingPower` has limited to do with voting on proposals. this is the percent of all stake delegated to this validator

src/components/staking/LiValidator.vue
faboweb: see @jbibla comment

src/components/staking/LiValidator.vue
Bitcoinera: ok, will change it. I didn't know that is what it actually meant

src/components/staking/LiValidator.vue
Bitcoinera: I can definitely move the `chooseFilter` thing to another PR. The `validator.website` improvement as well. Let's do it

PR #3281, opened 2019-12-10
Fix redelegation

### LINKED ISSUE #3096: unstaking and clicking redelegation sets the to and from fields incorrectly
if i go to unstake from a validator, but i click the redelegate button - the "to" field becomes the validator i was intending to unstake from. this should be the opposite. 

ideally, i'd propose we just use dropdowns for both "to" and "from" on the redelegation modal.

as well, maybe we should just put dropdowns for "to" and "from" for all staking modals.

mariopino:
Closes #3096 

**Description:**

- Implement redelegation as a specific case of undelegation.
- Show delegations in `From` field and list of active validators (excluding source validator) in `To` field.

**Pending tasks:**

- Delete redelegation logic in DelegationModal.vue
- Testing
- Code coverage
- Improve redelegation constraints handling, new PR: https://github.com/luniehq/lunie/issues/3325

image

Thank you! 🚀

---

For contributor:

For reviewer:


src/ActionModal/components/DelegationModal.vue
```diff
@@ -52,7 +58,37 @@
...
 
     <TmFormGroup
-      v-if="fromOptions.length > 1"
+      v-else
+      class="action-modal-form-group"
+      field-id="to"
+      field-label="To"
+    >
+      <TmField
+        id="to"
```
jbibla: i think the `from` field should be above the `to` field.

src/ActionModal/components/DelegationModal.vue
```diff
@@ -52,7 +58,37 @@
...
+        v-model="targetValidator.operatorAddress"
+        :options="validators"
+        type="select"
+      />
+      <TmFormMsg
+        v-if="targetValidator.status === 'INACTIVE' && !isRedelegation"
```
jbibla: instead of showing this, should we only list active validators in the dropdown?

src/ActionModal/components/DelegationModal.vue
```diff
@@ -374,6 +443,63 @@ export default {
...
+        ) {
+          validators(
+            networkId: $networkId
+            searchTerm: $searchTerm
+            activeOnly: $activeOnly
+          ) {
+            name
+            operatorAddress
+            consensusPubkey
+            votingPower
```
jbibla: why do we need all this validator data just to populate the list?

src/ActionModal/components/DelegationModal.vue
```diff
@@ -374,6 +443,63 @@ export default {
...
+                key: `${validator.name} - ${formatBech32(
+                  validator.operatorAddress,
+                  false,
+                  20
+                )}`,
+                value: validator.operatorAddress
+              }
+            })
+        )
+        // console.log(options)
```
jbibla: you can delete this comment 😉

src/ActionModal/components/DelegationModal.vue
```diff
@@ -52,7 +58,37 @@
...
     <TmFormGroup
-      v-if="fromOptions.length > 1"
+      v-else
+      class="action-modal-form-group"
+      field-id="to"
+      field-label="To"
+    >
+      <TmField
+        id="to"
+        v-model="targetValidator.operatorAddress"
```
jbibla: how much work would it be to show the picture (like we do in the activity page) in the dropdown 😄 ?

jbibla: and, would it be possible (how much work?) if we changed the page of the validator when a user selected a validator from this list? would that be strange ux? just a thought...

faboweb: #scopecreep

faboweb:
I don't like the double dropdown. It increases complexity as it increases the simultanious decisions a user needs to make. I know @jbibla proposed this as a solution and I am pretty late to protest.

mariopino:
> I don't like the double dropdown. It increases complexity as it increases the simultanious decisions a user needs to make. I know @jbibla proposed this as a solution and I am pretty late to protest.

Well, we can simplify user options just maintaining the To field and remove From. What do you think?

src/ActionModal/components/DelegationModal.vue
mariopino: > and, would it be possible (how much work?) if we changed the page of the validator when a user selected a validator from this list? would that be strange ux? just a thought...

mmm I think would be strange :-), but useful because then users could see helpful information about the possible validators to redelegate (like comission or voting power), half hided by the modal window thought xD.

mmm may be a dumb question but why we need modal windows? Shouldn't be easier to use the same layaout for all actions? For example in this case we are limited by the modal window size...

Just a thought! don't want to reinvent the world ;-)

jbibla: > #scopecreep

**Definitely, but sometimes it's nice to take these opportunities to discuss new ideas - however always good to assume these things will happen in another PR! 

> why we need modal windows

what do you mean @mariopino ?

jbibla:
> It increases complexity as it increases the simultanious decisions a user needs to make.

while this is true, i think it's ok because the user has already indicated that they want to do something more complicated (redelegate).

however, i don't feel super strongly about this. 

the thing i do care about is that the `to` and `from` are not backwards as they were before.

faboweb:
> the thing i do care about is that the to and from are not backwards as they were before.

can we revert to just this change?

mariopino:
> > the thing i do care about is that the to and from are not backwards as they were before.
> 
> can we revert to just this change?

Oki, I will just fix the to field.

jbibla:
yep you got it mario! 

i would leave the from field but have it disabled as we do for the delegation modal. users have in the past wanted to see this.

i will add:
- remove the "redelegation" link from the undelegation modal
- undelegation modal should include a "To" input, which is a dropdown with a default of "My Wallet" selected. the rest of the validators should appear as the other options in the dropdown.

this should be simple and will address the issue in the short term. long term we will have to re-assess the UX around these transactions. 

hope this is clear and makes sense.

faboweb:
Great! Almost done!
I would appreciate a test for the `update` functions in the 2 modals
And for the `setFromSelectedIndex` in `UndelegationModal`

mariopino:
> Great! Almost done!
> I would appreciate a test for the `update` functions in the 2 modals

Hey @faboweb! My plan was to keep redelegation logic only in `UndelegationModal.vue` and removing it from `DelegationModal.vue`, what do you think?

> And for the `setFromSelectedIndex` in `UndelegationModal`

Sure! I want to do some (quick) cleanup before and check variable names again to make sure their names provide valuable information.

faboweb:
test for `update` missing for DelegationModal

faboweb:
Looking good!!!

src/ActionModal/components/DelegationModal.vue
```diff
@@ -91,8 +76,7 @@
         />
       </TmFieldGroup>
       <span class="form-message">
-        {{ isRedelegation ? "Available to redelegate" : "Available to stake" }}
-        :
+        Available to stake:
         {{ maxAmount }}
```
jbibla: will `maxAmount` change to reflect your current delegation if you're doing a redelegation? or always shows available balance?

src/ActionModal/components/UndelegationModal.vue
```diff
@@ -4,9 +4,11 @@
     ref="actionModal"
     :validate="validateForm"
     :amount="0"
-    title="Unstake"
+    :title="isRedelegation ? 'Restake' : 'Unstake'"
```
jbibla: apologies for the all the miscommunication here.

i think we should remove redelegation from this component as well. both delegation and undelegation modals should include to and from fields that are dropdowns.

faboweb:
@mariopino there are some comments here, did you see them?

mariopino:
> @mariopino there are some comments here, did you see them?

Hey @faboweb, yes, I have to address them and finish the issue.

faboweb:
Success is not triggered for me. Is that an API issue or an issue with this PR?

There is some dropdown showing even if I can't use it:
image

mariopino:
> Success is not triggered for me. Is that an API issue or an issue with this PR?
> 
> There is some dropdown showing even if I can't use it:
> image

We can enable it to show all validators in `To` field, it's an easy modification.

faboweb:
> We can enable it to show all validators in To field, it's an easy modification.

but we don't right now. should we keep code we don't use in there?

PR #3295, opened 2019-12-13
Update proposals every new block

### LINKED ISSUE #3277: proposals page never refreshes
**Describe the bug**

- the proposal page will only load once per session
- this means you have to refresh the page to see new proposals
- on mobile, this means you have to refresh the app 

**Expected behavior**

- every block we should fetch proposals
- every x number of blocks we should poll for proposals
- whenever there is a new proposal we should push this data to lunie

mariopino:
Closes #3277

**Description:**

Update proposals every new block in PageProposals.

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/changes/mario_3277-proposals-page-never-refreshes
+++ b/changes/mario_3277-proposals-page-never-refreshes
@@ -1,0 +1,1 @@
+[Fixed] [#3277](https://github.com/cosmos/lunie/issues/3277) Update proposals every 10 blocks @mariopino
--- a/src/components/governance/PageProposal.vue
+++ b/src/components/governance/PageProposal.vue
@@ -417,11 +417,12 @@ export default {
           return !this.found
         },
         result() {
-          // Don't update passed or rejected proposals
           /* istanbul ignore next */
           if (
+            // Don't update passed or rejected proposals
             this.proposal.status !== "Passed" &&
-            this.proposal.status !== "Rejected"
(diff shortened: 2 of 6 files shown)


mariopino:
Should we need to do the same with single proposal page?

jbibla:
why not every block when you're on PageProposal?

mariopino:
> why not every block when you're on PageProposal?

Hey @jbibla I tried it every block but It's a bit annoying to see the data loading every 6 seconds, so decided to reload it every 10 blocks.

Best solution would be a proposal subcription then FE would get notified when a new proposal arrives (just like you proposed in the issue).

jbibla:
ok cool. i understand. 

the way we should solve the problem you're mentioning is to use a `loaded` property. after the first `loading` is done, we set `loaded` to true and then the page will not rerender, but the data will update (vue takes care of this).

mariopino:
> ok cool. i understand your thinking now!
> 
> the way we should solve the problem you're mentioning is to use a `loaded` property. after the first `loading` is done, we set `loaded` to true and then the page will not rerender, but the data will update (vue takes care of this). `TmPage` has this logic in place.

Aaam yes. that's the behaviour I was looking for. I'll change it to use the loaded logic ...

mariopino:
Ready to merge finally (I hope) :-)

PR #3304, opened 2019-12-14
Ana/fix scrollbar in side menu

Bitcoinera:
Closes #ISSUE
I just encountered this bug this morning

**Description:**
There was a weird scrollbar around the `AppMenu` component and also the `Help/Feedback` button was behaving strangely when resizing the screen. Here the screenshots:

Now it is fixed! :muscle: 

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/src/components/common/AppHeader.vue
+++ b/src/components/common/AppHeader.vue
@@ -159,6 +159,7 @@ export default {
   .app-header > .container {
     position: fixed;
     height: 100%;
+    overflow: auto;
     background: var(--app-nav);
   }
 }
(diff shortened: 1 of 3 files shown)


src/components/common/AppHeader.vue
```diff
@@ -159,7 +159,6 @@ export default {
   .app-header > .container {
     position: fixed;
     height: 100%;
-    overflow: scroll;
```
jbibla: this should be `overflow: auto`

i set it to `scroll` to fix the bug you reported about the menu not being visible on certain screen sizes. scroll forces scrollbars on windows and linux, but on mac os it only displays them if the content of the container is greater than the container (when scrollbars are actually needed).

Bitcoinera: Thank you for the explanation! Maybe my solution was a bit bold :sweat_smile:

PR #3337, opened 2019-12-18
Fabo/3316 cross account creation

### LINKED ISSUE #3316: Create addresses for different networks
It should be possible to create an address for different networks apart from Cosmos.

faboweb:
Closes #3316, #3342, #3341

**Description:**

Thank you! 🚀

---

For contributor:

For reviewer:


faboweb:
money on terra:
Bildschirmfoto von 2019-12-18 21-10-40

btw https://soju-faucet.terra.money/

faboweb:
Test by adding `to.meta.feature !== "Session" &&` to line 21 in `router.js`

src/vuex/modules/keystore.js
```diff
@@ -69,3 +69,37 @@ export default () => {
...
+    case "terra-mainnet": {
+      const addressCreator = await getCosmosAddressCreator(network)
+      return addressCreator(seedPhrase)
+    }
+    default:
+      throw new Error(
+        "Lunie doesn't support address creation for this network."
+      )
+  }
+}
```
Bitcoinera: Neat!

Bitcoinera:
> money on terra:
> 
> btw https://soju-faucet.terra.money/

I am having issues when sending... don't know if it is directly related with the new addresses. Looking into it right now.



Also happening now in `cosmos-testnet`:

Bitcoinera:
It looks like we are missing something. The API returns "0" when querying for the balance:



And in Hubble...

faboweb:
thanks for checking sending. I just checked account creation. will test this tomorrow.

faboweb:
need to: prevent users from sending in with a wrong prefix

Bitcoinera:
> need to: prevent users from sending in with a wrong prefix

Oops, good point. In `SendModal` it is only checking that the address sending to is valid bech32 address, nothing more

faboweb:
hhaha I misstyped. But that is actually also a good check ^^

I actually meant: Prevent users to sign in with a wrong prefix. / or give them the option to instead log in to the chain that this address is used for.

faboweb:
@Bitcoinera would you agree to merging this PR and creating a follow up which handles the issues with the balance in the API and the better checking for correct addresses?

Bitcoinera:
> @Bitcoinera would you agree to merging this PR and creating a follow up which handles the issues with the balance in the API and the better checking for correct addresses?

For sure, yes. We can do that

Bitcoinera:
I merged my first PR! Hehe, I feel so powerful now! :muscle:

PR #3353, opened 2019-12-23
WIP: Ana/livepeer handle getselfstake separately

Bitcoinera:
Closes #ISSUE

**Description:**
This PR should fix, together with the PR with the same name for the API (#238), the problem of the slow loading of validators in Livepeer.

The big issue I have encountered here is that it is kind of hard to handle Objects as inputs in GraphQL. So that is why this PR is a work in progress. I don't know how to write the `selfStake` query in `PageValidator` to allow a validator Object as a parameter.

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/src/components/staking/PageValidator.vue
+++ b/src/components/staking/PageValidator.vue
@@ -348,7 +348,7 @@ export default {
       query: gql`
         query SelfStakePageValidator(
           $networkId: String!
-          $validator: Validator!
+          $validator: Validator! --> This is what I am struggling to figure out
         ) {
           selfStake(networkId: $networkId, validator: $validator) {
             amount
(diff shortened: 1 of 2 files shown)


faboweb:
Have you considered handling everything on the server side?
Re object as parameter: You can just pass an ID and look for the object on the server side by using this ID.

Bitcoinera:
> Have you considered handling everything on the server side?
> Re object as parameter: You can just pass an ID and look for the object on the server side by using this ID.

But then we would still to update the query every now and then, no? And it is what goes slow.

Maybe I didn't get fully what you propose...

Bitcoinera:
It is still unclear to me how to handle this

faboweb:
> But then we would still to update the query every now and then, no? And it is what goes slow.

yes and I would solve, why the query is slow from the API. something I would try:

- load validators per block into the store
- asynchronously load the selfStake also in the store (adding it to the validator object)
- if updating the selfStake takes too long, thottle the selfStake updates to every x blocks

Bitcoinera:
Put in the freezer for now

PR #3357, opened 2019-12-24
Ana/fix address undefined error in Bech32

Bitcoinera:
Closes #ISSUE
I just discovered this bug today

**Description:**

This was giving me trouble today as I was debugging. The field "address" was set to `required: true` in the `Bech32.vue` component and this was causing the same error about hundred times: `Expected type String with value "undefined" but got Undefined`.

I don't think it makes sense to have `address` as `required` since we are already filtering with `formatBech32` and if `address` is blank we simply mark it as `Adress Not Found`.

I didn't look into yet why all addresses appear as `undefined` but I presume it is related to the recent Cosmos upgrade, although I don't know for certain.

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/changes/ana_fix-prop-address-Undefined
+++ b/changes/ana_fix-prop-address-Undefined
@@ -1,0 +1,1 @@
+[Fixed] [#3357](https://github.com/cosmos/lunie/pull/3357) Not require address anymore in the Bech32 component @Bitcoinera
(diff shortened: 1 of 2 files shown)


faboweb:
Hmmm, can we maybe investigate why this happens? Would you agree that this solution only hides the problem?

Bitcoinera:
> Hmmm, can we maybe investigate why this happens? Would you agree that this solution only hides the problem?

Yes, pretty much xD
That is why I just pushed the real solution, PR #3359

PR #3359, opened 2019-12-27
Ana/fix multi send to address

Bitcoinera:
Closes #ISSUE

**Description:**

This is the real fix the bug my PR #3357 intended to address

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/src/components/transactions/message-view/MultiSendMessageDetails.vue
+++ b/src/components/transactions/message-view/MultiSendMessageDetails.vue
@@ -11,7 +11,7 @@
       </template>
       <template v-else-if="sentFromSessionAddress">
         <span>To&nbsp;</span>
-        <Bech32 :address="transaction.value.to_address" />
+        <Bech32 :address="transaction.value.outputs[0].address" />
       </template>
       <template v-else-if="receivedToSessionAddress">
         <span>From&nbsp;</span>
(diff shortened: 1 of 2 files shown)


src/components/transactions/message-view/MultiSendMessageDetails.vue
```diff
@@ -11,7 +11,7 @@
       </template>
       <template v-else-if="sentFromSessionAddress">
         <span>To&nbsp;</span>
-        <Bech32 :address="transaction.value.to_address" />
+        <Bech32 :address="transaction.value.outputs[0].address" />
```
faboweb: Hmm what happens to the other output addresses?

faboweb: We can add this fix and discuss an improvement later. Want to open an issue?

Bitcoinera: > Hmm what happens to the other output addresses?

There are no other addresses :woman_shrugging: It works like the `inputs`, which is already implemented this way in this same file.

But yes, we can add this now and then see later. It should be fixed though, I think.

faboweb: Inputs and outputs are arrays as far as I see. Have you considered, that there could be multiple in and outputs in a multi send? Maybe the implementation before is already faulty.

Bitcoinera: Yes, I finally realized what you meant... It is really worth investigating

PR #3362, opened 2019-12-30
Ana/add back to validators button

Bitcoinera:
Closes #ISSUE

**Description:**

This is one of the simple improvements I have thought of lately. 

It is just a button in `PageValidator` to go back to the `ListValidators`.

Didn't take me almost any time to make it, no worries ;)

Some screenshots:

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/src/components/staking/PageValidator.vue
+++ b/src/components/staking/PageValidator.vue
@@ -9,6 +9,15 @@
     class="small"
   >
     <template v-if="validator.operatorAddress" slot="managed-body">
+      <button
+        class="validators-list-button"
+        color="secondary"
+        @click="$router.push(`/validators`)"
+      >
+        <img src="/img/icons/arrow-icon.svg" />
+        <div style="width: 20px;display: inline-block"></div>
+        Back to Validators
+      </button>
       <div class="status-button-container">
         <div class="status-container">
           <span :class="validator.status | toLower" class="validator-status">
@@ -21,12 +30,6 @@
             {{ validator.statusDetailed }}
           </span>
         </div>
-        <TmBtn
-          class="validators-list-button"
-          value="Back to Validators"
-          color="secondary"
-          @click.native="$router.push(`/validators`)"
-        />
       </div>
       <tr class="li-validator">
         <td class="data-table__row__info">
@@ -401,9 +404,20 @@ export default {
 </script>
 <style scoped>
 .validators-list-button {
-  position: absolute;
-  right: 1.25rem;
-  top: 1.25rem;
+  margin: 0 0 20px 10px;
+  width: 177px;
+  height: 40px;
+  background-color: #272b48;
+  font-size: 14px;
+  color: #7a88b8;
+  border: 1px solid rgb(122, 136, 184, 0.1);
+  border-radius: 5px;
+}
+
+.validators-list-button:hover {
+  background: #445381;
+  color: #f1f3f7;
+  border-color: #445381;
 }
 
 .li-validator {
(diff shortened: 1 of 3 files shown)


faboweb:
Have you considered, that the user could always click on the validators link on the left (on desktop)? On mobile this is not possible though. Maybe work with @jbibla and Charly on a concept for the user flow here?

Bitcoinera:
> Have you considered, that the user could always click on the validators link on the left (on desktop)? On mobile this is not possible though. Maybe work with @jbibla and Charly on a concept for the user flow here?

Is my button so bad? xD

faboweb:
It is very "functional" :D

Bitcoinera:
> It is very "functional" :D

Well, that is not too bad, right? Maybe we could just merge it until someone else comes up with something better :cat:

faboweb:
It is not critically needed right? So let's come up with a proper design before merging it.

Bitcoinera:
I just remastered the button according to Charlie designs.

Here is Charlie's design:



Here is the result:



Not too bad although I needed to do a little "trick" to create that space between arrow and text, by placing a `div` in the middle.

What do you think? :paintbrush: 

:thinking:  I just noticed Charlie wrote "validators" in lowercase...

faboweb:
Can we hide the button on Desktop?
How about adding a `cursor: pointer` so it feels like a button?

Bitcoinera:
> Can we hide the button on Desktop?
> How about adding a `cursor: pointer` so it feels like a button?

Yes, good point

Bitcoinera:
OK, suggestions applied :+1: 

Only I am afraid guys I discovered yet another bug :laughing:  But this one really minor. It only happens the screen is exactly 1023px wide:



Some orphan CSS line shouting for love around there.

src/components/staking/PageValidator.vue
```diff
@@ -535,4 +528,10 @@ span {
     width: 50%;
   }
 }
+
+@media screen and (min-width: 1024px) {
```
faboweb: not a problem this time. but it is better to stick with either `max-width` or `min-width` so it is better to grasp what the breakpoints are.

In this case we could have done:
```
.validators-list-button {
  display: block;
}
@media screen and (max-width: 1024px) {
  display: none;
}
```

src/components/staking/PageValidator.vue
```diff
@@ -9,13 +9,27 @@
...
-        </span>
-        <span v-if="validator.statusDetailed" class="validator-status-detailed">
-          {{ validator.statusDetailed }}
-        </span>
+      <button
+        class="validators-list-button"
+        color="secondary"
+        @click="$router.push(`/validators`)"
+      >
+        <img src="/img/icons/arrow-icon.svg" />
```
jbibla: @Bitcoinera @faboweb we already import material icons for this. we don't need to include separate svg's for this kind of stuff.

Bitcoinera: Charlie seemed very happy with this arrow. I don't know :woman_shrugging:

jbibla: sure. but we already import them as fonts - icons should be used like this (not imported as separate SVGs):

https://github.com/luniehq/lunie/blob/b9566b9f842c7060378bae337c134aa32896eaaf/src/components/transactions/message-view/SendMessageDetails.vue#L24

Bitcoinera: pushed: https://github.com/luniehq/lunie/pull/3388

PR #3364, opened 2019-12-31
Fabo/better handling of sequence in ActionModal

faboweb:
An ActionModal would sometimes not query the sequence number correctly on a consecutive tx. This PR refactors parts of the logic for creating the tx context and refetches the sequence on every tx.

Note: We should decouple account info from the overview. They should probably be loaded independently.

Please test before merging.


faboweb:
Test by sending tokens to yourself twice without refreshing the browser

Bitcoinera:
:sweat_smile:  I have some magic power that once I press "Approve" then I start seeing the errors.

I am still checking, for now there was an error with the Sentry part: an arrow function was needed in order to inject `this` inside the scope of the function.

Bitcoinera:
> Test by sending tokens to yourself twice without refreshing the browser

I get the following error with both Lunie Browser Extension and Ledger, even waiting for the "Success" Message and closing it:



Also the `Total ATOMs` balance is not updating. Might be related to the issue I opened yesterday.

src/ActionModal/components/ActionModal.vue
```diff
@@ -106,8 +106,12 @@
           />
         </div>
         <div v-else-if="step === signStep" class="action-modal-form">
+          <!-- fallback to prevent the user to submit a tx before the sequence was loaded -->
+          <TmDataLoading
```
jbibla: this doesn't solve the issue and could be removed from this PR

src/ActionModal/components/ActionModal.vue
```diff
@@ -106,8 +106,12 @@
...
+            v-if="!context.account || context.account.sequence === undefined"
```
```suggestion
            v-if="!context.account || !Number.isInteger(context.account.sequence)"
```

jbibla: showing a loading screen doesn't help and won't go away if it is undefined / null so i'm not sure i get the need for this.

faboweb: nicer then letting the user submit and then it crashes?

jbibla: i may be misunderstanding this, but this doesn't make sense to me.

package.json
```diff
@@ -60,6 +60,7 @@
     "moment": "^2.24.0",
     "no-scroll": "^2.1.1",
     "register-service-worker": "^1.6.2",
+    "serialize-javascript": ">=2.1.1",
```
jbibla: what is this for / is it needed?

faboweb: good catch. don't know where this comes from and will remove it

faboweb: this is also in the resolution (to prevent security vulnerable deps) will leave it in and create another PR

Issue #244, opened 2017-12-14
Switch chain-ids


faboweb:
If we switch chain-ids, Gaia will not be reinitialized.

faboweb:
Mock server was interfering with the request... Everything is fine...

PR #3365, opened 2020-01-03
Ana/multiple denoms in porfolio (missing better designs)

Bitcoinera:
Closes #ISSUE

**Description:**

In this PR I add new features to the component `TmBalance` to fit networks which have multiple tokens. 
Basically I add a dropdown component where users can peek at their total balances amounts and also a currency selector where they can choose in what fiat currency they'd like to see the total value displayed.

For this PR a better design/UX is missing. I already talked to Charlie and I'm waiting for his insights. Meanwhile here is the code.

Screenshots:

Thank you! 🚀

---

For contributor:

For reviewer:


src/components/common/TmBalance.vue
```diff
@@ -92,6 +148,85 @@ export default {
     },
     onSend() {
       this.$refs.SendModal.open()
+    },
+    async calculateTotalFiatValue() {
```
faboweb: As you are stating in your comments, this is highly tight to specific networks (e-money). But the FE should be stupid and don't know much about individual networks. We need it to be that way so we can scale to many networks. This whole logic needs to be abstracted and put into the API I think.

Bitcoinera: Yes, in fact I considered this. Then I thought, 'OK, but the user needs to select a fiat currency in the FE'... But yes, a simple call to API and that's it.

It is something I definitely thought.

Bitcoinera:
> Logic needs to move to backend

It is done & working now. Look at my PR #244 "emoney portfolio integration"

Bitcoinera:
How do you test the `skip()` and `variables()` functions inside an apollo query with jest?

src/components/common/TmBalance.vue
```diff
@@ -142,6 +202,45 @@ export default {
         return data.network.stakingDenom
       }
     },
+    totalFiatValue: {
+      query: gql`
+        query totalFiatValue(
```
faboweb: this produces a roundtrip. why not include this somehow in the overview. it will always be served together right?

faboweb:
This also currently doesn't work with any other Token. So we should probably not merge this.

src/components/common/TmBalance.vue
Bitcoinera: Didn't you say in the other PR you wanted to slice overview in separate queries?

Bitcoinera:
> This also currently doesn't work with any other Token. So we should probably not merge this.

:woman_shrugging:

src/components/common/TmBalance.vue
faboweb: true, but the idea was to serve fiat values directly with any amount

Bitcoinera:
I just adapted the TmBalance component to the recent changes we merged into develop.

I guess the only missing thing here are Charlie's final designs.

Bitcoinera:
I saw that in develop we deleted the `userTransactionAdded` subscription.

Isn't this how we kept the balance updated in TmBalance? Ah, but now we have the balances query, makes sense. Then I delete it.

src/components/common/TmBalance.vue
```diff
@@ -180,76 +193,97 @@ export default {
...
       },
-      update(data) {
+      skip() {
         /* istanbul ignore next */
-        return data.network.stakingDenom
+        return !this.address
       }
     },
-    totalFiatValue: {
+    balancesWithFiat: {
```
faboweb: is it correct, that we load the balances twice now?

src/components/common/TmBalance.vue
```diff
@@ -142,6 +149,16 @@ export default {
       ]
     }
   },
+  watch: {
+    balancesWithFiat: function() {
+      this.convertedBalances = this.balancesWithFiat.map(
```
faboweb: have you considered making this a computed instead?

src/components/common/TmBalance.vue
Bitcoinera: I don't see any other option. Unless there is some way to use one fragment or another conditionally (either
```
{
amount
denom
}
```
or 
```
{
amount
denom
fiatValue
}
```

faboweb: ?! why not just load the latter? it holds all the info right?

src/components/common/TmBalance.vue
Bitcoinera: the issue here is that I cannot query `balancesWithFiat` at the beginning, when the component loads, since I lack of the fiatCurrency parameter. That is why I went the watch way.

faboweb:
> I saw that in develop we deleted the userTransactionAdded subscription.

the balances should be updated every block I guess (blockAdded subscription). Even though I think that is suboptimal. But enough for now.

src/components/common/TmBalance.vue
Bitcoinera: because I don't have fiatCurrency at the beginning... 

Well, this is easily fixed by selecting a default fiatCurrency. And then I can also make `balancesWithFiat` a computed prop :thinking: 

I will go this way :v:

Bitcoinera:
Alright, I introduced your suggestions @faboweb .

Now there is no watch, only one balances query and `EUR` is the fiat currency shown by default.

faboweb:
image

Love it

Bitcoinera:
> image
> 
> Love it

Reinigung!

PR #3371, opened 2020-01-04
Ana/add better validation for cosmos addresses

Bitcoinera:
Closes #ISSUE

**Description:**

I was looking through the Sentry errors and found one was due to an user trying to log in as an address starting with `cosmospub`. [Here](https://monitoring.lunie.io:9000/lunie/frontend/issues/69/?referrer=slack) is the link to the error: 

This PR fixes similar future errors by improving the validation in the `TmSessionExplore` component.
 It is true that it seems that only the "cosmospub" addresses were missing, since I tried `cosmosvalcons`, `cosmosvaloperpub` and so on and the validation for `bech32` is enough in those other cases. Still, I added these other cases just to stay on the safe side and really make sure that users can only log in as regular `cosmos1` addresses.

Again, this is only cosmos focused and now that we will soon have `session` enabled for other networks maybe we need to tweak this a bit.

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/src/components/common/TmSessionExplore.vue
+++ b/src/components/common/TmSessionExplore.vue
@@ -58,9 +58,9 @@
           />
           <TmFormMsg
             v-else-if="
-              $v.address.$error && !$v.address.isNotAnyOtherCosmosAddress
+              $v.address.$error && !$v.address.isAWhitelistedBech32Prefix
             "
-            name="You can only sign in with a regular Cosmos address, starting with 'cosmos1'"
+            name="You can only sign in with a regular address"
             type="custom"
           />
         </TmFormGroup>
@@ -134,12 +134,13 @@ export default {
         return false
       }
     },
-    isNotAnyOtherCosmosAddress(param) {
+    isAWhitelistedBech32Prefix(param) {
       if (
-        param.substring(0, 9) !== "cosmospub" &&
-        param.substring(0, 13) !== "cosmosvalcons" &&
-        param.substring(0, 16) !== "cosmosvalconspub" &&
-        param.substring(0, 16) !== "cosmosvaloperpub"
+        param.substring(0, 7) === "cosmos1" ||
+        param.substring(0, 6) === "terra1" ||
+        param.substring(0, 5) === "xrn:1" ||
+        param.substring(0, 7) === "emoney1" ||
+        param.substring(0, 16) === "0x"
       ) {
         return true
       } else {
@@ -169,7 +170,7 @@ export default {
         required,
         bech32Validate: this.bech32Validate,
         isNotAValidatorAddress: this.isNotAValidatorAddress,
-        isNotAnyOtherCosmosAddress: this.isNotAnyOtherCosmosAddress
+        isAWhitelistedBech32Prefix: this.isAWhitelistedBech32Prefix
       }
     }
   }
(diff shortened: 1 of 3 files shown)


faboweb:
This PR doesn't scale to other networks. Can you write it in a way, that it scales to other Cosmos networks and other PoS networks like livepeer?

Bitcoinera:
> This PR doesn't scale to other networks. Can you write it in a way, that it scales to other Cosmos networks and other PoS networks like livepeer?

Maybe too much for one single PR? I mean, other Cosmos networks easy, but all other networks...

src/components/common/TmSessionExplore.vue
```diff
@@ -127,6 +134,18 @@ export default {
         return false
       }
     },
+    isNotAnyOtherCosmosAddress(param) {
```
faboweb: instead of blacklisting prefixes, how about checking if the prefix is whitelisted => a `bech32_prefix` in any network

jbibla:
awesome to see you investigating sentry errors and discovering problems this way!

src/components/common/TmSessionExplore.vue
Bitcoinera: OK, I think I just implemented it like you had in mind

src/components/common/TmSessionExplore.vue
```diff
@@ -58,9 +58,9 @@
           />
           <TmFormMsg
             v-else-if="
-              $v.address.$error && !$v.address.isNotAnyOtherCosmosAddress
+              $v.address.$error && !$v.address.isAWhitelistedBech32Prefix
             "
-            name="You can only sign in with a regular Cosmos address, starting with 'cosmos1'"
+            name="You can only sign in with a regular address"
```
jbibla: this is not a good error message. what does "regular" actually mean to a user?

src/components/common/TmSessionExplore.vue
```diff
@@ -135,12 +135,13 @@ export default {
...
       }
     },
-    isNotAnyOtherCosmosAddress(param) {
+    isAWhitelistedBech32Prefix(param) {
       if (
-        param.substring(0, 9) !== "cosmospub" &&
-        param.substring(0, 13) !== "cosmosvalcons" &&
-        param.substring(0, 16) !== "cosmosvalconspub" &&
-        param.substring(0, 16) !== "cosmosvaloperpub"
+        param.substring(0, 7) === "cosmos1" ||
```
jbibla: @Bitcoinera @faboweb this is another place where we should think about how to reduce frontend changes for new networks...

src/components/common/TmSessionExplore.vue
Bitcoinera: I know, but it is difficult to find a broad enough message to cover all networks... Any suggestions?

jbibla: > difficult to find a broad enough message to cover all networks

true! then the error messages should be network specific.

for now: "We don't currently support this type of address. Please try a different address format." ...?

src/components/common/TmSessionExplore.vue
jbibla: #underengineered @faboweb 😝

Issue #255, opened 2017-12-19
Tests have a lot of errors not failing the test suite


faboweb:
An example:
![image](https://user-images.githubusercontent.com/5869273/34155403-71473ae4-e4b9-11e7-878c-445dc2f9c61e.png)

I will make Jest fail on every console.error called to catched those errors.

jolesbi:
i'm pretty sure all these errors are because of nested dependencies. wallet.js needs node.js which needs rpc etc... 

a bunch of them are also missing the `config.js` module.

i spent some time trying to solve this but didn't have any luck.

lead me to thinking — maybe we ought to be using `shallow()` more often.

PR #3384, opened 2020-01-08
Ana/fix micro denoms

Bitcoinera:
Closes #ISSUE

**Description:**

Together with my other PR for the API, #255 , finishes fixing (for now) the issues we had in the action modals for sending transactions and staking in other networks apart from Cosmos.

It just add the other networks micro denoms in a dictionary. Thanks @faboweb for the "complete" hint :wink: .

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/src/scripts/common.js
+++ b/src/scripts/common.js
@@ -45,8 +45,11 @@ export const sleep = function(amount) {
 }
 
 export function toMicroDenom(denom) {
-  if (denom === "ATOM") {
-    return "uatom"
+  const lookup = {
+    ATOM: "uatom",
+    MUON: "umuon",
+    LUNA: "uluna",
+    TREE: "seed"
   }
-  return denom.toLowerCase()
+  return lookup[denom] ? lookup[denom] : denom.toLowerCase()
 }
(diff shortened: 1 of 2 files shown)


faboweb:
hahahaha thanks @Bitcoinera for the completion

Bitcoinera:
You are very welcome!

But hey, what about the tests? :woman_shrugging: 
I am going to make another PR then for them. I know, not so much to test, but still...

faboweb:
sure! the code you added was covered already though.

src/scripts/common.js
```diff
@@ -45,8 +45,11 @@ export const sleep = function(amount) {
 }
 
 export function toMicroDenom(denom) {
-  if (denom === "ATOM") {
-    return "uatom"
+  const lookup = {
+    ATOM: "uatom",
```
jbibla: will we need to add this for every network?

Bitcoinera: I tried using `substr` and just removing the first character, the problem are the "other" tokens... like `ukr` in Terra and all the others in e-Money... they also go through this function.

So right now I don't see any other solution :woman_shrugging:

jbibla: we should probably plan to store this in the db / api.

we want to get to a place where we don't need to update lunie frontend to add new networks.

(i think)

Bitcoinera: Yes, that would be the ideal

faboweb: #overengineering

Bitcoinera: hahaha. One thing is the ideal and another one the raw reality :laughing:

PR #3385, opened 2020-01-08
Fabo/cleanup dependencies

faboweb:
Removes some and updates some dependencies

Enforces linting prepush


jbibla:
approved, but netlify and yarn.lock should be fixed.

PR #3386, opened 2020-01-08
Ana/add tests for to microdenom and sleep

Bitcoinera:
Closes #ISSUE

**Description:**

Just a couple of boring tests for the `common` script.

Thank you! 🚀

---

For contributor:

For reviewer:


tests/unit/specs/scripts/common.spec.js
```diff
@@ -25,3 +25,22 @@ describe(`coinsToObject`, () => {
...
+
+describe(`toMicroDenom`, () => {
+  it(`returns the right micro denom of a given network`, () => {
+    const microDenom = toMicroDenom(`ATOM`)
+    expect(microDenom).toBe(`uatom`)
+  })
+})
+
+describe(`sleep`, () => {
+  it(`returns a true as a promise`, () => {
```
faboweb: hahahahahaha
this is not what this does
you should test if the promise resolves after a timeout

checkout jest.useFakeTimers and jest.runAllTimers

Bitcoinera: Using jest.runAllTimers was kind of a headache. I think it is still testing the main functionality though

faboweb: the thing you want to test is: `it should resolve after x seconds`. then, inside the test, you use fake timers. now you run sleep with x seconds and store the promise. to later know that the promise resolved or not, you react to it resolving by setting a value when it does `sleep(x).then(() => done = true)`. now that the promise is initialized, we want to go to the future! there is a function `jest.forwardAllTimersBy(x)` or similar. We are now in the future. Now we can check if the promise was resolved by checking `done`

tests/unit/specs/scripts/common.spec.js
```diff
@@ -1,4 +1,11 @@
-import { coinsToObject, calculateShares, toMicroDenom, sleep } from "scripts/common"
+import {
+  coinsToObject,
+  calculateShares,
+  toMicroDenom,
+  sleep
+} from "scripts/common"
+
+jest.useFakeTimers()
```
faboweb: this should go into a test and at the end of the test it should be removed by using `jest.useRealTimers()` just so we avoid side effects

Bitcoinera: Fixed

Bitcoinera:
Hey, I think we forgot about this one! ;)

tests/unit/specs/scripts/common.spec.js
faboweb: @Bitcoinera see this comment?

Bitcoinera: Oh, now I do :+1: missed this one

Bitcoinera:
OK, I actually wasn't able to combine the tests you mentioned in one, and had to split them into two:
One for testing that when it resolves the promise returns `true` and another one to test the timer with the FakeTimers.

I don't see how to combine both in one. Please, show me the way :pray: 

PS: This might be slowly becoming the most tested Promise function ever xD

tests/unit/specs/scripts/common.spec.js
```diff
@@ -39,11 +39,13 @@ describe(`toMicroDenom`, () => {
...
-    expect(setTimeout).toHaveBeenCalledTimes(1)
-    expect(sleep(10)).resolves.toEqual(10)
-    jest.useRealTimers()
+  it(`returns a true as a promise`, async () => {
+    const sleepCheck = () => {
+      return sleep(1000).then(() => {
+        return true
+      })
+    }
+    const res = await sleepCheck()
```
faboweb: you are now blocking all tests by 1s. if you use fakeTimers this test can pass in some milliseconds

Bitcoinera: Yes... but this was one of the first things I tried:

```
    jest.useFakeTimers()
    const sleepCheck = () => {
      return sleep(1000).then(() => {
        return true
      })
    }
    const res = sleepCheck()
    jest.runAllTimers()
    expect(res).toBe(true)
    jest.useRealTimers()
```
And it surprisingly enough returns `{}` instead of `true`. Very unexpected :eyes: 

No undefined, no Promise... empty object :woman_shrugging:

faboweb: No, in the `then` block you need to set an external variable (`finished`)

Bitcoinera: I am not really sure how you want me to do this... 

I have done this way:

```
  it(`the timer works properly, taking as many millisecons as we input and the Promise returns true`, done => {
    let check = false
    const cb = jest.fn(() => (check = true))

    jest.useFakeTimers()
    sleep(10000)
      .then(cb)
      .then(() => expect(cb).toHaveBeenCalled())
      .then(() => expect(check).toBe(true))
      .then(done)

    jest.runAllTimers()
    jest.useRealTimers()
  })
```

I think like this we are properly checking the function: both the timer and the resolved Promise.

faboweb: Looks good.

This didn't work?

```
it(`the timer works properly, taking as many millisecons as we input and the Promise returns true`, ()=> {
    let check = false

    jest.useFakeTimers()
    sleep(10000)
      .then(() => check = true)

    jest.runAllTimers()

    expect(check).toBe(true)

    jest.useRealTimers()
  })
```

Bitcoinera: > Looks good.
> 
> This didn't work?
> 
> ```
> it(`the timer works properly, taking as many millisecons as we input and the Promise returns true`, ()=> {
>     let check = false
> 
>     jest.useFakeTimers()
>     sleep(10000)
>       .then(() => check = true)
> 
>     jest.runAllTimers()
> 
>     expect(check).toBe(true)
> 
>     jest.useRealTimers()
>   })
> ```

Yes, this is one of the first things I tried I was quite weirded out when I saw it didn't work...

But yeah, `check` remains `false` because this part of the code outside the `then` block (here concretely the `expect` line) will execute all before things inside `then` complete. 

It doesn't really matter that is written after the `jest.runAllTimers()` line. Everything outside the `then` block will be executed before.

So we both learned a lot here thanks to this not-so-boring-after-all test :wink:

PR #3390, opened 2020-01-09
Fabo/fetch policy

### LINKED ISSUE #3391: API serves stale balances via GraphQL
Sometimes the ActionModals have stale data, if there was a liveupdate like a tx received.

faboweb:
Closes #3391

instead of setting the fetching policy individually per request this relys on a cache first strategy. the requests go first to the cache so data is there quick. but in the end they still go to the API and if the data changes the components get updated.

https://medium.com/@galen.corey/understanding-apollo-fetch-policies-705b5ad71980


src/ActionModal/components/ActionModal.vue
```diff
@@ -477,7 +477,6 @@ export default {
         userAddress: this.session.address,
         rewards: this.rewards,
         totalRewards: this.overview.totalRewards,
-        delegations: this.delegations,
```
jbibla: how come this is no longer needed?

src/ActionModal/components/ActionModal.vue
```diff
@@ -731,14 +730,14 @@ export default {
...
             liquidStake
-            totalStake
             accountInformation {
               accountNumber
               sequence
             }
           }
         }
       `,
+      fetchPolicy: "cache-and-network",
```
jbibla: very cool. can we set this globablly or does it have to be set per component?

src/components/staking/PageValidator.vue
```diff
@@ -257,21 +256,8 @@ export default {
...
-    },
-    clearUndelegationCache() {
-      this.$store.commit("invalidateCache", [
-        `overview`,
-        `delegations`,
-        `undelegations`,
-        `transactions`
-      ]) // TODO use more finegrained query string (network and address)
-    },
+    clearDelegationCache() {},
```
jbibla: what does this do?

src/vuex/modules/session.js
```diff
@@ -71,19 +71,6 @@ export default ({ apollo }) => {
     },
     setCurrrentModalOpen(state, modal) {
       state.currrentModalOpen = modal
-    },
-
-    // TODO to own store module?
-    // clear the cache manually so we can force reloading of stale data in not mounted components
-    invalidateCache(state, queries) {
```
jbibla: nice.

Bitcoinera:
It is working really well for Send but not so for Staking. It needs the user to pass by Portfolio in order to show updated available Tokens in the DelegateModal

src/ActionModal/components/ActionModal.vue
faboweb: we can, but may want to not do it for things like `networks`. but we could make it the default and set the fetch policy for networks to `cache`

src/components/staking/PageValidator.vue
faboweb: nothing, smarta** :P

src/ActionModal/components/ActionModal.vue
faboweb: I didn't find any use of delegations in the context. 5 withdraws are calculated based on the rewards.

Bitcoinera:
Still not working

faboweb:
This error does look strange. Why would the validator addresses be `cosmos1` based? I am checking it now

faboweb:
Just staked successfully on https://deploy-preview-3390--lunieio.netlify.com
Could you check again?

Bitcoinera:
Sorry, maybe I didn't explain so clearly (actually at all) what that error earlier with the `cosmosvaloper` was about.

The error happens when you try to delegate to a validator with you available balance in the DelegateModal not yet updated.

I am going to upload a video so you can see better the issue here. I just tested and it is still unresolved.

Bitcoinera:
Now it is all fixed! Updating every time :smile: Also in the last case I commented earlier.

Now there is only one issue left: 

on line 346 in PageValidator, this part (data is undefined):

Bitcoinera:
why don't we do it like so, like it is done in PageProposal and PageProposals?



Like this it also checks that the query has been loaded.

faboweb:
Greate idea

PR #3408, opened 2020-01-13
tests updated

iambeone:
Tests update to work with different networks.
Networks configuration in tests/e2e/networks.json
Call npm run test:e2e:local -- --network=1

Thank you! 🚀

---

For contributor:

For reviewer:


faboweb:
Does this test against a live version /deployment preview of Lunie?

tests/e2e/delegation.spec.js
```diff
@@ -1,138 +1,142 @@
...
-    },
-    [selector, option]
-  )
-}
+const { actionModalCheckout, nextBlock, waitForText } = require("./helpers.js")
+
+module.exports = {
+  "Delegate Action": async function(browser) {
+    // move to according page
+    browser.url(browser.launch_url + "/validators", async () => {
```
faboweb: browser.url doesn't return a promise?

tests/e2e/globals.js
```diff
@@ -1,125 +1,174 @@
...
+
+  async before() {
+    await schemaAvailable()
+    await apiUp()
+  },
+
+  beforeEach(browser, done) {
+    let network = 1
+    let networkData = {}
+    process.argv.map(arg => {
```
faboweb: refactor into `getNetworkFromInput`?

tests/e2e/globals.js
```diff
@@ -1,125 +1,174 @@
...
+  { address, network, wallet, password, name, stakeAmount, restakeAmount }
+) {
+  // clear parameters
+  browser.globals.availableAtoms = 0
+  browser.globals.denom = ""
+  browser.globals.password = password
+  browser.globals.stakeAmount = stakeAmount
+  browser.globals.restakeAmount = restakeAmount
+  // test if the test account was funded as we need the account to have funds in the tests
+  axios
```
faboweb: why did you not leverage `async` `await` syntax here?

tests/e2e/signin.spec.js
```diff
@@ -1,133 +1,151 @@
...
-  browser.waitForElementVisible(`body`, 10000, true)
-  browser.waitForElementVisible(`#app-content`, 10000, true)
-  signOut(browser)
-  signIn(browser)
-
-  browser.waitForElementVisible("#session-welcome", 10000, true)
-}
+module.exports = {
+  /*
+  "Sign in with local account": async function(browser) {
```
faboweb: this didn't work? insecure mode not set maybe?

tests/e2e/delegation.spec.js
faboweb: I guess it doesn't

tests/e2e/networks.json
```diff
@@ -0,0 +1,19 @@
+[
```
faboweb: I propose a structure:

```
{
  networkdID: {
    ...networkConfig
  }
}
```

so we can access the network configs directly by id

tests/e2e/globals.js
```diff
@@ -1,125 +1,174 @@
...
+
+  beforeEach(browser, done) {
+    let network = 1
+    let networkData = {}
+    process.argv.map(arg => {
+      if (arg.indexOf("--network=") !== -1)
+        network = arg.slice(arg.indexOf("=") + 1)
+    })
+    if (networks.length >= network) {
+      networkData = networks[network - 1]
```
faboweb: how about using the networkId string instead of an intransparent number

tests/e2e/globals.js
faboweb: I also think this could be refactored into a function `waitForFunds(network, address)`

tests/e2e/globals.js
```diff
@@ -1,125 +1,174 @@
...
+      })
+      // switching to homepage
+      browser.url(browser.launch_url, () => {
+        // beforeEach done
+        done()
+      })
+    })
+}
+
+async function apiUp() {
```
faboweb: if we live test, we don't need this anymore, right?

tests/e2e/globals.js
```diff
@@ -1,125 +1,174 @@
...
+      apiUp = true
+    } catch (err) {
+      console.log("Failed to check API", err)
+      console.log("Waiting 1s for API to be up")
+      await new Promise(resolve => setTimeout(resolve, 1000))
+    }
+  }
+}
+
+async function schemaAvailable() {
```
faboweb: if we live test, we don't need this anymore, right?

tests/e2e/networks.json
faboweb: this file should probably not be checked in or people might f*** with our tests by draining the accounts.

faboweb:
try to create individual PRs for inidivudal things. the api persitence could have been done in a separate PR

src/gql/apollo.js
```diff
@@ -12,8 +12,47 @@ import config from "src/../config"
...
+      "Your API version differ from normal. Query ?iwouldliketochangetheapipersistentlyandiknowwhatido=false to reset"
+    )
+    return localStorage.getItem(`persistentapi`)
+  }
+  return false
+}
+
 const graphqlHost = urlParams =>
-  (urlParams.graphql ? decodeURIComponent(urlParams.graphql) : false) ||
+  checkPersistentAPI(urlParams) ||
```
faboweb: needs to be replicated in `ActionManager`

tests/e2e/globals.js
```diff
@@ -47,9 +82,132 @@ module.exports = {
...
+    [networkData]
+  )
+  // wallet info
+  browser.globals.wallet = tempAcc.value.wallet
+  // address
+  browser.globals.address = tempAcc.value.address
+  // switching to master account
+  await switchToAccount(browser, networkData)
+
+  // funding temp account
```
faboweb: all those should be refactored into functions so this is better to read

tests/e2e/delegation.spec.js
```diff
@@ -103,7 +107,7 @@ module.exports = {
       "0"
     )
 
-    await nextBlock(browser)
+    //await nextBlock(browser)
```
faboweb: if we comment those out, should we remove them?

tests/e2e/delegation.spec.js
```diff
@@ -82,16 +84,18 @@ module.exports = {
   },
   "Undelegate Action": async function(browser) {
     // be sure that the balance has updated, if we don't wait, the baseline (balance) shifts
-    await nextBlock(browser)
+    //await nextBlock(browser)
```
faboweb: if we comment those out, should we remove them?

tests/e2e/globals.js
```diff
@@ -91,85 +45,192 @@ async function next(browser) {
...
+              first: validatorLIs[0].getAttribute("data-name"),
+              second: validatorLIs[1].getAttribute("data-name")
+            })
+          }
+          f()
+        })
+      },
+      [browser, networkData]
+    )
+    browser.globals.validatorOneName = validators.value.first
```
faboweb: this doesn't get deleted on a refresh?

iambeone: No, those variables are global for all tests during test sessions

package.json
```diff
@@ -19,7 +19,7 @@
     "test:e2e": "node tests/e2e/local-run.js",
     "test:e2e:build": "cross-env VUE_APP_E2E=true vue-cli-service build --mode production",
     "test:e2e:browserstack": "node tests/e2e/browserstack.runner.js -c  tests/e2e/browserstack.conf.js",
-    "test:e2e:local": "kill-port 9515 && nightwatch --config ./tests/e2e/local-chrome.conf.js",
+    "test:e2e:local": "nightwatch --config ./tests/e2e/local-chrome.conf.js",
```
faboweb: why did you remove this?

tests/e2e/signin.spec.js
```diff
@@ -17,43 +16,7 @@ module.exports = {
     await browser.waitForElementNotPresent(".session", 10000, true)
     openMenu(browser)
     await browser.waitForElementVisible("#sign-out", 10000, true)
-  },
-  */
-  "Create local account": async function(browser) {
```
faboweb: Why did you remove this?

package.json
iambeone: Sorry, my fault. kill-port always needs more than 3 minutes to be executed, so that's why i'm removing it always. Forgot to add it back

tests/e2e/signin.spec.js
iambeone: There is no "Sign in with local account" even with insecure flag. Don't know why

package.json
faboweb: on Windows. Maybe there is a better plugin.

faboweb: https://github.com/tiaanduplessis/kill-port/pull/39 :/

tests/e2e/helpers.js
```diff
@@ -37,6 +37,31 @@ async function waitFor(check, iterations = 10, timeout = 1000) {
   console.error("Condition was not meet in time")
   process.exit(2) // exiting here so the e2e tests actually fail. if not they pass
 }
+
+async function fundMasterAccount(browser, network, address) {
```
faboweb: how does this scale?? will you create a funding function for every network? why not stick with the wallet strategy from before?

tests/e2e/globals.js
```diff
@@ -274,6 +283,11 @@ async function createAccountAndFundIt(browser, done, networkData) {
   await storeAccountData(browser, networkData)
   // switching to master account
   await switchToAccount(browser, networkData)
+  // funding main account
+  console.log(browser.globals.availableAtoms)
```
faboweb: remove? or if needed add some information what this value is

tests/e2e/interface.spec.js
```diff
@@ -0,0 +1,96 @@
+module.exports = {
+  "Validators search": async function(browser) {
+    // move to according page
+    browser.url(browser.launch_url + "/validators", async () => {
+      // wait until validators list load
+      await browser
+        .execute(function() {
+          return new Promise(resolve => {
+            let attempts = 5
+            const f = () => {
```
faboweb: no abbreviations

tests/e2e/interface.spec.js
```diff
@@ -0,0 +1,96 @@
+module.exports = {
+  "Validators search": async function(browser) {
+    // move to according page
+    browser.url(browser.launch_url + "/validators", async () => {
+      // wait until validators list load
+      await browser
+        .execute(function() {
+          return new Promise(resolve => {
```
faboweb: use async await to avoid nesting hell

tests/e2e/interface.spec.js
```diff
@@ -0,0 +1,96 @@
+module.exports = {
```
faboweb: please refactor this file. this is really hard to read

tests/e2e/signin.spec.js
```diff
@@ -104,22 +114,60 @@ function signIn(browser) {
...
+          JSON.stringify({
+            name: `rich_account`,
+            address: `cosmos1ek9cd8ewgxg9w5xllq9um0uf4aaxaruvcw4v9e`,
+            wallet: `ae1d20a49e1085cca29a71e270c6f64f8f86794cb67c6922caea6bcba0ed9e60g+nSTgP8/wHpWaomDkhW/7g2Xldvno3VRFggvdpWIDrRV+n4BJtpk3UpLKo0K3SDL5dRzxz3NmGFnSA8znggFmtesdqu6jWJYzSNqaQhM/gCPTVabF7t1UHaybze1NRlYcm/wl5oOyXRpki6ugOHxNhF7+4wlzhYxMilAB7ekDB4+VVHoPMUinU4dsUdtC4XwDUA0rbX1TTmrh+i1eBp6UTQ+nHGiZXL1TkhhR1mE0fR3bLRunz5XagYtjoST33pecQWzqeaZZQ/mgm9QXu/i+ymfbnPQkh8ivx+J6/d2RfZuAV4NnwFZDUr7CzPX4TU`
+          })
+        )
+        return true
+      }, [])
+      await browser.refresh()
+      openMenu(browser).then(async () => {
```
faboweb: can't await here?

tests/e2e/signin.spec.js
```diff
@@ -104,22 +114,60 @@ function signIn(browser) {
...
+        // check if we are already singed in
+        const isSigned = await browser.execute(function() {
+          return new Promise(resolve => {
+            let attempts = 3
+            const f = () => {
```
faboweb: can't you make this more linear? like in a simple async while loop?

tests/e2e/signin.spec.js
```diff
@@ -104,22 +114,60 @@ function signIn(browser) {
...
+              const signOutElement = document.querySelector(
+                ".user-box-address #sign-out"
+              )
+              const singInElement = document.querySelector(".app-menu #sign-in")
+              if (!signOutElement < 2 && attempts-- > 0) {
```
faboweb: why are there 2 #sign-in elements?

faboweb:
There are conflicts on here and the unit tests are failing

tests/e2e/networks.json
faboweb: ok, we can leave the wallets in for now. I think it is cleaner though if we remove the whole file and add it via env variable later

tests/e2e/helpers.js
iambeone: Yes, I suppose to or it could just notify us somehow. This function only trigger when master account has a low balance. Newly created accounts are still funded from master, this function is only for funding master account

tests/e2e/globals.js
iambeone: Sorry, this needs to be removed

tests/e2e/interface.spec.js
iambeone: You mean f()? But this is a common name for a function that will be executed only in this block. Any suggestions about the name?

tests/e2e/signin.spec.js
iambeone: That is sign-it and sign-out. For some reasons, we can't just rely on getting sign-in element cause it's always showing before it will be replaced with sign-out. So we need before sign-out will show (or not)

Bitcoinera:
> There are conflicts on here and the unit tests are failing

Strangely I have seen no conflicts when merging and unit tests are fine in local :thinking:

tests/e2e/helpers.js
faboweb: this only works for gaia testnet is what I meant. so this should if we want to keep this be refactored into some structure that tells us this only handles gaia testnet.

tests/e2e/interface.spec.js
faboweb: see my refactoring removing the need for this. We have a code guideline to never use abbreviations. Always describe what this does. If you can't describe what this does, this probably needs to be refactored.

tests/e2e/signin.spec.js
faboweb: But we can rely on the sign-out element as it only shows when the user is signed in, right?

faboweb:
changelog is missing

PR #3421, opened 2020-01-14
Fabo/allow to sign messages for Terra

faboweb:
Architecture for creating messages cross networks


src/ActionModal/utils/MessageConstructor.js
```diff
@@ -62,174 +56,31 @@ export const getMultiMessage = async (context, messages) => {
...
-      amount: amounts.map(Coin)
-    }
-  }
-}
+async function getMessageFormatter(network, messageType) {
+  let networkMessages
+  try {
+    networkMessages = await import(`./networkMessages/${network}.js`)
+  } catch (err) {
+    throw new Error("Signing for this network is not enabled.")
```
jbibla: this error seems like it's for a user - but could a user encounter it? feels like they shouldn't get to the point where we're requesting networkMessage schema if transactions aren't enabled.

src/ActionModal/utils/MessageConstructor.js
```diff
@@ -62,174 +56,31 @@ export const getMultiMessage = async (context, messages) => {
...
-  return {
-    type: `cosmos-sdk/MsgDelegate`,
-    value: {
-      delegator_address: senderAddress,
-      validator_address: validatorAddress,
-      amount: Coin({ amount, denom })
-    }
+  if (!messageFormatter) {
+    throw new Error(
+      `Creating the message ${messageType} for the network ${network} is not supported`
```
jbibla:
```suggestion
      `Creating the '${messageType}' message for the ${network} network is not supported`
```

jbibla: 😸

jbibla: on second thought: these are user error messages so i think they should be helpful and actionable. maybe we could say "there's currently an issue with x transactions on this network. the lunie team is working on a fix. please try again later"

would that make sense?

src/ActionModal/utils/networkMessages/terra-testnet.js
```diff
@@ -0,0 +1,9 @@
+import * as CosmosMessages from "./cosmos-hub-mainnet.js"
```
jbibla: should this come from the API?

is `msg.type` the only thing that changed or are the data structures different?

jbibla:
this is awesome. i can feel us _scaling rapidly_ 🚀

src/ActionModal/utils/MessageConstructor.js
faboweb: True! This is not only for the user but also a good message for us developers. :)

src/ActionModal/utils/networkMessages/terra-testnet.js
faboweb: I don't know. Maybe there could be structural differences down the line. I didn't even check all Terra transactions yet.

src/ActionModal/utils/MessageConstructor.js
faboweb: Same here: We should prevent this from ever surfacing to the user. The error is a backup

src/ActionModal/components/SendModal.vue
```diff
@@ -271,20 +275,6 @@ export default {
         )
       }
     },
-    setTokenAndBalance() {
```
jbibla: isn't this needed for multi denom support?

faboweb: this is a computed now

PR #3424, opened 2020-01-15
Implement mobile Intercom using vuex store

### LINKED ISSUE #3423: Modify mobile Intercom implementation to use vuex store.

We should modify mobile Intercom implementation to use vuex store.

**Additional context**
https://github.com/luniehq/lunie/issues/3326
https://github.com/luniehq/lunie/pull/3419

mariopino:
Closes #3423 

**Description:**

Implement mobile Intercom using vuex store, Intercom api keys for Android and IOS should be at top level `capacitor.config.json` (platform specific capacitor config files get overwritten by top level one).

Thank you! 🚀

---

For contributor:

For reviewer:


android/app/src/main/assets/capacitor.config.json
```diff
@@ -7,7 +7,9 @@
   "plugins": {
     "IntercomPlugin": {
       "android-apiKey": "android_sdk-xxxx",
-      "android-appId": "yyyy"
+      "android-appId": "yyyy",
+      "ios-apiKey": "ios_sdk-xxxx",
```
jbibla: probably don't need to add the ios stuff (or anything) here if it gets overridden by the top level capacitor.config

src/vuex/modules/intercom.js
```diff
@@ -0,0 +1,33 @@
...
+  return {
+    state: {
+      intercom
+    },
+    mutations: {
+      displayMessenger(state) {
+        state.intercom.displayMessenger()
+      },
+      displayHelpCenter(state) {
+        state.intercom.displayHelpCenter()
```
jbibla: what does this do?

mariopino: It opens the help center (I guess is the same thing you talked about the last demo day conf). 

What do you think about? We can remove it if it's not going to be useful.

android/app/src/main/assets/capacitor.config.json
mariopino: That's the reason why that file has this content (the top-level config file has it).

src/vuex/modules/intercom.js
faboweb: as long as it is not used, we should not have it in code. But nice to know this is possible.

src/vuex/modules/intercom.js
```diff
@@ -0,0 +1,34 @@
...
+      }
+    },
+    actions: {
+      displayMessenger({ commit }) {
```
faboweb: I guess we only need mutations or actions for this ^^ I guess only actions

mariopino: It put the mutations there cause I think not having it was causing problems with `index.spec.js` test, but that wasn't the cause of the problems :-).

I needed to modify transformIgnorePatterns in jest config to whitelist `capacitor-intercom` package, the problem is that this modification causes the tests to run very slow (>5min to complete). I want to figure out how to fix that.

```
transformIgnorePatterns: ["/node_modules/?!(capacitor-intercom)"],
```

faboweb: Have you considered mocking the `capacitor-intercom` module with `jest.mock` where it is imported instead of ignoring the transform?

mariopino: Thanks, good point!

mariopino: This is done and great! testing back to normal speed ;-)

faboweb:
@mariopino do you also want to take the lead on refactoring my small intercom keys script to only use the top level config file?

mariopino:
> @mariopino do you also want to take the lead on refactoring my small intercom keys script to only use the top level config file?

Sure!

PR #3430, opened 2020-01-17
Fabo/emoney-signing

faboweb:
This enables signing on the emoney testnet


faboweb:
there is a pending added

Bitcoinera:
We just need to take into account the Martin said that denoms will change and won't have the "x" something at the beginning (the "exponents").

I guess they will end up looking like they do here, with just the "e": https://api.e-money.com/v1/rates.json

faboweb:
Thank you for clarifying Ana. Let's change that on the next testnet?

Issue #3402, opened 2020-01-10
Address prefixes should be received from API


faboweb:
**Is your feature request related to a problem? Please describe.**

To be able to scale networks without needing to update the client we should deliver as much data from the API as possible. Currently the bech 32 prefixes for the Ledger handling are in config.js. These should be received from the API (networks query). Same goes for identifying valid addresses in the explore input.

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

Bitcoinera:
Actually this is already working xD

I think we just didn't notice... :woman_shrugging: 

<img width="530px" src="https://user-images.githubusercontent.com/40721795/72824338-60bc1300-3c75-11ea-85cf-99b1ad68815a.png">


I guess will need to refactor a couple of things

faboweb:
But this is not used in the FE. We are hardcoding network to prefix lookup rn.

Bitcoinera:
> But this is not used in the FE. We are hardcoding network to prefix lookup rn.

Exactly. Today I am making a PR to refactor this

jbibla:
this is done already?

Bitcoinera:
> this is done already?

Yes, we just need to merge it. It is [here](https://github.com/luniehq/lunie/pull/3456)

PR #3438, opened 2020-01-19
Ana/filter addresses per network in session explore

### LINKED ISSUE #3389: When you sign in with an address, Lunie directly connects you to that address network

Now (or very very soon) we can sign in with different networks addresses.

Still, when we sign in with an address, if we weren't previously connected to that address network, we get very ugly errors that the user will not even see and it will certainly be utterly confusing for them.

It is the title of this feature request :smiley_cat: 

Warning messages when you are signing in with other networks addresses (which is not even a solution) or, another option would be that you can only sign in with addresses from the network you are currently connected to.

But this last option I don't like as much as the one I propose here, which just makes the UX a breeze. 

**Additional context**

This is what happens now when you sign in with another network address. Not ideal.

Bitcoinera:
Closes #3389 

**Description:**

This is just a temporary fix. I still think that the best UX is what I originally proposed to handle networks/sessions. That is, that when you choose an address, this directly connects you to the network it belongs to.

But yes, implementing this is definitely tricky :woman_shrugging: 

I think it requires a change on how we have the vuex modules working right now, on their architecture.

Another thing I realized was if it makes sense to distinguish between addresses from `cosmos-hub-testnet` and `cosmos-hub-mainnet`. Here I think it would be cool to have like some kind of "Testnet" tag on the address UI.

I can add this now :wink: 

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/src/components/common/TmSessionExplore.vue
+++ b/src/components/common/TmSessionExplore.vue
@@ -74,6 +74,7 @@
 
 <script>
 import { mapState, mapGetters } from "vuex"
+import config from "src/../config"
 import { required } from "vuelidate/lib/validators"
 import TmBtn from "common/TmBtn"
 import SessionFrame from "common/SessionFrame"
@@ -107,9 +108,10 @@ export default {
     ...mapState([`session`]),
     ...mapGetters([`network`]),
     filteredAddresses() {
-      const networkPrefix = this.network.split(`-`)[0]
       return this.session.addresses
-        .filter(address => address.address.startsWith(networkPrefix))
+        .filter(address =>
+          address.address.startsWith(config.bech32Prefixes[this.network])
+        )
         .slice(-3)
     }
   },
--- a/src/components/common/TmSessionExplore.vue
+++ b/src/components/common/TmSessionExplore.vue
@@ -7,7 +7,7 @@
 
       <div v-if="session.addresses.length > 0" class="session-list">
         <div
-          v-for="account in session.addresses.slice(-3)"
+          v-for="account in filteredAddresses"
           :key="account.address"
           :title="account.address"
           @click="exploreWith(account.address)"
@@ -73,7 +73,7 @@
 </template>
 
 <script>
-import { mapState } from "vuex"
+import { mapState, mapGetters } from "vuex"
 import { required } from "vuelidate/lib/validators"
 import TmBtn from "common/TmBtn"
 import SessionFrame from "common/SessionFrame"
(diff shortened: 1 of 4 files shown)


Bitcoinera:
I just realized that adding the testnet tag is a bit more complicated, since there is nothing in the `session.addresses` that indicates if it is coming from a testnet or not (all the Cosmos addresses are mixed up). Also maybe this feature doesn't make too much sense, since like in all the other blockchains, in Cosmos we can also use addresses from both mainnet and testnet interchangeably.

Too bad, I also had added some cool tags :sweat_smile:

src/components/common/TmSessionExplore.vue
```diff
@@ -104,7 +104,14 @@ export default {
     error: ``
   }),
   computed: {
-    ...mapState([`session`])
+    ...mapState([`session`]),
+    ...mapGetters([`network`]),
+    filteredAddresses() {
+      const networkPrefix = this.network.split(`-`)[0]
```
faboweb: this doesn't work reliably. It only be chance works with the current networks. What about Regen i.e.. Their address prefix is `xrn:`. You need to do an address lookup.

Btw.: The address lookup is needed in several cases. We should refactor it and put it someplace. Maybe the store?

Bitcoinera: Wow, you are right! How could I miss the regen addresses?

OK, let me refactor this. Yes, maybe it could be one of those `store.dispatch` functions

Bitcoinera: I found we already this address lookup in place. It exists in the `config.js` file and it is called `bech32Prefixes`

faboweb:
Does this validate the prefix for the currently selected network?

Bitcoinera:
> Does this validate the prefix for the currently selected network?

It uses network id as a whole to look up for the prefix. Nothing fancy, just a dictionary

faboweb:
I meant the input. If you add a Terra address in there while signing in to Cosmos

src/components/common/TmSessionExplore.vue
faboweb: Needs to move out of there: #3402

faboweb: But we can do that in another PR

Bitcoinera:
> I meant the input. If you add a Terra address in there while signing in to Cosmos

Ah ok. No, it doesn't. Right now it is working like you can create any address connected to any network. Which is one the reasons I am not sure about this flow...

Maybe for a future, when we have kind of an onboarding part of the app, where the user goes and clicks a "Create a new address" button and then chooses the network. For the extension when the Lunie app is not open this flow also makes sense. Also for Backup like you suggested.

Bitcoinera:
> I meant the input. If you add a Terra address in there while signing in to Cosmos

OK, gotcha (sorry for the confusion). 

Hey, brilliant idea, I will definitely include this right now in this PR :+1:

Bitcoinera:
Great. I think this is ready to be merged. Incredible what a few lines of code can do :smiley:

faboweb:
Nice! Got my approval. @jbibla anything about the copy?

faboweb:
Note: The filter needs to be also applied for the extension sign in step.

Issue #3213, opened 2019-11-27
add network title to urls in lunie


jbibla:
**Is your feature request related to a problem? Please describe.**

as we add more networks and links continue to be shared, we have to make sure they can resolve correctly. currently, there is no way to share a link for a specific network. if you share a proposal for example, but you have a different network selected it will just say "no data available". 

**Describe the solution you'd like**

let's add the network to the url for all supported networks
```
app.lunie.io/cosmos-hub/portfolio
app.lunie.io/cosmos-hub/validators
app.lunie.io/cosmos-hub/proposals/14
app.lunie.io/cosmos-hub (network page)
```

faboweb:
Thought: we also want to give a short url to validators in the future. Validators exist cross network so prefixing them with a chain name doesn't work.
Proposal: if a short url like app.lunie.io/cosmos-hub-mainnet is called the slug is analyzed and then checks if this is a network name or validator id, forwarding to the validators profile in case.
So: this is not a problem for this implementation.

jbibla:
good thought. i think what you're describing is a different feature.

> Validators exist cross network so prefixing them with a chain name doesn't work.

right now, profiles aren't "cross network" they will display data per network - so i think this should still be implemented as described.

faboweb:
Which I shared was just trying to access future scenarios and how they fit with this strategy

jbibla:
go away stalebot

PR #3439, opened 2020-01-19
Ana/add select network in signup flow

Bitcoinera:
Closes #ISSUE

It is related to [this proposal](https://github.com/luniehq/lunie-browser-extension/issues/124#issuecomment-575910068).
Although this one references the Lunie Browser Extension.
  
**Description:**

In this PR I add the `TmSelectNetwork` component into the sign up flow. It is working but the approach I used might not be the best. Still, this way we can discuss it.

How I did it: 
I used the routes to pass along the whole sign up flow the chosen network prefix. This is the only way to pass this data I could come up with. Suggestions and other ideas super welcome.

Things that are missing:
- Tests
- Connect automatically to the chosen network right after the account creation
- fix the `route` type in `TmLiSession`. Unfortunately we lack of type `any`, which I find super useful.

That's mostly it.

This is the new component I created :heart_eyes: 
(ready to add it to the Lunie Browser Extension! :muscle: )

EDIT: One thing that is not so good of this approach is that it always forces the user to choose a network, instead of defaulting to Cosmos Hub, for instance. So probably it is better like proposed by @faboweb and just present this choice when we are importing or recovering an account :thinking: 

Ok, so let's take this PR for the time being just like an experiment :laughing: 

Thank you! 🚀

---

For contributor:

For reviewer:


src/components/common/TmSelectNetwork.vue
```diff
@@ -0,0 +1,45 @@
...
+import gql from "graphql-tag"
+
+export default {
+  name: `select-network`,
+  components: {
+    SessionFrame,
+    LiSession
+  },
+  data: () => ({
+    networks: []
```
faboweb: can you sort the networks by putting the non-testnets first. And the `default` one on the top?

src/components/common/TmSessionSignUpPassword.vue
```diff
@@ -111,10 +111,13 @@ export default {
     }
   },
   methods: {
+    prefix() {
+      return this.$route.params.prefix
```
faboweb: how about passing along the network id. then here we do another lookup for network (using the same gql query). from that object you now extract the prefix.

src/components/common/TmSessionSignUpPassword.vue
```diff
@@ -111,10 +111,13 @@ export default {
     }
   },
   methods: {
+    prefix() {
```
faboweb: better make this a computed? this way you avoid needing to call a function and can just assume that this value is set.

src/vuex/modules/keystore.js
```diff
@@ -73,29 +73,38 @@ export default () => {
 }
 
 // creates a cosmos addres for the network desired
-function getCosmosAddressCreator(network) {
+function getCosmosAddressCreator(network, networkPrefix) {
```
faboweb: here you are duplicating code. why not default to cosmos
```suggestion
function getCosmosAddressCreator(network, networkPrefix = config.bech32Prefixes[network]) {
```

src/components/common/TmSessionSignUpPassword.vue
Bitcoinera: I first set it as a computed value in fact, but strangely enough it didn't work :woman_shrugging: 
Maybe `computed` executes even before that `this.$route` is set?

faboweb: computed executes on every change of the contained values. So you also need to account for the parameter not being set, which is ok as this would default to `undefined` in this case

src/vuex/modules/keystore.js
Bitcoinera: But there you are not defaulting to cosmos right?

To default to cosmos it would be `networkPrefix = 'cosmos'`

And then the other `network` parameter becomes superfluous

src/vuex/modules/keystore.js
```diff
@@ -73,29 +73,38 @@ export default () => {
...
   return async seedPhrase => {
     const { getNewWalletFromSeed } = await import("@lunie/cosmos-keys")
-    return getNewWalletFromSeed(seedPhrase, config.bech32Prefixes[network])
+    if (networkPrefix) {
```
faboweb:
```suggestion
    networkPrefix = networkPrefix || config.bech32Prefixes[network]
```

src/vuex/modules/keystore.js
faboweb: ur right, added another recommendation for slicker code

src/vuex/modules/keystore.js
Bitcoinera: Yes, that definitely looks good

src/components/common/TmSessionSignUpPassword.vue
Bitcoinera: Although computationally speaking it is better the URL solution, no?

This way we don't have weird URLs... but we still need to fetch twice or three times for the same thing along the flow :woman_shrugging:

src/components/common/TmSelectNetwork.vue
Bitcoinera: It seems we need to reimplement the `default` field. Got somehow lost along the way

Bitcoinera:
:heart_eyes:

src/components/common/TmLiSession.vue
```diff
@@ -20,14 +23,18 @@ export default {
...
+      type: String,
+      default: ``
     },
     title: {
       type: String,
       required: true
     },
     route: {
-      type: String,
+      type: String || Object, // this is not working. Where is type any
```
jbibla: there is a way to have `or` in types - but it smells like a bad design

src/components/common/TmSelectNetwork.vue
```diff
@@ -0,0 +1,64 @@
+<template>
+  <SessionFrame>
+    <div class="select-network">
+      <h2 class="session-title">Select a Network</h2>
+      <div class="select-networks-list">
+        <LiSession
```
jbibla: i think we should be re-using the network list and network item components instead of this.

src/components/common/TmLiSession.vue
faboweb:
```suggestion
      type: String
```

src/components/common/TmSelectNetwork.vue
jbibla: this logic is duplicated on the networks page. let's try to re-use.

src/components/common/TmSelectNetwork.vue
```diff
@@ -0,0 +1,64 @@
...
+  name: `select-network`,
+  components: {
+    SessionFrame,
+    LiSession
+  },
+  data: () => ({
+    networks: []
+  }),
+  computed: {
+    sortedNetworks() {
```
jbibla: this logic is duplicated on the networks page. let's try to re-use.

src/components/common/TmSessionSignUpPassword.vue
faboweb: apollo caches the data, so there is no refetch (if the apollo queries have `fetchPolicy: "cache-first",` set)

src/routes.js
```diff
@@ -118,7 +118,17 @@ export default [
     }
   },
   {
-    path: `/create`,
+    path: `/selectnetwork`,
```
jbibla:
```suggestion
    path: `/network`,
```
won't this resolve to `/session/network`?

src/routes.js
```diff
@@ -118,7 +118,17 @@ export default [
...
+    name: `/select-network`,
```
jbibla:
```suggestion
    name: `select-network`,
```

faboweb:
Great suggestions @jbibla

src/components/common/TmLiSession.vue
Bitcoinera: It will still throw an error, since I am sending an Object from `TmSelectNetwork`...

src/routes.js
```diff
@@ -128,7 +138,7 @@ export default [
     }
   },
   {
-    path: `/create/password`,
+    path: `/create/:networkId/password`,
```
jbibla: why only add `networkId` to password and create?

#3213 ?

src/components/common/TmLiSession.vue
faboweb: but you shouldn't

src/routes.js
faboweb: Didn't you have something on this? Mind to push a PR?

src/vuex/modules/keystore.js
```diff
@@ -73,29 +73,35 @@ export default () => {
...
-    case "regen-testnet":
-    case "regen-mainnet":
-    case "terra-testnet":
-    case "emoney-testnet":
-    case "emoney-mainnet":
-    case "terra-mainnet": {
-      const addressCreator = await getCosmosAddressCreator(network)
-      return addressCreator(seedPhrase)
+async function getWallet(seedPhrase, network, networkPrefix) {
+  if (networkPrefix) {
```
jbibla: can you explain what this is about?

Bitcoinera: this is about sending the right prefix to cosmos-keys to create the address

src/components/common/TmSelectNetwork.vue
Bitcoinera: Yes, makes a lot of sense

src/components/common/TmLiSession.vue
Bitcoinera: send an Object you mean? I could stringify it, yes, but what happens when we do send a string? :woman_shrugging:

src/vuex/modules/keystore.js
faboweb: but why are you moving around the network check in line 89?

src/components/common/TmLiSession.vue
faboweb: no you should not send an object at all. you should set an id and then query for the object in the route / component that you want the object to access from

Bitcoinera:
? Some comments just disappeared? What happened??

The ones about `keystore.js` :crying_cat_face: 

I don't know if it has been my fault... In any case, I was answering Fabo when he has asked about me moving around the check (`network` parameter) at line 89 in `keystore.js`:

"Ah ok. My thought here was like, if we are calling the function without passing through selectnetwork, and therefore we don't get the networkPrefix parameter, then do it this other way just by using the network (the logic we had before).

But as you have made clear, the user will always go through the SelectNetwork component in the flow of creating an account. So we can just delete those lines and keep networkPrefix as the second parameter and delete network.

Did I answer your question?"

src/vuex/modules/keystore.js
Bitcoinera: Ah ok. My thought here was like, if we are calling the function without passing through selectnetwork, and therefore we don't get the networkPrefix parameter, then do it this other way just by using the network (the logic we had before).

But as you have made clear, the user will always go through the SelectNetwork component in the flow of creating an account. So we can just delete those lines and keep networkPrefix as the second parameter and delete network.

Did I answer your question?

src/components/common/TmLiSession.vue
Bitcoinera: I fixed it a bit differently now. You can have a look. Thanks to @jbibla suggestion, I now kill two birds with one shot (WARNING: directly translated idiom):

When the user selects the network in TmSelectNetwork, they get connected to that network. So when I am creating the address in TmSessionSignUpSeed, I just grab `connection` and that it. Fixes passing along the info and getting connected to the right network in one shot :wink:

src/components/common/TmSelectNetwork.vue
Bitcoinera: I am afraid this is quite difficult to do otherwise... The logic you are referring to is located in the PageNetworks component. But then, how can I use `@click=$router.push(`/create`)` on NetworkItem?

I hope I can explain well enough where the difficulty lies... if I just import PageNetworks in its totality, then how can I modify the inner components PageNetworks brings within? Otherwise the user can just click anywhere in the component (without selecting any network) and then make it to create an account.

src/routes.js
Bitcoinera: ? Why would it? I mean, `/session/network` doesn't even exist as a route right?

src/routes.js
Bitcoinera: These routes are now back to "normal". I went for a completely different solution ([see](https://github.com/luniehq/lunie/pull/3439#discussion_r368981255))

src/components/common/TmLiSession.vue
jbibla: kill two birds with one ~shot~ stone 😉

src/components/common/TmSelectNetwork.vue
jbibla: i don't know what you mean about the router - but it's fine. `LiNetwork` should be used for sure.

src/components/common/TmSessionSignUpPassword.vue
Bitcoinera: Thank you, I didn't know that

Bitcoinera:
I've got a question: this way the signup flow is working now is really good... but thinking now when we add it to the Lunie Browser Extension... does the Lunie Extension handle `connection` like Lunie Core? What happens if the user doesn't have the Lunie app open in the browser at that time? Maybe we should rethink how this is working now?

Sorry, it was more than one question :laughing:

src/components/common/TmSelectNetwork.vue
Bitcoinera: I was meaning the `@click=$router.push`directive :wink:

src/routes.js
jbibla: ah my mistake. i thought theses were child routes...

src/vuex/modules/keystore.js
faboweb: checkout my version here: https://github.com/luniehq/lunie/pull/3456/files#diff-5bdea6c484046b75be42e1cf57c1167bR81
I expect the same flow for selecting the network. It will not pass the prefix around. It will ONLY pass the network ID around. The function that needs more information will then query for the network information like the prefix it needs.

faboweb:
What exactly do you mean by `connection`? The signup flow in the extension ALWAYS shows the network selector.

src/components/common/TmSelectNetwork.vue
faboweb: `NetworkList` could actually be used and refactored so it takes an ´onClick` handler. `PageNetworks` then implements the `onClick` handler

Bitcoinera: Done! :muscle: 

Now it looks like this (guess we need to tweak the styles a bit):

src/vuex/modules/keystore.js
Bitcoinera: It just refactored it and it is working now like in your PR :wink:

src/components/network/NetworkItem.vue
```diff
@@ -5,7 +5,7 @@
   >
     <div class="network-icon">
       <img
-        :src="`img/networks/${network.id}.png`"
+        :src="`${network.icon}`"
```
Bitcoinera: nice

faboweb: where is the icon comming from?

faboweb: nice, but remember to do smaller PRs and don't do everything in one PR. this now requires the API to be deployed before this PR

src/components/common/TmLiSession.vue
faboweb: what is the flow here, if you start the flow and the stop? you are now connected to the target network. isn't that confusing?

faboweb:
unit tests are broken and changelog is missing

faboweb:
added some refactors, added tests, added changelog

src/components/common/TmSelectNetwork.vue
```diff
@@ -47,12 +47,16 @@ export default {
       // sorts networks setting showing the current network first, mainnets at the top and the default one the first
       if (this.networks.length > 0) {
         return [
+          // current network first
           this.networks.find(({ id }) => id === this.networkId),
```
Bitcoinera: nice!

src/components/network/NetworkItem.vue
Bitcoinera: The API PR for this was really concise and has already been merged! :wink: 

You are right this PR has perhaps grown too big, but we needed it like this in order to have the submodule to work on the extension concurrently.

Bitcoinera:
I have one concern here... isn't creating an address in Cosmos Hub (mainnet) and in Gaia (testnet) actually the same thing? :woman_shrugging: 

I have two accounts I am using in both... at the end of the day it is just creating a Cosmos address, so the testnet thing I see it only makes sense as an informative remark when the network is still only a testnet (like all the other networks but Livepeer). And then to create an address in Cosmos/Gaia we just say "Cosmos".

What do you think about this?

faboweb:
I think these networks should actually have different prefixes to prevent users from fuck up by sending to the wrong addresses. But you are right. If you use the Ledger i.e. you will create the same address for all the networks (just encoded differently).

Bitcoinera:
> I think these networks should actually have different prefixes to prevent users from fuck up by sending to the wrong addresses.

No, but I already tried it and you don't f*** anything up there. The address exists the same in both gaia and Cosmos. I don't see where you can actually go wrong.

So, for example, you give me by error your gaia address and I send you money, no problem. You just connect to the cosmos-hub-mainnet (if you are using Lunie) and there you have your funds :wink: 

This is also how Ethereum (and I think also Bitcoin) works

faboweb:
I think your right. The only thing is though, that if we show the address without the according network, it might be confusing. You are right that the address can be used for all the networks, but maybe it is better to reduce this option for the user to provide a simpler user experience. But worth discussing this.

.circleci/config.yml
```diff
@@ -61,6 +61,10 @@ jobs:
       - attach_workspace:
           at: *WORKSPACE
       - checkout
+      - run:
+        name: "Setup GraphQL endpoint"
+        command: |
+          echo 'export VUE_APP_GRAPHQL_URL="https://production-api.lunie.io/"' >> $BASH_ENV
```
faboweb: this should come from an environment variable set in CircleCI not in plain text.

faboweb: the build is actually not triggered here. it is in Netlify. There this variable is already set. Why did you want to set it for testing?

Bitcoinera: Yes, I already realized these commits have been of little use :woman_shrugging:

jbibla:
> isn't creating an address in Cosmos Hub (mainnet) and in Gaia (testnet) actually the same thing?

yes. creating an address creates a public and private key pair which is valid on both mainnets and testnets.

> And then to create an address in Cosmos/Gaia we just say "Cosmos".

we are asking the user to select a network first. there are other advantages to doing this - like being able to know which network they intend to use and where to take them after they create an address.

i have a feeling grouping the create process together for testnets and mainnets will make our code more complicated and confuse the user. just a feeling though.

Bitcoinera:
> we are asking the user to select a network first. there are other advantages to doing this - like being able to know which network they intend to use and where to take them after they create an address.

We could add at the end of the flow a choice between `mainnet` and `testnet` so the user lands directly there. Like it does now. Only when the network has both mainnet and testnet.

Looking at Charlie's flow, I really like to see only one instance of each network. It is very visual and simple the way he has put it... but maybe it is not such a big improvement for the UX doing it like this.

jbibla:
> I really like to see only one instance of each network. It is very visual and simple the way he has put it

me too. we can leave testnets as second-class networks for advanced users. most people don't care / don't use them anyways.

Bitcoinera:
> me too. we can leave testnets as second-class networks for advanced users. most people don't care / don't use them anyways.

Should we implement an `Advanced` mode in Lunie, kind of like the one that Chrome/Brave has :smiley:  :rocket: ?

PR #3448, opened 2020-01-20
Fabo/handle some failing queries

### LINKED ISSUE #3447: Failed Fetch bug (Sentry)
I suppose most of these result from the `update` function not properly handling a case where there is no value, like when the fetch fails.

faboweb:
Closes #3447

**Description:**

Handles the case in the GQL update functions that the query fails

Thank you! 🚀

---

For contributor:

For reviewer:


src/ActionModal/components/ActionModal.vue
```diff
@@ -746,6 +742,11 @@ export default {
         }
       },
       update(data) {
+        if (!data.overview) {
+          return {
+            totalRewards: 0
```
jbibla: shouldn't the API always return this, instead of adding conditionals to all our queries?

src/ActionModal/components/DelegationModal.vue
```diff
@@ -385,6 +385,9 @@ export default {
           address: this.address,
           denom: this.denom
         }
+      },
+      update(data) {
+        return data.balance || { amount: 0 }
```
jbibla: shouldn't the API always return this, instead of adding conditionals to all our queries?

src/ActionModal/components/UndelegationModal.vue
```diff
@@ -324,6 +324,33 @@ export default {
         return data.delegations
       }
     },
+    balance: {
+      query: gql`
+        query BalanceModalPropos(
```
jbibla: what is `BalanceModalPropos` ?

src/ActionModal/components/UndelegationModal.vue
```diff
@@ -364,7 +391,7 @@ export default {
       },
       update(data) {
         /* istanbul ignore next */
-        return data.validators
+        return data.validators || []
```
jbibla: shouldn't the API always return this, instead of adding conditionals to all our queries?

src/ActionModal/components/ActionModal.vue
faboweb: what happens if the fetch from the API fails though?

src/ActionModal/components/UndelegationModal.vue
faboweb:
```suggestion
        query Balance (
```

src/ActionModal/components/ActionModal.vue
jbibla: we should be using the error option, no?

https://vue-apollo.netlify.com/api/smart-query.html#options

faboweb: did you experiment with that? how should we use this? does this skip the update function?

jbibla: no i haven't tried it. from an architecture perspective, the API should always return the data we expect. if it doesn't there is likely an error with the API / server which we should catch the _vue-apollo_ way. 

the approach in this PR seems to be handling backend issues in the frontend which seems a bit funny to me. but i am not certain about this. and if this is a temporary fix for some problem we have now then we can certainly merge and address the underlying issues later.

https://learn.hasura.io/graphql/vue/queries/3-handle-errors

faboweb: I think there is a misunderstanding:
This doesn't handle issues with the backend but with the client losing its connection to the internet or similar issues. In this case, the backend can't serve the correct data.

jbibla: yeah, i think i get what you're saying. 

i'm suggesting two things:
1. let's use this vue-apollo `error` option for when there is no internet or the API is broken
2. let's make sure the API returns what the FE needs:

if `!data.overview` than the API is not working and `error` would catch it. shouldn't `data.overview` return `data.overview.totalRewards = 0`?
```
if (!data.overview) {
          return {
            totalRewards: 0
```

faboweb: ok, so you want to show an error "Couldn't get the data to show here" if there is any error? Or you want to show the technical error?

faboweb: this also doesn't solve the Sentry issues if the update function is still triggered and `data.overview` is undefined

Bitcoinera: > this also doesn't solve the Sentry issues if the update function is still triggered and `data.overview` is undefined

But we need to know about these errors happening, to be able to fix them. Unless we are already successfully capturing the error in the API and this is just a duplicate.

faboweb: also should each component individually display an error? or a page displays an error?

Bitcoinera: It depends. I think in the case of "floating" components like ActionModal it should be shown on the component itself and for "embedded" components it should appear at the top of the page, with small red letters (like we have them now), saying query X failed. IMO

faboweb: > But we need to know about these errors happening, to be able to fix them. Unless we are already successfully capturing the error in the API and this is just a duplicate.

we need to know every failing internet connection of a user?

jbibla: so do you think what you're doing in this PR is the right strategy for `update` queries?

Bitcoinera:
I just added as muster the error handling described [here](https://learn.hasura.io/graphql/vue/queries/3-handle-errors) to TmBalance.

What do you think? Should we implement this in all our pages/components?



One thing I like about this approach is that it gives the user enough information so they can also help us fix bugs better and more efficiently.

faboweb:
@bitcoinera the errors should be human readable. I disagree with burdening our users with complex error messages. All errors should be caught by Sentry so we can evaluate them better then the user could.

Bitcoinera:
> @Bitcoinera the errors should be human readable. I disagree with burdening our users with complex error messages. All errors should be caught by Sentry so we can evaluate them better then the user could.

OK, yes, I agree this approach is most likely better. 

So we should always display then a simple "There was an error. Apologies"? Maybe we could add something like "the component Total Balance cannot be displayed correctly"... but I am not sure how we could implement something like that... maybe just a `switch` to select the right error message :woman_shrugging: 
Or is this overcomplicating?

faboweb:
I would simply go for "There was an error so we can't show you this data right now. If this error persists please contact our [Support].". This scales well. But giving some context could also be good.

PR #3449, opened 2020-01-20
Fabo/set version for sentry

faboweb:
Sets the version to be displayed in Sentry along the error. So we see which version the user is executing.


Bitcoinera:
Brilliant! But why did the netlify deployment failed?

Bitcoinera:
Cool. Good catch @jbibla

PR #3456, opened 2020-01-21
Fabo/get network info from api

### LINKED ISSUE #3402: Address prefixes should be received from API

faboweb:
Closes #3402

Needs https://github.com/luniehq/lunie-api/pull/278 to be released first

Queries information about the used ledger app and address creator from the API so we don't need to have this information in the FE code (to scale better)


Bitcoinera:
Now it is working! :wink: And I love the whole refactor :+1:

Bitcoinera:
I'm OK merging this. What about you @jbibla ?

src/components/common/TmSessionHardware.vue
```diff
@@ -135,7 +135,7 @@ export default {
       this.status = `detect`
       this.address = null
       try {
-        this.address = await getAddressFromLedger(this.networkId)
+        this.address = await getAddressFromLedger(this.networkId, this.$apollo)
```
jbibla: why do we have to pass `$apollo` as an argument?

Bitcoinera: it needs it to query for the address prefix in `ledger.js`, lines 48-59. 
Before we were using the hardcoded dictionary and now we are using the API.

PR #3463, opened 2020-01-24
Define network by address prefix and selecting it on login with exten…

iambeone:
Closes #ISSUE

Define network by address prefix and switching to this network by pressing "use account" on extension step

Thank you! 🚀

---

For contributor:

For reviewer:


src/components/common/AccountList.vue
```diff
@@ -1,77 +1,77 @@
...
-
-.account-button {
-  padding: 0.25rem 0.5rem;
-}
-
-.bech32-address {
-  font-size: 14px;
-}
-</style>
+<template>
```
faboweb: here again the lines are all showing as changed. did you figure out why this is happening? This makes it very hard to read the PRs

src/vuex/modules/connection.js
```diff
@@ -18,6 +18,16 @@ export default function({ apollo }) {
   }
 
   const actions = {
+    async getNetworkByAddress(store, address) {
+      const { data } = await apollo.query({
```
faboweb: This is not a 1:1 mapping. Usually testnets and mainnets have the same prefix.

faboweb: @jbibla what do you think we should do here? Offer a mainnet / testnet selection?

src/vuex/modules/connection.js
```diff
@@ -18,6 +18,16 @@ export default function({ apollo }) {
...
+        query: Networks,
+        fetchPolicy: "cache-first"
+      })
+      const network = data.networks.find(
+        network => address.indexOf(network.address_prefix) == 0
+      )
+      return network || false
```
faboweb: you are returning 2 different types here. Instead of returning false you should return undefined and check in the receiving function to make the API predictable

src/components/common/AccountList.vue
iambeone: Something with lint. Before pushing with git diff i can see changes. Trying to figure out what is happen

src/vuex/modules/connection.js
iambeone: @faboweb I know, it's just for testing logging in with network change.

jbibla: > what do you think we should do here? Offer a mainnet / testnet selection?

no, not a selection for the user. probably, we should store and pass the network id with the address when sending from the extension.

PR #3475, opened 2020-01-26
Ana/add feature not present component

Bitcoinera:
Closes #ISSUE

**Description:**

In this PR I add the FeatureNotPresent component, thinking to have a way to tell users when a feature just doesn't exist in the given network.

This I think is actually just a temporary solution/patch, since the ideal would be that the link doesn't appear altogether in these cases... Now that I was writing this realization came all of the sudden :sweat_smile: 

In any case, if we still go for this solution, adding a FeatureNotPresent component, there is some issue now with this PR that the code enters in a loop when rendering the component that ends killing the computer. This has happened to me already 5 times and I haven't yet really seen why.

In the meantime, I will leave it here so perhaps another pair of eyes can see what is wrong (I added a `return` in line 35 in `router.js`, the component is super simple with no props and no nothing... what is going on)

Perhaps it should be wrapped inside TmPage like FeauteNotAvailable. I don't know because I wasn't able to see it yet :laughing: 

EDIT: OK, I think I just saw the problem xD I just needed to push to realize what was wrong. My conditional... Fixing it! :muscle: 

Thank you! 🚀

---

For contributor:

For reviewer:


jbibla:
should we hide the links from the sidebar?

Bitcoinera:
> should we hide the links from the sidebar?

I think this is actually the best way to go. But what do you all think?

Bitcoinera:
Now it is working.

faboweb:
This contradicts our design strategy (show all and educate). I would like to have an opinion from Charlie first. This is also a classic feature where a design first would have been nice.
Personally I would stick with the current strategy.

faboweb:
I would also propose a standardized way of handling this and not distinguish between us not having enabled the feature or the feature not existing on the chain.

Bitcoinera:
> I would also propose a standardized way of handling this and not distinguish between us not having enabled the feature or the feature not existing on the chain.

But if our goal is to educate, I don't see how we can standardize this :woman_shrugging:  Either we say it is a feature we are working on or it is a feature that simply doesn't exist, no?

In the first case, users have available a link to request the feature. In the second, not.

faboweb:
I understand now. let's discuss a strategy further. this seems like a bigger fish to fry.

Bitcoinera:
Maybe I should close this one for now? Set is as 'archived'.

When we come back to fry this, we should implement what Jordan suggested with the forums and social networks links to replace the governance section.

Bitcoinera:
OK, I am going to put this one in the freezer, but it is something will need to undoubtedly revisit. And soon I think.

Bitcoinera:
Trying to make it work together with luniehq/lunie-api#510

Bitcoinera:
This is how the component is looking now after applying @jrmoreau and @jbibla 's suggested changes

image

faboweb:
Can we make it look like this: https://sketch.cloud/s/Oy9lm/a/V7r9z7 ?

Bitcoinera:
> Can we make it look like this: https://sketch.cloud/s/Oy9lm/a/V7r9z7 ?

I made some comments there

src/router.js
```diff
@@ -16,14 +16,23 @@ export const routeGuard = (store, apollo) => async (to, from, next) => {
...
     to.meta.feature &&
     !(store.state.connection.network === "testnet") && // TODO remove once we have Hasura integrated in e2e tests
-    !(await featureAvailable(apollo, store.state.connection.network, to))
+    !(await featureAvailable(apollo, store.state.connection.network, to)) &&
+    !(
+      (await featureAvailable(apollo, store.state.connection.network, to)) ===
+      null
+    )
   ) {
     next(`/feature-not-available/${to.meta.feature}`)
```
faboweb: we should embed these and not reroute. that was a bad decision of me before. on a refresh you are stuck there

src/router.js
```diff
@@ -45,9 +54,13 @@ export default Router
...
-  const feature = `feature_${to.meta.feature.toLowerCase()}`
-  const { data } = await apollo.query({
-    query: NetworkCapability(networkId)
-  })
-  return NetworkCapabilityResult(feature)(data)
+  if (to.meta.feature === undefined) {
+    return
+  } else {
+    const feature = `feature_${to.meta.feature.toLowerCase()}`
+    const { data } = await apollo.query({
```
faboweb: you should get this from the networks object and not do another query

src/router.js
faboweb: but later

src/router.js
Bitcoinera: I didn't add this query. It was already [there ](https://github.com/luniehq/lunie/blob/5648abce07c8746d6856ea6d21ac9e7b544e3b51/src/router.js#L49)

Bitcoinera: Also, there is not network object right now in this file. Are you suggesting using the two getters (networks and network) to get it?

faboweb: > Are you suggesting using the two getters (networks and network) to get it?

Yes, like everywhere else

Bitcoinera: Done

jbibla:
unit tests are failing - lets merge this and change to the new designs later.

Bitcoinera:
> unit tests are failing - lets merge this and change to the new designs later.

I didn't update all snapshots it seems. Because I used `--bail` 🤦‍♀

src/router.js
```diff
@@ -16,17 +21,21 @@ export const routeGuard = store => async (to, from, next) => {
...
+        return
+      }
+      case "MISSING": {
+        next(`/feature-not-present/${to.meta.feature}`)
+        return
+      }
+      default: {
+        // if the feature is enabled just continue
+      }
+    }
```
Bitcoinera: Ah yeah, this looks actually quite better! Good job!

PR #3484, opened 2020-01-27
Ana/fix missing denom when no balances

Bitcoinera:
Closes #ISSUE

**Description:**

@iambeone just discovered this bug happening throughout all our platforms. When an account has no balances, the denom in the SendModal is missing.

This PR fixing it by querying "network" from apollo to get its `stakingDenom`.

When there are no balances, it will default to the stakingDenom as `selectedToken`. 

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/src/ActionModal/components/SendModal.vue
+++ b/src/ActionModal/components/SendModal.vue
@@ -186,7 +186,8 @@ export default {
     editMemo: false,
     isFirstLoad: true,
     selectedToken: ``,
-    balances: []
+    balances: [],
+    denom: ``
   }),
   computed: {
     ...mapGetters([`network`]),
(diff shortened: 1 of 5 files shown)


src/ActionModal/components/SendModal.vue
```diff
@@ -340,6 +345,27 @@ export default {
...
+      fetchPolicy: "cache-first",
+      /* istanbul ignore next */
+      variables() {
+        return {
+          networkId: this.network
+        }
+      },
+      /* istanbul ignore next */
+      update(data) {
+        return data.network.stakingDenom
```
faboweb:
```suggestion
        if (data.network) return data.network.stakingDenom
        return ''
```

Bitcoinera: :thinking: then I should apply this to the other modals too...

Shouldn't this be solved once we add the `error` field in all apollo queries?

PR #3488, opened 2020-01-27
Ana/add universal signin for cross network

Bitcoinera:
Closes #ISSUE

**Description:**

Trying the Android app I realized that when displaying the "This address doesn't belong to the network you are currently connected to" error, we probably leave the average user quite clueless. So I decided to add a second line showing them the way to the networks section so they realize they need to select the right network.

But while I was doing this I also noticed that the TmSessionSignIn component had not this network validation implemented :scream:  So I also added this in this PR.

Now, there is still a bug left in this component but I will fix it tomorrow :wink: 

That's all for now

Thank you! 🚀

---

For contributor:

For reviewer:


faboweb:
tests failing

Bitcoinera:
Please, don't merge yet. There is still a bug in "Sign in with account" I need to fix. 

This is going to be second bug I am going to fix in this PR... maybe I should add them to the Kanban board so there is a record :smirk:

Bitcoinera:
OK, all bugs fixed now! 🤩 Ready to merge!!

faboweb:
There is a css issue:
image

Bitcoinera:
> There is a css issue:
> image

 I think this should be addressed in another PR and this one is good to merge.

jbibla:
i see the problem you're trying to address here but i don't really think this is the right solution.

1) can't we just switch networks for them if we're already checking what network the prefix is from?
2) if no, and we have to ask them to switch, can't we send them to the session network screen instead of network page in the app?
3) if no, let's show them the session network screen _before_ they sign in to an account

faboweb:
1) only works if we ignore testnets

jbibla:
true. so we should just ask them to select a network before signing in using the `selectnetwork` session modal.

Bitcoinera:
> true. so we should just ask them to select a network before signing in using the `selectnetwork` session modal.

Or we just ignore testnets in case there is a mainnet for that prefix and go all in for 1.

I agree with you this is by far the best UX.

If someone still wants to sign in to the testnet instead of mainnet, they just go to Networks.

faboweb:
> If someone still wants to sign in to the testnet instead of mainnet, they just go to Networks.

so you are proposing, that if the user enters an address that fits the current network, they enter the network. if it is of another network we sign them into the mainnet of that network?
that sounds like very intransparent logic to me.

how about a checkbox "use testnet"?

Bitcoinera:
> > If someone still wants to sign in to the testnet instead of mainnet, they just go to Networks.
> 
> so you are proposing, that if the user enters an address that fits the current network, they enter the network. if it is of another network we sign them into the mainnet of that network?
> that sounds like very intransparent logic to me.
> 
> how about a checkbox "use testnet"?

The checkbox idea is not a bad one.

I mean, once that a network, like e-Money, for example, has its mainnet released, I think there won't be many users interested in their testnet.

But yes, we could add this "use testnet" checkbox, to make it more transparent :+1:

jbibla:
a checkbox is a very nice and simple solution for this.

Bitcoinera:
Please help me with the tests for `selectNetworkByAddress`. I don't really know how to handle `localStorage` in there 🤷‍♀

src/components/common/TmSessionExplore.vue
```diff
@@ -170,18 +178,24 @@ export default {
...
+      let selectedNetwork = this.addressPrefixes.filter(({ address_prefix }) =>
+        address.startsWith(address_prefix)
       )
       // handling query not loaded yet or failed
-      if (!selectedNetwork) return false
-
-      if (param.startsWith(selectedNetwork.address_prefix)) {
-        return true
+      if (!selectedNetwork) {
+        console.error("Connecting to the selected network failed")
```
faboweb: This should throw to prevent the process to continue, right?

src/components/common/TmSessionExplore.vue
```diff
@@ -170,18 +178,24 @@ export default {
...
+        return
+      }
+      // handling when there are both mainnet and testnet networks
+      if (selectedNetwork.length > 1) {
+        selectedNetwork = selectedNetwork.filter(({ id }) =>
+          this.testnet ? id.slice(-7) === `testnet` : id.slice(-7) === `mainnet`
```
faboweb: Networks have a testnet property so we don't need this funky string destructuring

src/components/common/TmSessionExplore.vue
```diff
@@ -170,18 +178,24 @@ export default {
...
+        )[0]
       } else {
-        return false
+        selectedNetwork = selectedNetwork[0]
```
faboweb: Do we need the else?

src/components/common/TmSessionSignIn.vue
```diff
@@ -125,11 +144,66 @@ export default {
...
         this.$el.querySelector(`#sign-in-name`).focus()
       }
+    },
+    async selectNetworkByAddress(address) {
+      let selectedNetwork = this.addressPrefixes.filter(({ address_prefix }) =>
+        address.startsWith(address_prefix)
+      )
+      // handling query not loaded yet or failed
+      if (!selectedNetwork) {
+        console.error("Connecting to the selected network failed")
```
faboweb: See comments above

src/components/common/TmSessionSignIn.vue
```diff
@@ -85,6 +100,9 @@ export default {
       }))
     }
   },
+  created() {
+    this.$store.dispatch("loadAccounts")
```
faboweb: Why do we need this now?

Bitcoinera: This fixes the other bug, that accounts were not loaded after refreshing.

It has nothing to do with the universal sign in. It is just an additional thing, fixing a bug (yes, it should have been done in another PR but it is a really easy fix, 3 lines)

src/components/common/TmSessionExplore.vue
Bitcoinera: I like funky 🤖

src/components/common/TmSessionExplore.vue
Bitcoinera: Yes, this is what the next line, the `return`, does

src/components/common/TmSessionExplore.vue
Bitcoinera: Yes. For all the networks that only have testnet (the vast majority right now).

I think I should change the selectedNetwork to plural to make it more clear.

src/components/common/TmSessionExplore.vue
faboweb: ah right! but it never displays an error to the user, or I don't see it.

src/components/common/TmSessionExplore.vue
faboweb: hahahah but we don't need complex code ^^
this also assumes that the ids of the networks always have `testnet` or something in it.

src/components/common/TmSessionExplore.vue
faboweb: but if the length of `selectedNetwork` is 1 you can still use the same logic of filtering it. There should be a handling if there is no testnet or mainnet for the network but the user selected it.

src/components/common/TmSessionExplore.vue
faboweb: ah no, the return will not prevent succeeding of line 145

src/components/common/TmSessionSignIn.vue
faboweb: I think this is solving the issue at the wrong position. But let's keep in the back of our head.

Bitcoinera: It is only fixing what I told you xD 

What is this wrong position bug?

Bitcoinera:
How can I handle in the TmSessionExplore and TmSessionSignIn localStorage correctly?

It will only return `null`... How can I do `getItem` from the external localStorage?

faboweb:
Why do you even want to manipulate localStorage shouldn't you rather manipulate the store?

Bitcoinera:
> Why do you even want to manipulate localStorage shouldn't you rather manipulate the store?

Maybe my approach wasn't the right one then... In fact I started trying the `store.dispatch` thing. Will give it another try

Bitcoinera:
I just fixed tests!! 😸 

I really feel like I'm starting to get the grip of jest, finally!

Thanks @faboweb for helping me changing my approach 😉

PS: the cat in the Windows Unicode's representation actually looks like a tiger

PS2: Let's get this merged!

faboweb:
let's merge, but SessionExplore needs a serious refactor

Bitcoinera:
Wow, you refactored a lot! 👀

faboweb:
and I didn't even start ^^

Bitcoinera:
> and I didn't even start ^^

😆

PR #3500, opened 2020-01-29
Ana/adding checks. tentative

### LINKED ISSUE #3496: Refactor all update fields in SmartQueries
All `update` functions inside SmartQuery should now add a check to see if the data was succesfully fetched:

```
      update(data) {
        if (!data.overview) {
          return {
            totalRewards: 0
          }
        }
        /* istanbul ignore next */
        return {
          ...data.overview,
          totalRewards: Number(data.overview.totalRewards)
        }
      }
```

In case it wasn't, we return either "", 0 or {} depending on the case.

As I understood it from the meeting we just had, this should now be applied to <b>every</b> SmartQuery, doesn't matter if it has nested values or not.

Bitcoinera:
Closes #3496 

**Description:**

In this PR I aim to refactor all update functions inside Smartqueries so we prevent all the errors when a query fails.

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/src/components/network/PageBlock.vue
+++ b/src/components/network/PageBlock.vue
@@ -104,7 +104,13 @@ export default {
           networkId: this.network
         }
       },
+      /* istanbul ignore next */
       update: function(result) {
+        if (!result) {
+          return {
+            validators: []
+          }
+        }
         return result.validators
       }
     },
--- a/changes/ana_refactor-all-update-in-smartqueries
+++ b/changes/ana_refactor-all-update-in-smartqueries
@@ -1,0 +1,1 @@
+[Code Improvements] [#3500](https://github.com/cosmos/lunie/pull/3500) Refactors all update funcions inside smartqueries to check if the query has failed or not @Bitcoinera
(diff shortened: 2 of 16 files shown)


jbibla:
nice! defensive programming 🛡 

tests are failing...

Bitcoinera:
> nice! defensive programming 
> 
> tests are failing...

Yes, thank you. I will first see if I am going the right way :pray:

src/components/network/PageBlock.vue
```diff
@@ -104,7 +104,13 @@ export default {
           networkId: this.network
         }
       },
+      /* istanbul ignore next */
       update: function(result) {
+        if (!result) {
```
faboweb: this is wrong right? shouldn't this return just an array as the good path is returning `result.validators`

Bitcoinera: yes, looks quite wrong... :eyes:  thx

Bitcoinera: I think we better wait and see if we adopt @iambeone 's solution.

He has being demoing it to me earlier and it really does look far superior to this one.

I am tempted to close this PR quite frankly.

Bitcoinera:
I close this as I think Aleksei's solution is by far better than this one

PR #3502, opened 2020-01-29
show all validators in multi claim rewards

### LINKED ISSUE #3244: Claiming transactions only considers the first claiming tx (of 5 usually)
**Describe the bug**

Claiming transactions in In Activity page only considers the first claiming tx (of 5 usually).

Example tx:

`
https://lcd.nylira.net/txs/DFEAD48410631A06D6EA42F458346EEE71259C02A3F5B33FE6A225E078B131A7`

**To Reproduce**
1. Sign in with address `cosmos18da7h02z3e46ulee5prg5zl6pr9va0ytv6w46k` in explore mode
2. Go to Activity page.

**Expected behavior**

The tx should show all the validators you get your rewards off.  If there are too many,  we can show the message `From x validators`.

**Screenshots**

image

Bitcoinera:
Closes #3244 
**Description:**

This is looking good. Now all validators of a multi-claiming reward transaction are being displayed.

There is just an issue with a very concrete case I don't yet really understand why it is being problematic. It is in very specific transactions of Terra like this [one](https://terra-testnet.lunie.io/txs/D61B64A349BA0550D621245159F321720AA683C5B1898DE2A6D3C9BC0729AD42), where we get inside messages both a transaction of "MsgWithdrawDelegationReward" and "MsgDelegate"... I had to hack my way there. I get somehow a `[{...}, [Observer] ]` structure instead of only `[Observer]`.

The rest is working quite well and with 0 issues. 

EDIT: This PR <b>only</b> works together with luniehq/lunie-api#285

image

Thank you! 🚀

---

For contributor:

For reviewer:


faboweb:
How does this look like for 5 validators? How about just adding them vertically? @mariopino implemented this already in the past.

Tests are missing.

mariopino:
> How does this look like for 5 validators? How about just adding them vertically? @mariopino implemented this already in the past.
> 
> Tests are missing.

I remember Charlie made new designs to handle that: I think the solution was to only show the icon to save space and display them in a column in mobile view.

Bitcoinera:
> I remember Charlie made new designs to handle that: I think the solution was to only show the icon to save space and display them in a column in mobile view.

Can you pass the link of the design?

Also, there are many validators which don't have yet an image :woman_shrugging:

faboweb:
We set a default calculated image for validators which don't have an image right?

Bitcoinera:
> We set a default calculated image for validators which don't have an image right?

Not from the API. What you mean is the identicons present in PageValidator and PageValidators.

jbibla:
https://app.zeplin.io/project/5e14d2907a541a559c09d01b/screen/5e26d0282b5bdc7f201ba2e3
https://app.zeplin.io/project/5e14d2907a541a559c09d01b/screen/5e26d028683ce156ce016d47

faboweb:
> Not from the API. What you mean is the identicons present in PageValidator and PageValidators.

yes, that is what I was referring to

Bitcoinera:
> https://app.zeplin.io/project/5e14d2907a541a559c09d01b/screen/5e26d0282b5bdc7f201ba2e3
> https://app.zeplin.io/project/5e14d2907a541a559c09d01b/screen/5e26d028683ce156ce016d47

Cool. I will apply those changes and what Fabo commented on displaying the identicon as default

jbibla:
awesome! thank you ana!

src/components/transactions/message-view/WithdrawDelegationRewardMessageDetails.vue
```diff
@@ -58,9 +84,28 @@ export default {
     }
   },
   computed: {
-    validator() {
-      return this.validators[this.transaction.value.validator_address] || false
+    getValidators() {
```
jbibla: i think usually we name computer properties the thing that's returned - not like the name of a function. but just a little note / thought.

jbibla:
this is not really working for me at all but seems like the API is just returning a string with one delegator / validator pair...

src/components/transactions/message-view/WithdrawDelegationRewardMessageDetails.vue
Bitcoinera: The problem here is that there is already a `validators` as a prop

Bitcoinera:
OK, this is how it is looking now

image

src/components/transactions/message-view/WithdrawDelegationRewardMessageDetails.vue
jbibla: ah ok. i didn't see that. still, should the prop be `allValidators` maybe?

i find it weird to have a computer property named like a function. but maybe just me!

jbibla:
how does this relate to: https://github.com/luniehq/lunie-api/pull/169/files ?

src/components/transactions/message-view/WithdrawDelegationRewardMessageDetails.vue
```diff
@@ -58,9 +107,56 @@ export default {
...
+  flex-direction: row;
+}
+.multi-claim-reward-row {
+  display: flex;
+}
+.multi-claim-reward-h3 {
+  margin-right: 20px;
+}
+.row-validator-image {
+  float: left;
```
jbibla: best not to use `floats` and `flex` - what are you trying to achieve here?

src/components/transactions/message-view/WithdrawDelegationRewardMessageDetails.vue
```diff
@@ -58,9 +107,56 @@ export default {
...
+  padding: 5px;
+}
+.validator-image {
+  border-radius: 100%;
+  padding: -1px;
```
jbibla: why negative pixels? `-1px`?

src/components/transactions/message-view/WithdrawDelegationRewardMessageDetails.vue
```diff
@@ -58,9 +107,56 @@ export default {
...
+  height: 1rem;
+  width: 1rem;
+  vertical-align: middle;
```
jbibla: this probably isn't necessary. if you're trying to have them centered in the row (which they currently aren't) you can use flex styles like `align-items: center`

Bitcoinera:
> how does this relate to: https://github.com/luniehq/lunie-api/pull/169/files ?

It doesn't.

He used a different solution that simply consists of sending the whole of the transactions messages to the FE. 
A solution I am against to, since the FE should be as dummy as possible and receive all data correctly processed from the API.

Bitcoinera:
This now look glorious :heart_eyes: 

(Only the tiny detail of squared identicon)

Bitcoinera:
I just noticed that Charlie also made a screen for when the transaction div is open. Just implemented this too. Here you have a video.


LunieSimpleStakingWallet_Staking


Let me know if you think we should add some CSS animation additionally... I'd need to brush up my CSS skills but I can give it a try 😅

src/components/transactions/message-view/WithdrawDelegationRewardMessageDetails.vue
```diff
@@ -122,22 +122,18 @@ export default {
   },
   computed: {
     getValidators() {
-      if (this.transaction.value.length > 1) {
+      if (this.transaction.withdrawValidators) {
         let validators = []
-        this.transaction.value.forEach(msg => {
+        JSON.parse(this.transaction.withdrawValidators).forEach(msg => {
```
faboweb: why does `withdrawValidators` not just hold all validator addresses?

Bitcoinera: What do you mean? Why do you say it doesn't?

faboweb: `this.transaction.withdrawValidators` is returning a list of messages which is not necessary as the FE only needs the validator addresses. It should return just an array of validator addresses.

Bitcoinera: Oh, I see what you mean... `withdrawValidators` PR has but been already merged and we are going to refactor all this in no time though...

So I think it is OK to pass this for this one time and merge this ^^

faboweb: ok

faboweb:
e2e tests are consistently failing on this

PR #3507, opened 2020-01-29
Ana/possible fix for chunks

### LINKED ISSUE #3442: Fetching chunk bug (Sentry)
Might be solved by the index.html now not caching

https://monitoring.lunie.io:9000/lunie/frontend/issues/65

Currently still appears. But added #3449 to see if this is from outdated Lunies.

Bitcoinera:
Closes #3442 
**Description:**

A tentative solution. Found [this article](https://medium.com/@suixo/configure-sentry-for-vue-js-project-6aa07efe2c80) while searching for answers.

In the last paragraph of it we can read:

> **Important note:** by switching productionSourceMap to true in vue.config.js , we now generate source maps for all the chunks (app + vendors) and also add them to the dist folder. Depending on your deployment pipeline, this may deploy them to your production server, which may not be desirable. Ideally, we would only generate and upload source maps for the application chunk, but this requires tweaking the Webpack configuration further and is out of scope of this post.

Checking the [documentation](https://cli.vuejs.org/config/#productionsourcemap) (and also the comments of this article), we see that this field defaults to `true`, so perhaps this means we are sending chunks to production that we don't want to be there.

image

My fix is to explicitly set it to `false`.

Thank you! 🚀

---

For contributor:

For reviewer:


faboweb:
As seen in this bug report, the issue is not, that a sourcemap is not loaded: https://monitoring.lunie.io:9000/lunie/frontend/issues/65/?query=is%3Aunresolved

Here we see that the user has an outdated version:
image

The current version is 1.0.161. Therefor the files he requests do not exist anymore.

Bitcoinera:
> The current version is 1.0.161. Therefor the files he requests do not exist anymore.

So we need to create a script that automatically deletes old chunks and fetched the new... Sounds like a cool challenge! :muscle:

faboweb:
No, we need a script that at the beginning of the app checks if the version still matches. If not force a refresh before loading the chunks.

faboweb:
Or we keep the old files somehow alive. But I am not sure that is possible with netlify. THis way the user can still load the old version. But we still need to update the app so he doesn't see stale data.

Bitcoinera:
Ok, I have tried now another different solutions. I think simpler than what you were proposing @faboweb . We don't need to check versions and there we don't need to add the version JSON to the dist folder either.

We simply reload the page very time there is a "loading chunk failed" error. If we don't keep any cache in our web server, there shouln't be any "refresh hell" here.

Theoretically it plays nice but I have been trying for a while and I am not sure how I can check it is properly working. Have been playing around in the netlify version of this PR and also with the `dist` folder, but no luck :woman_shrugging:

src/router.js
```diff
@@ -40,6 +40,13 @@ const router = new Router({
   routes
 })
 
+router.onError(error => {
+  // if we fail fetching any chunk, we refresh the page to fetch the new version chunks
+  if (/loading chunk \d* failed./i.test(error.message)) {
+    window.location.reload()
```
jbibla: this feels like a weird place to handle caching / chunk issues. 

is window.location.reload even a hard refresh?

faboweb:
Do these errors get thrown inside the application or from the js imports in index.html?

faboweb:
Ahhhhh no, I think @Bitcoinera found something. 👏 The problem I guess comes from the the async loads of pages. The user still has stale data but moves to a page. The chunk for the page is not there anymore so this error appears. The solution provided here could actually work well.

I propose on top: Ping on app start the created version file and show a "App is outdated warning, click here to update".

jbibla:
> The best solution I can think of is to have the site javascript check the version of the deployed site (using the etag header for some file, for example /index.html) and force an update of all assets when the etag changes.

https://community.netlify.com/t/keep-js-chunks-from-old-deploys/5991/2

Bitcoinera:
> > The best solution I can think of is to have the site javascript check the version of the deployed site (using the etag header for some file, for example /index.html) and force an update of all assets when the etag changes.
> 
> this would be pro-active instead of reactive and seems like the right solution.
> 
> https://community.netlify.com/t/keep-js-chunks-from-old-deploys/5991/2

But why would we want to keep the old js chunks? Or you mean just using this solution with `lazy` imports as mentioned by the OP in the link you shared? [This](https://dev.to/goenning/how-to-retry-when-react-lazy-fails-mb5) solution.

jbibla:
no no.

this solution:

> have the site javascript check the version of the deployed site (using the etag header for some file, for example /index.html) and force an update of all assets when the etag changes.

Bitcoinera:
> no no.
> 
> this solution:

OK. Then I guess I need to use the `fs` module in the browser in the end. Like [this one](https://www.npmjs.com/package/fs-web)

faboweb:
why do you want `fs` in the browser?
did you consider querying for `/version.json?` and parsing the version?

Bitcoinera:
> why do you want `fs` in the browser?
> did you consider querying for `/version.json?` and parsing the version?

true xD

faboweb:
We could refresh here: https://github.com/luniehq/lunie/blob/bd2f8bf25f7f9491dc2a07a4e0a9f9a8fa8e9db0/src/registerServiceWorker.js#L23

But maybe we want to do that even earlier

Bitcoinera:
> But maybe we want to do that even earlier

What steps are you following concretely to test the chunks updating? Just changing code while `yarn serving` is enough? Did you do `yarn build`?

Sorry if these are obvious questions. The other day I tried modifying the chunks names here but quite unsuccessfully :sweat_smile: 



TL;DR: How do you test the refreshing on updating content?

faboweb:
you need to run `yarn build` with production mode.
now load the page.
then run `yarn build` again. 
now load the page again.

the console should state the updating behavior

Bitcoinera:
> the console should state the updating behavior

Awesome. Yes, for sure. Every time you run `yarn build` you get different chunk names! I didn't fully realize until now. Thanks :wink:

faboweb:
if we refresh the page to early, we might not even fix the problem as the SW is not updated yet so we will just serve the same stale data again ending in a loop.

Bitcoinera:
I will close this now since the action has clearly moved on to #3518

PR #3514, opened 2020-01-31
Aleksei/signin fixes

iambeone:
Closes #ISSUE

**Description:**

Different changing in sign-in flow:
Add warning if there are no accounts in extension
Remove select network step in FE on create new account step
Add network switch feature on sign-in with extension account

Thank you! 🚀

---

For contributor:

For reviewer:


jbibla:
can you share some screenshots?

> Add warning if there are no accounts in extension

what was displayed previously?

> Remove select network step in FE on create new account step

why? how do we know which network they are creating an account for?

> Add network switch feature on sign-in with extension account

i think the plan was to just to sign in to the network based on the prefix, right?

iambeone:
First fix is about this screen:
image
Not cool)

> why? how do we know which network they are creating an account for?
The flow was like this
1. Create new address
image
2. Select network
image
3. Become sad
image
I have removed the step with the network selection. Only for non mobiles and non insecure modes

> i think the plan was to just to sign in to the network based on the prefix, right?
It's a bit incorrect, cause testnet and mainnet has the same prefix. So now we are using network value from wallet

changes/aleksei_signin_fixes
```diff
@@ -0,0 +1,4 @@
+[Changed] Add no accounts allert on use extension step @iambeone
```
faboweb: classic case of better splitting up the PR ;)

jbibla:
```suggestion
[Changed] Add no accounts alert on use extension step @iambeone
```

src/components/common/TmSelectNetwork.vue
```diff
@@ -43,32 +44,30 @@ export default {
...
-      } else {
-        return []
-      }
+    }
+  },
+  mounted: async function() {
+    await this.$apollo.queries.networks.refetch()
+    if (this.networks.length > 0) {
+      this.sortedNetworks = [
+        // current network first
```
jbibla: i would prefer for the network list to appear the same way always. the order never needs to switch.

src/components/common/TmSessionExtension.vue
```diff
@@ -29,6 +29,13 @@
           :button-text="`Use Account`"
         />
       </div>
+
+      <div v-else class="session-main">
+        <p class="extension-message">
+          Looks like you don't have any accounts yet. Head over to
+          <router-link to="selectnetwork">create one</router-link>.
```
faboweb: that leads the user to create an account in the browser right?

src/components/common/TmSelectNetwork.vue
```diff
@@ -43,32 +44,30 @@ export default {
...
+          // show all mainnets next
+          .sort((a, b) => {
+            return a.testnet - b.testnet
+          })
+          // show the default network on the top
+          .sort((a, b) => {
+            return b.default - a.default
+          })
+      ]
+        // filter out undefineds (happens if this.networkId is not set like in the extension)
```
jbibla: why would networkId not be set? shouldn't we set it instead of filtering out undefineds?

faboweb: in the extension this is not set, as the extension isn't signing in to any network

src/components/common/TmSelectNetwork.vue
faboweb: if we don't show this on desktop, it is anyway not needed

src/components/common/TmSelectNetwork.vue
```diff
@@ -43,32 +44,30 @@ export default {
...
-          ]
-            // filter out undefineds (happens if this.networkId is not set like in the extension)
-            .filter(x => !!x)
-        )
-      } else {
-        return []
-      }
+    }
+  },
+  mounted: async function() {
```
faboweb: what was the problem and why did you choose this imperative approach?

src/vuex/modules/session.js
```diff
@@ -128,8 +128,12 @@ export default () => {
           connection: { network }
         }
       },
-      { address, sessionType = `ledger` }
+      { address, sessionType = `ledger`, networkId }
     ) {
+      if (networkId && network !== networkId) {
+        await commit(`setNetworkId`, networkId)
```
faboweb: why not `dispatch("setNetwork", { id: networkId })`?

src/components/common/TmSelectNetwork.vue
iambeone: The problem was that the order of the networks was changed after selecting the network. What is the imperative approach?

Bitcoinera:
I really like this PR. Great job @iambeone ! I give it my approval :smiley:

src/components/common/TmSelectNetwork.vue
faboweb: the problem was that networkId changed and therefor the list changed. Why not just remember the networkId on start instead of changing the whole computed (reactive) selectedNetwork list?

faboweb: but not super important

src/components/common/TmSelectNetwork.vue
jbibla: why?

faboweb: because it requires "current network first" requires a "current network". and in the extension we don't have a network as you are never in the context of a session

Bitcoinera:
> unit tests are failing and several comments are unaddressed

Maybe I got overexcited :sweat_smile:

src/components/common/TmSelectNetwork.vue
iambeone: But this would be a strange solution for a computed property, don't you think so? Computed property depends on its variables. Moreover is will cause a warning about side effects of mounted property. And the network list could be changed during this step. Based on those thoughts i decided to move in to mounted, as we need only the initial value of this list.

src/components/common/TmSessionExtension.vue
iambeone: That will only redirect to the create an account page. Now one this page we have a warning about creating accounts in browser with the links to apps, extension and ledger. 
This one 

But with insecure mode the this link will lead to an account creation step, like it should

faboweb: Shouldn't from this the account creation happen in the extension and not locally even for insecure mode?

iambeone: It's a proposal or a suggestion? Anyway, i kept all logic of creation, just added a link there.

src/components/common/TmSelectNetwork.vue
faboweb: exactly. and with a computed property we leverage the asynchronicity of the network request without explicitly waiting for the network request to have passed. we would fix one of the inputs of the computed property `networkId` by setting it once. that's all you wanted to fix.

src/components/common/TmSessionExtension.vue
faboweb: suggestion. right now this is wrong. when the user (in insecure mode) clicks on "sign in with extension" > "create account" they will be directed to create an account locally instead of in the extension.

PR #3518, opened 2020-02-01
Fabo/force reload

faboweb:
My version of refreshing the page when the version is outdated

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -58,6 +58,8 @@
   <!-- built files will be auto injected -->
 
   <script>
+    // check for the current version on the server and refresh the page if that is the local version is outdated
+    // in plain JS so this is served with the index.html ans executed immediatly
     if (process.env.NODE_ENV === "production") {
       const currentVersion = "<%= process.env.LUNIE_VERSION %>"
 
--- a/public/index.html
+++ b/public/index.html
@@ -57,6 +57,23 @@
   <div id="app"></div>
   <!-- built files will be auto injected -->
 
+  <script>
+    if (process.env.NODE_ENV === "production") {
+      const currentVersion = "<%= process.env.LUNIE_VERSION %>"
+
+      async function main() {
+        const { version } = await fetch("/version.json", { cache: "no-store" }).then(res => res.json())
+
+        if (version !== currentVersion) {
+          location.reload(true)
+        }
+      }
+
+      main()
+    }
+
+  </script>
+
   <% if (!process.env.VUE_APP_E2E) { %>
   <script async src='https://www.google-analytics.com/analytics.js'></script>
 
--- a/public/index.html
+++ b/public/index.html
@@ -60,7 +60,7 @@
   <script>
     // check for the current version on the server and refresh the page if that is the local version is outdated
     // in plain JS so this is served with the index.html ans executed immediatly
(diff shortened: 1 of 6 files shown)


Bitcoinera:
> this seems like a good solution

Did you test it? 😆

Bitcoinera:
Ok, I just tested this and I have a couple of things to remark.

1. the `process` variable in the index.html is not working. That means the solution as it is is not working. I found [this netlify thread](https://community.netlify.com/t/uncaught-referenceerror-process-is-not-defined-how-to-use-environment-variables-in-html/1726) where they explain that it is not possible to access `process` from the client side since it is coming from Node. The solution recommended here is to use an script to inject this values in the client side.

2. I wanted to test it anyway so I changed it a little bit. This is what I used for the `ìndex.html` check:

image

I mean, first I did `yarn build --mode production` with `currentVersion` set as `1.0.161` and then, after having it loaded, I changed it for `1.0.162`.

This is the result:

Maybe my way of testing was not the most orthodox one and it is just a matter of fixing the `process.env` being undefined. I don't know. Worth checking.

faboweb:
@Bitcoinera which environment variable are you referencing?
You are correct that normally this is not possible, Aber the index html is post processed by vue-cli and the process.env.x occurances are replaced with the values.

public/index.html
```diff
@@ -57,6 +57,25 @@
   <div id="app"></div>
   <!-- built files will be auto injected -->
 
+  <script>
+    // check for the current version on the server and refresh the page if that is the local version is outdated
+    // in plain JS so this is served with the index.html ans executed immediatly
+    if (process.env.NODE_ENV === "production") {
```
faboweb:
```suggestion
    if ("<%= process.env.NODE_ENV %>" === "production") {
```

faboweb: @Bitcoinera this should fix it. Good find!

faboweb: I used the template syntax in line 64 but forgot it here

Bitcoinera: Great! I thought the fix would be much harder. Tomorrow I'll retest this 😉

Bitcoinera:
> You are correct that normally this is not possible, Aber the index html is post processed by vue-cli and the process.env.x occurances are replaced with the values.

Thank you for explaining all this to me 🙏

Bitcoinera:
Hmm, I am not this is fixing the chunks issue... While I have the page loaded with `yarn serve:dist`, if I do `yarn build --mode production` in another terminal, I always get this:



The first chunk to fail is the very registerServiceWorker's one, so it is true we need to check even earlier, but just checking the version doesn't seem to help here... I have tried changing the version every time before doing `yarn build --mode production` and I still get "Correct version. All good ;)"... 

The reason is that when we run `build`, we then already create the `version.json` with the new version and send it updated to the `dist` folder 🤷‍♀

Bitcoinera:
OK, we can try merging this and looking at the Sentry.

I am not really sure if I'm testing this correctly anyway. Workbox creates too much noise and it is not related to the actual production environment at all 🤷‍♀

faboweb:
Would be interesting to show a message "New version of Lunie loaded, checkout changelog here"
But maybe overkill

PR #3525, opened 2020-02-03
RTM: Prevent bad amounts to be send

### LINKED ISSUE #3476: Prevent bad amounts to be send
Sentry Issue: [PRODUCTION-API-B](https://monitoring.lunie.io:9000/lunie/production-api/issues/164/?referrer=github_integration)

```
Error: {"error":"math/big: cannot unmarshal \"NaN\" into a *big.Int"}
  File "/var/www/server/lib/controller/transaction/index.js", line 83, in null.<anonymous>
    throw new Error(await res.text())
  ?, in runMicrotasks
  File "/var/www/server/lib/controller/transaction/index.js", line 73, in async broadcastTransaction
    const res = await fetch(`${url}/txs`, {
  File "/var/www/server/lib/controller/transaction/index.js", line 39, in async broadcast
    const hash = await broadcastTransaction(
...
(1 additional frame(s) were not displayed)
```

There was also an amount send in uatoms that was a decimal: https://monitoring.lunie.io:9000/lunie/production-api/issues/164/events/234719bfe4ca4f2ca59cb950bc3c1973/

mariopino:
Closes #3476 

**Description:**

* Prevent sending to the backend txs with NaN amount values.
* Modify `uatom()` function to return an integer value.

Should we implement checks for all parameters needed to build the transaction in all `transactionData()` computed fields?

I think it's a good idea to avoid pass to backend tx that will fail anyway. Maybe we can also send a sentry error if any of the tx params it's not correct, just to help us to figure out what's happening here.

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/src/ActionModal/components/ActionModal.vue
+++ b/src/ActionModal/components/ActionModal.vue
@@ -653,8 +653,7 @@ export default {
       this.submissionError = null
 
       if (this.transactionData === {}) {
-        const error = new Error(`Error in transaction data`)
-        this.onSendingFailed(error)
+        this.onSendingFailed(new Error(`Error in transaction data`))
         return
       }
 
--- a/src/ActionModal/components/ActionModal.vue
+++ b/src/ActionModal/components/ActionModal.vue
@@ -708,6 +708,7 @@ export default {
       this.$apollo.queries.overview.refetch()
     },
     onSendingFailed(error) {
+      /* istanbul ignore next */
       Sentry.withScope(scope => {
         scope.setExtra("signMethod", this.selectedSignMethod)
         scope.setExtra("transactionData", this.transactionData)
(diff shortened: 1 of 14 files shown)


src/ActionModal/components/ActionModal.vue
```diff
@@ -790,45 +800,39 @@ export default {
...
           }
         },
+        /* istanbul ignore next */
         skip() {
-          /* istanbul ignore next */
           return !this.txHash
         },
         query: UserTransactionAdded,
+        /* istanbul ignore next */
         result({ data }) {
```
faboweb: I actually think we should test this as it is more complex. but maybe for another time

src/ActionModal/components/ActionModal.vue
```diff
@@ -651,6 +651,15 @@ export default {
     },
     async submit() {
       this.submissionError = null
+
+      if (
```
faboweb: different idea I want to share: How about returning `{error: true}` to avoid this fancy checking if the object is empty? :D

PR #3529, opened 2020-02-04
Fabo/fix releasing

faboweb:
Somehow the release branch is not merged into develop. Only a release commit is created. This works around it.

### DIFF
--- a/tasks/publish.sh
+++ b/tasks/publish.sh
@@ -7,5 +7,5 @@ then
     git remote add bot https://${GIT_BOT_TOKEN}@github.com/luniehq/lunie.git
     git push origin develop:master
 else
-    echo "No release detected found so not publishing"
+    echo "No release detected, so not publishing"
 fi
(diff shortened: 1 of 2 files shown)


tasks/publish.sh
```diff
@@ -1,10 +1,11 @@
...
+if [[ $COMMIT_MESSAGE == "[Simsala] automatic release"* ]]
 then
     echo "Publishing"
     git config user.email "bot@lunie.io"
     git config user.name "MergeBack Lunie Bot"
     git remote add bot https://${GIT_BOT_TOKEN}@github.com/luniehq/lunie.git
     git push origin develop:master
 else
-    echo "No tag found so not publishing"
+    echo "No release detected found so not publishing"
```
jbibla:
```suggestion
    echo "No release detected, so not publishing"
```

faboweb: Thank you

PR #3531, opened 2020-02-05
Aleksei/apollo error handling

iambeone:
Closes #ISSUE

Error handling and preventing of triggering of the update function

Thank you! 🚀

---

For contributor:

For reviewer:


debug.log
```diff
@@ -0,0 +1,5 @@
+[0204/164418.758:ERROR:prune_crash_reports.cc(35)] PruneCrashReportDatabase: Failed to get pending reports
```
jbibla: feels like this should be in the gitignore?

src/gql/apollo.js
```diff
@@ -35,17 +37,50 @@ const makeWebSocketLink = () => {
...
+        sub = forward(operation).subscribe({
+          next: result => {
+            // check if we have errors
+            if (!result.errors) {
+              observer.next(result)
+            } else {
+              // pass errors to sentry
+              result.errors.map(err => {
+                Sentry.captureException(err, {
+                  level: "error"
```
faboweb: why do we need to set the level?

iambeone: Don't know for sure, just used as I have found in the example

faboweb: we don't use this anywhere else, does it provide any additional value? if not let's remove it to stay coherent in the codebase.

iambeone: Updated

PR #3540, opened 2020-02-06
Modify formatting of validator rewards to display very big numbers nicely

mariopino:
**Description:**

Modify formatting of validator rewards to display more nicely very big numbers like the ones we can find in e-money, so `71.900.626.967.285.360,00%` becomes  `71.900,62 T %` (T is for trillion).

Func:
* Display T (trillion), B (billion), M (million) and K (kilo) suffixes
* Used `Intl.NumberFormat` in the same way as the other percent filters.
* Used US billions and trillions (respectively 1e9 and 1e12)

Thank you! 🚀

---

For contributor:

For reviewer:


src/components/staking/LiValidator.vue
```diff
@@ -49,9 +49,14 @@
       </div>
     </td>
     <td :class="{ 'hide-xs': showOnMobile !== 'expectedReturns' }">
-      {{
-        validator.expectedReturns ? percent(validator.expectedReturns) : `--`
-      }}
+      <span v-if="network === `emoney-testnet`">
```
faboweb: nooooooop, no network specific fixes please. why not apply the fix for all?

mariopino: I just want to hear from you :-)

Bitcoinera:
OK, tests were missing so I add then and while doing that I noticed that the function's behavior was kind of unpredictable, so had to do a minor fix in the `percentBig` function (actually I think it should be called `bigPercent` better).

Hopefully now all tests are in place and the function works in a more predictable manner

Bitcoinera:
Actually, why not apply this same function to all these huge numbers? Why does it need to be specifically for percents only?

src/components/staking/LiValidator.vue
jbibla: i may have mislead mario here 🙃

jbibla:
> why not apply this same function to all these huge numbers

what other numbers are you referring to?

src/scripts/num.js
```diff
@@ -80,6 +80,32 @@ export function percent(number = 0) {
   )
 }
 
+// Needed for e-money. This func uses short scale billions and trillions (respectively 1e9 and 1e12)
```
jbibla: fabo's probably right. we can remove the network specific reference.

tests/unit/specs/scripts/num.spec.js
```diff
@@ -80,6 +81,22 @@ describe(`number helper`, () => {
     expect(percent(0.2612)).toBe(`26.12%`)
   })
 
+  it(`should format a very big percent. Convert it to trillions`, () => {
+    expect(percentBig(`8719010560462937.9`)).toBe(`8,719.01 T %`)
```
jbibla: this is an internationalized number - is that going to be ok in the tests?

tests/unit/specs/scripts/num.spec.js
```diff
@@ -80,6 +81,22 @@ describe(`number helper`, () => {
     expect(percent(0.2612)).toBe(`26.12%`)
   })
 
+  it(`should format a very big percent. Convert it to trillions`, () => {
+    expect(percentBig(`8719010560462937.9`)).toBe(`8,719.01 T %`)
+  })
+
+  it(`should format a very big percent. Convert it to billions`, () => {
+    expect(percentBig(`719010560462.9379`)).toBe(`719.01 B %`)
```
jbibla: what number is this?

719 billion? if so, do we need the `.01` at the end?

tests/unit/specs/scripts/num.spec.js
```diff
@@ -80,6 +81,22 @@ describe(`number helper`, () => {
...
+  it(`should format a very big percent. Convert it to trillions`, () => {
+    expect(percentBig(`8719010560462937.9`)).toBe(`8,719.01 T %`)
+  })
+
+  it(`should format a very big percent. Convert it to billions`, () => {
+    expect(percentBig(`719010560462.9379`)).toBe(`719.01 B %`)
+  })
+
+  it(`should format a very big percent. Convert it to millions`, () => {
+    expect(percentBig(`7010560.46`)).toBe(`7.01 M %`)
```
jbibla: should probably just be `7 M`

i think 1 decimal place is customary for giant trunacted numbers like this (for all `bigPercent`)

Bitcoinera:
> what other numbers are you referring to?

The new e-Money testnet is a world of millionaires utopia:



What I wanted to mean in the end is, why not we just format all these numbers with BigNumber (ellipsis), including percentages?

I mean, we should be consistent right? 
Either we use BigNumber or `bigPercent` to format big numbers. Or is BigNumber more suitable for long decimals and this one for big integers? I guess that can work out.

tests/unit/specs/scripts/num.spec.js
Bitcoinera: For now it looks like yes

tests/unit/specs/scripts/num.spec.js
Bitcoinera: OK, let me change that

src/scripts/num.js
```diff
@@ -80,6 +80,31 @@ export function percent(number = 0) {
...
+    suffix = "M"
+  } else if (Math.abs(Number(formatted)) >= 1e3) {
+    formatted = Number(formatted) / 1e3
+    suffix = "K"
+  }
+  return (
+    new Intl.NumberFormat(language, {
+      minimumFractionDigits: 0,
+      maximumFractionDigits: 1
+    }).format(formatted) + ` ${suffix} %`
```
faboweb: removing '%' you could use the same logic for all the values.

faboweb: like `setDecimalLength`

Bitcoinera: Yes, this is what I was thinking actually

Bitcoinera:
Hey, how do you like this?



To be able to apply both shortDecimals and bigFigure filters (that is percentBig renamed) I had to create a new filter called `bigFigureOrShortDecimals` (I'm open to suggestions) where I just check the number of positive digits and in case it is over `1e6` then I apply bigFigure, otherwise apply shortDecimals.

tests/unit/specs/scripts/num.spec.js
```diff
@@ -82,19 +82,15 @@ describe(`number helper`, () => {
...
+    expect(bigFigure(`719010560462.9379`)).toBe(`719 B`)
   })
 
   it(`should format a very big percent. Convert it to millions`, () => {
-    expect(percentBig(`7010560.46`)).toBe(`7 M %`)
-  })
-
-  it(`should format a very big percent. Convert it to k-thousands`, () => {
-    expect(percentBig(`719010.56`)).toBe(`719 K %`)
+    expect(bigFigure(`7010560.46`)).toBe(`7 M`)
```
faboweb: missing the percent ;)

src/scripts/num.js
```diff
@@ -170,5 +201,7 @@ export default {
   percent,
   percentInt,
   prettyDecimals,
-  roundObjectPercentages
+  roundObjectPercentages,
+  bigFigure,
```
jbibla: why `figure` as opposed to `bigNumber` or `truncateBigNumber`?

jbibla:
this is really cool!

tests/unit/specs/scripts/num.spec.js
Bitcoinera: Well, actually I see a percent too much in there xD

src/scripts/num.js
Bitcoinera: I chose `bigFigure` to distinguish it from the `BigNumber` lib. It could also be called `truncatedBigNumber` but then the naming of the selector filter would also need to change. I think `truncatedBigNumberOrShortDecimals` can already hurt the eyes

src/components/staking/LiValidator.vue
```diff
@@ -49,7 +49,7 @@
       </div>
     </td>
     <td :class="{ 'hide-xs': showOnMobile !== 'expectedReturns' }">
-      {{ validator.expectedReturns | bigFigureOrShortDecimals }}
+      {{ validator.expectedReturns | bigFigureOrShortDecimals }} %
```
faboweb: expected returns are in percent value. so they are delivered as `0.01` = `1%`. this would display only `0.01 %`

faboweb: you need a bigFigureOrPercent function

Bitcoinera: thanks

src/scripts/num.js
jbibla: `truncatedBigNumberOrShortDecimals`

it's odd to me to have an `Or` in this. not really sure when to use it or what it will do.

Bitcoinera: How would you call it then?

jbibla: its ok for now.

jbibla:
still WIP?

what account are you using for those crazy numbers?

Bitcoinera:
> still WIP?

Oh sorry. This is not WIP anymore xD

I just forgot to remove it 😆

PR #3542, opened 2020-02-06
Ana/add create account in testnets

Bitcoinera:
Closes #ISSUE

**Description:**

All the intellectual property rights go to @jbibla 😜

OK, you probably want my solution here too much either because I used _that_ "funky" string destructuring to tell if a network is a testnet or not.

But I mean, to do this check in this component I would need to call Apollo and this is much cheaper computaniolly speaking...

Another option would be to pass this info (testnet or not) as a param through routing.

Do you have any other better solution?

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/src/components/common/TmSessionSignUp.vue
+++ b/src/components/common/TmSessionSignUp.vue
@@ -109,8 +109,10 @@ export default {
       }
     },
     isTestnet() {
-      // this.networks is undefined here O.o why??
-      return this.networks.filter(({ id }) => id === this.network)[0].testnet
+      const selectedNetwork = this.networks.find(
+        ({ id }) => id === this.network
+      )
+      return selectedNetwork ? selectedNetwork.testnet : false
     }
   },
   methods: {
--- a/src/components/common/TmSessionSignUp.vue
+++ b/src/components/common/TmSessionSignUp.vue
@@ -5,11 +5,7 @@
         Create a new address
       </h2>
       <div
-        v-if="
-          !session.insecureMode &&
-            !session.mobile &&
-            !(network.slice(-7) === `testnet`)
-        "
+        v-if="!session.insecureMode && !session.mobile && !isTestnet"
         class="session-main"
       >
         <InsecureModeWarning />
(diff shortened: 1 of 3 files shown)


src/components/common/TmSessionSignUp.vue
```diff
@@ -4,7 +4,14 @@
       <h2 class="session-title">
         Create a new address
       </h2>
-      <div v-if="!session.insecureMode && !session.mobile" class="session-main">
+      <div
+        v-if="
+          !session.insecureMode &&
+            !session.mobile &&
+            !(network.slice(-7) === `testnet`)
```
jbibla: 🤢

jbibla: i think @iambeone made a helper function to check if it's a network - maybe we can use that or something?

faboweb: we have the `testnet` property for checking if this is a testnet

Bitcoinera: > 🤢

xD a bit exaggerated no?

src/components/common/TmSessionSignUp.vue
```diff
@@ -107,6 +107,10 @@ export default {
       set(value) {
         this.$store.commit(`updateField`, { field: `signUpName`, value })
       }
+    },
+    isTestnet() {
+      // this.networks is undefined here O.o why??
+      return this.networks.filter(({ id }) => id === this.network)[0].testnet
```
faboweb:
```suggestion
      return this.networks.find(({ id }) => id === this.network).testnet
```

faboweb: don't really know why this.networks is undefined tough

Bitcoinera: yes, I know. But your solution works. I don't understand it either 🤷‍♀

Bitcoinera:
Codecov is not working?

src/components/common/TmSessionSignUp.vue
jbibla: yes, very exaggerated.

jbibla:
might be waiting for the changelog 🙃

Bitcoinera:
> might be waiting for the changelog 🙃

xD

PR #3550, opened 2020-02-08
Ana/possible how to handle contact us

Bitcoinera:
Closes #ISSUE

**Description:**

Since the Intercom thing is not working in desktop in the last two merged PRs, I went on and made this possible alternative. Very simple, it `mobileApp` is true, then we show Intercom. But if it is false (Desktop), then `mailto` is fired.

Unfortunately, the tests for the intercom triggering are failing. I don't understand why, but `mobileApp` is always false. It seems I cannot override this value even by doing

```
    jest.mock(`src/../config.js`, () => ({
      mobileApp: true
    }))
```

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/changes/ana_possible-how-to-handle-contact-us
+++ b/changes/ana_possible-how-to-handle-contact-us
@@ -1,0 +1,1 @@
+[Changed] [#3550](https://github.com/cosmos/lunie/pull/3550) To contact us, intercom will be displayed if it is from a  mobile app and otherwise mailto will be fired @Bitcoinera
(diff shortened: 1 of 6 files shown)


jbibla:
ana, thanks for looking into the solution for this. 

intercom does work as expected on desktop, just not on netlify links.

we can just remove the `config.mobileApp` and it will work as expected.

Bitcoinera:
> intercom does work as expected on desktop, just not on netlify links.

Wow, I need to look into this. It seems it doesn't work on localhost either or maybe I just need to set it to 'production'.

Thank you for the screenshot and clarification 🙏 

Now I also would like to look into why these two messages in both PageValidator and PageTransactions are now gone in `develop`...

But yes, I will close this for now seeing that this is not needed

PR #3558, opened 2020-02-12
Aleksei/fix multisendmessage

iambeone:
Closes #ISSUE

**Description:**

Fixed multiSendMessage

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/src/components/transactions/message-view/MultiSendMessageDetails.vue
+++ b/src/components/transactions/message-view/MultiSendMessageDetails.vue
@@ -9,18 +9,13 @@
       <template v-if="toYourself">
         <span>To yourself —&nbsp;</span>
       </template>
-      <template v-else-if="sentFromSessionAddress">
+      <template v-else-if="isSent">
         <span>To&nbsp;</span>
-        <Bech32 :address="transaction.value.outputs[0].address" />
-      </template>
-      <template v-else-if="receivedToSessionAddress">
-        <span>From&nbsp;</span>
-        <Bech32 :address="transaction.value.inputs[0].address" />
+        <Bech32 :address="recipient" />
       </template>
       <template v-else>
         <span>From&nbsp;</span>
-        <Bech32 :address="transaction.value.inputs[0].address" /> to
-        <Bech32 :address="sessionAddress" />
+        <Bech32 :address="sender" />
       </template>
     </div>
     <div class="tx__content__right">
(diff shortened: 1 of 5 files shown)


src/components/transactions/message-view/MultiSendMessageDetails.vue
```diff
@@ -60,12 +55,35 @@ export default {
     }
   },
   computed: {
+    recipient() {
+      let index = this.transaction.value.inputs.findIndex(
+        input => input.address == this.sessionAddress
```
Bitcoinera: I didn't know this `findIndex` function. Wow, looks powerful!

src/components/transactions/message-view/MultiSendMessageDetails.vue
```diff
@@ -60,12 +55,35 @@ export default {
     }
   },
   computed: {
+    recipient() {
```
faboweb: what are you doing here? why the conditional that the user address is in the inputs?

src/components/transactions/message-view/MultiSendMessageDetails.vue
```diff
@@ -60,12 +55,35 @@ export default {
...
+      let index = this.transaction.value.inputs.findIndex(
+        input => input.address == this.sessionAddress
+      )
+      if (index !== undefined && this.transaction.value.outputs[index]) {
+        return this.transaction.value.outputs[index].address
+      }
+      return false
+    },
+    sender() {
```
faboweb: same here, why is the sender necessarily in the outputs?

iambeone: Only my thoughts... Like if the user is in inputs (e.g. senders) he is a sender, if he is in outputs (e.g. recipients) he is a recipient

Bitcoinera: Isn't it actually the opposite?

iambeone: True, my fault, but the code is right) We are looking for a sender if our address is a recipient and we are looking for a recipient if our address is sender.

faboweb: I get it. Deserves comments though

faboweb:
A lot of effort for us wanting to abstract transactions

faboweb:
Should we just drop this? Mario is already working on the abstraction

Bitcoinera:
> Should we just drop this? Mario is already working on the abstraction

What do you think @iambeone ? Seems reasonable

Issue #351, opened 2018-01-12
the component file structure doesn't make sense anymore


jolesbi:
the file structure doesn't match the app anymore. 

- [ ] reorganize files and folders to match app

one option...
```
/common
/balances
/transactions
/delegates
/proposals
/validators
/blocks
```

vs.
```
/common
/govern
/monitor
/staking
/wallet
```

faboweb:
The organization seems to be domain/module specific instead of navigation specific. But I have no strong opinion. Only thing: Don't remove the govern folder. :)

faboweb:
Is this still an issue? How do you feel about this @nylira ?

nylira:
Let's leave this as-is for now. It's not important, and we have to switch it again later.

PR #3559, opened 2020-02-12
Ana/emoney-add multidenom rewards to tmbalance & livalidator

Bitcoinera:
Closes #ISSUE

**Description:**

Making the necessary changes to accommodate multidenom rewards.

Please give me all the feedback you can come up with to improve the UI. (Remember that Mario's PR for big numbers will also be added)

The responsiveness is still really bad. It's a work in progress.

Thank you! 🚀

---

For contributor:

For reviewer:


Bitcoinera:
OK, this is how it is going now. The only thing is that at the end I'm getting the pyramid (below 667px) but I don't know how to do this any way better... I mean, the scrolling how I had it was breaking the styles. 

Maybe have the pyramid but just make the fonts for rewards smaller? What do you think?

Also need to look into the last bit, why is showing that strange margin on the right...

GIF:

Bitcoinera:
Hmm... not working for Terra 🤷‍♀ 

So probably not that good. Something is failing. Also still compatibility problems with other networks.

I will set it back to "WIP". Not ready yet.

Bitcoinera:
OK, this branch should be call now "multidenom-format-rewards" 😎 

It's also going to include Terra and hopefully all the ones to come

Bitcoinera:
OK, now it should be just the test missing and this one is good to go

src/components/staking/LiValidator.vue
```diff
@@ -94,10 +94,30 @@ export default {
...
+      default: ""
     }
   },
   methods: {
-    percent
+    percent,
+    shortDecimals,
+    selectMostRelevantReward(rewards) {
+      rewards.sort((a, b) => b.amount - a.amount)
+      // we return the reward with the highest amount
```
jbibla: we should choose the staking reward, not the reward with the highest amount.

the "highest amount" has no meaning because we don't know the underlying value of the tokens.

we want to display the staking rewards next to the number of tokens staked.

src/components/staking/LiValidator.vue
```diff
@@ -94,10 +94,30 @@ export default {
...
+      return shortDecimals(rewards[0].amount)
+        .toString()
+        .concat(
+          this.isMultiDenomReward
+            ? ` ${rewards[0].denom.slice(-3).toUpperCase()}`
```
jbibla: who is TOM 😉 

i don't think this is good logic. we don't need to change the denom.

Screen Shot 2020-02-14 at 8 42 32 AM

src/scripts/common.js
```diff
@@ -54,3 +54,9 @@ export function toMicroDenom(denom) {
   }
   return lookup[denom] ? lookup[denom] : denom.toLowerCase()
 }
+
+export function removeUFromMicroDenom(denom) {
```
jbibla: should this be done in the API?

do we have uATOM's anywhere in the FE?

src/components/common/TmBalance.vue
```diff
@@ -31,23 +31,42 @@
...
-              <div v-if="overview.totalRewards" class="rewards">
-                <h3>Total Rewards</h3>
+              <div
+                v-if="overview.totalRewards && overview.totalRewards > 0.001"
+                class="rewards"
+              >
+                <h3>
+                  {{
+                    isMultiDenomReward
+                      ? `Total Rewards in ${stakingDenom}`
```
jbibla:
```suggestion
                      ? `Rewards in ${stakingDenom}`
```

src/scripts/common.js
Bitcoinera: Sorry, forgot to remove. This is from the previous approach, `selectMostRelevantReward`

src/components/staking/LiValidator.vue
Bitcoinera: hahaha. Yes, I also saw this one. It is outdated though. I am not showing now any denom for single-denom networks. It is not needed.

src/components/staking/LiValidator.vue
Bitcoinera: Yes, I already changed it

src/scripts/common.js
Bitcoinera: Oops, just noticed this is actually needed. Could be done in the API though, although it would require changing stuff in the FE (filters for `uatoms)`.

The thing is that in e-Money these tokens are called `eCHF`, `eEUR`, and in Terra `uMNT` and so on... so in e-Money it makes sense to display the whole thing, but not in Terra (since we are not displaying the micro unit).

In e-Money these ones should be called `ueCHF` actually. I don't know why they have it like that. 

I think it is better to wait for transaction standardization, since doing this in the API would invove removing filters in the FE

Bitcoinera:
I call this one ready! 💪

Bitcoinera:
OK, maybe it is not so ready. The problem I'm having here is with styles of the positioning of alt-tokens in Desktop view.

You can see here if there are only a few it looks kind of weird:

image

Whereas if there are a lot of them it looks OK:

image

I am thinking of using some kind of "conditional" CSS

jbibla:
Let's merge this amazing PR @Bitcoinera!!

not as nice as charlie would make it — but totally great for now! 

Screen Shot 2020-02-18 at 11 24 29 PM
Screen Shot 2020-02-18 at 11 24 40 PM
Screen Shot 2020-02-19 at 12 05 06 AM
Screen Shot 2020-02-19 at 12 04 54 AM
Screen Shot 2020-02-19 at 12 04 41 AM
Screen Shot 2020-02-19 at 12 05 23 AM

src/scripts/common.js
jbibla: amazing function name, for the record.

src/components/common/TmBalance.vue
```diff
@@ -469,30 +481,6 @@ export default {
...
-.token-balance {
-  font-weight: bold;
-}
-
-.currency-selector.tm-form-group {
-  position: absolute;
-  right: 1.25rem;
-  top: -0.7rem;
-}
-
```
Bitcoinera: Yep. Yesterday I was looking at all this CSS code and thinking "oh god, I need to clean all this up!"

Thanks for doing it @jbibla !

Bitcoinera:
OK, this one is definitely ready to go now!

Please somebody approve and merge and let's deliver 💪 !

src/components/common/TmBalance.vue
```diff
@@ -124,6 +158,7 @@ export default {
     ModalTutorial
   },
   filters: {
+    removeUFromMicroDenom,
```
faboweb: this already tells me something is wrong. the UI should not have to handle micro denoms in views

src/components/common/TmBalance.vue
```diff
@@ -233,6 +269,15 @@ export default {
       } else {
         return [this.stakingDenom]
       }
+    },
+    isMultiDenomReward() {
+      if (this.overview.rewards && this.overview.rewards.length > 0) {
+        return this.overview.rewards[0].denom !== this.overview.rewards[1].denom
```
faboweb: are multiple rewards with the same denom possible?

tests/unit/specs/components/common/TmBalance.spec.js
```diff
@@ -40,7 +40,21 @@ describe(`TmBalance`, () => {
       overview: {
         totalStake: 3210,
         liquidStake: 1230,
-        totalRewards: 1000.45
+        totalRewards: 1000.45,
+        rewards: [
+          {
+            amount: 1,
+            denom: `utoken1`
```
faboweb: here is the issue. you implemented the API wrong. the rewards need to be delivered in a view token

src/components/common/TmBalance.vue
Bitcoinera: yes, because of rewards from different validators.

But I don't like this logic either. This is why I want this PR luniehq/lunie-api#343 merged

src/components/common/TmBalance.vue
Bitcoinera: I thought I had deleted this one... but good that is signaling something wrong 👍

src/components/common/TmBalance.vue
faboweb: why does the overview query receive all rewards? doesn't only need a sum per denom?

src/components/common/TmBalance.vue
jbibla: 😇

tests/unit/specs/components/common/TmBalance.spec.js
Bitcoinera: It is in the works luniehq/lunie#351

I will change now tests

PR #3564, opened 2020-02-14
Aleksei/network in url

iambeone:
Closes #ISSUE

**Description:**

Initial fix to change network by url requests like /cosmos-hub-mainnet/validators

Thank you! 🚀

---

For contributor:

For reviewer:


Bitcoinera:
Interesting. But why do you think we need this? What is the purpose?

We are already connecting to each network every time the user selects it & logs in...

iambeone:
@Bitcoinera this one https://github.com/luniehq/lunie/issues/3213

src/components/common/AppMenu.vue
```diff
@@ -33,8 +33,8 @@
       @click.native="signIn()"
     />
     <div class="app-menu-main">
-      <router-link
-        class="app-menu-item hide-xs"
+      <routerLinkNetwork
+        cclass="app-menu-item hide-xs"
```
jbibla:
```suggestion
        class="app-menu-item hide-xs"
```

src/components/common/RouterLinkNetwork.vue
```diff
@@ -0,0 +1,50 @@
+<template>
+  <router-link
+    :class="cclass"
```
jbibla: this is very strange

src/routes.js
```diff
@@ -299,7 +221,92 @@ export default [
...
     path: `/feature-not-available/:feature`,
     component: () => import(`./components/common/PageFeatureNotAvailable`),
+    name: `fns`,
     props: true
   },
-  { path: `*`, component: () => import(`./components/common/Page404`) }
+  {
+    path: `/:networkId`,
+    component: () => import(`./components/common/setNetwork`),
+    children: [
```
jbibla: nice.

src/routes.js
```diff
@@ -299,7 +221,92 @@ export default [
   {
     path: `/feature-not-available/:feature`,
     component: () => import(`./components/common/PageFeatureNotAvailable`),
+    name: `fns`,
```
jbibla: prefer not to use abbreviations

src/components/common/RouterLinkNetwork.vue
iambeone: Could you explain? We can't use class property there

src/routes.js
iambeone: This is not needed, i forgot to remove

src/components/common/AppMenu.vue
jbibla: can you use `:class` here?

src/components/common/RouterLinkNetwork.vue
```diff
@@ -0,0 +1,50 @@
...
+    title: {
+      type: String,
+      default: null
+    }
+  },
+  computed: {
+    ...mapGetters([`network`]),
+    url() {
+      if (this.network.length) {
+        return `/${this.network}${this.to}`
```
jbibla: can we remove `-mainnet` from the link? so just `/terra/` or `/cosmos/`?

for testnets, we can leave `-testnet` in.

src/components/common/RouterLinkNetwork.vue
jbibla: i guess i would rather use a different word or change it with a symbol or something. i'm not sure. 

`:class="class_"` 

all feels strange.

src/components/common/RouterLinkNetwork.vue
Bitcoinera: I just left `cosmos-hub` and yes, `terra` for terra-mainnet. But I used slice, so you might not like it xD (please don't use the vomit emoji)

src/components/common/RouterLinkNetwork.vue
Bitcoinera: I changed it for `routerClass`. What do you think?

iambeone: I don't know why i have chosen cclass for this variable, so any changes will be good)

src/components/common/RouterLinkNetwork.vue
iambeone: Rewritten a bit

src/components/common/NetworkSetter.vue
```diff
@@ -8,15 +8,10 @@ export default {
...
-        )
-      } else {
-        return networks.find(
-          network => network === this.$route.params.networkId.concat(`-mainnet`)
-        )
-      }
+      return networks.find(
+        network =>
+          network.replace(/-mainnet$/, ``) === this.$route.params.networkId
+      )
```
Bitcoinera: Nice! Didn't know this method replace either. Great stuff

src/components/common/NetworkSetter.vue
```diff
@@ -35,9 +30,13 @@ export default {
       /* istanbul ignore next */
       update(data) {
         const networks = data.networks.map(network => network.id)
-        if (this.$route.params.networkId && this.findNetwork(networks)) {
+        let networkTitle
+        if (
+          this.$route.params.networkId &&
+          (networkTitle = this.findNetwork(networks))
```
Bitcoinera:
```suggestion
          (networkTitle === this.findNetwork(networks))
```

src/components/common/RouterLinkNetwork.vue
```diff
@@ -36,9 +36,7 @@ export default {
     ...mapGetters([`network`]),
     url() {
       if (this.network.length) {
-        const networkName = this.network.includes(`mainnet`)
-          ? this.network.slice(0, -8)
-          : this.network
+        const networkName = this.network.replace(/-mainnet$/, ``)
```
Bitcoinera: Cool

src/components/common/NetworkSetter.vue
Bitcoinera: Ok, this actually works. I am learning so much looking at your code @iambeone !

src/routes.js
```diff
@@ -301,5 +223,89 @@ export default [
     component: () => import(`./components/common/PageFeatureNotAvailable`),
     props: true
   },
-  { path: `*`, component: () => import(`./components/common/Page404`) }
+  {
+    path: `/:networkId`,
+    component: () => import(`./components/common/NetworkSetter`),
+    children: [
```
jbibla: should we make session screens children of network?

Bitcoinera: Doesn't make too much sense right? They are they same no matter which network you are connected to

src/components/common/NetworkSetter.vue
faboweb: this might work, but is confusing, this assignes networkTitle and returns true if networkTitle is not undefined. Let's make the code more readable by using findNetwork in line 33 and using a this.$route.params.networkId check inside of findNetwork

src/components/common/RouterLinkNetwork.vue
faboweb: this forces us to have the same id schema for every network. :/

faboweb: I think we should avoid string manipulations on IDs wherever possible

src/components/common/RouterLinkNetwork.vue
```diff
@@ -0,0 +1,51 @@
...
+    },
+    title: {
+      type: String,
+      default: null
+    }
+  },
+  computed: {
+    ...mapGetters([`network`]),
+    url() {
+      if (this.network.length) {
```
faboweb: is `network` an array?

faboweb:
Why not use a routeGuard as a solution? This way we need to implement the RouterLinkNetwork everywhere and new devs might forget.

src/components/common/RouterLinkNetwork.vue
iambeone: > I think we should avoid string manipulations on IDs wherever possible

Why?

src/components/common/RouterLinkNetwork.vue
iambeone: Nope, just checking if it's not empty

iambeone:
> Why not use a routeGuard as a solution? This way we need to implement the RouterLinkNetwork everywhere and new devs might forget.

How we can use it? What should it replace?

src/components/common/RouterLinkNetwork.vue
faboweb: because IDs are IDs and this forces us to a naming schema. IDs should not need to follow a naming schema.

src/components/common/RouterLinkNetwork.vue
faboweb: well then not use `length` but check if it is empty. this code suggests it is an array...

src/components/common/RouterLinkNetwork.vue
iambeone: But since then, how we are going to remove useless parts like mainnet?

faboweb: probably doing a lookup

faboweb:
we already use route guards. check initializeApp line 39. there you can rewrite the request and prefix the network name if needed.

faboweb:
but, let's ignore the centralization for now and take your solution for now. please follow up on the other comments though.

src/components/common/RouterLinkNetwork.vue
Bitcoinera: I am not sure how to do this. 

If the ID doesn't follow any naming schema/pattern, then how can we know which is the lexeme of the network and which the "useless" parts.

Maybe I didn't understand you correctly, because I don't see any possible solution to this

faboweb: there should not be "parts" in an ID. it could be random. that is why I forced everyone to use the `testnet` variable from the networks response. This is the same case. If we want the URL to be different then the ID, we should do a dictionary lookup. Or we just pass the `route` from the API. which is preferable as we don't want network specific logic on the client.

faboweb:
e2e tests failed

Bitcoinera:
> e2e tests failed

When I push nearly 90% of the times I get "Max attempts reached" in the e2e tests

src/components/common/RouterLinkNetwork.vue
iambeone: Maybe we can just store path in our db?

iambeone: Updated

faboweb:
unit and e2e tests are failing

Bitcoinera:
We should fix the e2e tests... they fail since they don't correctly wait for the netlified version to be deployed. It is a pity.

faboweb:
They are failing in here because of this PR not because of netlify unavailability (that happens only on first push)

Bitcoinera:
> They are failing in here because of this PR not because of netlify unavailability (that happens only on first push)

Weird, you are right that the netlified version is already there. Why is it failing with "Max attempts reached" though?

faboweb:
Yep, it'S actually a bug ^^'

faboweb:
I confirm this is not working

src/components/common/RouterLinkNetwork.vue
jbibla: #slugs

src/routes.js
jbibla: my thinking is that it would simpler because the user wouldn't have to choose a network before a session flow. 

we could also share links to create / recovering / signing in / exploring for specific networks which may be valuable for support questions and for customers who want to share a link to creating an address on their own network.

... just a thought. doesn't have to be done here / now.

iambeone:
Fixed some redirect issues and rewrite to routeGuard, but e2e tests still need to be fixed. Working on it

src/routes.js
```diff
@@ -6,306 +7,312 @@
...
-    name: `welcome`,
-    components: {
-      session: () => import(`./components/common/TmSessionWelcome`)
+export default (apollo, store) => {
+  return [
+    {
+      path: `/`,
+      beforeEnter: (to, from, next) =>
+        setNetwork({ to, from, next }, apollo, store),
+      component: () => import(`./components/common/NetworkSetter`)
```
faboweb: This component doesn't exist anymore

src/routes.js
```diff
@@ -6,306 +7,312 @@
...
+      component: () => import(`./components/common/PageCareers`)
+    },
+    {
+      path: `/feature-not-available/:feature`,
+      component: () => import(`./components/common/PageFeatureNotAvailable`),
+      props: true
+    },
+    {
+      path: `/:networkId`,
+      component: () => import(`./components/common/NetworkSetter`),
```
faboweb: same

src/scripts/setNetwork.js
```diff
@@ -0,0 +1,54 @@
...
+    fetchPolicy: "cache-first"
+  })
+  let path = to.path
+  if (path == "/") {
+    path = "/portfolio"
+  }
+  // current network slug
+  let network
+  if (
+    store.state.connection.networkSlug != to.params.networkId &&
```
faboweb:
```suggestion
    store.state.connection.networkSlug !== to.params.networkId &&
```

src/scripts/setNetwork.js
```diff
@@ -0,0 +1,54 @@
...
+    network = networks.find(network => network.slug === to.params.networkId)
+    if (!network) {
+      network = networks.find(network => network.default === true)
+    }
+    store.dispatch(`setNetwork`, network)
+    next(`/${network.slug}${path}`)
+  } else if (!to.params.networkId) {
+    // swithing to current network
+    if (store.state.connection.networkSlug) {
+      network = networks.find(
```
faboweb: these lines are doublicated with lines 30+, refactor

src/scripts/setNetwork.js
```diff
@@ -0,0 +1,54 @@
...
+        network => network.slug === store.state.connection.networkSlug
+      )
+      if (!network) {
+        network = networks.find(network => network.default === true)
+      }
+      store.dispatch(`setNetwork`, network)
+      next(`/${network.slug}${path}`)
+    } else {
+      network = networks.find(network => network.default === true)
+      next(`/${network.slug}${path}`)
```
faboweb: why is there no store.dispatch(`setNetwork`, network) in this case?

faboweb:
Looking good! Smaller issues.

src/routes.js
Bitcoinera: ? it does still exist. And still necessary it seems. It is located in `src/components/common`

src/routes.js
Bitcoinera: it's still alive! xD

src/scripts/setNetwork.js
Bitcoinera: Done

src/scripts/setNetwork.js
Bitcoinera: Refactored

src/scripts/setNetwork.js
Bitcoinera: Yes, I bad fixed it. Now it's good again

Bitcoinera:
> Looking good! Smaller issues.

Not so small.

Things currently failing:

- The `Back to validators` button still goes to the old url, `/validators`, thus failing.

- When you sign in with extension, we get the old bug of being always connected to Cosmos Hub 3, and thus failing too.

- I had to do a strange hack to pass the unit tests. Concretely the one about the fallback network. You can see my hack at line 54 of `vuex/modules/connection.js`. Otherwise it wouldn't get inside the `setNetwork` function 🤷‍♀  Maybe it's fine like this. It seems to be working this way.

src/vuex/modules/connection.js
```diff
@@ -34,14 +34,14 @@ export default function({ apollo }) {
         ? data.networks.find(network => network.id === persistedNetwork)
         : false
       if (persistedNetwork && storedNetwork) {
-        await commit(`setNetworkId`, storedNetwork.id)
+        await dispatch(`setNetwork`, storedNetwork)
```
faboweb: why was this necessary?

Bitcoinera: Because this is the proper way, otherwise it fails in execution. But I changed it yesterday fast to pass the tests 😅

faboweb:
we can simplify this more if we always use the network slug for all routes I belive

src/scripts/setNetwork.js
```diff
@@ -20,11 +20,33 @@ export const setNetwork = async ({ to, next }, apollo, store) => {
...
+  if (to.params.networkId) {
+    const subPathRegexp = /\/.+(\/.+)/ //extracts the subpath
+    const match = subPathRegexp.exec(to.path)
+    if (match && match.length >= 2) {
+      // the path is is of form /cosmos-hub/portfolio
+      path = match[1]
+    } else {
+      // the path is is of form /cosmos-hub so we forward to portfolio
+      path = "/portfolio"
+    }
```
Bitcoinera: There is something here I don't get. Wasn't the reloading happening in production?

Or it was also happening here? In what cases? I didn't notice 👀

Bitcoinera:
> we can simplify this more if we always use the network slug for all routes I belive

Also the session ones right? And then just centralize it

src/routes.js
faboweb: this is useless, need to remove

PR #3407, opened 2020-01-12
WIP: Ana/clear transactions amount handling

Bitcoinera:
Closes #ISSUE

**Description:**

This PR goes together with #264 in the API.

Basically just getting rid of all the filters for amount handling in the FE.

Especially challenging for the modals. Still looking into it.

Thank you! 🚀

---

For contributor:

- [ ] Added changes entries. Run `yarn changelog` for a guided process.
- [ ] Reviewed `Files changed` in the github PR explorer
- [ ] Attach screenshots of the UI components on the PR description (if applicable)
- [ ] Scope of work approved for big PRs

For reviewer:

- [ ] Manually tested the changes on the UI

PR #3567, opened 2020-02-17
Aleksei/fee in extension

iambeone:
Closes #ISSUE

**Description:**

@Bitcoinera knows the issue. If the fee is lower than 1 denom item it wasn't shown in extension on sign step. Not the solution, just indication of the reason for the problem.

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/src/ActionModal/utils/ActionManager.js
+++ b/src/ActionModal/utils/ActionManager.js
@@ -1,7 +1,7 @@
 import config from "src/../config"
 import { getSigner } from "./signer"
 import transaction from "./transactionTypes"
-import { uatoms } from "scripts/num"
+import { gasMilliUnits } from "scripts/num"
 import { toMicroDenom } from "src/scripts/common"
 import { getGraphqlHost } from "scripts/url"
 import {
@@ -237,13 +237,13 @@ export default class ActionManager {
 
 function convertCurrencyData(amounts) {
   return amounts.map(({ amount, denom }) => ({
-    amount: toMicroAtomString(amount),
+    amount: toGasMilliUnitString(amount),
     denom: toMicroDenom(denom)
   }))
 }
 
-function toMicroAtomString(amount) {
-  return String(uatoms(amount))
+function toGasMilliUnitString(amount) {
+  return String(gasMilliUnits(amount))
 }
 
 // limitation of the Ledger Nano S, so we pick the top 5 rewards and inform the user.
--- a/src/ActionModal/utils/ActionManager.js
+++ b/src/ActionModal/utils/ActionManager.js
@@ -1,7 +1,7 @@
 import config from "src/../config"
 import { getSigner } from "./signer"
 import transaction from "./transactionTypes"
-import { gasMilliUnits } from "scripts/num"
+import { multiplyToGasDecimals } from "scripts/num"
 import { toMicroDenom } from "src/scripts/common"
 import { getGraphqlHost } from "scripts/url"
 import {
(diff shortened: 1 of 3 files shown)


src/scripts/num.js
```diff
@@ -135,7 +135,7 @@ export function atoms(number = 0) {
 export function uatoms(number = 0) {
   return BigNumber(number)
     .times(1e6)
-    .toFixed(0)
+    //.toFixed(0) // removing this will solve the issue with fee amount if extension
```
faboweb: this function converts view values of tokens into internal values. i.e. converts 1.5atoms into 1500000 uatoms. uatoms never have decimals.

If adding decimals to uatoms. I guess the problem lies in a different location.

iambeone: Ok, then we need to rewrite fee defining function in signRequest

src/ActionModal/utils/ActionManager.js
```diff
@@ -237,13 +237,13 @@ export default class ActionManager {
...
   return amounts.map(({ amount, denom }) => ({
-    amount: toMicroAtomString(amount),
+    amount: toGasMilliUnitString(amount),
     denom: toMicroDenom(denom)
   }))
 }
 
-function toMicroAtomString(amount) {
-  return String(uatoms(amount))
+function toGasMilliUnitString(amount) {
```
faboweb: this will probably need to take in a denom to determin the currency conversion. we could serve the currency conversions from the API

src/ActionModal/utils/ActionManager.js
```diff
@@ -1,7 +1,7 @@
 import config from "src/../config"
 import { getSigner } from "./signer"
 import transaction from "./transactionTypes"
-import { uatoms } from "scripts/num"
```
faboweb: is uatoms still used?

src/ActionModal/utils/ActionManager.js
Bitcoinera: you mean to show all values in fiat right? I think this would make another PR though

src/ActionModal/utils/ActionManager.js
Bitcoinera: no, it was only being used for that particular function (converting the gas fee)

faboweb: then let'S remove the function?

src/ActionModal/utils/ActionManager.js
faboweb: no, to convert different tokens with different milli unit precision

src/ActionModal/utils/ActionManager.js
faboweb: but yeah let's do it another PR

src/ActionModal/utils/ActionManager.js
faboweb: but yeah let's do it another PR

src/ActionModal/utils/ActionManager.js
Bitcoinera: this we will only be able to do once we have fully integrated the Transaction Abstraction.

For that we already have my PR #3407 waiting for the right time.

src/scripts/num.js
```diff
@@ -139,6 +139,12 @@ export function uatoms(number = 0) {
     .toString()
 }
 
+export function gasMilliUnits(number = 0) {
+  return BigNumber(number)
```
faboweb: why is this working? this is just removing the fixed part. but milliunits should never have decimals

Bitcoinera: because we are sending them as decimals. Should I change the name or should we change the format and use proper milliunits for gas fees? 

As an user I think it is easier to understand decimals, since it is the same unit they are using for their transaction and therefore doesn't lead to confusion.

I need to change the naming then

faboweb: this is used to convert view value to milli unit values, right?

Bitcoinera: No. That is what is confusing. It is simply doing `times(1e6)`. That's is all it has to do.

I'm changing the name

Bitcoinera:
I think this is now fixed right? Probably related to having the incorrect denoms (in milliunits) coming from the API.

Can someone confirm?

PR #3581, opened 2020-02-20
Ana/drop claim rewards gasprice temporary hack

Bitcoinera:
Closes #ISSUE

**Description:**

~~(WARNING: horrible network-specific logic inside. Have your eyes ready 🙈).~~
(no, poor monkey! What am I actuallydoing 😆 )

~~Actually a very rudimentary attempt~~

~~I see here a good use case for the multi field in the Network schema to avoid this kind of super ugly hacks.~~ Ok, actually this doesn't make sense as the problem is how much gas NGM is somehow demanding to be transferred, which is not happening with LUNA for example...

How to fix this then? 🤔  I guess add some gas prices dictionary to the FE

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/src/ActionModal/components/ActionModal.vue
+++ b/src/ActionModal/components/ActionModal.vue
@@ -492,20 +492,15 @@ export default {
         account: this.overview.accountInformation
       }
     },
-    // TODO: delete in favor of the multi field in network
-    isMultiDenomNetwork() {
-      return this.balances.length > 1 &&
-        this.balances[0].denom !== this.balances[1].denom
-        ? true
-        : false
-    },
     getDenom() {
       return this.selectedDenom || this.network.stakingDenom
     },
     selectedBalance() {
+      // Here is a good use case for the multi field I proposed for the network schema...
       const defaultBalance = {
         amount: 0,
-        gasPrice: 4e-7 // TODO: temporary fix for claim rewards for NGM
+        // awful network-specific logic. But how to do it otherwise?
+        gasPrice: this.network.stakingDenom === `NGM` ? 4e-7 : config.default_gas_price
       }
       if (this.balances.length === 0 || !this.network) {
         return defaultBalance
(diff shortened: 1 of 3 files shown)


src/scripts/common.js
```diff
@@ -54,3 +54,13 @@ export function toMicroDenom(denom) {
...
 }
+
+export function gasPricesDictionary(denom) {
+  const lookup = {
+    ATOM: "6.5e-9",
+    MUON: "6.5e-9",
+    LUNA: "6.5e-9",
+    NGM: "4e-7"
+  }
+  return lookup[denom] ? lookup[denom] : `6.5e-9`
```
iambeone: You can do it just like that
```
const lookup = {
     NGM: "4e-7"
   }
return lookup[denom] || `6.5e-9`
```

src/ActionModal/components/ActionModal.vue
```diff
@@ -492,20 +493,14 @@ export default {
...
-    },
     getDenom() {
       return this.selectedDenom || this.network.stakingDenom
     },
     selectedBalance() {
       const defaultBalance = {
         amount: 0,
-        gasPrice: 4e-7 // TODO: temporary fix for claim rewards for NGM
+        // awful network-specific logic. But how to do it otherwise?
+        gasPrice: gasPricesDictionary(this.getDenom)
```
faboweb: don'T we get this from the API for all the denoms?

Bitcoinera: The problem is the claim's reward case, where there is no balance selected (soon to change with transaction abstraction).

So [this ](https://github.com/luniehq/lunie/blob/7e078d4e345dd2162327e7e9da584e28977a806f/src/ActionModal/components/ActionModal.vue#L503)is the problematic line:

faboweb: if there is no balance selected you can assume the stakingDenom right?

Bitcoinera: That is what I am doing. This is why there are only staking denoms in the dictionary

faboweb: but you are replicating the data in the FE code. you can just use the same data you added to the API here. just default `selectedDenom` to `stakingDenom`. this also reduces the logic here.

Bitcoinera: this makes sense actually. Let me try that

Bitcoinera: I did try it in #3592. There is something a bit weird with one test though.

PR #3582, opened 2020-02-21
handle infinity numbers

Bitcoinera:
Closes #ISSUE

**Description:**

Just added the case for numbers bigger to 99 trillion in the `bigFigure` filter, so it returns infinity (∞)

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/changes/ana_add-infinity-to-bigfigure
+++ b/changes/ana_add-infinity-to-bigfigure
@@ -1,0 +1,1 @@
+[Changed] [#3582](https://github.com/cosmos/lunie/pull/3582) Adds a return infinity symbol case in the bigFigure filter when number is over 99 T @Bitcoinera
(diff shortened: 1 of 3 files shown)


faboweb:
🙈  does this actually happen or is it just a conversion error?

Bitcoinera:
> 🙈 does this actually happen or is it just a conversion error?

It was a conversion error but I thought it was a good idea to implement this too

Bitcoinera:
Agh, e2e still gets stuck in the first run...

faboweb:
I don't think this will ever happen and only appeared by a sever bug. Let's not add not needed code.

Bitcoinera:
> I don't think this will ever happen and only appeared by a sever bug. Let's not add not needed code.

We never thought we would deal with billions and here we are. Especially in testnets anything is possible 🤷‍♀  But OK (plus the infinity symbol is mega cool)

PR #3583, opened 2020-02-22
Ana/fix transaction item in extension

Bitcoinera:
Closes #ISSUE

**Description:**

Uploaded to try in the extension

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/config.js
+++ b/config.js
@@ -16,6 +16,7 @@ export default {
   // Ledger
   CosmosAppTestModeAllowed: false,
   mobileApp: Boolean(process.env.MOBILE_APP),
+  isExtension: Boolean(process.env.EXTENSION),
 
   graphqlHost: graphql,
 
--- a/src/components/transactions/TransactionItem.vue
+++ b/src/components/transactions/TransactionItem.vue
@@ -25,6 +25,7 @@
 <script>
 import { messageType } from "./messageTypes.js"
 import TransactionMetadata from "./TransactionMetadata"
+import config from "src/../config"
 import Bech32 from "common/Bech32"
 
 import {
@@ -73,7 +74,8 @@ export default {
     }
   },
   data: () => ({
-    show: false
+    show: false,
+    isExtension: config.isExtension
   }),
   computed: {
     messageTypeComponent: function() {
@@ -107,7 +109,9 @@ export default {
   },
   methods: {
     toggleDetail(event) {
-      if (event.target.className !== `address`) {
+      if (this.isExtension) {
+        return
+      } else if (event.target.className !== `address`) {
         this.show = !this.show
       }
     }
--- a/src/components/transactions/TransactionItem.vue
+++ b/src/components/transactions/TransactionItem.vue
@@ -109,6 +109,7 @@ export default {
     }
   },
   methods: {
+    /* istanbul ignore next */
     toggleDetail(event) {
       if (this.isExtension) {
         this.showValidators = !this.showValidators
--- a/src/components/transactions/TransactionItem.vue
+++ b/src/components/transactions/TransactionItem.vue
@@ -110,11 +110,11 @@ export default {
   },
   methods: {
     /* istanbul ignore next */
-    toggleDetail(event) {
+    toggleDetail() {
       if (this.isExtension) {
         this.showValidators = !this.showValidators
         return
-      } else if (event.target.className !== `address`) {
+      } else {
         this.show = !this.show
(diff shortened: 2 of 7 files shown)


src/components/transactions/TransactionItem.vue
```diff
@@ -73,7 +74,9 @@ export default {
     }
   },
   data: () => ({
-    show: false
+    show: false,
+    showValidators: false,
```
jbibla: why do we need this boolean?

src/components/transactions/TransactionItem.vue
```diff
@@ -73,7 +74,9 @@ export default {
     }
   },
   data: () => ({
-    show: false
+    show: false,
```
jbibla: do we still need this is `show-validators`?

config.js
```diff
@@ -16,6 +16,7 @@ export default {
   // Ledger
   CosmosAppTestModeAllowed: false,
   mobileApp: Boolean(process.env.MOBILE_APP),
+  isExtension: Boolean(process.env.EXTENSION),
```
jbibla: 👍

jbibla:
@Bitcoinera cha cha cha... changelog 💃

Bitcoinera:
> i like the `isExtension` flag. however, this feels like a lot of extra logic for something that should just work as expected (validators should display in transactions). maybe we should just wait for transaction abstraction? is that the issue here?

Well, I couldn't find any other solution apart from adding this variable and the issue doesn't lie there really... Since this is actually not solving the validators not displaying issue but the empty box issue (tx metada).

If you click on the TransactionTiem the Metadata slide pops out. How can you fix it otherwise but telling the slide not to pop when it is in Extension?

The validators not displaying issue is a different one and have it almost fixed it. I am just getting the pic from the DB. There is no other alternative (transaction abstraction doesn't help either here. This is just how the extension works)

src/components/transactions/TransactionItem.vue
Bitcoinera: this is to make the show-all-validators thing in the WithdrawRewardsMessageDetails component independent from the `show` boolean coming from TransactionItem (so in the extension you actually get the cool effect of validators going vertical with their names)

jbibla:
ok i think i understand. thanks for the clarification.

> If you click on the TransactionItem the Metadata slide pops out. How can you fix it otherwise but telling the slide not to pop when it is in Extension?

why can't we display the metadata, like we do in the FE? why does the tx in the extension need to be different?

Bitcoinera:
> why can't we display the metadata, like we do in the FE? why does the tx in the extension need to be different?

The problem is that metadata doesn't exist yet at that point since you are about to sign for that transaction to be broadcast to the network

src/components/transactions/TransactionItem.vue
Bitcoinera: no, this one is the `show` for the transaction metadata, so yes, we still need it here

jbibla: so this should be called `showMetaData` probably 😸

jbibla:
> The problem is that metadata doesn't exist yet at that point since you are about to sign for that transaction to be broadcast to the network

ah! brilliant! thank you! i did not understand the problem.

you are absolutely correct. this is great.

src/components/transactions/TransactionItem.vue
jbibla: wait, there is already a prop called `showMetaData` on the `TransactionMetaData` component.

src/components/transactions/TransactionItem.vue
```diff
@@ -106,9 +109,14 @@ export default {
...
   },
   methods: {
+    /* istanbul ignore next */
     toggleDetail(event) {
-      if (event.target.className !== `address`) {
+      if (this.isExtension) {
+        this.showValidators = !this.showValidators
+        return
+      } else if (event.target.className !== `address`) {
         this.show = !this.show
```
jbibla: this isn't being used anymore, right?

Bitcoinera: No, you are right. I just deleted it

faboweb:
horribly complex. why not just add a `v-if="!isExtension"` to line 17?

Bitcoinera:
> horribly complex

xD

Bitcoinera:
Closed on favor of #3590

PR #3584, opened 2020-02-22
Ana/add currency selector back and apply Charlie's design

Bitcoinera:
Closes #ISSUE

**Description:**

Applying Charlie's design (concretely [this ](https://sketch.cloud/s/Oy9lm/a/1Mp1Zp)one) to TmBalance and adding back the currency selector logic.

Now it looks like this:

image

Thank you! 🚀

---

For contributor:

For reviewer:


Bitcoinera:
The design has only been partially applied. Maybe we should go for the whole thing?

src/components/common/TmBalance.vue
```diff
@@ -143,13 +189,14 @@
 </template>
 <script>
 import { bigFigureOrShortDecimals } from "scripts/num"
+import { removeLastCharacter } from "src/scripts/common"
```
jbibla: what makes this script necessary?

src/components/common/TmBalance.vue
```diff
@@ -19,6 +19,24 @@
...
+                class="currency-flag"
+                :src="
+                  '/img/icons/currencies/' +
+                    selectedFiatCurrency.toLowerCase() +
+                    '.png'
+                "
+                alt="USA flag"
+              />
+              <select v-model="selectedFiatCurrency">
+                <option>EUR</option>
```
jbibla: should we populate this list dynamically or is hardcoded going to work for now?

src/components/common/TmBalance.vue
```diff
@@ -19,6 +19,24 @@
...
             </div>
+            <div class="currency-selector">
+              <img
+                class="currency-flag"
+                :src="
+                  '/img/icons/currencies/' +
+                    selectedFiatCurrency.toLowerCase() +
+                    '.png'
+                "
+                alt="USA flag"
```
jbibla: the alt tag should change with the `selectedFiatCurrency`

src/components/common/TmBalance.vue
```diff
@@ -457,6 +479,55 @@ export default {
...
+}
+
+select option {
+  background: var(--app-bg);
+  color: var(--txt);
+  font-family: var(--sans);
+}
+
+.currency-flag {
+  width: 15px;
```
jbibla: `1rem` (in lunie) is `16px` so probably `1rem` is fine! 
```suggestion
  width: 1rem;
```

src/components/common/TmBalance.vue
```diff
@@ -457,6 +479,55 @@ export default {
...
+  margin-right: 3px;
```
jbibla: `0.25rem` (in lunie) is `4px` which is probably fine!

```suggestion
  margin-right: 0.25rem;
```

src/components/common/TmBalance.vue
```diff
@@ -457,6 +479,55 @@ export default {
...
+}
+
+.currency-selector {
+  display: flex;
+  align-items: center;
+}
+
+.total-fiat-value {
+  min-width: 30px;
```
jbibla:
```suggestion
  min-width: 2rem;
```

src/components/common/TmBalance.vue
```diff
@@ -457,6 +479,55 @@ export default {
...
+  margin-top: 0.25rem;
+}
+
+.user-box {
```
jbibla: what makes this a good class name?

src/components/common/TmBalance.vue
```diff
@@ -457,6 +479,55 @@ export default {
...
+  border-radius: 1.25rem;
+  display: inline-block;
+  cursor: pointer;
+}
+
+.user-box:hover {
+  color: var(--link-hover);
+}
+
+.value-div {
```
jbibla: maybe this class could be more descriptive? what value is it actually for?

src/components/common/TmBalance.vue
```diff
@@ -457,6 +479,55 @@ export default {
...
+  border: 1px solid var(--primary-alpha);
+  padding: 5px;
```
jbibla: would `0.25rem` work here?

src/components/common/TmBalance.vue
```diff
@@ -457,6 +479,55 @@ export default {
...
+  margin-right: 10px;
```
jbibla: would `0.5rem` work here?

src/components/common/TmBalance.vue
```diff
@@ -457,6 +479,55 @@ export default {
...
+  border-radius: 5px;
```
jbibla: would `0.25rem` work here?

src/components/common/TmBalance.vue
```diff
@@ -170,8 +217,7 @@ export default {
       stakingDenom: "",
       sentToGA: false,
       balances: [],
-      selectedTokenFiatValue: `Tokens Total Fiat Value`,
-      selectedFiatCurrency: `EUR`, // EUR is our default fiat currency
+      selectedFiatCurrency: `EUR`, // EUR is our default fiat currency,
```
jbibla: is it necessary to have a default fiat? i imagined we would default to `crypto` and the user could select a fiat they like which would persist for them.

src/components/common/TmBalance.vue
Bitcoinera: that I am sending `fiatValues` from the API with the currency symbol (default `€`) at the very end

src/components/common/TmBalance.vue
Bitcoinera: I think we can leave it hardcoded for now, since we are sending this fixed list of fiatValues from the API

src/components/common/TmBalance.vue
Bitcoinera: oh true, I forgot!

src/components/common/TmBalance.vue
Bitcoinera: it is where I copied most of the CSS from xD

OK, I change it 😅

src/components/common/TmBalance.vue
Bitcoinera: this is each currency  block (staking denom on the left and the alt-tokens ones on the right).

Should be called `currency-div`, now that I think about it 🤔

src/components/common/TmBalance.vue
Bitcoinera: Hm, interesting. Maybe a default fiat currency works for the API but not so for the FE.

So what should we show then in the fiat value bubbles by default? If we show nothing by default, how can the users know by a simple glimpse they can activate that display by selecting a currency in the selector?

Maybe add a `span` next to the currency selector when it is the first time and the user hasn't selected yet any?

Really good idea storing the preferred fiat currency! 👏

src/components/common/TmBalance.vue
jbibla: seems like this could be improved in the API. maybe something like this would work:
```
denom: 'USD'
denomSymbol: '$'
```
...so that the frontend doesn't have to re-format things that we have control of in the API.

Bitcoinera: 🤔 so fiatValue would be an Object then. Yes, I think it is a pretty good idea actually

Bitcoinera:
This one is ready! Let's get it merged together with luniehq/lunie-api#360 ⭐️

faboweb:
The required API changes are on production now

src/components/common/TmBalance.vue
```diff
@@ -298,6 +344,12 @@ export default {
           })
         return rewardsAccumulator
       }
+    },
+    preferredCurrency() {
+      return localStorage.getItem(`preferredCurrency`)
```
faboweb: this is badly implemented. it will always ask the localStorage. it should instead be a value on data which is set only when the currency is selected and on mounted

faboweb: after this anyone is free to merge

Bitcoinera: Done!

And thank you for showing the correct implementation. I wasn't too convinced with my method either

src/components/common/TmBalance.vue
```diff
@@ -345,11 +349,9 @@ export default {
         return rewardsAccumulator
       }
     },
-    preferredCurrency() {
-      return localStorage.getItem(`preferredCurrency`)
-    },
     setPreferredCurrency() {
       localStorage.setItem(`preferredCurrency`, this.selectedFiatCurrency)
+      this.preferredCurrency = localStorage.getItem(`preferredCurrency`)
```
faboweb:
```suggestion
      this.preferredCurrency = this.selectedFiatCurrency
```

src/components/common/TmBalance.vue
```diff
@@ -320,6 +321,9 @@ export default {
       }
     }
   },
+  mounted() {
+    this.preferredCurrency = localStorage.getItem(`preferredCurrency`)
```
faboweb:
```suggestion
    this.setPreferredCurrency()
```

src/components/common/TmBalance.vue
```diff
@@ -351,7 +351,7 @@ export default {
     },
     setPreferredCurrency() {
       localStorage.setItem(`preferredCurrency`, this.selectedFiatCurrency)
-      this.preferredCurrency = localStorage.getItem(`preferredCurrency`)
+      this.preferredCurrency = this.selectedFiatCurrency
```
Bitcoinera: OK, I get it. Yes. Like this we avoid retrieving the info from the localStorage at all. Makes so much sense

PR #3585, opened 2020-02-23
Aleksei/send statistics to api

iambeone:
Closes #ISSUE

**Description:**

Sending stat data to the API

Thank you! 🚀

---

For contributor:

For reviewer:


src/components/common/TmBalance.vue
```diff
@@ -358,6 +358,7 @@ export default {
           )
           this.sentToGA = true
         }
+        // sending to ga only once
```
jbibla: 1. are we still sending to GA?
2. is this comment in the right place? shouldn't it be on line 359?

tests/unit/specs/scripts/db_store.spec.js
```diff
@@ -0,0 +1,47 @@
...
+              action: "Action",
+              label: "Label",
+              value: "Value",
+              browserid: "browserid-uuid"
+            }
+          }
+        })
+      }
+      store_in_db(customObject, `Category`, `Action`, `Label`, `Value`)
+      expect(fetch).toHaveBeenCalledWith(`api/stat/store`, options)
```
jbibla: based on proposed changes in the API: https://github.com/luniehq/lunie-api/pull/359
```suggestion
      expect(fetch).toHaveBeenCalledWith(`api/stats/store`, options)
```

src/components/common/TmBalance.vue
iambeone: 1. Why not? )
2. Ooops

We can do it in API only, but without browser-id

src/scripts/db_store.js
```diff
@@ -0,0 +1,27 @@
+import browserID from "browser-id"
```
faboweb: move this logic to API to reduce surface?

src/scripts/google-analytics.js
```diff
@@ -46,9 +47,11 @@ function customToNum(custom) {
   return false
 }
 
-module.exports.sendEvent = function event(customObject, ...args) {
+export const sendEvent = function event(customObject, ...args) {
   if (window.ga) {
     let newKey
+    // sending data before converting
+    store_in_db(customObject, ...args)
```
faboweb: nice way to reuse the existing logic 👍

src/scripts/db_store.js
faboweb: clarification: proxy this request through the API as the API already is connected to the DB. we don't need to expose the DB to the FE.

src/scripts/db_store.js
```diff
@@ -0,0 +1,27 @@
...
+import { getGraphqlHost } from "scripts/url"
+
+export const store_in_db = (customObject, ...args) => {
```
faboweb:
```suggestion
export const uploadStatistics = (customObject, ...args) => {
```

Also: let's rename the file to something similar like `upload_statics.js`

package.json
```diff
@@ -45,6 +45,7 @@
     "axios": "^0.19.0",
     "bech32": "^1.1.3",
     "bignumber.js": "^8.1.1",
+    "browser-id": "1.1.0",
```
faboweb: this repo has 0 stars

faboweb: let's find an alternative or implement the logic ourselves

faboweb:
How about:
Send a browser ID (maybe GA ID) as a header in GQL requests.
API requests trigger analytics writing.
Problem is only with tx volume? We could handle that case by adding meta data to our broadcasting requests.

faboweb:
How about:
Send a browser ID (maybe GA ID) as a header in GQL requests.
API requests trigger analytics writing.
Problem is only with tx volume? We could handle that case by adding meta data to our broadcasting requests.

PR #3586, opened 2020-02-24
Ana/add exchange memo warning

Bitcoinera:
Closes #ISSUE

**Description:**

Customers requested this and I think it is super necessary. 

Currently the edit memo field can be easily overseen and the use case of sending to an exchange is critical and a warning has to be on place.

This is the one I made:

 

Thank you! 🚀

---

For contributor:

For reviewer:


Bitcoinera:
> 1. big red warnings are scary and negative - should be a last case solution

Yes, I completely get you. Maybe I went a bit too far with all those red signs 😆 

I was just having in mind a poor user sending their funds to Binance and then going crazy because they never received them... But yeah, this is a bit exaggerated now xD

I think the problem is that they didn't know they had to send the funds with a memo. I can relate because this has also happened to me. 

Maybe we could offer a small tutorial when the user goes through the Send flow for the first time instead? Or maybe add a small hint at the end of the flow before sending the transaction to the Ledger/Extension if they haven't modified the memo?

Bitcoinera:
What if we add a checkbox on SendModal? "Are you sending to an exchange?"

If they are not they won't check it. If they are, they will check and see an explanatory message (not red) so they know perhaps they need to edit their memo field.

I think checkboxes are in general quite noticeable.

jbibla:
i think a checkbox would be better. do we leave the memo question too?

Bitcoinera:
> do we leave the memo question too?

Yes, for sure. People might still want to edit memo for whatever other reason. Not only exchanges

Bitcoinera:
This is how it is looking now:



Am I moving in the good direction? I still think it is important to add some kind of warning & clarification here.

faboweb:
Did you have a call with Charlie about this?

Bitcoinera:
> Did you have a call with Charlie about this?

just on the chat. I asked him for advice

PR #3592, opened 2020-02-25
Ana/delete gas prices dictionary

Bitcoinera:
Closes #ISSUE

**Description:**

Following @faboweb's wise advice, I saw that actually the `gasPricesDictionary` is not necessary.

Still, I don't know exactly why I had to slightly change a test, but I think it is fine allowing user have 0 fee transactions (when they transact with their whole balance), since they already get a warning.

I don't know why it changed now though 😅 

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/src/ActionModal/components/ActionModal.vue
+++ b/src/ActionModal/components/ActionModal.vue
@@ -288,7 +288,6 @@ import TmDataMsg from "common/TmDataMsg"
 import TableInvoice from "./TableInvoice"
 import Steps from "./Steps"
 import { mapState, mapGetters } from "vuex"
-import { gasPricesDictionary } from "src/scripts/common"
 import { viewDenom, prettyInt } from "src/scripts/num"
 import { between, requiredIf } from "vuelidate/lib/validators"
 import { track, sendEvent } from "scripts/google-analytics"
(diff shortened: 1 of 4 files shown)


src/ActionModal/components/ActionModal.vue
```diff
@@ -499,8 +498,11 @@ export default {
     selectedBalance() {
       const defaultBalance = {
         amount: 0,
-        // awful network-specific logic. But how to do it otherwise?
-        gasPrice: gasPricesDictionary(this.getDenom)
+        gasPrice:
```
faboweb: ur replicating the same logic as lines 501+
why is line 511 not delivering a balance is the question here

Bitcoinera: Yes, you were right. The problem was exactly there, why those lines were not working.

And I found out. It was because this strange thing doing `({ denom: balanceDenom }) => balanceDenom === denom`.

Much more simple to rename the const `denom` and have it more readable (and working). Plus `balanceDenom` wasn't being used anywhere else.

src/ActionModal/components/ActionModal.vue
```diff
@@ -498,20 +498,14 @@ export default {
...
       }
       if (this.balances.length === 0 || !this.network) {
         return defaultBalance
       }
       // default to the staking denom for fees
-      const denom = this.selectedDenom || this.network.stakingDenom
-      let balance = this.balances.find(
-        ({ denom: balanceDenom }) => balanceDenom === denom
-      )
+      const stakingDenom = this.selectedDenom || this.network.stakingDenom
```
faboweb: why the renaming? it's even wrong. this isn't always the stakingDenom

Bitcoinera: oops, true...

But it needs to be renamed because of what I mentioned in my commend above. To do the `denom ===` comparison from the balances array.

Let's call it `preferredDenom` then? Or do you have any suggestion?

faboweb: feeDenom?

Bitcoinera: Done

PR #3622, opened 2020-02-28
Fabo/fix extension deep link

faboweb:
Fixes issue with sometimes the extension deep link not signing in the user

### DIFF
--- a/src/routes.js
+++ b/src/routes.js
@@ -174,9 +174,10 @@ export default (apollo, store) => {
     {
       path: `/extension/:address/:network`,
       name: `extension-signin`,
-      /* istanbul ignore next */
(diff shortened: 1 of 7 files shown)


src/routes.js
```diff
@@ -171,18 +172,16 @@ export default (apollo, store) => {
...
     },
     {
-      path: `/extension`,
-      name: `extension`,
-      components: {
-        session: () => import(`./components/common/TmSessionExtension`)
-      },
-      meta: {
-        feature: "Session"
+      path: `/extension/:address/:network`,
```
jbibla: it's odd to me that address comes before network

faboweb: hard to change now. but we can do it

PR #3625, opened 2020-03-01
Fabo/address component

faboweb:
To allow Polkadot addresses


Bitcoinera:
I think this is deprecated right? `longForm` is not being used anymore

```
    longForm: {
      type: Boolean,
      required: false,
      default: false
    }
```
lines 30-34 in Address.vue (GitHub doesn't allow me to suggest any change there)

and lines 28-31 in Address.spec.js (the test for `longForm`)

faboweb:
Good catch. Fixed.

tests/unit/specs/components/ActionModal/components/__snapshots__/DelegationModal.spec.js.snap
```diff
@@ -10,7 +10,7 @@ exports[`DelegationModal Submission Data for Redelegating should set the subtota
       </span></div>
   </tmformgroup-stub>
   <tmformgroup-stub fieldid="to" fieldlabel="To" class="action-modal-form-group">
-    <tmfield-stub type="text" value="undefined - cosmosvaladdr…aladdr1kjisjsd862323" id="to" readonly=""></tmfield-stub>
+    <tmfield-stub type="text" value="undefined - cosmosvaladdr1kjisjs…aladdr1kjisjsd862323" id="to" readonly=""></tmfield-stub>
```
jbibla: this doesn't look right

tests/unit/specs/components/common/__snapshots__/Address.spec.js.snap
```diff
@@ -0,0 +1,25 @@
...
+
+exports[`Address Component should show a short address 1`] = `
+<div
+  class="bech32-address"
+>
+  <div
+    class="address"
+  >
+    
+    cosm…6789
```
jbibla: this doesn't look right

tests/unit/specs/components/common/__snapshots__/TmSessionExplore.spec.js.snap
```diff
@@ -140,7 +132,7 @@ exports[`TmSessionExplore shows a form to sign in with an address 1`] = `
               class="tm-li-session-title"
             >
               <span>
-                cosmos…a7m64khj8xyz
+                cosmos1vxkye…a7m64khj8xyz
```
jbibla: doesn't look right

jbibla:
bunch of snapshots don't look correct in this PR

Bitcoinera:
> bunch of snapshots don't look correct in this PR

@faboweb what do you think?

faboweb:
looks like it. will fix it. this looks like it has to do with the length parameter

faboweb:
the addresses are not bech32 this is why they get formatted in a standard way.

PR #3628, opened 2020-03-01
Fabo/select signer per network

faboweb:
Select signing method based on network type

Merge https://github.com/luniehq/lunie-api/pull/378 first


src/ActionModal/components/ActionModal.vue
```diff
@@ -486,6 +486,7 @@ export default {
         url: this.network.api_url,
         networkId: this.network.id,
         chainId: this.network.chain_id,
+        networkType: this.network.network_type,
```
Bitcoinera: How did you get this to work without having `network_type` in the Network schema? (I just added it to your PR)
O.o

src/ActionModal/utils/ActionManager.js
```diff
@@ -192,7 +193,7 @@ export default class ActionManager {
       txMessages.push(txMessage)
     }
 
-    const createSignedTransaction = await getTransactionSigner(context)
+    const createSignedTransaction = await signedTransactionCreator(context)
```
Bitcoinera: I think the naming of this function can be improved. Why not `getSignedByTransactionCreator` or `signByTransactionCreator`? I think both are more descriptive names

src/ActionModal/utils/signer.js
```diff
@@ -11,57 +11,75 @@ export async function cancelSign(submitType = "", { address, network }) {
...
+    )
+
+    return {
+      signature,
+      publicKey: Buffer.from(wallet.publicKey, "hex")
+    }
+  }
+}
+async function getCosmosLedgerSigner(config) {
+  // TODO show which properties of config are actually needed
```
Bitcoinera: What properties of config? Why is this now needed?

src/ActionModal/utils/signer.js
```diff
@@ -11,57 +11,75 @@ export async function cancelSign(submitType = "", { address, network }) {
...
+          "Something went wrong connecting to your Ledger. Please refresh your page and try again."
+        )
       }
+      throw err
     }
-  } else if (submitType === `extension`) {
-    return signMessage => {
-      return signWithExtension(signMessage, address, network)
+
+    // cleanup. if we leave this open, the next connection will brake for HID
```
Bitcoinera:
```suggestion
    // cleanup. if we leave this open, the next connection will break for HID
```

src/components/common/TmSessionExtension.vue
```diff
@@ -62,15 +62,6 @@ export default {
...
   },
-  async created() {
-    if (this.$route.params.address && this.$route.params.network) {
-      await this.signIn({
-        address: this.$route.params.address,
-        network: this.$route.params.network
-      })
-      this.$router.push("/portfolio")
-    }
-  },
```
Bitcoinera: Deleting this is not related to these PR changes no? Or it is?

src/routes.js
```diff
@@ -316,3 +315,27 @@ export default (apollo, store) => {
...
+// handle direct sign in from the extension via deeplink
+export async function extensionSignIn({ to, next }, apollo, store) {
+  const {
+    data: { network }
+  } = await apollo.query({
+    query: gql`
+      query Network {
+        network(id: "${to.params.network}") {
+          id
+          slug
```
Bitcoinera: Why are we not sending here `network_type` as well?

tests/unit/specs/components/ActionModal/utils/ActionManager.spec.js
```diff
@@ -29,8 +29,8 @@ const mockMsgWithdraw = jest.fn(() => ({
   send: () => ({ included: () => async () => true })
 }))
 
-const mockGetTransactionSigner = jest.fn(() => {
-  console.log("mockGetTransactionSigner executed")
+const mocksignedTransactionCreator = jest.fn(() => {
```
Bitcoinera: I advice changing the function name

Bitcoinera:
It looks quite good though. Going through with the debugger and everything seems to be working as expected 💃

Bitcoinera:
e2e tests keep failing, but if we are going for #3629 it is not worth it to fix them for this one

src/ActionModal/utils/ActionManager.js
faboweb: Why are those `signedBy`?

src/ActionModal/utils/signer.js
faboweb: This was in the code before. This is just a refactor. `config` is passed through to the library. It would be nice to show here which properties of `config.js` are actually needed. This would help with search and replace or when we want to delete things from config.js.

src/components/common/TmSessionExtension.vue
faboweb: Nop, sorry might have slipped in.

src/routes.js
faboweb: not needed here, is it?

tests/unit/specs/components/ActionModal/utils/ActionManager.spec.js
faboweb: why?

src/routes.js
Bitcoinera: to tell the extension the network type that need to sign on it, no? I guess in the extension we will also introduce this `getCosmosSigner` & `getWhateverNetworkSigner` logic? or I got that wrong?

src/ActionModal/utils/signer.js
Bitcoinera: Oh right. Thank you so much for the clarification

src/ActionModal/components/ActionModal.vue
Bitcoinera: Magic xD

Bitcoinera:
The e2e tests seem to be failing in exactly the same thing that @iambeone  had problems with. The `:nth-of-type(2)` element being not visible.

src/components/common/TmBalance.vue
```diff
@@ -109,6 +109,7 @@
                     <span
                       v-if="
                         isMultiDenomNetwork &&
+                          stakingBalance.fiatValue &&
```
Bitcoinera: hey, this looks familiar to me! hehe

src/components/common/TmBalance.vue
```diff
@@ -166,7 +167,11 @@
                     </h2>
                   </div>
                   <div
-                    v-if="balance.fiatValue.amount > 0 && preferredCurrency"
+                    v-if="
+                      balance.fiatValue &&
```
Bitcoinera: same 😛

PR #3630, opened 2020-03-02
Bring back the fingerprint header

faboweb:
Reverts luniehq/lunie#3623


Bitcoinera:
Please, revert until we don't have Aleksei's fix (luniehq/lunie-browser-extension#160) in the Chrome Web Store. Otherwise extension will turn non-functional again.

PR #3634, opened 2020-03-02
Fabo/display subtotal correctly

faboweb:
It still showed the whole amount


faboweb:
breaks undelegation e2e tests

faboweb:
sign in for unclear reasons breaks now

PR #3637, opened 2020-03-02
Aleksei/subscribe to more fixed

iambeone:
Closes #ISSUE

**Description:**

Added subscribeToMore to PageTransactions

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/src/components/wallet/PageTransactions.vue
+++ b/src/components/wallet/PageTransactions.vue
@@ -197,6 +197,31 @@ export default {
           ]
         }
         return this.loadedTransactions
+      },
+      subscribeToMore: {
+        document: gql`
+          subscription($networkId: String!, $address: String!) {
+            userTransactionAddedV2(networkId: $networkId, address: $address) {
+              ${txFields}
+            }
+          }
+        `,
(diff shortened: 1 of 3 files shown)


src/components/wallet/PageTransactions.vue
```diff
@@ -197,6 +197,31 @@ export default {
           ]
         }
         return this.loadedTransactions
+      },
+      subscribeToMore: {
+        document: gql`
+          subscription($networkId: String!, $address: String!) {
+            userTransactionAddedV2(networkId: $networkId, address: $address) {
```
Bitcoinera: Looking at `resolvers.js`, the code for both `userTransactionAdded` and `userTransactionAddedV2` are exactly the same.

What I would like to ask here is: we don't really need to use V2 to add this multiple subscription or we do? Please, I'd love to know

iambeone: I think it's a temporary solution, after switching to transaction abstraction we will replace every v2 in function names

faboweb: the returned schema is different. check schema.js

PR #3643, opened 2020-03-03
Fabo/fix addres

faboweb:
Mostly for @jbibla showing correct addresses in tests


tests/unit/specs/components/common/__snapshots__/TmSessionExplore.spec.js.snap
```diff
@@ -132,7 +132,7 @@ exports[`TmSessionExplore shows a form to sign in with an address 1`] = `
               class="tm-li-session-title"
             >
               <span>
-                cosmos1vxkye…a7m64khj8xyz
+                cosmos15ky9d…rm5ctpesxxn9
```
jbibla: ?

Bitcoinera: why did we change the format to a much longer one?

faboweb: ups, this is still not a valid bech32 address

tests/unit/specs/components/common/__snapshots__/TmSessionExplore.spec.js.snap
```diff
@@ -132,7 +132,7 @@ exports[`TmSessionExplore shows a form to sign in with an address 1`] = `
               class="tm-li-session-title"
             >
               <span>
-                cosmos15ky9d…rm5ctpesxxn9
+                cosmos…gc05sslwesuy
```
Bitcoinera: I think it is better like it was before. Showing four at the beginning, after the `cosmos` prefix, and four at the end.

Like this is much easier to identify imo

faboweb: would be a change though. cosmos…gc05sslwesuy is how we displayed it all the time

PR #3657, opened 2020-03-07
HACK Ana/fix terra fee amount the hacky way

Bitcoinera:
Closes #ISSUE

**Description:**

Beware ⚠️ !

Your eyeballs will probably hurt and you will hate so much my hacky solution.

It contains **awful network-specific logic**.

That said, I don't know how to fix this otherwise. Being able to account for terra fee behavior.

PS: I also will retest e-Money once more in case this is also applying there.

Please don't use too many dislike emojis 🙈 

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/src/ActionModal/components/ActionModal.vue
+++ b/src/ActionModal/components/ActionModal.vue
@@ -424,7 +424,12 @@ export default {
   computed: {
     ...mapState([`extension`, `session`]),
     ...mapGetters([`connected`, `isExtensionAccount`]),
-    ...mapGetters({ networkId: `network` }),
+    // hack to avoid computed property in data error
+    ...mapGetters({ networkID: `network` }),
+    // hack for tests
+    networkId() {
+      return this.network.id
+    },
     checkFeatureAvailable() {
       const action = `action_` + this.featureFlag
       return this.network[action] === true
@@ -437,7 +442,7 @@ export default {
     },
     estimatedFee() {
       // hack
-      return this.networkId.includes(`terra`)
+      return this.networkId.startsWith(`terra`)
         ? Number(this.gasPrice) * Number(this.gasEstimate) * Number(this.amount) // already in atoms
         : Number(this.gasPrice) * Number(this.gasEstimate)
     },
@@ -446,7 +451,7 @@ export default {
     },
     invoiceTotal() {
       // hack
-      return this.networkId.includes(`terra`)
+      return this.networkId.startsWith(`terra`)
         ? Number(this.subTotal) +
             Number(this.gasPrice) *
               Number(this.gasEstimate) *
@@ -696,7 +701,7 @@ export default {
         gasEstimate: this.gasEstimate,
         gasPrice: {
           // hack
-          amount: this.networkId.includes(`terra`)
+          amount: this.networkId.startsWith(`terra`)
             ? this.gasPrice * this.amount
             : this.gasPrice,
           denom: this.getDenom
--- a/src/ActionModal/components/ActionModal.vue
+++ b/src/ActionModal/components/ActionModal.vue
@@ -436,15 +436,23 @@ export default {
       )
     },
     estimatedFee() {
-      return Number(this.gasPrice) * Number(this.gasEstimate) // already in atoms
+      // hack
+      return this.networkId.includes(`terra`)
+        ? Number(this.gasPrice) * Number(this.gasEstimate) * Number(this.amount) // already in atoms
+        : Number(this.gasPrice) * Number(this.gasEstimate)
     },
     subTotal() {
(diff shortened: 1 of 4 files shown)


faboweb:
Aren't the fees actually:
Gas estimate * gas price + amount * tax
Where amount * tax is only applied for sending transactions?

Bitcoinera:
> Aren't the fees actually:
> Gas estimate * gas price + amount * tax
> Where amount * tax is only applied for sending transactions?

Apparently it was always gas estimate * gas price. I cannot find anywhere the logic you are describing.

Only [here ](https://github.com/luniehq/lunie/blob/cf7d822b44830908fa66508c371068dd7588cf75/src/ActionModal/utils/ActionManager.js#L92)in ActionManager there is separate logic for "WITHDRAW" transactions. But all the rest follow the same logic.

Maybe I should change this hack and put in ActionManager now that I think about it... Oops, if this is something that all networks need, I mean. Then it would be much much cleaner.

Will also test with e-Money to see if the same is also happening there.

But then again if this was an issue in Cosmos we'd already know by now 🤔  

Maybe something only affecting alt-tokens? It could still go in ActionManager. Really worth investigating.

faboweb:
No, this is only Terra and might effect EMoney. As the stable coins tax there tokens.

faboweb:
Read: https://docs.terra.money/docs/node-terracli#query-tax-rate

The amount needs to be multiplied with the tax-rate: https://fcd.terra.dev/treasury/tax_rate we can hardcode to `0.008` for now

Bitcoinera:
>  we can hardcode to `0.008` for now

I just tried it and it is too low 🤷‍♀ I don't really know how to apply that number in the tax_rate call

Bitcoinera:
> No, this is only Terra as far as I know

Yeah, I just checked all e-Money tokens and it is not happening. So I guess in this case there is no workaround other than network-specific logic

faboweb:
I can send KRT to myself and others on soju. How can I reproduce this error in the first place?

Bitcoinera:
> I can send KRT to myself and others on soju. How can I reproduce this error in the first place?

First send a small quantity, like 10. Then try sending 100. With 100 it should fail already. You will see the required gas in the error increasing as you try to send more.

faboweb:
could replicate. refactored

src/ActionModal/components/ActionModal.vue
```diff
@@ -442,10 +444,19 @@ export default {
...
+      // terra uses a tax on all send txs
+      if (
+        this.networkId.startsWith(`terra`) &&
+        this.transactionData.type === transactionTypes.SEND
+      ) {
+        // hardcoding terra tax here until we have it in the API
+        const terraTax = 0.008
+        return (
+          Number(this.gasEstimate) * Number(this.gasPrice) +
+          Number(this.amount) * terraTax
```
Bitcoinera: Oh, now I see what you meant! 

Still, for me it was working without tax... Maybe something only present in mainnet then? I am really curious

src/ActionModal/components/ActionModal.vue
```diff
@@ -694,12 +705,7 @@ export default {
       const feeProperties = {
         gasEstimate: this.gasEstimate,
         gasPrice: {
-          // hack
-          amount:
-            this.networkId.startsWith(`terra`) &&
-            this.selectedBalance.denom !== this.network.stakingDenom
-              ? this.gasPrice * this.amount
-              : this.gasPrice,
+          amount: this.estimatedFee / this.gasEstimate,
```
Bitcoinera: 👀

src/ActionModal/components/ActionModal.vue
faboweb: nope, could reproduce in soju. you set the gas price in two places in the code. that was confusing to me, maybe happened to you as well. reduced to just setting the estimate in one place.

Bitcoinera:
> could replicate. refactored

Cool. Will test it. What do you think about it working without tax at least in soju? Is it only applicable in mainnet 🤷‍♀  ?

src/ActionModal/components/ActionModal.vue
faboweb: fees = gasPrice * gasEstimate (in comos-api)
we want: fees = gasPrice * gasEstimate + amount * tax
ergo: gasPriceDesired * gasEstimate = gasPrice * gasEstimate + amount * tax
ergo: gasPriceDesired = (gasPrice * gasEstimate + amount * tax) / gasEstimate

faboweb: added a comment

Bitcoinera: yes, now I saw it. Thank you for this amazing Math clarification 🙏

PR #3662, opened 2020-03-09
Ana/claim any token rewards

Bitcoinera:
Closes #ISSUE

**Description:**

Enable claiming rewards from any token, even when there is not staking denom reward.

Also adds a logic where by default we pay fees with the staking denom for claim rewards (like it was before) but, in case there is no available balance for the staking denom, then we will pay fees with the token we are claiming rewards from.

Of course, changes readyToWithdraw too in TmBalance.

I can split this PR if it is more convenient.

Thank you! 🚀

---

For contributor:

- [ ] Added changes entries. Run `yarn changelog` for a guided process.
- [ ] Reviewed `Files changed` in the github PR explorer
- [ ] Attach screenshots of the UI components on the PR description (if applicable)
- [ ] Scope of work approved for big PRs

For reviewer:

- [ ] Manually tested the changes on the UI

PR #3663, opened 2020-03-09
Ana/claim rewards from any token, also when staking denom rewards are 0

Bitcoinera:
Closes #ISSUE

**Description:**

First part of the split PR #3662 . 

This one only contains the logic to allow claiming rewards from any token, also when there are no staking denom rewards.

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/src/ActionModal/components/ModalWithdrawRewards.vue
+++ b/src/ActionModal/components/ModalWithdrawRewards.vue
@@ -20,7 +20,9 @@
       field-id="amount"
       field-label="Amount"
     >
-      <span class="input-suffix">{{ denom }}</span>
+      <span v-if="claimedReward" class="input-suffix">{{
+        claimedReward.denom
+      }}</span>      
       <TmField
         id="amount"
         v-model="totalRewards"
@@ -63,10 +65,11 @@ export default {
       }
     },
     totalRewards() {
-      return this.rewards
+      const stakingRewards = this.rewards
         .filter(({ denom }) => denom === this.denom)
         .reduce((sum, { amount }) => sum + Number(amount), 0)
         .toFixed(6)
+      return stakingRewards > 0 ? stakingRewards : this.claimedReward.amount
     },
     notifyMessage() {
       return {
@@ -76,6 +79,18 @@ export default {
     },
     validatorsWithRewards() {
       return this.rewards.length > 0
+    },
+    claimedReward() {
+      if (this.denom && this.rewards && this.rewards.length > 0) {
+        // we return the staking denom reward if it has any. Otherwise, we return the first reward from the other tokens
+        return (
+          this.rewards.filter(
+            reward => reward.denom === this.denom && reward.amount > 0
+          )[0] || this.rewards.filter(reward => reward.amount > 0)[0]
+        )
+      } else {
+        return ""
+      }
     }
   },
   methods: {
--- a/src/ActionModal/components/ModalWithdrawRewards.vue
+++ b/src/ActionModal/components/ModalWithdrawRewards.vue
@@ -83,12 +83,12 @@ export default {
     claimedReward() {
       if (this.denom && this.rewards && this.rewards.length > 0) {
         // we return the staking denom reward if it has any. Otherwise, we return the first reward from the other tokens
-        const rewardGreaterThanZero = array =>
-          array.filter(reward => reward.amount > 0)[0]
+        const rewardsGreaterThanZero = this.rewards.filter(
+          reward => reward.amount > 0
+        )
         return (
-          rewardGreaterThanZero(
-            this.rewards.filter(reward => reward.denom === this.denom)
-          ) || rewardGreaterThanZero(this.rewards)
+          rewardsGreaterThanZero.find(reward => reward.denom === this.denom) ||
+          rewardsGreaterThanZero[0]
         )
       } else {
         return ""
--- a/src/ActionModal/components/ModalWithdrawRewards.vue
+++ b/src/ActionModal/components/ModalWithdrawRewards.vue
@@ -22,7 +22,7 @@
     >
       <span v-if="claimedReward" class="input-suffix">{{
         claimedReward.denom
-      }}</span>      
+      }}</span>
       <TmField
         id="amount"
         v-model="totalRewards"
--- a/src/ActionModal/components/ModalWithdrawRewards.vue
+++ b/src/ActionModal/components/ModalWithdrawRewards.vue
@@ -83,10 +83,12 @@ export default {
     claimedReward() {
       if (this.denom && this.rewards && this.rewards.length > 0) {
         // we return the staking denom reward if it has any. Otherwise, we return the first reward from the other tokens
+        const rewardGreaterThanZero = array =>
+          array.filter(reward => reward.amount > 0)[0]
         return (
-          this.rewards.filter(
-            reward => reward.denom === this.denom && reward.amount > 0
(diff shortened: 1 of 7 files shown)


src/ActionModal/components/ModalWithdrawRewards.vue
```diff
@@ -76,6 +79,18 @@ export default {
     },
     validatorsWithRewards() {
       return this.rewards.length > 0
+    },
+    claimedReward() {
```
faboweb: So much effort for a temporary solution...

src/components/common/TmBalance.vue
```diff
@@ -299,7 +305,16 @@ export default {
     // only be ready to withdraw of the validator rewards are loaded and the user has rewards to withdraw
     // the validator rewards are needed to filter the top 5 validators to withdraw from
     readyToWithdraw() {
-      return this.overview.totalRewards > 0
+      if (this.overview.rewards && this.overview.rewards.length > 0) {
```
faboweb: Why so complicated? Can't u just check if there is any record in rewards?

Bitcoinera: ? This is to prevent errors from overview & rewards not being loaded yet.

I am not sure what you mean

faboweb: I mean the other lines in this function. Why do you need to calculate "allTotalRewards"? Any reward is enough to allow withdraw right?

Bitcoinera: Because rewards are coming from the API divided by validator at the moment

faboweb: but you can withdraw if you have any validator with any amount of rewards, right?

Bitcoinera: nope. That is what `readyToWithdraw` is precisely doing ^^

faboweb: replace 309-314 with a check if there are any records in rewards. no need for all this logic...

Bitcoinera: Look at my commits. I started like that. But the problem is that rewards are spread across validators... Do you see the issue? You'd need to wait for rewards to be above 0.001 on a single validator.

You can also do filter rewards with positive amounts, I agree. But users could withdraw then "invisible" rewards they cannot see in the UI, and it doesn't make sense since you are paying more fees than rewards received.

Bitcoinera:
Still, maybe we should close this PR since, as we clarified in the meetings, this idea doesn't make too much sense.

Only in rare cases like an account at the beginning of staking in Terra and with KRT tokens.

image

What do you think @faboweb ?

src/components/common/TmBalance.vue
faboweb: I see. Good thinking. Can you comment those thoughts?

src/ActionModal/components/ModalWithdrawRewards.vue
```diff
@@ -76,6 +79,18 @@ export default {
...
+      if (this.denom && this.rewards && this.rewards.length > 0) {
+        // we return the staking denom reward if it has any. Otherwise, we return the first reward from the other tokens
+        return (
+          this.rewards.filter(
```
faboweb: refactor the filter for `reward.amount >0`. then use that in the next to searches

src/components/common/TmBalance.vue
```diff
@@ -299,7 +305,16 @@ export default {
...
+        const allTotalRewards = this.overview.rewards.map(reward =>
+          this.calculateTotalRewardsDenom(reward.denom)
```
faboweb: are rewards by validator? then `allTotalRewards` would hole multiple records per denom right?

Bitcoinera: yes, rewards are by validators. And yes, allTotalRewards returns an array like this, just with the total amounts per token: `[0.0623, 0.5566, 3.2456 ... ]`

src/ActionModal/components/ModalWithdrawRewards.vue
Bitcoinera: Done

src/ActionModal/components/ModalWithdrawRewards.vue
```diff
@@ -76,6 +79,20 @@ export default {
...
+    claimedReward() {
+      if (this.denom && this.rewards && this.rewards.length > 0) {
+        // we return the staking denom reward if it has any. Otherwise, we return the first reward from the other tokens
+        const rewardGreaterThanZero = array =>
+          array.filter(reward => reward.amount > 0)[0]
+        return (
+          rewardGreaterThanZero(
+            this.rewards.filter(reward => reward.denom === this.denom)
+          ) || rewardGreaterThanZero(this.rewards)
+        )
```
faboweb:
```suggestion
        const rewardsGreaterThanZero = array.filter(reward => reward.amount > 0)
        return (
          rewardsGreaterThanZero.find(reward => reward.denom === this.denom)
          ) || rewardsGreaterThanZero[0]
        )
```

src/components/common/TmBalance.vue
```diff
@@ -299,7 +305,16 @@ export default {
...
+        )
+        return allTotalRewards.length > 0
+          ? allTotalRewards.find(reward => parseFloat(reward) > 0.001)
+          : null
```
faboweb:
```suggestion
        return allTotalRewards.find(reward => parseFloat(reward) > 0.001)
```

src/components/common/TmBalance.vue
faboweb: it will return multiple records per denom. doesn't really matter but it is badly coded...
you also call `calculateTotalRewardsDenom` which is costly
instead do a computed, that will create one map `aggregatedRewards` with the structure `{ denom: amount }`

Bitcoinera: Yes, I noticed what you said. It returns the same amount per repeated denom in rewards. I guess better to change it for good engineering

faboweb: not suuuuper important as your solution works. but maybe you can take the extra half an hour to practice proper code.

Bitcoinera: xDDDDD

Bitcoinera: OK, I was a long while trying to practice "proper" code, but at the end and after unsuccessfully trying to do it with a `reduce`, I came to the conclusion the best is to just filter the repeated denoms before using `calculateTotalRewardsDenom`. 

I don't know, having this function already in place, I think it is overcomplicating things adding a new computed to do almost the same thing 🤷‍♀

src/components/common/TmBalance.vue
```diff
@@ -306,8 +307,12 @@ export default {
     // the validator rewards are needed to filter the top 5 validators to withdraw from
     readyToWithdraw() {
       if (this.overview.rewards && this.overview.rewards.length > 0) {
-        const allTotalRewards = this.overview.rewards.map(reward =>
-          this.calculateTotalRewardsDenom(reward.denom)
+        const uniqRewardsDenoms = uniqBy(
+          this.overview.rewards,
+          reward => reward.denom
+        ).map(reward => reward.denom)
+        const allTotalRewards = uniqRewardsDenoms.map(denom =>
```
faboweb: please check out `.reduce` for the future

src/components/common/TmBalance.vue
faboweb: please remember me to do this with you in the 1:1

Bitcoinera: OK👌

PR #3669, opened 2020-03-10
Ana/fix concat of undefined in ClaimRewardsTxDetails

Bitcoinera:
Closes #ISSUE

**Description:**

Another easy check fix. Don't know how I didn't notice while coding this. Probably it was all cached already.

I think it is a good practice to turn the vue server down and then restart to catch this kind of silly bugs.

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/src/components/transactions/message-view/ClaimRewardsTxDetails.vue
+++ b/src/components/transactions/message-view/ClaimRewardsTxDetails.vue
@@ -41,7 +41,11 @@
     >
       <div class="tx__content__left multi-claim-reward-row">
         <h3 class="multi-claim-reward-h3">{{ caption }}</h3>
-        <div class="multi-claim-reward-row" :class="{ validatorsToggle: show }">
+        <div
+          v-if="getValidators.length > 0"
+          class="multi-claim-reward-row"
+          :class="{ validatorsToggle: show }"
+        >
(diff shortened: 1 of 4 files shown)


src/components/transactions/message-view/ClaimRewardsTxDetails.vue
```diff
@@ -41,7 +41,11 @@
     >
       <div class="tx__content__left multi-claim-reward-row">
         <h3 class="multi-claim-reward-h3">{{ caption }}</h3>
-        <div class="multi-claim-reward-row" :class="{ validatorsToggle: show }">
+        <div
+          v-if="getValidators.length > 0"
```
faboweb: why is this necessary? line 50 handles an empty array gracefully

Bitcoinera: You mean line 39? true... I didn't notice. So this is then no the fix probably.

OK, but please don't close this PR as this is still an issue. Will look into it

Bitcoinera: OK, just fixed it the real way. But we really need to fix API taking too much time to load validators

PR #3671, opened 2020-03-10
Fabo/polkadot signer

faboweb:
Draft of tx generation and signing in polkadot

Missing:
- extract fees

Needs: https://github.com/luniehq/lunie-api/pull/461


src/components/common/TmSessionExtension.vue
```diff
@@ -62,15 +62,6 @@ export default {
...
   },
-  async created() {
-    if (this.$route.params.address && this.$route.params.network) {
-      await this.signIn({
-        address: this.$route.params.address,
-        network: this.$route.params.network
-      })
-      this.$router.push("/portfolio")
-    }
-  },
```
Bitcoinera: Shouldn't this be another PR 😛 ?

src/routes.js
```diff
@@ -316,3 +315,27 @@ export default (apollo, store) => {
...
+    fetchPolicy: "cache-first"
+  })
+
+  await store.dispatch(`signIn`, {
+    sessionType: `extension`,
+    address: to.params.address,
+    networkId: to.params.network
+  })
+  next(`/${network.slug}/portfolio`)
+}
```
Bitcoinera: This side change with the routing should be another PR no? ;)

Or why is it needed here?

src/components/common/TmSessionExtension.vue
faboweb: left overs, I can clean it up if necessary

Bitcoinera: Please, split PRs to make more readable & understandable💃

faboweb: sure when I clean this up

faboweb:
one strange test error

src/ActionModal/utils/networkMessages/polkadot-testnet.js
```diff
@@ -1,4 +1,19 @@
...
+
+// will only be inited once per session
+let api
+async function getAPI() {
+  if (api) {
+    return api
+  }
+  api = new ApiPromise({
+    provider: new WsProvider(
+      "wss://kusama-rpc.polkastats.io/apikey/HunRG7eUwMTjaBAkz1A6GU1MqBJapaYsYfmU5EVzpAebMr8/"
```
jbibla: APIkey should be `ENV` variable?

faboweb: whole thing should just be in the api. will need to consider rotating the apikey @mariopino

mariopino: No problem as that was only for testing and I'll remove the exposed RPC, now we are using parity RPC.

btw tested and works awesome @faboweb!!

PR #3681, opened 2020-03-11
Fabo/polkadot raw seed

faboweb:
Now we can import Marios generated accounts

### DIFF
--- a/src/components/common/TmSessionImport.vue
+++ b/src/components/common/TmSessionImport.vue
@@ -23,7 +23,7 @@
             type="required"
           />
           <TmFormMsg
-            v-else-if="$v.seed.$error && !$v.seed.words24"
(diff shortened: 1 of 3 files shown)


src/components/common/TmSessionImport.vue
```diff
@@ -81,7 +86,10 @@ export default {
     }
   },
   validations: () => ({
-    seed: { required, words24 }
+    seed: {
+      required,
+      validSeed: param => words24(param) || polkadotRawSeed(param)
```
Bitcoinera: In the future I guess this will need to be refactored to a dictionary that takes the network_type property in network and then selects the proper seed function

faboweb: we will also need to change the wording and error messages

faboweb: rn this is just for us

PR #3688, opened 2020-03-12
Ana/fix terra not enough balance to pay for fees

Bitcoinera:
Closes #ISSUE

**Description:**

Fixes this bug:

Asking for extra fees when we are sending max amount. It only affects Terra on Send because of their tax * amount system.

I solved it passing a `isMaxAmount` boolean prop from SendModal to ActionModal, but I am curious if there is anyone has a better idea. I think this fix is good though.

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/src/ActionModal/components/ActionModal.vue
+++ b/src/ActionModal/components/ActionModal.vue
@@ -340,10 +340,6 @@ const sessionType = {
 // hardcoding terra tax here until we have it in the API
 const terraTax = 0.008
 
-const maxDecimals = (value, decimals) => {
-  return Number(BigNumber(value).toFixed(decimals)) // TODO only use bignumber
-}
-
 export default {
   name: `action-modal`,
   components: {
@@ -457,7 +453,7 @@ export default {
         this.networkId.startsWith(`terra`) &&
         this.transactionData.type === transactionTypes.SEND
       ) {
-        return maxDecimals(
+        return this.maxDecimals(
           Number(this.gasEstimate) * Number(this.gasPrice) +
             Number(this.amount) * terraTax,
           6
@@ -793,6 +789,9 @@ export default {
       this.submissionError = `${this.submissionErrorPrefix}: ${error.message}.`
       this.trackEvent(`event`, `failed-submit`, this.title, error.message)
       this.$apollo.queries.overview.refetch()
+    },
+    maxDecimals(value, decimals) {
+      return Number(BigNumber(value).toFixed(decimals)) // TODO only use bignumber
     }
   },
   validations() {
--- a/src/ActionModal/components/ActionModal.vue
+++ b/src/ActionModal/components/ActionModal.vue
@@ -9,11 +9,17 @@
       >
         <i class="material-icons notranslate">arrow_back</i>
       </div>
-      <div id="closeBtn" class="action-modal-icon action-modal-close" @click="close">
+      <div
+        id="closeBtn"
+        class="action-modal-icon action-modal-close"
+        @click="close"
+      >
         <i class="material-icons notranslate">close</i>
       </div>
       <div class="action-modal-header">
-        <span class="action-modal-title">{{ requiresSignIn ? `Sign in required` : title }}</span>
+        <span class="action-modal-title">{{
+          requiresSignIn ? `Sign in required` : title
+        }}</span>
         <Steps
           v-if="
             [defaultStep, feeStep, signStep].includes(step) &&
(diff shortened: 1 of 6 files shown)


src/ActionModal/components/ActionModal.vue
```diff
@@ -451,10 +455,10 @@ export default {
...
         // hardcoding terra tax here until we have it in the API
         const terraTax = 0.008
-        return (
-          Number(this.gasEstimate) * Number(this.gasPrice) +
-          Number(this.amount) * terraTax
-        )
+        const terraExtraFees = this.isMaxAmount
+          ? 0
+          : Number(this.amount) * terraTax
+        return Number(this.gasEstimate) * Number(this.gasPrice) + terraExtraFees
```
faboweb: will this work? your just not sending the required fees
better solution to calculate the extra fees into max
or just ignoring this for now...

Bitcoinera: Yes, actually I am realizing this is a terrible solution anyway, because what happens if we are sending the max amount minus a little bit, like 0.0000001. 

Then we would still need to pay for the tax * amount and get the same error message...

Ah, this is what you are saying anyway

Bitcoinera: Yeah, I am going to change this to your solution. This is a bad one

Bitcoinera:
O.o what...?

src/ActionModal/components/SendModal.vue
```diff
@@ -256,7 +256,13 @@ export default {
       this.sending = false
     },
     setMaxAmount() {
-      this.amount = this.selectedBalance.amount
+      if (this.network.startsWith(`terra`)) {
+        const terraTax = 0.008
+        this.amount =
```
faboweb: ignores fees but we are ignoring that everywhere?

Bitcoinera: ? what do you mean?

Bitcoinera:
Strange. It works using https://staging-api.lunie.io but not localhost:4000 (local API)

src/ActionModal/components/SendModal.vue
faboweb: this sets the max to available balance - tax. but the actual max is available balance - tax - fees

faboweb:
Fixed the logic. Needs refactors. Needs tests.

src/ActionModal/components/SendModal.vue
```diff
@@ -258,8 +263,10 @@ export default {
     setMaxAmount() {
       if (this.network.startsWith(`terra`)) {
         const terraTax = 0.008
-        this.amount =
-          this.selectedBalance.amount - this.selectedBalance.amount * terraTax
+        this.amount = maxDecimals(
+          this.selectedBalance.amount / (1 + terraTax),
+          6 // TODO get precision fro API
+        )
```
Bitcoinera: Oh, good one! I didn't think of the decimals

Bitcoinera:
Something is wrong with the linting. I got 50 warnings... mostly quotes related.

Did you change your linting config @faboweb ?

src/ActionModal/components/ActionModal.vue
```diff
@@ -682,9 +685,17 @@ export default {
 
       // limit fees to the maximum the user has
       if (this.invoiceTotal > this.selectedBalance.amount) {
+        let payable = Number(this.subTotal)
+        // in terra we also have to pay the tax
+        // TODO refactor using a `fixedFee` property
```
Bitcoinera: Where should we include this property? Inside the network JSON or better in balances, inside gasPrice I presume? And then gasPrice wouldn't be gasPrice anymore and would look like this:

```
fees: {
gasPrice: String,
fixedFee: String
}
```

Or what where your thoughts here?

faboweb:
> Did you change your linting config @faboweb ?

I did not. But need to investigate on that!

src/ActionModal/components/ActionModal.vue
faboweb: I don't now yet how to properly architect this. I would probably do al the calculation server side and just return a final fees object. Probably not handling all the edge cases client side. Just showing a warning: Your balance is not enough to pay for the transaction and fees.

Bitcoinera: Yes, makes sense. We don't need to be so accurate. Just a warning/error message would be enough, I think too.

But we need to refactor a lot to do it like you suggest. Get simulate() and estimatedGasFee all those in the API. But sounds like a good approach.

PR #3691, opened 2020-03-13
Aleksei/sentry add arch to link

iambeone:
Closes #ISSUE

**Description:**

Fixed arch value replaced with arch result

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/tasks/sentry.sh
+++ b/tasks/sentry.sh
@@ -1,6 +1,6 @@
 # install sentry cli
 echo "Installing and configuring sentry cli"
-curl -L -o sentry-cli https://github.com/getsentry/sentry-cli/releases/download/1.40.0/sentry-cli-Linux-x86_64
+curl -L -o sentry-cli https://github.com/getsentry/sentry-cli/releases/download/1.40.0/sentry-cli-Linux-`arch`
 chmod u+x sentry-cli
 ./sentry-cli --version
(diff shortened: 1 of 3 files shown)


faboweb:
did this fix it?

tasks/sentry.sh
```diff
@@ -1,6 +1,6 @@
 # install sentry cli
 echo "Installing and configuring sentry cli"
-curl -L -o sentry-cli https://github.com/getsentry/sentry-cli/releases/download/1.40.0/sentry-cli-Linux-x86_64
+curl -L -o sentry-cli https://github.com/getsentry/sentry-cli/releases/download/1.40.0/sentry-cli-Linux-`arch`
```
Bitcoinera: Why does it need to be a string like that for `arch` and not for `x86_64`?

iambeone: It's not just a string, it's a command. Placing it in ` means that we need to put here its result

Bitcoinera: oh I see. And what does the command `arch` then do? Fetch the `arch` version?

iambeone: Yap, it prints computer architecture.

Bitcoinera: Nice!

PR #3698, opened 2020-03-14
Fabo/adjusted release script

faboweb:
Release script adjusted

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "changelog": "simsala log --pending-path ./changes",
     "lint": "vue-cli-service lint",
     "lint:style": "vue-cli-service lint:style",
-    "release": "git fetch --all && git checkout origin/develop -B develop && git pull && git merge origin/master && git push -u origin develop && git push origin develop:release",
+    "release": "git fetch --all && git checkout origin/develop -B develop && git pull && git checkout origin/release -B release && git pull && git merge origin/develop && git push",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -10,7 +10,7 @@
     "changelog": "simsala log --pending-path ./changes",
     "lint": "vue-cli-service lint",
     "lint:style": "vue-cli-service lint:style",
-    "release": "git fetch --all && git checkout origin/develop -B develop && git pull && git merge origin/master && git push -u origin develop && git push origin develop:release",
+    "release": "git fetch --all && git checkout origin/develop -B develop && git pull && git checkout origin/release -B release && git pull && git merge origin/develop && git push",
```
Bitcoinera: I am sure this works, but going through it I see the danger that the `origin/release` branch would be possibly outdated in comparison with develop, and then `git pull` will not be enough (since the branch is outdated anyway.

Anyway, I guess I am wrong and you solve it by doing `git checkout origin/develop -B develop && git pull` first?

Another question I actually wanted to ask you, but forgot. When do you create the automatic simsala thing? That part is not here, right

faboweb: it gets triggered in `.circleci/config.yaml` and the code for the PR creation is in https://github.com/faboweb/simsala

Bitcoinera: You created simsala? O.o

faboweb: for sure! mostly for us to extract the release script from Lunie

Bitcoinera: Wow, saucool!

PR #3704, opened 2020-03-16
Aleksei/add extension getter to differ pathes

iambeone:
Closes #ISSUE

**Description:**

We are redirecting to the portfolio after creating/restoring accounts. But we don't have such a route in extension. 
Added an isExtension getter to identify the extension and redirecting to different url

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/src/components/common/TmSelectNetwork.vue
+++ b/src/components/common/TmSelectNetwork.vue
@@ -45,10 +45,8 @@ export default {
       }
     }
   },
-  watch: {
-    networks: networks => {
-      this.updateSelectedNetwork(networks)
-    }
+  mounted() {
+    this.updateSelectedNetwork(this.networks)
   },
   methods: {
     async selectNetworkHandler(network) {
--- a/src/components/common/TmSessionSignUpSeed.vue
+++ b/src/components/common/TmSessionSignUpSeed.vue
@@ -78,7 +78,7 @@ export default {
   }),
   computed: {
     ...mapState([`session`, `signup`]),
-    ...mapGetters([`network`, `networkSlug`]),
+    ...mapGetters([`network`, `networkSlug`, `isExtension`]),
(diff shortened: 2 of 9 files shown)


src/components/common/TmSessionSignUpSeed.vue
```diff
@@ -115,12 +115,16 @@ export default {
           name: this.signup.signUpName,
           network: this.network
         })
-        this.$router.push({
-          name: "portfolio",
-          params: {
-            networkId: this.networkSlug
-          }
-        })
+        if (this.isExtension) {
```
faboweb: better solution would be make the path a property, but that would require a huge refactor so this solution is good.

src/vuex/getters.js
```diff
@@ -7,6 +7,7 @@ export const address = state => state.session.address
 export const network = state => state.connection.network
 export const networkSlug = state => state.connection.networkSlug
 export const addressType = state => state.connection.addressType
+export const isExtension = () => false
```
faboweb: use existing `isExtension` in config

src/components/common/TmSelectNetwork.vue
```diff
@@ -45,10 +45,8 @@ export default {
       }
     }
   },
-  watch: {
-    networks: networks => {
-      this.updateSelectedNetwork(networks)
-    }
+  mounted() {
+    this.updateSelectedNetwork(this.networks)
```
faboweb: This will work but the better fix and to keep this async is to use `immediate: true` in the watcher. Please search the repo for it

Bitcoinera: oki

Bitcoinera: It doesn't work 🤷‍♀

PR #3705, opened 2020-03-16
Aleksei/remove network loading

iambeone:
Closes #ISSUE

**Description:**

Replace networks load from checkForPersistedNetwork function with state.networks

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/src/vuex/modules/connection.js
+++ b/src/vuex/modules/connection.js
@@ -1,5 +1,5 @@
 import config from "src/../config"
-import { Networks, NetworksAll } from "../../gql"
+import { NetworksAll } from "../../gql"
 
 export default function({ apollo }) {
   const state = {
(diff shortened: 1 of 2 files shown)


src/vuex/modules/connection.js
```diff
@@ -32,19 +32,14 @@ export default function({ apollo }) {
   const actions = {
     async checkForPersistedNetwork({ dispatch, commit }) {
```
faboweb: probably chain `checkForPersistedNetwork` after `preloadNetworks` then?

iambeone: Why? They are not logically connected. If you are about to avoid await on preloadNetworks that will no do any matter

faboweb: they apparently are logically connected if `checkForPersistedNetwork` needs the networks to have been loaded

iambeone: But many functions need the, to be loaded. Why do we need to chain only this one?

PR #3706, opened 2020-03-16
Aleksei/account creation fix

iambeone:
Closes #ISSUE

**Description:**

Commented an isTestnet function logic due to wrong account creation behavior

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/changes/aleksei_account_creation_fix
+++ b/changes/aleksei_account_creation_fix
@@ -1,0 +1,1 @@
+[Fixed] account creation steps should be disabled by default  @iambeone
--- a/src/components/common/TmSessionSignUp.vue
+++ b/src/components/common/TmSessionSignUp.vue
@@ -106,14 +106,10 @@ export default {
       }
     },
     isTestnet() {
-      return false
-      /*
-      left for future use
       const selectedNetwork = this.networks.find(
         ({ id }) => id === this.network
       )
       return selectedNetwork ? selectedNetwork.testnet : false
-      */
     }
   },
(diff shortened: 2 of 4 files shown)


faboweb:
What's the exact issue here?

iambeone:
The one with showing the account name window on account creation flow in secure mode

faboweb:
And why does this fix it?

iambeone:
For some reason (i don't know why) account name step is shown if current network is a testnet. Probably it was made for a reason, but not finished. I have updated this function to always return false, just to keep someone's work. Maybe we will enable account creation for testnets in future

faboweb:
so the issue is that the intended functionality is: if this is a testnet allow account creation. I think @Bitcoinera worked on that. Instead of this hack, can we fix that this intended functionality is not working?

Bitcoinera:
> Instead of this hack, can we fix that this intended functionality is not working?

OK, let me take over. I'm going to look into it

iambeone:
If we need to do it right now, why not)

Bitcoinera:
OK, now it is all working as expected. I also tested the mainnets and trying copy pasting the link and it would go to the InsecureModeWarning component as expected.

I don't know how all those InsecureModeWarnings got into the flow (TmSessionSignUpPassword and TmSessionSignUpSeed) but definitely not necessary and causing this bug.

Yes, this solution consisted of actually deleting code! 🍾 

Let's celebrate

Issue #3700, opened 2020-03-15
Activity/PageTransactions: Invariant Violation: Must contain a query definition


Bitcoinera:
**Describe the bug**

Anytime you go to the Activity section, you get greeted by this error coming from the `subscribeToMore` query.

![image](https://user-images.githubusercontent.com/40721795/76706261-6cacda00-66e6-11ea-87ab-4ab3ec9b7754.png)

**Additional Context**

Skipping the query "fixes" it.

Also I noticed there is almost the same query repeated. One inside the transactions query (lines 201 - 226) and the other one outside as a separate query (this is the failing one) in lines 228 - 263).

I don't know if this is an error or if the one outside is just there for tests (in that case, we could skip it in execution)

Bitcoinera:
I don't understand how but it seems the same error is still present. 

While testing Aleksei's PR #3735 the error was one and this one has been merged, so it is quite surprising 🤷‍♀

Bitcoinera:
This is how the error looks like now. The message changed a bit:

<img width="380px" src="https://user-images.githubusercontent.com/40721795/77227170-7a4adf80-6b7e-11ea-8645-59f16a788faa.png">

faboweb:
I saw the same. Maybe I bad fixed something. What was the initial solution?

faboweb:
did you update the backend? did you run against staging?

Bitcoinera:
I am running it against staging yes. The solution was just merged yesterday, in #3735 

While testing this PR these errors weren't appearing. 

Why do you think you were the one causing this?

faboweb:
There were merge conflicts

faboweb:
At least there are no visible errors

jbibla:
on develop and production i'm not getting any errors anymore - but I'm also not seeing any transactions (cosmos17njs9mgaszmhvanzx7dpuv7cyngs7wffk6g4s4)

Bitcoinera:
Seems to closed by #3751

PR #3710, opened 2020-03-16
Fabo/persist cache

faboweb:
Adds an apollo cache to local storage persistence


faboweb:
Hot hot hot 🐎 🔥

iambeone:
Dangerous thing...

faboweb:
Why dangerous?

Bitcoinera:
Will this have any effect on our chunks issue?

iambeone:
We can't manage this cache, so we can't rely on it) But the only issue i have noticed is with activity page, it doesn't show latest transactions

faboweb:
@Bitcoinera no I don't think so
@iambeone ok let's check why. And I think we can still manipulate that cache I guess.

iambeone:
I have made two stake and the first one appears in activity, but not the second

faboweb:
Does the request still happen? What's the result of the request?

Bitcoinera:
> But the only issue i have noticed is with activity page, it doesn't show latest transactions

Remember that we still have that bug that is probably preventing `subscribeToMore` from working... So maybe that is why.

This needs to be fixed before testing: #3700

faboweb:
I confirm the activity issue. let's check if the update works.

faboweb:
I also got issues with networks preloading

iambeone:
This issue is fixed (i will make a pr today), but you can try to update the page and activity page will remain the same

Bitcoinera:
Yes, Aleksei and me were discussing about this PR and the only thing we could come up for now is having a `localStorage.clear()` in ActionManager to handle new transactions.

Still, if this happens in Activity I bet it happens with rewards and validators too... so it seems quite unsafe (and putting `localStorage.clear()` everywhere kind of defeats the purpose too).

We could research if there is a way we can better control this cache. But probably not using localStorage.

faboweb:
hmm ok. I am doubting a bit that this is an unsolvable problem.

faboweb:
Can you recheck this PR. Seems to work fine.

Bitcoinera:
Oh God!!!

We need to merge this!!! It's A-M-A-Z-I-N-G!!

I just tried and it is boosting the performance like nothing else! This is this week's PR I'd say xD

And yeah, working like a charm.

So fast!

PR #3715, opened 2020-03-17
Fabo/remove service worker

faboweb:
Removes the service worker completely to avoid chunk errors


Bitcoinera:
@iambeone Did you test the chunks thing and confirmed this is fixing it?

Bitcoinera:
I followed all the steps in our Lunie Weekly document to reproduce the chunks error but ultimately couldn't... in the normal develop branch.

I still think this is a good idea to merge, since the service worker doesn't seem to be anything at all

iambeone:
Yap, i have checked and it works for me

jbibla:
> The networks icons are squares!

this is fixed already somewhere 😄

PR #3731, opened 2020-03-18
Ana/apply terra suggestion on LUNA fees

Bitcoinera:
Closes #ISSUE

**Description:**

Apply suggestions given by the Terra team to handle LUNA fees differently.

EDIT: To take into account, now we'd allow people to send LUNA with 0 fee too, although I believe this probably would fail on mainnet. Need to check. 

We could add the 1.5 multiplication safe measure there and maybe it would then work. Not sure.

Thank you! 🚀

---

For contributor:

For reviewer:


faboweb:
Would love to discuss this on a call before merging

Bitcoinera:
> Would love to discuss this on a call before merging

Sure. But first stay safe 😉

src/ActionModal/components/ActionModal.vue
```diff
@@ -451,11 +451,16 @@ export default {
...
         this.transactionData.type === transactionTypes.SEND
       ) {
-        return this.maxDecimals(
-          Number(this.gasEstimate) * Number(this.gasPrice) +
-            Number(this.amount) * terraTax,
-          6
-        ) // TODO get precision from API
+        return this.getDenom === this.network.stakingDenom
+          ? this.maxDecimals(
+              Number(this.gasEstimate) * Number(this.gasPrice) * 1.5,
```
faboweb: refactor 1.5 into a `const GAS_ADJUSTMENT`

src/ActionModal/components/ActionModal.vue
```diff
@@ -451,11 +451,16 @@ export default {
...
+              6
+            )
+          : this.maxDecimals(
+              Number(this.gasEstimate) * Number(this.gasPrice) +
```
faboweb: `GAS_ADJUSTMENT` needs to be applied here as well I guess.

src/ActionModal/components/ActionModal.vue
faboweb: is our gasPrice also `0.015` (per uluna)?

Bitcoinera: no, it is `0.001` for uluna (same than for uatom)

faboweb: Let's fix that?

Bitcoinera: ah true. Forgot about that recommendation. Let's do this

Bitcoinera: ...wait actually it is much higher no?

He said

```
const GAS_PRICE = '0.015'
const fee = Math.ceil(gas_estimate * 1.5 * GAS_PRICE)
```

whereas now we have 1e-8... something is off here

src/ActionModal/components/ActionModal.vue
Bitcoinera: I will ask Sim

faboweb:
Decision:
We don't use the adjustment. We only use fixed gasEstimates per tx and network served from the API.
We add `postAppliedFee` to ActionModal. This will receive an optional value. This will be added to the total fees. But it will not be used as the actual fee in the tx. Those fees are added by the network. In Polkadot we will have something similar.

Bitcoinera:
> But it will not be used as the actual fee in the tx. Those fees are added by the network. In Polkadot we will have something similar.

So that `postAppliedFee ` we only display it right? Ah, ok, now I also think understand the `terraTax` thing. That comes later and it is automatically done by the network. 

Is that right?

faboweb:
that is how I understand it

Bitcoinera:
I guess I need to add the `postAppliedFee` to the API together with the gasEstimates

src/ActionModal/components/ActionModal.vue
```diff
@@ -448,22 +451,17 @@ export default {
...
-          : this.maxDecimals(
-              Number(this.gasEstimate) *
-                TERRA_GAS_ADJUSTMENT *
-                Number(this.gasPrice) +
-                Number(this.amount),
-              6
-            ) // TODO get precision from API
+        // Terra tax is only a displayed value in TableInvoice
+        this.setTerraTax(this.amount)
+        return this.maxDecimals(
```
faboweb: This formula is weird. GasEstimate * GasPrice * amount?

Bitcoinera:
Please, allow me to experiment just a bit more (plus fixing tests and and solving conflicts ofc)

src/ActionModal/components/SendModal.vue
```diff
@@ -271,7 +274,18 @@ export default {
...
+        this.selectedBalance.denom !== `LUNA`
+      ) {
+        const terraTax = Math.min(
+          Number(this.selectedBalance.amount) * TERRA_TAX_RATE,
+          TERRA_TAX_CAP
+        )
+        this.amount = this.selectedBalance.amount - terraTax
+      } else {
+        this.amount = this.selectedBalance.amount
+      }
```
faboweb:
```suggestion
      let chainAppliedFees = 0
      if (
        this.network.startsWith(`terra`) &&
        this.selectedBalance.denom !== `LUNA`
      ) {
        const terraTax = Math.min(
          Number(this.selectedBalance.amount) * TERRA_TAX_RATE,
          TERRA_TAX_CAP
        )
        chainAppliedFees = terraTax
      }
      this.amount = this.selectedBalance.amount - 0
```

src/ActionModal/components/SendModal.vue
```diff
@@ -276,12 +274,15 @@ export default {
...
+        this.selectedBalance.denom !== `LUNA`
+      ) {
+        const terraTax = Math.min(
+          Number(this.selectedBalance.amount) * TERRA_TAX_RATE,
+          TERRA_TAX_CAP
         )
+        this.amount = this.selectedBalance.amount - terraTax
       } else {
         this.amount = this.selectedBalance.amount
       }
```
faboweb:
```suggestion
      let chainAppliedFees = 0
      if (
        this.network.startsWith(`terra`) &&
        this.selectedBalance.denom !== `LUNA`
      ) {
        const terraTax = Math.min(
          Number(this.selectedBalance.amount) * TERRA_TAX_RATE,
          TERRA_TAX_CAP
        )
        chainAppliedFees = terraTax
      }
      this.amount = this.selectedBalance.amount - chainAppliedFees
```

Bitcoinera: I have refactored everything and now terraTax is a prop coming from SendModal. Since there is no point or use case for the other actions... except perhaps claim rewards when there is only alt-tokens to claim 🤔 

I'm looking into it

src/ActionModal/components/SendModal.vue
```diff
@@ -11,6 +11,7 @@
     :notify-message="notifyMessage"
     feature-flag="send"
     :disabled="sendingNgm"
+    :terra-tax="getTerraTax(false)"
```
faboweb: let's not make this network specific and use `chain-applied-fees`

faboweb: ah I see you use the tax in different ways. then make fees 0? doe we need to send the terra tax or is applied on chain?

src/ActionModal/components/ActionModal.vue
faboweb: naaa, that's only if you get the gas estimation from the Cosmos SDK node. sorry for not seeing this earlier

Bitcoinera:
I am now dealing with ClaimRewards... The special case when there are no staking denom tokens available

Bitcoinera:
I'd call this ready. Let's get this merged 💪 !

src/ActionModal/components/ActionModal.vue
Bitcoinera: Ok, so then we are good no? We are using the gas price the said for uluna, `0.015`. This is what you were asking, right? 

(hard to keep track of these async conversations)

PR #3735, opened 2020-03-19
Aleksei/subscribe to more fix

iambeone:
Closes #ISSUE

**Description:**

Fixes https://github.com/luniehq/lunie/issues/3700

Thank you! 🚀

---

For contributor:

For reviewer:


iambeone:
Should be merged with https://github.com/luniehq/lunie-api/pull/476

Bitcoinera:
I got one of the transactionAdded duplicated. Will look into it

src/components/wallet/PageTransactions.vue
```diff
@@ -195,6 +195,10 @@ export default {
             ...this.loadedTransactions,
             ...result.transactionsV2
           ]
+          // sorting transactions
+          this.loadedTransactions = this.loadedTransactions.sort(
+            (a, b) => b.height - a.height
```
faboweb: This could lead to shuffling of messages in one tx as we have no way to sort them. Added an issue to lunie-api.

iambeone: Why? They are sorted in a component

faboweb: there can be multiple transactions per height even per hash right? How do you sort the transactions then?

iambeone: But they are already sorting somewhere in transaction list component. Even after this sort method.
I have added it just for one purpose. We have two possible update calls: to bottom when older transactions are loaded and to top — when new ones are loaded from subscribeToMore.
And in both cases we make result = [...result, ...newTransactions] so the new ones are on the bottom of the list.
But transactions are shown with limitation, so only the first ones are sending to transactions list. 
It's ok when we are adding older records, but wrong when we are added new records.
That's why i have to sort them.

faboweb: This is a correct implementation I think. That's why I did approve the PR already. But it surfaced that there is not enough sorting in the data model. See this issue https://github.com/luniehq/lunie-api/issues/477

Bitcoinera:
> Updates doublicate txs
> 
> image
> 
> Reproduce:
> Go to activity until txs are loaded
> Do a transaction
> Quickly go back to activity and wait until tx is in a block

This is what I said xDD

iambeone:
Can't confirm duplication and even claimed tx appearance on subscription)

faboweb:
I used the deployment preview...
image

iambeone:
With the right api?

faboweb:
production api

iambeone:
That will not work with current production

faboweb:
> That will not work with current production

would be good to write that ;)

Bitcoinera:
Using staging I am still getting duplicates 🤷‍♀

Bitcoinera:
Just fixed duplicates. Although could probably refactor without lodash (using reduce)

Bitcoinera:
> Not a good idea. Need to understand why it's duplicated

this.loadedTransactions contains all the old ones and result.transactionsV2 contains the old and the new one... but yes, I agree not a 'real' fix

src/components/wallet/PageTransactions.vue
```diff
@@ -195,8 +196,13 @@ export default {
             ...this.loadedTransactions,
             ...result.transactionsV2
           ]
+          // avoid duplicate transactions
+          const filteredLoadedTransactions = uniqWith(
```
faboweb: nooooop. as mentioned several times `hash` is not unique

package.json
```diff
@@ -61,6 +61,7 @@
     "graphql-tag": "2.10.1",
     "https-proxy-agent": ">=3.0.0",
     "jdenticon": "2.2.0",
+    "lodash": "4.17.15",
```
faboweb: we import the lodash packages individually as you can see in line 65+

src/components/wallet/PageTransactions.vue
Bitcoinera: true, woah, you are so right. I forgot about multi messages once more

package.json
Bitcoinera: oh, sure. I even used it once..

Bitcoinera:
@iambeone I was doing a bit more research and the only other alternative I can find to fix this is adding this conditional right before

```
this.loadedTransactions = [
...this.loadedTransactions,
...result.transactionsV2
]
```
which would be this one:

```
          if (
            this.loadedTransactions.length === result.transactionsV2.length ||
            this.loadedTransactions.length + 1 === result.transactionsV2.length
          )
            this.loadedTransactions = []
```
I think that using `uniqWith` here is much more elegant than this other fix.

I have been testing a good deal with other possibilities (using skips and others) and nothing else works.

Bitcoinera:
Codecov just went crazy 🤷‍♀ it is marking just an import

PR #3736, opened 2020-03-19
Ana/send claim rewards amount to extension

Bitcoinera:
Closes #ISSUE

**Description:**

The TransactionData being returned from ModalWithdrawRewards was incomplete and the getter for stakingDenom there was broken too.

Includes a hack I'll need to refactor to include a last message in Claim rewards with the amount (for extension).

This is a work in progress.

Thank you! 🚀

---

For contributor:

For reviewer:


Bitcoinera:
Another possibility would be, instead of adding amount as the last message in the signMessage for extension, to add it in every message together with validators, like this:

(lines 103-109)
```
          txMessage = {
            ...txMessage,
            value: {
              ...txMessage.value,
              amount: transactionProperties.amounts
            }
          }
```

But there is not much difference and I think it is better to add it as the last one

Bitcoinera:
It is actually working quite well

src/ActionModal/utils/ActionManager.js
```diff
@@ -103,6 +103,12 @@ export default class ActionManager {
           txMessages.push(txMessage)
         })
       )
+      // HACK until this is refactored in a proper way
+      const claimAmountMessage = {
+        type: "/".concat(type),
+        value: { amount: transactionProperties.amounts }
+      }
+      txMessages.push(claimAmountMessage)
```
faboweb: lol I get it now. very hacky... ^^
you remember that it is a bad idea to mix data types. same goes for adding pseudo messages to an actual message array.

faboweb: Proposal:

extend `getSigner` in `signer.js` to return functions of the format `(signMessage, metaData) => ...` where `metaData = { claimableRewards }` then pass `metaData` to `signWithExtension` and send it to the extension. The extension now reads this prop and uses it to display this information.

Bitcoinera: OK, so then I don't throw this into the messages... it feels weird to call this a metaData though. Especially when there is already a `txMetaData` already dancing around in ActionManager (can lead to confusion)

faboweb: yeah this whole thing needs to be heavily refactored. you can call it anything else. how about `displayingData`.

Bitcoinera: Well, I will added it right there then. To `txMetadata`

Bitcoinera: What about https://github.com/luniehq/lunie/pull/3736#issuecomment-601202588?

Bitcoinera:
Oops, you were right. The blockchain is not liking my message xD



I was wrong. It never worked

faboweb:
Looks better! Why are e2e failing?

Bitcoinera:
> Looks better! Why are e2e failing?

Some problem with fetching from browserstack

src/ActionModal/components/ActionModal.vue
```diff
@@ -737,6 +737,11 @@ export default {
         submitType: this.selectedSignMethod,
         password: this.password
       }
+      const txMetadata = {
```
jbibla:
```suggestion
      const txMetaData = {
```

src/ActionModal/components/ActionModal.vue
```diff
@@ -753,7 +758,7 @@ export default {
           type,
           memo,
           properties,
-          feeProperties
+          txMetadata
```
jbibla:
```suggestion
          txMetaData
```

src/ActionModal/utils/ActionManager.js
```diff
@@ -78,12 +78,19 @@ export default class ActionManager {
...
     txMetaData
   ) {
-    const { gasEstimate, gasPrice, submitType, password } = txMetaData
+    const {
+      gasEstimate,
+      gasPrice,
+      submitType,
+      password,
+      displayedProperties
+    } = txMetaData
```
jbibla: 😉

src/ActionModal/components/ActionModal.vue
```diff
@@ -737,6 +737,11 @@ export default {
         submitType: this.selectedSignMethod,
         password: this.password
       }
+      const txMetaData = {
+        ...feeProperties,
+        displayedProperties:
+          this.title === "Claim Rewards" ? properties.amounts : null
+      }
```
```suggestion
        displayedProperties: {
             amounts: properties.amounts
          }
      }
```

Bitcoinera:
Give me a bit and I also implement [this](https://github.com/luniehq/lunie-browser-extension/pull/179#discussion_r395429409)

src/ActionModal/components/ModalWithdrawRewards.vue
```diff
@@ -60,8 +60,15 @@ export default {
   computed: {
     ...mapGetters([`address`, `network`, `stakingDenom`]),
     transactionData() {
+      if (!this.claimedReward) return {}
```
faboweb: what is this for? will this break if `claimedReward` is not there?

Bitcoinera: I just copied from the other components 🤷‍♀ 

Yes, it would break but it would in any case. I can just get rid of it

By the way, we should merge this now that luniehq/lunie-browser-extension#179 is in, since otherwise it will break sowieso with the first claim rewards transaction

PR #3739, opened 2020-03-19
Jordan/improve portfolio loading

jbibla:
Closes #ISSUE

**Description:**

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/src/components/staking/Undelegations.vue
+++ b/src/components/staking/Undelegations.vue
@@ -8,7 +8,7 @@
     class="undelegations"
   >
     <h1>
-      Unstaked Validators
+      Pending
     </h1>
     <TableUndelegations :undelegations="undelegations" />
   </div>
--- a/src/components/staking/Undelegations.vue
+++ b/src/components/staking/Undelegations.vue
@@ -5,10 +5,11 @@
         undelegations &&
         undelegations.length > 0
     "
+    class="undelegations"
   >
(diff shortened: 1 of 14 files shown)


src/components/staking/Undelegations.vue
```diff
@@ -5,10 +5,11 @@
         undelegations &&
         undelegations.length > 0
     "
+    class="undelegations"
   >
-    <h3 class="tab-header">
-      Pending Undelegations
-    </h3>
+    <h1>
+      Unstaked Validators
```
faboweb:
```suggestion
      Unstaking Validators
```

faboweb: Pending Unstaking makes somehow more sense though. Also because you are not necessarily unstaking everything from a validator

jbibla: i know what you mean - but "Pending Unstaking" doesn't work grammatically. I changed it because we are moving away from "delegations / undelegations"...

We could just do "Pending" maybe?

@jrmoreau what do you think?

jrmoreau: Trying to visualize this. 

If the page or area of the page gives context for the user that they're looking at where their un-staking is done, I think simple "pending" is best. But without any context, I think we need "pending unstaking"

PR #3777, opened 2020-03-25
DO NOT MERGE Ana/fix extension redirecting. add testnet checkbox

Bitcoinera:
Closes #ISSUE

**Description:**

Adds a critical testnet checkbox for extension signin

Thank you! 🚀

---

For contributor:

For reviewer:


Bitcoinera:
Let's get this merged 💃 !

src/components/common/TmSessionExtension.vue
```diff
@@ -68,20 +79,45 @@ export default {
...
         sessionType: `extension`,
         address: account.address,
-        networkId: account.network ? account.network : "cosmos-hub-mainnet" // defaulting to cosmos-hub-mainnet
+        networkId: network.id
       })
     },
     async signInAndRedirect(account) {
       await this.signIn(account)
+      const accountNetwork = this.getSignInNetwork(account)
+      console.log(accountNetwork)
```
jbibla: remove console.log

Bitcoinera: xD

src/components/common/TmSessionExtension.vue
```diff
@@ -37,22 +37,33 @@
         </p>
       </div>
     </div>
+    <TmFormGroup class="field-checkbox" field-id="sign-up-warning" field-label>
+      <div class="field-checkbox-testnet">
+        <label class="field-checkbox-label" for="select-testnet">
+          <input id="select-testnet" v-model="testnet" type="checkbox" />
+          This is a testnet address</label
```
faboweb: @jbibla How about something like: Until we allow to copy addresses from testnet to mainnet you can decide the target network here.

src/components/common/TmSessionExtension.vue
```diff
@@ -68,20 +79,45 @@ export default {
...
       this.$store.dispatch("getAddressesFromExtension")
     },
+    getSignInNetwork(account) {
+      // defaulting to cosmos-hub-mainnet
+      const accountNetwork = account.network || "cosmos-hub-mainnet"
+      if (this.testnet) {
+        return (
+          this.networks
+            .filter(({ testnet }) => testnet)
+            .find(({ slug }) => accountNetwork.startsWith(slug)) ||
```
faboweb: 🤮 but ok for now

src/components/common/TmSessionExtension.vue
faboweb:
```suggestion
```

src/components/common/TmSessionExtension.vue
Bitcoinera: 😿

Bitcoinera: I mean, I also tried `.find(({ slug, testnet }) => accountNetwork.startsWith(slug) && testnet)`, but that somehow didn't work 🤷‍♀

faboweb:
Will stay open to provide preview to users in need

src/components/common/TmSessionExtension.vue
jbibla: 🤮 is much worse than 🤢 fabo

🌈 is the lunie way 🌈

Bitcoinera:
Oops, forgot about this one. What is the plan with this? Do we have the new designs?

jbibla:
we can probably close it and re-open if someone needs it.

Bitcoinera:
But right now it is not very ideal how the extension login is working... Why not just add this checkbox for now?

I think it is already better UX than what we have now. Even for us developers

Bitcoinera:
Why don't we merge this? I mean, we have the exact same logic for insecure Sign in, and this Sign In we will need to "open" it anyways once we release Polkadot on production...

What is the plan here?

jbibla:
i think this is very unique edge case and should not happen again in the future. the proper solution from my perspective is to allow seed recovery and add the mainnet / testnet tabs so it’s clearer.

jbibla:
i think this is very unique edge case and should not happen again in the future. 

the proper solution from my perspective is to allow seed recovery #3873, one seed to rule them all #3782 and add the mainnet / testnet tabs #3872.

PR #3778, opened 2020-03-25
Fabo/fix polkadot amount conversion

faboweb:
Fix conversion incorrect for Polkadot amounts


Bitcoinera:
Linting changes make readability here very difficult

faboweb:
https://github.com/luniehq/lunie/issues/3785

jbibla:
@faboweb cha cha cha changelog

src/ActionModal/components/SendModal.vue
```diff
@@ -205,7 +205,11 @@ export default {
         toAddress: this.address,
         amounts: [
           {
-            amount: uatoms(+this.amount),
+            amount: toMicroUnit(
+              +this.amount,
```
jbibla: is this `+` on purpose?

src/ActionModal/utils/networkMessages/polkadot-testnet.js
```diff
@@ -11,7 +11,7 @@ export async function MsgSend(
 ) {
   return await getSignMessage(senderAddress, "balances", "transfer", [
     toAddress,
-    amounts[0].amount * 1000000 // FIXME! Need to clarify why this conversion factor
```
jbibla: 🎉

src/scripts/num.js
```diff
@@ -141,6 +141,15 @@ export function uatoms(number = 0) {
   )
 }
 
+export function toMicroUnit(number, viewDenom, network) {
```
jbibla: it feels wrong to me that we're dealing with number conversions in the FE. I'm not up to date on the state of the transaction service but it feels like the API should be handling this change because the user never needs to see microunits in the frontend.

src/ActionModal/components/SendModal.vue
faboweb: would convert to a number. don't know if that is needed. wil lcheck

src/scripts/num.js
faboweb: you need to sign client side. so visual representation needs to be converted into sth signable which requires a conversion.

we could prepare the sign message in the API but that could cause man in the middle attacks where someone serves a wrong message and let's the user sign it without the user knowing.

src/ActionModal/components/SendModal.vue
faboweb: removed

src/scripts/num.js
michielmulders: Agree with this point!

tests/unit/specs/components/ActionModal/components/ModalDeposit.spec.js
```diff
@@ -12,15 +12,23 @@ describe(`ModalDeposit`, () => {
...
       dispatch: jest.fn(),
       getters: {
         userAddress: "cosmo1",
-        network: "testnet"
+        network: "testnet",
+        networks: [
+          {
+            id: "testnet",
+            coinLookup: [
+              { viewDenom: "STAKE", chainToViewConversionFactor: 0.000001 }
```
michielmulders: Nice!

tests/unit/specs/components/ActionModal/utils/ActionManager.spec.js
```diff
@@ -71,8 +71,11 @@ describe("ActionManager", () => {
   // values passed in from the ActionModal component
   const defaultContext = {
     chainId: "cosmos",
-    networkId: "cosmos-hub-testnet",
-    networkType: "cosmos",
+    network: {
+      id: "cosmos-hub-testnet",
+      coinLookup: [{ viewDenom: "uatom" }],
```
michielmulders: Here we want to check for `uatom`?

src/ActionModal/utils/networkMessages/polkadot-testnet.js
mariopino: bye bye uncertain! :-)

src/scripts/num.js
faboweb: which point?

tests/unit/specs/components/ActionModal/utils/ActionManager.spec.js
faboweb: what do you mean?

PR #3779, opened 2020-03-26
Aleksei/extension sign slug error fix

iambeone:
Closes #ISSUE

**Description:**

Fixed defaulting to cosmos-hub slug

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/src/components/common/TmSessionExtension.vue
+++ b/src/components/common/TmSessionExtension.vue
@@ -76,10 +76,10 @@ export default {
       })
     },
     async signInAndRedirect(account) {
-      let accountNetwork = await this.$store.dispatch("getNetworkByAccount", {
-        network: account.network,
-        address: account.address
-      })
+      let accountNetwork = await this.$store.dispatch(
+        "getNetworkByAccount",
+        account
+      )
       if (!account.network && accountNetwork) {
         account.network = accountNetwork.id
       }
--- a/src/components/common/TmSessionExtension.vue
+++ b/src/components/common/TmSessionExtension.vue
@@ -76,7 +76,6 @@ export default {
       })
     },
     async signInAndRedirect(account) {
-      await this.signIn(account)
       let accountNetwork = this.networks.find(
         ({ id }) => id === account.network
       )
@@ -94,6 +93,10 @@ export default {
           )
         }
       }
+      if (!account.network && accountNetwork) {
+        account.network = accountNetwork.id
+      }
+      await this.signIn(account)
       this.$router.push({
         name: "portfolio",
         params: {
--- a/src/components/common/TmSessionExtension.vue
+++ b/src/components/common/TmSessionExtension.vue
@@ -56,7 +56,7 @@ export default {
   }),
   computed: {
     ...mapState([`extension`]),
-    ...mapGetters([`networks`]),
+    ...mapGetters([`networks`, `findNetwork`]),
     accounts() {
       return this.extension.accounts
     }
@@ -76,23 +76,10 @@ export default {
       })
     },
     async signInAndRedirect(account) {
-      let accountNetwork = this.networks.find(
-        ({ id }) => id === account.network
-      )
-      // if no network, trying to identify by network_prefix
-      if (!accountNetwork) {
-        // select mainnet first
-        accountNetwork = this.networks.find(
-          ({ address_prefix, testnet }) =>
-            account.address.startsWith(address_prefix) && !testnet
-        )
-        if (!accountNetwork) {
-          // check for testnet
-          accountNetwork = this.networks.find(({ address_prefix }) =>
-            account.address.startsWith(address_prefix)
-          )
-        }
-      }
+      let accountNetwork = await this.$store.dispatch("getNetworkByAccount", {
+        network: account.network,
+        address: account.address
+      })
       if (!account.network && accountNetwork) {
         account.network = accountNetwork.id
       }
--- a/src/components/common/TmSessionExtension.vue
+++ b/src/components/common/TmSessionExtension.vue
@@ -80,9 +80,12 @@ export default {
       this.$router.push({
         name: "portfolio",
         params: {
-          networkId:
-            this.networks.find(({ id }) => id === account.network).slug ||
-            "cosmos-hub" // defaulting to cosmos-hub-mainnet
+          networkId: (
+            this.networks.find(({ id }) => id === account.network) || {
+              slug: "cosmos-hub"
+            }
+          ).slug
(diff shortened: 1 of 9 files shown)


src/components/common/TmSessionExtension.vue
```diff
@@ -80,9 +80,12 @@ export default {
...
         params: {
-          networkId:
-            this.networks.find(({ id }) => id === account.network).slug ||
-            "cosmos-hub" // defaulting to cosmos-hub-mainnet
+          networkId: (
+            this.networks.find(({ id }) => id === account.network) || {
+              slug: "cosmos-hub"
+            }
+          ).slug
+          // defaulting to cosmos-hub-mainnet
```
Bitcoinera: Isn't this returning exactly the same, `"cosmos-hub"`, for the default?

src/components/common/TmSessionExtension.vue
```diff
@@ -76,13 +76,31 @@ export default {
       })
     },
     async signInAndRedirect(account) {
+      let accountNetwork = this.networks.find(
```
faboweb: this should reallly really really really become a getter^^

src/components/common/TmSessionExtension.vue
```diff
@@ -76,13 +76,31 @@ export default {
...
+        ({ id }) => id === account.network
+      )
+      // if no network, trying to identify by network_prefix
+      if (!accountNetwork) {
+        // select mainnet first
+        accountNetwork = this.networks.find(
+          ({ address_prefix, testnet }) =>
+            account.address.startsWith(address_prefix) && !testnet
```
faboweb: doesn't work with Polkadot

faboweb: create helper function `getNetworkFromAddress`

Bitcoinera: I thought Polkadot had now an address_prefix, 2

faboweb: but the prefix is not a prefix. it is a value to calculate the address from the private key

Bitcoinera: I see. It is a "fake" prefix

src/components/common/TmSessionExtension.vue
iambeone: Converted to universal getter, defaulting to id

src/components/common/TmSessionExtension.vue
iambeone: This is not the issue with Polkadot. Mainly this change should help users with old extension, having no network field in their accounts. But those cases are numerous.

Bitcoinera: oh sure, and those ones cannot have Polkadot accounts logically

src/components/common/TmSessionExtension.vue
```diff
@@ -76,13 +76,18 @@ export default {
       })
     },
     async signInAndRedirect(account) {
+      let accountNetwork = await this.$store.dispatch("getNetworkByAccount", {
+        network: account.network,
+        address: account.address
+      })
+      if (!account.network && accountNetwork) {
+        account.network = accountNetwork.id
+      }
```
Bitcoinera: This looks a bit strange no? How is it possible that there is no `account.network` but yes accountNetwork if you are assigning `account.network` inside `accountNetwork`?

Guess it would look like 
```
accountNetwork {
network: undefined,
address: "blablabla"
}
```
Or does it error? Did you test this?

iambeone: It tries to define a network by prefix

src/vuex/getters.js
```diff
@@ -16,6 +16,16 @@ export const stakingDenom = state => {
   )
   return filteredNetwork ? filteredNetwork.stakingDenom : ``
 }
+export const findNetwork = state => value => {
```
faboweb: this can be a computed getter. so we don't need to call

faboweb: please read up on getters. this should be a computed value not a function. how it is written rn it should just move to connection.js

iambeone: But how is it possible? We need to select by value.

faboweb: > how it is written rn it should just move to connection.js

faboweb: what I mentioned before was changing the `network` getter to actually be the networks object and not just the network id. This can be a getter based on the network id defined in connection.js

src/components/common/TmSessionExtension.vue
faboweb: But the logic is confusing. Why didn't you use the same logic already existing in `TmSessionEplore`?

iambeone: Cause we have no testnet property in this component and we don't have addressPrefixes too.

iambeone: Yap, you are right...

faboweb:
```
      // HACK as polkadot addresses don't have a prefix
      if (isPolkadotAddress(this.address) && this.testnet) {
        return this.networks.find(({ id }) => id === "polkadot-testnet")
      }

      const selectedNetworksArray = this.networks.filter(({ address_prefix }) =>
        this.address.startsWith(address_prefix)
      )
```

Now find the network `selectedNetworksArray` based on the testnet prop.

src/components/common/TmSessionExtension.vue
```diff
@@ -76,13 +76,31 @@ export default {
...
       await this.signIn(account)
       this.$router.push({
         name: "portfolio",
         params: {
-          networkId:
-            this.networks.find(({ id }) => id === account.network).slug ||
-            "cosmos-hub" // defaulting to cosmos-hub-mainnet
+          networkId: accountNetwork ? accountNetwork.slug : "cosmos-hub" // defaulting to cosmos-hub-mainnet
```
faboweb: in which case is `accountNetwork` undefined?

iambeone: If someone will use a network account that was deleted

faboweb: so that should instead throw an error then "Network you are trying to sign into is not active on Lunie anymore".

iambeone: There is no error system in fe rn (

faboweb: but the error can be caught here in the component, right?

iambeone: Yap, but need to show something to user, cause now it's acting like broken link

PR #3795, opened 2020-03-27
Ana/get gas estimate from API using networksFee query

Bitcoinera:
Closes #ISSUE

**Description:**

Now we don't need those network specific hacks anymore. We just query for the specific network gas estimates directly

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/src/ActionModal/components/ActionModal.vue
+++ b/src/ActionModal/components/ActionModal.vue
@@ -300,7 +300,6 @@ import * as Sentry from "@sentry/browser"
 
 import ActionManager from "../utils/ActionManager"
 import BigNumber from "bignumber.js"
-import transactionTypes from "../utils/transactionTypes"
 
 const defaultStep = `details`
 const feeStep = `fees`
@@ -451,14 +450,6 @@ export default {
       )
     },
     estimatedFee() {
-      // another hack. e-Money doesn't neet such a high gas estimate for sending
-      if (
-        this.networkId.startsWith(`emoney`) &&
-        this.transactionData.type !== transactionTypes.WITHDRAW
-      ) {
-        this.updateEmoneyGasEstimate()
-      }
-      // hack
       // terra uses a tax on all send txs
       if (this.chainAppliedFees > 0) {
         return this.chainAppliedFees
@@ -560,12 +551,6 @@ export default {
     }
   },
   methods: {
-    updateTerraGasEstimate() {
-      this.gasEstimate = 300000
-    },
-    updateEmoneyGasEstimate() {
-      this.gasEstimate = 200000
-    },
     confirmModalOpen() {
       let confirmResult = false
       if (this.session.currrentModalOpen || !this.queueEmpty) {
--- a/src/ActionModal/components/ActionModal.vue
+++ b/src/ActionModal/components/ActionModal.vue
@@ -414,7 +414,6 @@ export default {
     password: null,
     sending: false,
     gasEstimate: null,
-    simulateGasEstimate: null,
     gasPrice: 0,
     submissionError: null,
     show: false,
@@ -651,9 +650,7 @@ export default {
           if (!this.isValidChildForm) {
             return
           }
-          this.sending = true
-          await this.simulate() // simulate to get gas estimation
-          this.sending = false
+          this.step = feeStep
           return
         case feeStep:
           if (!this.isValidInput(`gasPrice`)) {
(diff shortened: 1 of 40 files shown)


src/ActionModal/components/ActionModal.vue
```diff
@@ -402,6 +402,10 @@ export default {
     chainAppliedFees: {
       type: Number,
       default: 0
+    },
+    transactionType: {
```
faboweb: nice

src/ActionModal/components/ActionModal.vue
```diff
@@ -410,6 +414,7 @@ export default {
     password: null,
     sending: false,
     gasEstimate: null,
+    simulateGasEstimate: null,
```
faboweb: why both?

src/ActionModal/components/ActionModal.vue
```diff
@@ -447,21 +452,11 @@ export default {
...
     },
     estimatedFee() {
-      // another hack
-      if (this.networkId.startsWith(`emoney`)) {
-        this.updateEmoneyGasEstimate()
-      }
-      // hack
       // terra uses a tax on all send txs
       if (this.chainAppliedFees > 0) {
         // Terra gas estimate // TODO: get this from the API
```
```suggestion
```

src/ActionModal/components/ActionModal.vue
```diff
@@ -863,7 +852,7 @@ export default {
       },
       /* istanbul ignore next */
       skip() {
-        return !this.address
+        return !this.session.address
```
faboweb: use `address` getter instead

src/ActionModal/components/ActionModal.vue
```diff
@@ -903,6 +892,38 @@ export default {
...
+      },
+      /* istanbul ignore next */
+      update(data) {
+        if (data.networkFees) {
+          return data.networkFees.gasEstimate
+        }
+      },
+      /* istanbul ignore next */
+      skip() {
+        return !this.session.address || !this.transactionData
```
faboweb: same: use `address` getter

faboweb: I like the idea of making this reactive. Might need to check if this is not called too often (like if transactionData changes on every change of the amount input).

src/ActionModal/components/ModalDeposit.vue
```diff
@@ -7,6 +7,7 @@
     title="Deposit"
     class="modal-deposit"
     submission-error-prefix="Depositing failed"
+    transaction-type="DepositTx"
```
faboweb: use transactionType enum from `transactionTypes.js`

Bitcoinera: I wouldn't know then the transaction type 🤷‍♀ I need that prop to get the right gas Estimate.

src/ActionModal/components/ActionModal.vue
Bitcoinera: to keep the simulate logic. Although now it seems to be useless. I can also delete / comment it

src/ActionModal/components/ActionModal.vue
Bitcoinera: wait, why?

src/ActionModal/components/ModalDeposit.vue
faboweb: Yes, you should use the enum value. Like a variable. Instead of hardcoding the string value in here

Bitcoinera: Oops, transactionTypes are the "old" values, MsgSubmitProposal and so on... I need to make a dictionary to convert them

Bitcoinera: Well, actually it would be just much easier to change the values in `network-fees` and use the normal Cosmos-SDK messages. Or hardcode them. Dictionary feels like it is adding too much if it is not already there

src/ActionModal/components/ActionModal.vue
faboweb: because we have the getter for the address. So for keeping the code uniform we should use the same patterns everywhere. This will help making the code more readable and understandable for us and others.

src/ActionModal/components/ModalDeposit.vue
faboweb: the correct types are in `src\components\transactions\messageTypes.js`. Use those and remove `transactionTypes.js` (will cause some refactoring)

src/ActionModal/components/ActionModal.vue
Bitcoinera: IMO it is adding unnecessary complexity in this case, in an already super complex component

src/ActionModal/components/ModalDeposit.vue
faboweb: checking this rn

faboweb: ok doing it in another PR

src/ActionModal/components/ActionModal.vue
faboweb: I disagree

faboweb: let's do the refactor in another PR as well

src/ActionModal/components/DelegationModal.vue
```diff
@@ -232,15 +236,15 @@ export default {
 
       if (this.isRedelegation) {
         return {
-          type: transaction.REDELEGATE,
+          type: transactionTypes.REDELEGATE,
```
faboweb: those need to switch to the new message types as well. but in another PR.

PR #3797, opened 2020-03-30
Enable Akash portfolio

michielmulders:
Added the following changes:
- Added Akash to validation for address (bech32 prefix)
- Added images for Akash (both the same picture for akt.png and kash.png)


src/components/common/TmSessionExplore.vue
```diff
@@ -206,6 +206,7 @@ export default {
         param.substring(0, 6) === "terra1" ||
         param.substring(0, 5) === "xrn:1" ||
         param.substring(0, 7) === "emoney1" ||
+        param.substring(0, 6) === "akash1" ||
```
faboweb: see https://github.com/luniehq/lunie/pull/3779

PR #3810, opened 2020-03-31
Fabo/fix explore

faboweb:
Explore mode would not select address


tests/unit/specs/store/connection.spec.js
```diff
@@ -138,6 +138,13 @@ describe(`Module: Connection`, () => {
           address_prefix: "cosmos",
           network_type: "cosmos",
           testnet: true
+        },
+        {
+          id: "terra-testnet",
```
jbibla: should include e-money here too, no? and terra-mainnet?

faboweb: this is a unit test nor our whole networks list

PR #3811, opened 2020-03-31
Fabo/explore cleanup

faboweb:
Cleans up the explore component moving the address logic to the store


jbibla:
i don't see anything _added_ to the store here... was the logic already in the store and you're using it here now and deleted the duplicated logic?

faboweb:
The logic is already in the store. I am just removing the douplicate checks from this component

PR #3813, opened 2020-03-31
Fabo/fix e2e

faboweb:
Working
Removed some smelly code and fixed internationalization

### DIFF
--- a/tests/e2e/local-chrome.conf.js
+++ b/tests/e2e/local-chrome.conf.js
@@ -28,7 +28,7 @@ const nightwatch_config = {
         "disable-web-security",
         "ignore-certificate-errors",
         "no-sandbox",
-        "headless",
+        // "headless", // headless is not possible as it doesn't allow to fix the internationalization
         "window-size=1920,1080"
       ],
       prefs: {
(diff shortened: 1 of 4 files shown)


Bitcoinera:
It seems they are already fixed anyway? 🤷‍♀ 😅

tests/e2e/local-chrome.conf.js
```diff
@@ -28,7 +28,7 @@ const nightwatch_config = {
         "disable-web-security",
         "ignore-certificate-errors",
         "no-sandbox",
-        "headless",
+        // "headless", // headless is not possible as it doesn't allow to fix the internationalization
```
Bitcoinera, line 31: O.o Could you please explain this bit?

faboweb: a number like `0.4` is displayed in Germany `0,4` so when you compare both in the UI it would fail.

jbibla:
> Removed some smelly code 

👃

jbibla:
> It seems they are already fixed anyway? 

still some improvements in here!

tests/e2e/local-chrome.conf.js
Bitcoinera: oh so interesting!

PR #3814, opened 2020-03-31
Fabo/switch to github actions for releases

faboweb:
Allow to release via Github to allow circle to only run on PRs

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -5,8 +5,8 @@ on:
     branches:
       - release
 jobs:
-  test:
-    name: Compile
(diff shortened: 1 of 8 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,19 @@
+name: Create a release PR
+
+on:
+  push:
+    branches:
+      - release
+jobs:
+  test:
```
faboweb, line 8:
```suggestion
  createReleasePR:
```

faboweb:
Needs to be tested on next release

PR #3819, opened 2020-04-01
Ana/fix top 5 validator list

Bitcoinera:
Closes #ISSUE

**Description:**

Right now the version in develop&production is buggy, repeating the same validator multiple times in the list, and therefore not claiming rewards it could and also not matching the numbers displayed on the Claim Rewards modal.

Here I aim to fix the validator picking logic.

FIXES this:

And this (numbers not matching from modal and effectively claimed):

image

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/src/ActionModal/utils/ActionManager.js
+++ b/src/ActionModal/utils/ActionManager.js
@@ -157,7 +157,16 @@ export function getTop5RewardsValidators(rewards) {
         (Number(all[reward.validator.operatorAddress]) || 0)
     }
   }, {})
-  return Object.keys(rewardsPerValidatorObject).reverse()
-  // const rewardsPerValidatorAddresses = Object.entries(rewardsPerValidatorObject)
-  // return rewardsPerValidatorAddresses.map(([a, b]) => {a, b}).sort((a, b) => b - a)
+  const rewardsPerValidatorAddresses = Object.keys(rewardsPerValidatorObject)
+  let rewardsPerValidatorArray = []
+  rewardsPerValidatorAddresses.forEach((validatorAddress, index) => {
+    rewardsPerValidatorArray.push({
+      validator: validatorAddress,
+      totalRewardAmount: Object.values(rewardsPerValidatorObject)[index]
+    })
+  })
+  return rewardsPerValidatorArray
+    .sort((a, b) => b.totalRewardAmount - a.totalRewardAmount)
+    .slice(0, 5)
+    .map(rewardPerValidator => rewardPerValidator.validator)
 }
--- a/src/ActionModal/utils/ActionManager.js
+++ b/src/ActionModal/utils/ActionManager.js
@@ -149,20 +149,15 @@ function convertCurrencyData(amounts, network) {
 
 // limitation of the Ledger Nano S, so we pick the top 5 rewards and inform the user.
 export function getTop5RewardsValidators(rewards) {
-  const mostClaimableToken = rewards.reduce((topRewardToken, reward) => {
-    if (Number(reward.amount) > Number(topRewardToken.amount)) {
-      return reward
+  const rewardsPerValidatorObject = rewards.reduce((all, reward) => {
+    return {
+      ...all,
+      [reward.validator.operatorAddress]:
+        Number(reward.amount) +
+        (Number(all[reward.validator.operatorAddress]) || 0)
     }
-    return topRewardToken
-  })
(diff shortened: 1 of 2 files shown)


Bitcoinera:
Now it is working great! 😸

src/ActionModal/utils/ActionManager.js
```diff
@@ -149,20 +149,24 @@ function convertCurrencyData(amounts, network) {
...
 // limitation of the Ledger Nano S, so we pick the top 5 rewards and inform the user.
 export function getTop5RewardsValidators(rewards) {
-  const mostClaimableToken = rewards.reduce((topRewardToken, reward) => {
-    if (Number(reward.amount) > Number(topRewardToken.amount)) {
-      return reward
+  const rewardsPerValidatorObject = rewards.reduce((all, reward) => {
+    return {
+      ...all,
+      [reward.validator.operatorAddress]:
+        Number(reward.amount) +
```
faboweb, line 156: this sums up all rewards. but is ok I guess

src/ActionModal/utils/ActionManager.js
```diff
@@ -149,20 +149,24 @@ function convertCurrencyData(amounts, network) {
...
     }
-    return topRewardToken
+  }, {})
+  const rewardsPerValidatorAddresses = Object.keys(rewardsPerValidatorObject)
+  let rewardsPerValidatorArray = []
+  rewardsPerValidatorAddresses.forEach((validatorAddress, index) => {
+    rewardsPerValidatorArray.push({
+      validator: validatorAddress,
+      totalRewardAmount: Object.values(rewardsPerValidatorObject)[index]
+    })
```
faboweb, line 166: Do you know about `Object.entries`?

Bitcoinera: I should since I copied it from you in my PR 🤣 

But yeah, here it doesn't work since they don't have "keys" (so I cannot do:
```
const rewardsPerValidatorArray = Object.entries(rewardsPerValidatorObject)
rewardsPerValidatorArray.map(([a, b]) => {a, b})
```
, since rewardsPerValidatorObject just has <validatorAddress>:<totalRewardsAmoun>, and no "a" nor "b" key or none at all for that matter)

src/ActionModal/utils/ActionManager.js
Bitcoinera: Yes. I also consider this since eJPY for example has a very bulky reward and is not so valuable. But I think it is OK for the time being (we don't really need that much precision for now)

src/ActionModal/utils/ActionManager.js
faboweb:
```
const rewardsPerValidatorArray = Object.entries(rewardsPerValidatorObject)
rewardsPerValidatorArray.map(([validator, totalRewardAmount]) => ({validator, totalRewardAmount}))
```

Bitcoinera: Yes, but for that I needed this forEach

faboweb: where do you see a forEach in my code? ^^

Bitcoinera: I mean in my code xD

this one:
```
  rewardsPerValidatorAddresses.forEach((validatorAddress, index) => {
    rewardsPerValidatorArray.push({
      validator: validatorAddress,
      totalRewardAmount: Object.values(rewardsPerValidatorObject)[index]
    })
```
This is where I introduce those keys. They weren't there before

faboweb: noooo
check my code...
the "keys" are just names you give to the "first" and "second" item of the array which is the first argument of the funtion. It's called array destructuring: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment

Bitcoinera: image

It doesn't work 🤷‍♀ 

Maybe we can go through this in the next 1:1?

faboweb: you forgot the `()`. when you return an object from an arrow function you need to do `() => ({})`

Bitcoinera: 🤦‍♀ something so simple!

Wow, thank you so much for going through this with me Fabo! 😸  It was sure an important lesson

PR #3820, opened 2020-04-01
Ana/magical fix for undelegations 1 day short

### LINKED ISSUE #3796: Undelegation End Time seems to be off by 1 day (i.e. 1 day too short)
**Describe the bug**

We have already got two users reporting this. One of them was trying to Redelegate during the day he thought he could, but he actually needed 24h to ellapse to be able to do it.

**To Reproduce**

We can monitor if we have any pending undelegation, although the waiting time is kind of long

**Expected behavior**

21 days equals to 504h. We need to make sure that the calculation in the FE is correct and that pending undelegations keep being displayed while an undelegation is pending

Bitcoinera:
**Description:**

A small line change (almost), but a big step onward for Lunie 🤣 

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/src/components/staking/DelegationsOverview.vue
+++ b/src/components/staking/DelegationsOverview.vue
@@ -1,11 +1,7 @@
 <template>
   <div class="table-container">
     <div
-      v-if="
-        $apollo.queries.delegations.loading &&
-          !delegationsLoaded &&
-          !delegations.length
-      "
+      v-if="$apollo.queries.delegations.loading && !delegationsLoaded"
       class="loading-image-container"
     >
       <img
--- a/src/components/staking/DelegationsOverview.vue
+++ b/src/components/staking/DelegationsOverview.vue
@@ -1,7 +1,11 @@
 <template>
   <div class="table-container">
     <div
-      v-if="$apollo.queries.delegations.loading && !delegations.length"
+      v-if="
+        $apollo.queries.delegations.loading &&
+          !delegationsLoaded &&
+          !delegations.length
+      "
       class="loading-image-container"
     >
       <img
@@ -48,7 +52,8 @@ export default {
     TmDataMsg
   },
   data: () => ({
-    delegations: []
+    delegations: [],
+    delegationsLoaded: false
   }),
   computed: {
     ...mapGetters(["address", `network`, `networks`])
@@ -76,8 +81,9 @@ export default {
           delegatorAddress: this.address
         }
       },
+      /* istanbul ignore next */
       update(data) {
-        /* istanbul ignore next */
+        this.delegationsLoaded = true
         return data.delegations || []
       }
     },
--- a/src/components/staking/Undelegations.vue
+++ b/src/components/staking/Undelegations.vue
@@ -1,7 +1,11 @@
 <template>
   <div class="table-container">
     <div
-      v-if="$apollo.queries.undelegations.loading && !undelegations.length"
+      v-if="
+        $apollo.queries.undelegations.loading &&
+          !undelegationsLoaded &&
+          !undelegations.length
+      "
       class="loading-image-container"
     >
       <img
@@ -31,7 +35,8 @@ export default {
     TableUndelegations
   },
   data: () => ({
-    undelegations: []
+    undelegations: [],
+    undelegationsLoaded: false
   }),
   computed: {
     ...mapGetters([`address`, `network`])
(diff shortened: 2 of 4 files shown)


Bitcoinera:
To check it actually works (I know it is difficult to believe looking at the code) you can check the address `cosmos1rfg85mh5pmm5f0m05wmqg5qaknv3u7rfyqp36t` in Gaia testnet in develop/production and then in this branch.

Surprise!

jbibla:
so....... what do you suggest?

Bitcoinera:
> so....... what do you suggest?

Merge this! 🤩

And reopen the issue if we get any other complaints. But is seems the days number is actually good

jbibla:
if the days number is not the issue, why would we remove the loading logic? this doesn't fix any problems.

if you think the issue was that it loaded too slowly, we can add a loading svg to the `TableUndelegation` component.

src/components/staking/Undelegations.vue
```diff
@@ -1,12 +1,5 @@
 <template>
-  <div
-    v-if="
-      !$apollo.queries.undelegations.loading &&
-        undelegations &&
-        undelegations.length > 0
-    "
-    class="undelegations"
-  >
+  <div v-if="undelegations && undelegations.length > 0" class="undelegations">
```
faboweb, line 2: this just skips the loading check 🤷‍♂

Bitcoinera:
> if you think the issue was that it loaded too slowly, we can add a loading svg to the `TableUndelegation` component.

Yeah, true, like in `DelegationsOverview`.

I just wonder, why does it work so well without the loading logic? Is it really needed?

I am remembering something about loading being deprecated somehow or am I wrong? 🤔

src/components/staking/Undelegations.vue
Bitcoinera: yes, but I think this might be the issue, what got users confused...

jbibla:
>  Is it really needed?

yes, we should account for loading.

>  am remembering something about loading being deprecated somehow or am I wrong?

not sure.

Bitcoinera:
OK, I just copied now how it looks in the DelegationOverview so now this PR should be 100% safe.

Sorry if I have disappointed you with the PR title & bug and you were expecting real magic 😅

src/components/staking/Undelegations.vue
```diff
@@ -1,9 +1,21 @@
 <template>
-  <div v-if="undelegations && undelegations.length > 0" class="undelegations">
-    <h1>
-      Pending
-    </h1>
-    <TableUndelegations :undelegations="undelegations" />
+  <div class="table-container">
+    <div
+      v-if="$apollo.queries.undelegations.loading && !undelegations.length"
```
faboweb, line 4: you need a `undelegationsLoaded` property here. if no, this will show loading all the time when some update is triggered but you have no undelegations

Bitcoinera:
I updated the PR description since I don't think this will really close #3796...

Instead, I will make a new 3 days undelegation, very well documented, and without DST in the middle to better understand what is really happening here.

Bitcoinera:
Still, I think this is a good PR to merge

src/components/staking/DelegationsOverview.vue
```diff
@@ -1,7 +1,11 @@
 <template>
   <div class="table-container">
     <div
-      v-if="$apollo.queries.delegations.loading && !delegations.length"
+      v-if="
+        $apollo.queries.delegations.loading &&
+          !delegationsLoaded &&
+          !delegations.length
```
faboweb, line 7:
```suggestion
          !delegationsLoaded
```

src/components/staking/Undelegations.vue
```diff
@@ -1,16 +1,25 @@
...
-    <h1>
-      Pending
-    </h1>
-    <TableUndelegations :undelegations="undelegations" />
+  <div class="table-container">
+    <div
+      v-if="
+        $apollo.queries.undelegations.loading &&
+          !undelegationsLoaded &&
+          !undelegations.length
```
faboweb, line 7:
```suggestion
          !undelegationsLoaded
```

src/components/staking/Undelegations.vue
```diff
@@ -1,16 +1,25 @@
...
+      "
+      class="loading-image-container"
+    >
+      <img
+        class="loading-image"
+        src="/img/portfolio-loading.svg"
+        alt="geometric placeholder shapes"
+      />
+    </div>
+    <div v-if="undelegations.length > 0" class="undelegations">
```
faboweb, line 17:
```suggestion
    <div v-else-if="undelegations.length > 0" class="undelegations">
```

PR #3824, opened 2020-04-01
Ana/3804 validation error is misleading

### LINKED ISSUE #3804: validation error message is misleading
**Describe the bug**

when a user has a balance of 0 - the amount validation should say "you don't have any tokens" or something like that.
Screen Shot 2020-03-30 at 4 21 44 PM

**To Reproduce**
use lunie with an address that has no balance — click send and try to move to the next step of the send flow. this screenshot is from Kusama.

Bitcoinera:
Closes #3804 

**Description:**

The issue described in #3804 was due to the fact that Polkadot balances are coming as strings, not as numbers. So this is why the correct validation error message wasn't being triggered.

An alternative solution would be to send balances as numbers. Probably better solution.

What do you think @mariopino ?

Thank you! 🚀

---

For contributor:

For reviewer:


src/ActionModal/components/ActionModal.vue
```diff
@@ -73,7 +73,7 @@
               min="0"
             />
             <TmFormMsg
-              v-if="overview.liquidStake === 0"
+              v-if="overview.liquidStake == 0"
```
faboweb, line 76: why did you use '=='. that's always dangerous

faboweb:
do a string comparison instead?

faboweb:
We can also check max numbers and see if we could in theory just transport numbers

Bitcoinera:
Close this as I did the real fix in the API

Bitcoinera:
> We can also check max numbers and see if we could in theory just transport numbers

The fix for this was actually much more simple 🤩

Check luniehq/lunie-api#566

PR #3826, opened 2020-04-01
Fabo/3804 not enough tokens

### LINKED ISSUE #3804: validation error message is misleading

### LINKED ISSUE #3839: Deprecate `between` check on amount inputs

The error shown for the `between` validation is partially confusing.

Deprecate the between check in favor of `max` `min` and `noBalance` checks so we can show better error messages.

Create the error messages in JS instead of using validations

**Additional context**

faboweb:
Closes #3804
Closes #3839 

**Description:**

image

Note: Needs to be added to other amount inputs as well.

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/src/ActionModal/components/ActionModal.vue
+++ b/src/ActionModal/components/ActionModal.vue
@@ -84,11 +84,15 @@
               type="required"
             />
             <TmFormMsg
-              v-else-if="$v.gasPrice.$error && !$v.gasPrice.between"
-              :max="$v.gasPrice.$params.between.max"
+              v-if="$v.gasPrice.$error && !$v.gasPrice.max"
+              type="custom"
+              :msg="`You don't have enough ${selectedDenom}s to proceed.`"
+            />
+            <TmFormMsg
+              v-else-if="$v.gasPrice.$error && !$v.gasPrice.min"
               :min="0"
               name="Gas price"
-              type="between"
+              type="min"
             />
           </TmFormGroup>
           <TableInvoice
@@ -97,11 +101,15 @@
             :bond-denom="getDenom"
           />
           <TmFormMsg
-            v-if="$v.invoiceTotal.$invalid"
+            v-if="$v.invoiceTotal.$invalid && !$v.invoiceTotal.max"
+            type="custom"
+            :msg="`You don't have enough ${selectedDenom}s to proceed.`"
+          />
+          <TmFormMsg
+            v-else-if="$v.invoiceTotal.$invalid && !$v.invoiceTotal.min"
+            :min="smallestAmount"
             name="Total"
-            type="between"
-            min="0"
-            :max="selectedBalance.amount"
+            type="min"
           />
         </div>
         <div v-else-if="step === signStep" class="action-modal-form">
@@ -291,8 +299,8 @@ import TmDataMsg from "common/TmDataMsg"
 import TableInvoice from "./TableInvoice"
 import Steps from "./Steps"
 import { mapState, mapGetters } from "vuex"
-import { viewDenom, prettyInt } from "src/scripts/num"
-import { between, requiredIf } from "vuelidate/lib/validators"
+import { viewDenom, prettyInt, SMALLEST } from "src/scripts/num"
+import { requiredIf } from "vuelidate/lib/validators"
 import { track, sendEvent } from "scripts/google-analytics"
 import { UserTransactionAdded } from "src/gql"
 import config from "src/../config"
@@ -436,7 +444,8 @@ export default {
     isMobileApp: config.mobileApp,
     balances: [],
     queueEmpty: true,
-    includedHeight: undefined
+    includedHeight: undefined,
+    smallestAmount: SMALLEST
   }),
   computed: {
     ...mapState([`extension`, `session`]),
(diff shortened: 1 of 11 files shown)


Bitcoinera:
Aren't we overlapping here? Just saw this PR 👀 

Also, why not use the message that is already in place? In all modals, also Send: `doesn't have any ${network.stakingDenom}s`

The issue here is that Polkadot balances are coming as String in that check

src/ActionModal/components/SendModal.vue
```diff
@@ -360,6 +366,7 @@ export default {
       amount: {
         required: x => !!x && x !== `0`,
         decimal,
+        max: x => Number(x) <= this.maxAmount,
```
Bitcoinera, line 369: Why add this if we already have between?

faboweb: To show a better error message if the max has passed. Maybe instead we change the between error to show the better error

faboweb:
> doesn't have any ${network.stakingDenom}s

The message should show when the user has tokens but not enough. The message you posted appears when the user has no tokens of that denom

Bitcoinera:
> > doesn't have any ${network.stakingDenom}s
> 
> The message should show when the user has tokens but not enough. The message you posted appears when the user has no tokens of that denom

OK, but then the idea is to deprecate between? Because the `between` validation is exactly doing that

src/ActionModal/components/SendModal.vue
Bitcoinera: This is why I was asking. They overlap

faboweb: yep! we should probably remove `between` in favor of individual ones to show better errors?

Bitcoinera: but actually just with `max` we would get all the possible errors covered right?

faboweb:
I created a follow up: #3839

Bitcoinera:
Oops, completely forgot about negative numbers! Guess will need a `min` after all.

And `maxDecimals` will just be a separate error

Bitcoinera:
Good to go! 💪

src/ActionModal/components/ActionModal.vue
```diff
@@ -812,10 +821,12 @@ export default {
...
         // we don't use SMALLEST as min gas price because it can be a fraction of uatom
         // min is 0 because we support sending 0 fees
-        between: between(0, this.selectedBalance.amount)
+        max: x => Number(x) <= this.selectedBalance.amount,
+        min: x => Number(x) >= 0
       },
       invoiceTotal: {
-        between: between(0, this.selectedBalance.amount)
+        max: x => Number(x) <= this.selectedBalance.amount,
+        min: x => Number(x) >= SMALLEST
```
faboweb, line 833: does this make sense?

src/ActionModal/components/DelegationModal.vue
```diff
@@ -312,7 +322,13 @@ export default {
       amount: {
         required: x => !!x && x !== `0`,
         decimal,
-        between: between(SMALLEST, this.maxAmount)
+        max: x => Number(x) <= this.maxAmount,
+        min: x => Number(x) >= SMALLEST,
+        maxDecimals: x => {
+          return x.toString().split(".").length > 1
```
faboweb, line 328: just making sure: this serializes always to american number format?

src/ActionModal/components/ModalPropose.vue
```diff
@@ -217,7 +226,17 @@ export default {
       amount: {
         required: x => !!x && x !== `0`,
         decimal,
-        between: between(SMALLEST, this.balance.amount)
+        max: x => Number(x) <= this.balance.amount,
+        min: x => Number(x) >= SMALLEST,
```
faboweb, line 230: do we need a min if we have `maxDecimals` and `required`?

src/components/common/TmFormMsg.vue
```diff
@@ -53,6 +53,12 @@ export default {
         case `between`:
           msg = `must be between ${prettyDecimals(this.min)} and ${this.max}`
           break
+        case `maxDecimals`:
+          msg = `can have maximum 6 decimals`
```
faboweb, line 57: this depends on the network, but I guess we fixing to 6 decimals for now

src/ActionModal/components/ModalPropose.vue
Bitcoinera: yes because of negative numbers

src/ActionModal/components/DelegationModal.vue
Bitcoinera: oops... let me check that

src/ActionModal/components/ActionModal.vue
Bitcoinera: yes

src/ActionModal/components/DelegationModal.vue
Bitcoinera: Yeah, it is good. It is impossible to input it with commas 👍

Bitcoinera:
E2E failing because of a new thing... `url` hash or something weird like that O.o

src/ActionModal/components/ActionModal.vue
faboweb: hahahah why? can you help me understand with an example?

src/ActionModal/components/ModalPropose.vue
faboweb: ahhhh

src/ActionModal/components/ActionModal.vue
faboweb: I see! negative numbers

Bitcoinera:
This one is good to go too!

PR #3827, opened 2020-04-01
WIP Fabo/remove total rewards from am

faboweb:
Should help speed up modal loading times when the reward splitting PR is merged in the API

### DIFF
--- a/src/ActionModal/components/ActionModal.vue
+++ b/src/ActionModal/components/ActionModal.vue
@@ -73,8 +73,8 @@
               min="0"
             />
             <TmFormMsg
-              v-if="overview.liquidStake === 0"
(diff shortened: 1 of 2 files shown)


src/ActionModal/components/ActionModal.vue
```diff
@@ -846,8 +846,6 @@ export default {
       query: gql`
         query OverviewActionModal($networkId: String!, $address: String!) {
           overview(networkId: $networkId, address: $address) {
-            totalRewards
-            liquidStake
             accountInformation {
```
faboweb, line 849: this should move into an own query imo

Bitcoinera:
Why did you change the title to WIP? It is working great as it is!

Let's merge!

Issue #3760, opened 2020-03-24
Let FE know if it is outdated


faboweb:
**Is your feature request related to a problem? Please describe.**

Sometimes our FE is outdated as the files keep being cached.

**Describe the solution you'd like**

Based on:
https://medium.com/hypefactors/busting-spa-chunk-cache-for-active-users-after-a-deploy-a9a06242edb1

We may do this as we already talked about once:
- [ ] on deploy in Netlify trigger a mutation in Apollo to write in the table `versions` a record for the latest deployed version in `production` and ` staging`
- [ ] clients load/subscribe to the version
- [ ] clients show a reload message if there is new version while the app is open
- [ ] clicking on the outdated warning should force refresh the app
- [ ] clients refresh when the version is different then the current version on inital load

**Describe alternatives you've considered**

We already tried shipping the version together with the app on `/version.json` which has similar caching issues.

We can also figure out, why the index.html is not properly rechecked on every load.

**Additional context**

atharva3010:
Have you tried leveraging the service worker’s `updated()` method?

Bitcoinera:
> Have you tried leveraging the service worker’s `updated()` method?

We removed our service worker right @faboweb ?

faboweb:
Yep, I think it's still there for old users. I think we still need to remove old ones. https://love2dev.com/blog/how-to-uninstall-a-service-worker/

Bitcoinera:
This is going to be fixed also by #3837

PR #3832, opened 2020-04-02
Fabo/write version to db

faboweb:
To be tested
Prepares for #3760


.github/workflows/update_version.yml
```diff
@@ -0,0 +1,16 @@
+name: Update version in DB
```
faboweb, line 1: hmmm maybe better put into netlify to wait for the UI being deployed

PR #3834, opened 2020-04-03
Ana/fix 3825, staking max amount not working as expected

### LINKED ISSUE #3825: Staking max amount not working as it should
**Describe the bug**

When in the Stake modal you select "Max" to stake all available token, still there are fees being added on top of that max amount, in a way that forces the user to go back and change the amount to stake manually. 

Cannot just go and  try pay 0 fees.

**To Reproduce**

Try to stake max amount in any network. You will see the validation error just in the second screen

**Expected behavior**

When you select max amount, it should not add any fees on top

**Screenshots**

Bitcoinera:
Closes #3825 

**Description:**

This is a possible fix. Another possibility:

Add a fee service/helper in the FE, so the fees are calculated separately (not only in ActionModal like now), and then we can just import the function to the Delegation and SendModal (the only ones needing this actually). 

Note: I still need to fix this on the SendModal. There the same is happening.

Note2: it would be also interesting to allow 0 fees in testnets.

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/src/ActionModal/components/ActionModal.vue
+++ b/src/ActionModal/components/ActionModal.vue
@@ -473,6 +473,12 @@ export default {
       return this.featureFlag === "undelegate" ? 0 : this.amount
     },
     invoiceTotal() {
+      if (
+        Number(this.subTotal) + this.estimatedFee >
+        this.selectedBalance.amount
+      ) {
+        this.adjustFeesToMaxPayable()
+      }
       return Number(this.subTotal) + this.estimatedFee
     },
     isValidChildForm() {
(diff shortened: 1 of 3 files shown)


Bitcoinera:
I think I should better do the other proposed solution right? Move `estimatedFee` to `utils` inside the `ActionModal` folder. Call it `feeManager` or so. 

Otherwise too much code repetition (in SendModal, DelegationModal and ActionModal)

Bitcoinera:
Good to go!

src/ActionModal/components/ActionModal.vue
```diff
@@ -685,16 +691,14 @@ export default {
     },
     // limit fees to the maximum the user has
     adjustFeesToMaxPayable() {
-      if (this.invoiceTotal > this.selectedBalance.amount) {
```
faboweb, line 688: why did you move this check to `invoiceTotal`?

Bitcoinera: because this.invoiceTotal doesn't exist yet. 

I mean, I could have done the `Number(this.subTotal) + this.estimatedFee >
        this.selectedBalance.amount` in here, true. But I thought this way is much more readable and also it's clear the hack against negative gasPrices will be always run

faboweb:
missing changelog

faboweb:
I think we should test this case: https://codecov.io/gh/luniehq/lunie/compare/645946c2f8ab193ea660aedfcdbddadca80fc2d7...775b8323890821b24a76d3f9aa2687fd3c9c1b25/diff#D1-698

faboweb:
Looks good!

Bitcoinera:
Let's merge this!

PR #3837, opened 2020-04-03
Ana/fix chunk errors for real

### LINKED ISSUE #3760: Let FE know if it is outdated

Bitcoinera:
Closes #3760

**Description:**

Still work in progress although not looking bad

Thank you! 🚀

---

For contributor:

For reviewer:


tasks/poll-index.js
```diff
@@ -0,0 +1,33 @@
...
+    .then(result => {
+        if (!loaded) {
+            currentEtag = result.headers._headers.etag[0]
+            loaded = true
+        }
+        if (currentEtag !== result.headers._headers.etag[0]) {
+            console.log('Outdated Lunie version detected! Refreshing...')
+            // the client has an outdated version. We need to refresh the browser to get the latest changes (and chunks)
+            // when the reload method receives true as an argument, it will always reload the page from the server
+            window.location.reload(true)
```
Bitcoinera, line 29: Here is where I have some doubts... Is the `window` object reachable here?

faboweb: on web always

tasks/poll-index.js
```diff
@@ -0,0 +1,33 @@
+/*
+ * this script polls the Lunie site every 10 seconds and extracts its etag to check if there has been 
+ * any new Lunie versions deployed on Netlify while the user had the app open on the browser
+ * 
+ * Usage:
+ *
+ * node tasks/poll-index.js
+ */
+
+const fetch = require("node-fetch")
```
faboweb, line 10: why node fetch?

tasks/poll-index.js
```diff
@@ -0,0 +1,33 @@
+/*
```
faboweb, line 1: this should not go into tasks as it is part of the app, right?

tasks/poll-index.js
Bitcoinera: It is the one we are using in the other scripts 🤷‍♀

tasks/poll-index.js
Bitcoinera: But it should be a separate node process running along in the same instance right?

tasks/poll-index.js
faboweb: lol yeah but they run in Node.js. This runs in the browser where `fetch` is native

tasks/poll-index.js
faboweb: noooo, this runs in the browser not in node

faboweb: or do you want to run it in the API?

faboweb: we could also run this in the API and on an update trigger an apollo subscription

Bitcoinera: no, I think in the browser it makes more sense & it is more simple

tasks/poll-index.js
Bitcoinera: oh I see 😮

Bitcoinera: I tried removing it and it didn't work 🤷‍♀ it said `fetch undefined`

faboweb: Ur still executing the script with node...

Bitcoinera:
I think at this point it should work... problem is that https://app.lunie.io does not have the Access-Control-Allow-Origin header, so until then I cannot make requests there.

That would be how it would work. I can try and change the request to the deploy itself just to test

Bitcoinera:
I certainly don't know how to test this...

Using the preview url it gets automatically translated to https://app.lunie.io. Without the headers set in production, it seems impossible to test this 🤷‍♀

public/_headers
```diff
@@ -2,6 +2,7 @@
   X-Frame-Options: DENY
   X-XSS-Protection: 1; mode=block
   Content-Security-Policy: frame-ancestors 'self'
+  Access-Control-Allow-Origin: *
```
faboweb, line 5: is that needed?

src/scripts/check-for-new-lunie-versions.js
```diff
@@ -0,0 +1,36 @@
+/*
+ * this script polls the Lunie site every minute and extracts its etag to check if there has been
+ * any new Lunie versions deployed on Netlify while the user had the app open on the browser
+ *
+ * Usage:
+ *
+ * node tasks/poll-index.js
+ */
+
+const fetch = require("node-fetch")
```
faboweb, line 10:
```suggestion
```

src/scripts/check-for-new-lunie-versions.js
```diff
@@ -0,0 +1,36 @@
+/*
+ * this script polls the Lunie site every minute and extracts its etag to check if there has been
+ * any new Lunie versions deployed on Netlify while the user had the app open on the browser
+ *
+ * Usage:
+ *
+ * node tasks/poll-index.js
```
faboweb, line 7:
```suggestion
```

src/scripts/check-for-new-lunie-versions.js
```diff
@@ -0,0 +1,36 @@
...
+        currentEtag = result.headers._headers.etag[0]
+        loaded = true
+      }
+      console.log("Client Etag", currentEtag)
+      if (currentEtag !== result.headers._headers.etag[0]) {
+        console.log("Outdated Lunie version detected! Refreshing...")
+        // the client has an outdated version. We need to refresh the browser to get the latest changes (and chunks)
+        // when the reload method receives true as an argument, it will always reload the page from the server
+        window.location.reload(true)
+        loaded = false
```
faboweb, line 29:
```suggestion
```

public/_headers
Bitcoinera: it seems so

src/scripts/check-for-new-lunie-versions.js
Bitcoinera: Then it won't update the client etag

src/scripts/check-for-new-lunie-versions.js
Bitcoinera: Yeah, you are right. This was because I was testing at the beginning with node 🤦‍♀

Bitcoinera:
I cannot go past this. If I try using the preview url, it directly translates it to https://app.lunie.io...

image

Any idea how to work around this?

My solution would be to merge the _headers with the Access-Control-Allow-Origin header into production and then test this

public/_headers
faboweb: noooooo

src/scripts/check-for-new-lunie-versions.js
faboweb: why? `window.location.reload(true)` will reload the whole page

faboweb:
don't specify an URL at all. just get `/index.html` it will resolve to the current URL + `/index.html`

faboweb:
If this doesn't work, you can use `window.location.hostname`

faboweb:
querying app.lunie.io is also bad for previews as you would compare your local build against the production build

Bitcoinera:
Still getting the very same CORS error. And I already tried using the "/index.html" solution before.

The issue is that we don't have that header set in production and that the preview URL is directly transalated to https://app.lunie.io for some reason

faboweb:
Did you test using window.location.host?

Bitcoinera:
> Did you test using window.location.host?

Wow, this was it!

Bitcoinera:
It seems it didn't work so well window.location.host... Look at the URL it is calling 👀 

image

Bitcoinera:
I think I am getting an opaque response due to CORS... `headers` is just shown as a prototype 🤔

Bitcoinera:
Na, will look more into this tomorrow. Getting that opaque response

Bitcoinera:
Amazing, this works! Good bye chunk errors!

I'm so happy and have learnt a lot along the way! 🤩

src/initializeApp.js
```diff
@@ -34,6 +35,8 @@ function setOptions(urlParams, store) {
 export default async function init(urlParams, env = process.env) {
   // add error handlers in production
   if (env.NODE_ENV === `production`) {
+    // check every minute if new Lunie versions have been deployed
```
jbibla, line 38: what do you mean "every minute"

public/_headers
faboweb: so we can remove this, right?

src/scripts/check-for-new-lunie-versions.js
Bitcoinera: oh, and then it reruns the whole function making this unnecessary you mean? 🤔 

Makes sense...

public/_headers
Bitcoinera: Maybe you can give a better explanation than "nooooo" 😅 

I will try

Bitcoinera:
OK, @faboweb applied your suggestions.

Let's merge this!

Bitcoinera:
Why is pendingUpdated failing so much on this PR?

faboweb:
> Why is pendingUpdated failing so much on this PR?

because there is no changelog entry

Bitcoinera:
xD I would have sworn I had added it! 37 commits and no changelog ..

tests/e2e/helpers.js
```diff
@@ -252,8 +252,8 @@ async function getAccountBalance(browser) {
         browser.globals.denom = total[1]
         browser.globals.totalAtoms = total[0]
       })
-      /*await browser.getText(".total-atoms h2", result => {
-        browser.globals.totalAtoms = result.value.replace(",", "")
+      /*await browser.getText(".total-atoms h2", result => {
+        browser.globals.totalAtoms = result.value.replace(",", "")
       })*/
```
faboweb, line 257:
```suggestion
```

PR #3840, opened 2020-04-05
Fabo/netlify config location

faboweb:
Before the config was not picked up


faboweb:
Should disable Sentry in previews

faboweb:
Sourcemaps are still not picked up correctly

PR #3841, opened 2020-04-05
Fabo/lint staged

faboweb:
Only lint staged files and also apply style linting


package.json
```diff
@@ -128,7 +129,7 @@
   },
   "husky": {
     "hooks": {
-      "pre-push": "yarn lint --no-fix"
+      "pre-push": "lint-staged"
```
Bitcoinera, line 132: Shouldn't `lint-staged` be a script?

Bitcoinera: Oh, so by doing `npx mrm lint-staged` then it should be fine right?

But do we all need to do this in our machines in order for it to work?

faboweb: the package is already installed: https://github.com/luniehq/lunie/pull/3841/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R114

Bitcoinera: Did you try this? Making a push with this?

faboweb: yep, works. to test just fork, change a file (disabling vscode auto format before) and push

Bitcoinera: It's fine, I believe you 😉

Bitcoinera:
Just missing changelog!

faboweb:
image

Bitcoinera:
> Weird but ur right

Should E2E also test if there is a changelog included?

PR #3857, opened 2020-04-08
WIP Ana/add Polkadot inactive delegations

Bitcoinera:
Closes #ISSUE

**Description:**

Works together with luniehq/lunie-api#577

Work in progress since I'm not a super designer and we still need to figure out couple of things (do we include a countdown like in undelegations for the delegation to go active? I think it's a good idea)

This is how it is looking now: inactive delegations appear as faded

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/changes/ana_add-inactive-delegations-polkadot
+++ b/changes/ana_add-inactive-delegations-polkadot
@@ -1,0 +1,1 @@
+[Changed] [#3857](https://github.com/cosmos/lunie/pull/3857) Adds handling inactive delegations, necessary for Polkadot @Bitcoinera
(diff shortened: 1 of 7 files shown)


Bitcoinera:
Ah yes, and the amount... I still need to look how to get it from the API for inactive. Although if it is too hard we could hack that in the FE for now (until the next Polkadot runtime ✌️ )

jbibla:
i really like this ana. i just think we need to do the simplest thing first. in this case, I don't think there's an issue with the validator appearing like normal and not earning rewards for a while. users will not notice. this is how it works in cosmos if you don't stake too much. you won't see rewards for a while.

PR #3860, opened 2020-04-10
Ana/refactor create account to avoid using store

Bitcoinera:
Closes #ISSUE

**Description:**

To make luniehq/lunie-browser-extension#191 work. Currently still failing

Concretely Polkadot address creation is "broken" again

Thank you! 🚀

---

For contributor:

For reviewer:


src/vuex/modules/keystore.js
```diff
@@ -35,27 +35,30 @@ export default () => {
...
+      const wallet = await getWallet(seedPhrase, networkObject)
 
       storeWallet(wallet, name, password, network)
 
-      state.externals.track(`event`, `session`, `create-keypair`)
+      store.state.externals.track(`event`, `session`, `create-keypair`)
 
       // reload accounts as we just added a new one
-      dispatch("loadAccounts")
+      store.dispatch("loadAccounts")
```
faboweb, line 59: why the switch to `store`?

Bitcoinera: well, I thought the store-free file should be just wallet.js.. But yeah, Polkadot account creation is still not working there. Maybe because of this?

Bitcoinera: so I should remove store from the whole account creation process then

src/vuex/modules/keystore.js
```diff
@@ -35,19 +35,28 @@ export default () => {
...
-      return wallet.cosmosAddress
-    },
-    async getNetworkInfo(networkId, store) {
-      return store.getters.networks.find(({ id }) => id === networkId)
+      switch (networkObject.network_type) {
+        case "cosmos": {
+          return wallet.cosmosAddress
+        }
+        case "polkadot": {
+          return JSON.stringify(wallet.publicKey)
```
faboweb, line 45: o.O

Bitcoinera: 😅 I don't even understand how it was working before, always returning wallet.cosmosAddress O.o

faboweb: the naming is always `cosmosAddress` even for Polkadot. Needs to be adjusted in `cosmos-keys`. But ths line is probably why it doesn't work

src/vuex/modules/keystore.js
faboweb: all right. it's just weird as we in all the other parts destructure store to `{ commit, dispatch, getters }`

src/vuex/modules/keystore.js
```diff
@@ -35,27 +35,30 @@ export default () => {
...
+    },
+    async createKey(store, { seedPhrase, password, name, network }) {
       // TODO extract the key storage from the key creation
       const { storeWallet } = await import("@lunie/cosmos-keys")
 
+      // get current network
+      const networkObject = await this.getNetworkInfo(network, store)
       // create a new key pair
-      const wallet = await getWallet(seedPhrase, network, { getters })
+      const wallet = await getWallet(seedPhrase, networkObject)
```
faboweb, line 61: nice

src/vuex/modules/wallet.js
```diff
@@ -0,0 +1,48 @@
+// creates a cosmos addres for the network desired
```
faboweb, line 1: move to `src/crypto`? we could move their all those crypto scripts

src/vuex/modules/keystore.js
faboweb: `getWallet` should actually return `{ address, publicKey, seed }` I think.
For now reverting to `return wallet.cosmosAddress` should work

Bitcoinera: this is actually what `getAddressFromSeed` returns. Let me revert

src/vuex/modules/wallet.js
Bitcoinera: you mean `src/scripts`? I cannot find any `src/crypto`

src/vuex/modules/keystore.js
Bitcoinera: Let me revert

Bitcoinera:
Working good together with luniehq/lunie-browser-extension#191 💪

src/vuex/modules/wallet.js
faboweb: was a proposal for a new location, we can do that later

faboweb:
e2e fails

PR #3863, opened 2020-04-13
Ana/3738 & fix address is null

### LINKED ISSUE #3738: Loading delay on opening action modals
There is still a lot of loading strangely in places where the networks should already loaded. Like when opening the ActionModal (i.e. Withdraw Modal) there should not be any loading

### LINKED ISSUE #3855: Object.a
Sentry Issue: [FRONTEND-8T](https://monitoring.lunie.io:9000/lunie/frontend/issues/373/?referrer=github_integration)

```
Error: Non-Error exception captured with keys: extensions, locations, message, path
  at Object.a (/bundles/c31fcd422fdbfb1f8688748e08295f8af352cf03.js:45:128486)
  at None (/bundles/c31fcd422fdbfb1f8688748e08295f8af352cf03.js:74:51446)
  at Array.map (<anonymous>)
  at Object.next (/bundles/c31fcd422fdbfb1f8688748e08295f8af352cf03.js:74:51393)
  at w (/bundles/c31fcd422fdbfb1f8688748e08295f8af352cf03.js:65:10227)
...
(17 additional frame(s) were not displayed)
```

Bitcoinera:
Fixes address is null in Action Modal
Closes #3738 (this one for sure)

**Description:**

The error I encountered when following the steps from [this sentry error](https://monitoring.lunie.io:9000/lunie/frontend/issues/373/?referrer=github_integration) is slightly different than anything with an `Object.a`, but it was `variable $address is null` or something similar.

The reason is because acting on the `open` method, the ActionModal was still triggering the smartqueries despite the skips.

I have found some code there with `skipAll` kind of cumbersome (and actually here was the bug for the skips not being triggered) and also have not seen the need of the `loaded` logic at all. Removing the `loaded` logic makes the ActionModals open automatically (fixes #3738)

The only problem I have left is the test for the TmDataLoading. But maybe it failing like this means something is still wrong.

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/src/ActionModal/components/ActionModal.vue
+++ b/src/ActionModal/components/ActionModal.vue
@@ -45,7 +45,7 @@
       <template v-if="!checkFeatureAvailable">
         <FeatureNotAvailable :feature="title" />
       </template>
-      <TmDataLoading v-else-if="!loaded" />
+      <TmDataLoading v-else-if="$apollo.loading && !$apollo.skipAll" />
       <template v-else>
         <div v-if="requiresSignIn" class="action-modal-form">
           <p class="form-message notice">
@@ -555,17 +555,11 @@ export default {
         }
       }
     },
-    "$apollo.loading": function(loading) {
-      this.loaded = this.loaded || !loading
-    },
     selectedBalance: {
       handler(selectedBalance) {
         this.gasPrice = selectedBalance.gasPrice
       }
     }
-  },
-  created() {
-    this.$apollo.skipAll = true
   },
   updated: function() {
     if (
@@ -597,7 +591,11 @@ export default {
       }
     },
     async open() {
-      this.$apollo.skipAll = false
+      if (!this.address) {
+        this.$apollo.skipAll = true
+      } else {
+        this.$apollo.skipAll = false
+      }
       // checking if there is something in a queue
       const queue = await this.actionManager.getSignQueue(
         this.selectedSignMethod
@@ -630,7 +628,6 @@ export default {
       this.show = false
       this.sending = false
       this.includedHeight = undefined
-      this.$apollo.skipAll = true
 
       // reset form
       // in some cases $v is not yet set
--- a/src/ActionModal/components/ActionModal.vue
+++ b/src/ActionModal/components/ActionModal.vue
@@ -932,7 +932,7 @@ export default {
         },
         /* istanbul ignore next */
         skip() {
-          return !this.txHash
+          return !this.session.address || !this.txHash
         },
         query: UserTransactionAdded,
         /* istanbul ignore next */
--- a/src/ActionModal/components/ActionModal.vue
+++ b/src/ActionModal/components/ActionModal.vue
@@ -45,7 +45,11 @@
       <template v-if="!checkFeatureAvailable">
         <FeatureNotAvailable :feature="title" />
       </template>
-      <TmDataLoading v-else-if="$apollo.loading && !$apollo.skipAll" />
+      <TmDataLoading
+        v-else-if="
+          $apollo.queries.overview.loading || $apollo.queries.balances.loading
+        "
+      />
       <template v-else>
         <div v-if="requiresSignIn" class="action-modal-form">
           <p class="form-message notice">
@@ -593,8 +597,6 @@ export default {
     async open() {
       if (!this.address) {
         this.$apollo.skipAll = true
-      } else {
-        this.$apollo.skipAll = false
       }
       // checking if there is something in a queue
       const queue = await this.actionManager.getSignQueue(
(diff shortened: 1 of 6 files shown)


src/ActionModal/components/ActionModal.vue
```diff
@@ -45,7 +45,7 @@
       <template v-if="!checkFeatureAvailable">
         <FeatureNotAvailable :feature="title" />
       </template>
-      <TmDataLoading v-else-if="!loaded" />
+      <TmDataLoading v-else-if="$apollo.loading && !$apollo.skipAll" />
```
faboweb, line 48: why?

src/ActionModal/components/UndelegationModal.vue
```diff
@@ -375,13 +375,13 @@ export default {
       `,
       /* istanbul ignore next */
       skip() {
-        return !this.userAddress
+        return !this.address
```
faboweb, line 378: ok nice find

src/ActionModal/components/DelegationModal.vue
```diff
@@ -358,6 +358,9 @@ export default {
       /* istanbul ignore next */
       update(data) {
         return data.validators
+      },
+      skip() {
+        return !this.address
```
faboweb, line 363: top

src/ActionModal/components/ActionModal.vue
```diff
@@ -597,7 +591,11 @@ export default {
       }
     },
     async open() {
-      this.$apollo.skipAll = false
+      if (!this.address) {
+        this.$apollo.skipAll = true
```
faboweb, line 595: do we need this?

src/ActionModal/components/ActionModal.vue
Bitcoinera: `loaded` was introducing too much unnecessary complexity and was responsible for the never loading modals bug. 
Also, if we set `skipAll` to true, it doesn't make sense to wait for data. I think this is the most logical.

src/ActionModal/components/ActionModal.vue
Bitcoinera: I tried getting rid of `skipAll` altogether, but then a bunch of tests were failing... I can look further into that

src/ActionModal/components/ActionModal.vue
faboweb: loaded != $apollo.loading. loading changes every loading call loaded changes only after the first load

Bitcoinera:
Good to go! Fixes two bugs in one 😉 And one (ActionModal getting hung) is an old one

src/ActionModal/components/ActionModal.vue
faboweb: ping

Bitcoinera: I changed to `$apollo.queries.<query>.loading`, but it is probably the same.

I will just copy from Undelegations table. There I remember it had the proper logic

faboweb: using "loading" will show the loader on every apollo load of balances or overview

Bitcoinera: Yes, I was thinking. Thank you for confirming. I just fixed the logic of it all

src/ActionModal/components/ActionModal.vue
```diff
@@ -47,7 +47,8 @@
       </template>
       <TmDataLoading
         v-else-if="
-          $apollo.queries.overview.loading || $apollo.queries.balances.loading
+          $apollo.loading &&
```
faboweb, line 50: whyyyy? still loading there

faboweb: hmm maybe your right, need to try it out. as we load some stuff on every step, right?

Bitcoinera: I can confirm this works, but you're free to try it out. Like you said, `$apollo.loading` is true if any of the three queries (balances, overview and gasEstimate) is still loading

src/ActionModal/components/DelegationModal.vue
jbibla: what does "top" mean?

src/ActionModal/components/ActionModal.vue
jbibla: this is not a good strategy from a ux perspective. the step should always load instantly. if there is data that is being loaded we should show a loader inline with the data and restrict the user from going forward. blocking this components like this is no bueno.

Bitcoinera: Sounds cool, I like the idea. So basically what you are proposing is loading as you go, show what you have at each point, right?

Bitcoinera:
Please, let's merge 😆

src/ActionModal/components/DelegationModal.vue
faboweb: 🔝 👍

PR #3868, opened 2020-04-17
Fabo/3835 note about txs not showing

### LINKED ISSUE #3835: blocks page does not display transactions on terra
**Describe the bug**

when you visit a blocks page on terra, no transactions appear.

**To Reproduce**

1. visit https://app.lunie.io/terra/blocks/1454879
2. see this: 
Screen Shot 2020-04-03 at 10 53 10 AM

**Expected behavior**

we should show the transactions in the block.

faboweb:
Closes #3835 
Simple solution with adding a note about the missing txs:

image

### DIFF
--- a/changes/fabo_3835-note-about-txs-not-showing
+++ b/changes/fabo_3835-note-about-txs-not-showing
@@ -1,0 +1,1 @@
+[Added] [#3835](https://github.com/cosmos/lunie/issues/3835) Not about not showing some transactions @faboweb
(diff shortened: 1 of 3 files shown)


Bitcoinera:
> we should make a decision about the future of the blocks page and add the tx page IMO

Are we going then all-in for the BE plan?

jbibla:
> Are we going then all-in for the BE plan?

we need to discuss the document! Tuesday seems like a good time.

PR #3869, opened 2020-04-17
Fabo/add development flag in gql

faboweb:
Adds headers so we can ignore statistics from development builds


Bitcoinera:
We can merge this. The error from E2E is the same one I got.  Rerunning anyways

PR #3871, opened 2020-04-17
Ana/add missing security headers

Bitcoinera:
Closes #ISSUE

**Description:**

Following the mozilla scan advices. [Here ](https://observatory.mozilla.org/analyze/app.lunie.io)is the current report.

I'm only unsure about the `Content-Security-Policy` header. This one has a lot of possible combinations that improve a lot security. But I'm uncertain if we should include 'unsafe-inline' for example... In local everything is working fine setting everything to `self` it seems.

Here are the `Content-Security-Policy` header [docs](https://infosec.mozilla.org/guidelines/web_security#content-security-policy). I followed the fifth example

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/public/_headers
+++ b/public/_headers
@@ -1,5 +1,0 @@
-/*
-  X-Frame-Options: DENY
-  X-XSS-Protection: 1; mode=block
-  X-Content-Type-Options: nosniff
-  Referrer-Policy: same-origin
--- a/public/_headers
+++ b/public/_headers
@@ -1,18 +1,19 @@
 /*
   X-Frame-Options: DENY
   X-XSS-Protection: 1; mode=block
-  Content-Security-Policy: frame-ancestors 'self';
-  Content-Security-Policy: object-src 'none'; 
-  Content-Security-Policy: base-uri 'self'; 
-  Content-Security-Policy: default-src 'unsafe-inline'; 
-  Content-Security-Policy: img-src 'self' https://www.google-analytics.com;
+  Content-Security-Policy: default-src 'self' https://*.lunie.io;
+  Content-Security-Policy: child-src 'self' https://*.lunie.io https://*.netlify.app;
+  Content-Security-Policy: connect-src 'self' https://api-iam.intercom.io;
+  Content-Security-Policy: img-src 'self' data: https://*.lunie.io https://*.netlify.app https://www.google-analytics.com https://s3.amazonaws.com/keybase_processed_uploads;
+  Content-Security-Policy: media-src 'self' https://*.lunie.io https://*.netlify.app blob:;
+  Content-Security-Policy: object-src 'self' data: blob: https://*.lunie.io https://*.netlify.app https://www.google-analytics.com https://s3.amazonaws.com/keybase_processed_uploads;
+  Content-Security-Policy: script-src 'self' 'unsafe-inline' 'unsafe-eval' https://*.lunie.io https://*.netlify.app https://www.google-analytics.com https://api-iam.intercom.io;
+  Content-Security-Policy: style-src 'self' 'unsafe-inline' https://*.lunie.io https://*.netlify.app;
   Content-Security-Policy: manifest-src 'self' https://*.lunie.io https://*.netlify.app;
-  Content-Security-Policy: script-src-elem 'unsafe-inline';
-  Content-Security-Policy: script-src 'unsafe-inline' https://*.lunie.io https://*.netlify.app; 
-  Content-Security-Policy: worker-src 'none'; 
-  Content-Security-Policy: style-src 'unsafe-inline';
+  Content-Security-Policy: worker-src 'none';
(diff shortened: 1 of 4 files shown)


Bitcoinera:
[More docs](https://content-security-policy.com/) on the `Content-Security-Policy` header (it is a very complex one)

public/_headers
```diff
@@ -1,7 +1,9 @@
 /*
   X-Frame-Options: DENY
   X-XSS-Protection: 1; mode=block
-  Content-Security-Policy: frame-ancestors 'self'
+  Content-Security-Policy: frame-ancestors 'self'; default-src 'self'; script-src 'self'
```
faboweb, line 4: did you consider https://github.com/luniehq/lunie/blob/develop/vue.config.js#L55 ?

Bitcoinera: I see, it is really well configured there... But checking app.lunie.io in Network I only get the header present in `headers`:



Also, I don't understand, why don't we put all those options directly in `headers`? Why do we need to do it through webpack?

Maybe it is not working

faboweb: let'S move it all over to the netlify config then

faboweb:
the file is invalid now according to Netlify

Bitcoinera:
> the file is invalid now according to Netlify

Fixed. This playground was useful: https://play.netlify.com/headers

Bitcoinera:
> doesn't work

Oops, should have checked earlier. Will fix it tomorrow 👌

public/_headers
```diff
@@ -5,7 +5,7 @@
   Content-Security-Policy: object-src 'none'; 
   Content-Security-Policy: base-uri 'self'; 
   Content-Security-Policy: default-src 'self; 
-  Content-Security-Policy: script-src 'self' https://*.lunie.io; 
+  Content-Security-Policy: script-src 'self' https://*.lunie.io https://deploy-preview-*--lunieio.netlify.app;
```
faboweb:
```suggestion
  Content-Security-Policy: script-src 'self' https://*.lunie.io https://*--lunieio.netlify.app; 
```

Bitcoinera: yeah, that wasn't working 👍 

Now that seems fixed, but the problem is that I'm getting a `Note that 'img-src' was not explicitly set, so 'default-src' is used as a fallback` when the `img-src` is already set and, in fact, even before the actual `default-src`.

Really weird.

Will continue looking into this tomorrow. Cannot explain why this happens

public/_headers
```diff
@@ -4,9 +4,9 @@
   Content-Security-Policy: frame-ancestors 'self';
   Content-Security-Policy: object-src 'none'; 
   Content-Security-Policy: base-uri 'self'; 
-  Content-Security-Policy: img-src * 'self' https://www.google-analytics.com;
+  Content-Security-Policy: img-src 'self' https://www.google-analytics.com data:;
   Content-Security-Policy: manifest-src 'self' https://*.lunie.io https://*.netlify.app;
-  Content-Security-Policy: script-src-elem * 'unsafe-inline';
+  Content-Security-Policy: script-src-elem 'unsafe-inline' data:;
```
faboweb, line 9: `data:` is considered dangerous I think lets investigate and maybe change

Bitcoinera: not so easy to set these headers. That is for sure

faboweb:
still doesn't work

Bitcoinera:
> still doesn't work

I know xD I was fixing Polkadot bugs

faboweb:
validator images are not loading

Bitcoinera:
> validator images are not loading

Yes, and loading is somehow so slow... CSP are tricky, but also powerful. Maximum security 🔐

Bitcoinera:
Now this is the only thing "failing"

image

Bitcoinera:
Can you also check the load speed? Seems to have decreased on first run... or maybe this is how deploys are anyways?

faboweb:
> Now this is the only thing "failing"

that is a different error

faboweb:
image

jbibla:
thank you for your initiative here @Bitcoinera

Bitcoinera:
> thank you for your initiative here @Bitcoinera

🙏 

This is actually ready to be merged! Let's do it! 💃

faboweb:
image

Bitcoinera:
I cannot quote reply, but here you have the explanation to that mysterious error: it comes from the Grammarly extension: https://stackoverflow.com/questions/45366744/refused-to-load-the-font-datafont-woff-it-violates-the-following-content

Bitcoinera:
I cannot find any other errors. Intercom I guess we can only test on production right? But should be fine

faboweb:
But we need to work away from the data attribute: https://security.stackexchange.com/questions/94993/is-including-the-data-scheme-in-your-content-security-policy-safe/95011#95011

PR #3874, opened 2020-04-17
Ana/create and send lunietransaction to extension

Bitcoinera:
Closes #ISSUE

**Description:**

Part of the master plan to refactor extension in a much more meaningful way. Send a complete TransactionV2 to extension (like proposed in luniehq/lunie-browser-extension#194)

Thank you! 🚀

---

For contributor:

- [ ] Added changes entries. Run `yarn changelog` for a guided process.
- [ ] Reviewed `Files changed` in the github PR explorer
- [ ] Attach screenshots of the UI components on the PR description (if applicable)
- [ ] Scope of work approved for big PRs

For reviewer:

- [ ] Manually tested the changes on the UI

PR #3875, opened 2020-04-17
Fabo/signing refactor

faboweb:
Refactors message creation and signing
message creator function now receive the same arguments as in TransctionV2 TransactionDetails
Simplified the logic on how to create messages and sign in ActionManager

Tests need to be adjusted and all messages need to be tested

`sendTxAPI` interface needs still some love

We can now use the same logic in the extension. We can pass around the messages which are very TransactionV2 (missing some properties) so we can show the transaction easily and just also send `transactionData` on top for message creation and signing.

Also changes the submitted messages types when broadcasting to lunie transaction types.


src/ActionModal/utils/cosmos-signing.js
```diff
@@ -0,0 +1,121 @@
+// adds the signature object to the tx
```
faboweb, line 1: I extracted a bunch of the functions from the `cosmos-api` library. Can instead export the needed functions from there.

faboweb:
@Bitcoinera do we still need `displayedProperties` in that case?

Bitcoinera:
> @Bitcoinera do we still need `displayedProperties` in that case?

No. Actually we don't even need `transactionData` anymore actually. With `lunieTransaction` we have all we need.

I removed `displayedProperties` completely in both luniehq/lunie#3874 and luniehq/lunie-browser-extension#198

src/ActionModal/utils/cosmos-signing.js
Bitcoinera: I'm a bit confused here. I thought we were going to create this in the extension? 

Also, are we going then to use my approach in #3874 to create a lunieTransaction (TransactionV2) or is that not needed having this?

Wouldn't like to write too much code that won't be later used 😓

src/ActionModal/components/DelegationModal.vue
```diff
@@ -251,25 +250,17 @@ export default {
...
           type: transactionTypes.REDELEGATE,
-          validatorSourceAddress: this.from,
-          validatorDestinationAddress: this.targetValidator.operatorAddress,
-          amount: toMicroUnit(
-            this.amount,
-            this.stakingDenom,
-            this.networks.find(({ id }) => id === this.network)
-          ),
-          denom: toMicroDenom(this.stakingDenom)
+          from: [this.from],
```
faboweb, line 253: @bitcoinera here you see that the transactionData already contains all the things that you will need for displaying

src/ActionModal/components/ActionModal.vue
```diff
@@ -728,41 +723,32 @@ export default {
...
-            userAddress: this.session.address,
-            rewards: this.rewards,
-            account: this.overview.accountInformation
-          },
-          type,
-          memo,
-          properties,
-          txMetaData
-        )
+        let transactionData
```
faboweb, line 729: @bitcoinera here you see the simplified flow in the ActionModal

src/signing/transaction-manager.js
```diff
@@ -0,0 +1,192 @@
...
+      body: JSON.stringify({ payload })
+    }
+    const graphqlHost = getGraphqlHost()
+
+    return fetch(`${graphqlHost}/transaction/broadcast`, options).then(result =>
+      result.json()
+    )
+  }
+
+  async getSignQueue(submitType) {
```
faboweb, line 34: we should deprecate the queue

src/signing/transaction-manager.js
```diff
@@ -0,0 +1,192 @@
...
+      {
+        address: senderAddress,
+        password,
+        network: network.id,
+        networkType: network.network_type
+      },
+      config // only needed for Ledger
+    )
+
+    const broadcastableObject = await createAndSign(
```
faboweb, line 102: `createAndSign´ together with `getSigner` can be used in the extension when we send to the extension

```
    messageType,
    message,
    transactionData,
    senderAddress,
    network,
```

faboweb:
Fees need to be calculated though in the extension from the gasEstimate and gasPrice for now.
Even though: We could just also send the calculated fees for displaying purposes. But then a malicious FE could hide way to high gas costs by sending wrong fees. The data always needs to be the same source of true for creating and signing as for displaying.

src/ActionModal/components/DelegationModal.vue
Bitcoinera: yes, I'm aware of it 😉

src/signing/transaction-manager.js
Bitcoinera: Isn't it useful in any way?

Bitcoinera:
> But then a malicious FE could hide way to high gas costs by sending wrong fees. The data always needs to be the same source of true for creating and signing as for displaying.

I don't see the point of such an attack. Also the user has already seen the true fee in the ActionModal before sending. They would know if the cost is too high from ActionModal.

But anyway, if we calculate that in the extension in the end, then we don't need to send them for displaying purposes. I can change that in #3874, remove fees from lunieTransaction

src/ActionModal/components/ActionModal.vue
Bitcoinera: nice

src/ActionModal/components/DelegationModal.vue
Bitcoinera: O.o oh, I thought we needed toMicroUnit and toMicroDenom to build the transactions! noooo 😨 

I took the time to add a `displayAmount` in my PR 🤦 will need to remove it then

Why did we have it then with the toMicroUnit and MicroDenom???

Bitcoinera:
This is one hell of a PR xD

Better split?

src/ActionModal/components/DelegationModal.vue
faboweb: now I just convert all `amount, denom` in the network message creators instead of in the ActionModal. The ActionModal creates values in the TransactionV2 exchange format

src/signing/transaction-manager.js
faboweb: how?

faboweb:
> Better split?

Probably

faboweb:
> I don't see the point of such an attack. Also the user has already seen the true fee in the ActionModal before sending. They would know if the cost is too high from ActionModal.

The malicious FE could also display wrong data in the ActionModal.

Imagine a different case where we just by accident send the wrong gasPrice/ gasEstimate and the user can't know because he can't see what he is signing.

src/ActionModal/components/ActionModal.vue
```diff
@@ -867,34 +854,6 @@ export default {
         return !this.session.address
       }
     },
-    overview: {
```
faboweb, line 870: done now in the transaction-manager

src/signing/cosmos-signing.js
```diff
@@ -0,0 +1,121 @@
+// adds the signature object to the tx
```
faboweb, line 1: maybe rename to `cosmos-transactions.js`

src/signing/transaction-manager.js
```diff
@@ -0,0 +1,209 @@
...
+
+async function createAndSign(
+  messageType,
+  message,
+  transactionData,
+  senderAddress,
+  network,
+  signer
+) {
+  const messages = await getMessage(
```
faboweb, line 166:
```suggestion
  const chainMessages = await getMessage(
```

src/signing/transaction-manager.js
```diff
@@ -0,0 +1,209 @@
...
+    network,
+    messageType,
+    senderAddress,
+    message
+  )
+
+  const { getSignableObject, getBroadcastableObject } = await import(
+    `./${network.network_type}-signing.js`
```
faboweb, line 174: consider polkadot and cosmos renaming to `polkadot-transaction...`

src/signing/polkadot-signing.js
```diff
@@ -7,18 +7,31 @@ const curvePrefixes = {
...
+  const keypair = keyring.createFromUri(wallet.seedPhrase)
 
   const rawSignature = keypair.sign(hexToU8a(payload.toRaw().data))
-  const signature = getSignature(rawSignature)
+
+  return { payload, transaction, rawSignature }
+}
+
+export function getBroadcastableObject(
+  _,
```
faboweb, line 30: rename

src/vuex/modules/keystore.js
```diff
@@ -35,8 +35,7 @@ export default () => {
       return getSeed()
     },
     async getAddressFromSeed(store, { seedPhrase, network }) {
```
faboweb, line 37: passing around the whole network object saves us from needing to search for the network in nested functions

tests/unit/specs/store/keystore.spec.js
```diff
@@ -123,35 +123,29 @@ describe(`Module: Keystore`, () => {
...
-          {
-            id: "polkadot-testnet",
-            network_type: "polkadot",
-            address_prefix: "2"
-          }
-        ]
-      }
-    }
-    const address = await actions.getAddressFromSeed(store, {
+    jest.setTimeout(10000)
```
faboweb, line 139: needs to be done because Polkadot address creation is partially very slow

Bitcoinera:
😨 



I actually already got this error in my closed PR #3874

src/signing/networkMessages/cosmos-hub-mainnet.js
```diff
@@ -15,39 +15,40 @@ export function SendTx(senderAddress, { to, amount }, network) {
 }
 
 // Staking
-export function StakeTx(senderAddress, { to, amount }, network) {
+export function StakeTx(senderAddress, { to, amount, denom }, network) {
```
faboweb, line 18: @Bitcoinera this is wrong. the schema for a StakeTx detail looks like this:

``` 
type StakeTx {
    amount: Coin!
    to: [String]!
  }
```

Bitcoinera: But this is how you pass it in transactionData:

```
        return {
          type: messageType.STAKE,
          to: [this.targetValidator.operatorAddress],
          amount: this.amount,
          denom: this.stakingDenom
        }
```

faboweb: then passing the value from the actionmodal is wrong. sorry

PR #3900, opened 2020-04-23
Ana/fix polkadot import account wasm crypto

### LINKED ISSUE #3901: Polkadot import account with Schnorrkel fails on first run
It is not waiting for the `wasm-crypto` package to be loaded so there is no `addFromUri` function

Bug screenshot:

image

Bitcoinera:
Closes #3901

**Description:**

Fixes this (wasn't waiting for the wasm-crypto package to be initialized):

image

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/src/vuex/modules/wallet.js
+++ b/src/vuex/modules/wallet.js
@@ -10,6 +10,7 @@ function getCosmosAddressCreator(bech32Prefix) {
 async function createPolkadotAddress(seedPhrase, addressPrefix) {
   const [{ Keyring }] = await Promise.all([
     import("@polkadot/api"),
+    import("@polkadot/wasm-crypto"),
     import("@polkadot/util-crypto").then(async ({ cryptoWaitReady }) => {
       // Wait for the promise to resolve, async WASM or `cryptoWaitReady().then(() => { ... })`
       await cryptoWaitReady()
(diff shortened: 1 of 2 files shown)


src/vuex/modules/wallet.js
```diff
@@ -10,6 +10,7 @@ function getCosmosAddressCreator(bech32Prefix) {
 async function createPolkadotAddress(seedPhrase, addressPrefix) {
   const [{ Keyring }] = await Promise.all([
     import("@polkadot/api"),
+    import("@polkadot/wasm-crypto"),
```
faboweb, line 13: is that also needed for signing or only for address creation?

faboweb: for signing it is not needed?

Bitcoinera: Let me check if I get it there too

Bitcoinera: no, signing works just fine (for Schnorrkel accounts).

For Edwards accounts that get "crunched" I still get BadProof though

faboweb: what does crunched mean?

Bitcoinera: Well, in this case I want to mean the addresses that get "destroyed".

Like I had Mario's one that looked like this `JGVLUnieotWLyg25D1uF88EahxmGjc7Dhe2fHUjA5ziBd7i`, but after the Schnorrkel thing it looks like this: `16hApVhutE942rs6G9FrVKbPHjgBANM4qpXmRvC8ENojcxkv` 

And now I cannot sign any transaction with that transformed address. I always get `BadProof`

PR #3905, opened 2020-04-26
Fabo/fix extension accounts not read

faboweb:
THe warning about extension account not available would always show


Bitcoinera:
Ah nice! I thought I was the only one xD

PR #3908, opened 2020-04-26
Ana/enable insecure signups in development

Bitcoinera:
Closes #ISSUE

**Description:**

Again, very small changes, but I think a significant improvement.

It was getting kind of monotonous to have to enter our secret flag all the time. This way it saves up both time and mental power.

Now we have all the signups options in development, and if we want to try things out in production, we just need to set `produtctionMode` to true by using the `?production=true` flag

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/src/components/common/CardSignInRequired.vue
+++ b/src/components/common/CardSignInRequired.vue
@@ -5,11 +5,7 @@
 
     <div class="session-list">
       <LiSession
-        v-if="
-          session.insecureMode ||
-            (session.developmentMode && !session.productionMode) ||
-            (isMobileApp && !accountExists)
-        "
+        v-if="session.insecureMode || (isMobileApp && !accountExists)"
         id="create-new-address"
         icon="person_add"
         title="Create a new address"
@@ -31,12 +27,7 @@
       >
       </LiSession>
       <LiSession
-        v-if="
-          accountExists &&
-            (session.insecureMode ||
-              (session.developmentMode && !session.productionMode) ||
-              isMobileApp)
-        "
+        v-if="accountExists && (session.insecureMode || isMobileApp)"
         id="sign-in-with-account"
         icon="lock"
         title="Sign in with account"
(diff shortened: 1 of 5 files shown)


Bitcoinera:
This bit doesn't seem to be working as expected...?

It always routes to the SelectNetwork component. Or is that expected?

```
    createAccountLink() {
      return this.isMobileApp ||
        this.session.insecureMode
        ? `/select-network/create`
        : `/create`
    }
```

src/components/common/CardSignInRequired.vue
```diff
@@ -5,7 +5,11 @@
 
     <div class="session-list">
       <LiSession
-        v-if="session.insecureMode || (isMobileApp && !accountExists)"
+        v-if="
+          session.insecureMode ||
+            (session.developmentMode && !session.productionMode) ||
```
faboweb, line 10: this looks weird

faboweb:
why not just set insecureMode to true if in development?

Bitcoinera:
> why not just set insecureMode to true if in development?

true 😅 ...

Bitcoinera:
Anyway, I was just thinking, maybe this soon won't be necessary right? Because as soon as we release Kusama, we will need to make these ones available in production too.

Or what is the plan here? Make available only for the Kusama network?

src/initializeApp.js
```diff
@@ -27,6 +27,9 @@ function setOptions(urlParams, store) {
   if (urlParams.insecure === `true`) {
     store.commit(`setInsecureMode`)
   }
+  if (urlParams.production === `true`) {
```
faboweb, line 30: remove production mode switch?

Bitcoinera: But then we cannot test the `/create` endpoint in development. I mean this bit:

image

faboweb: In doubt you can just manually change the value in code. This is a very rare case

Bitcoinera: OK, then I will remove it

PR #3912, opened 2020-04-27
Modify CSP for intercom wss error

michielmulders:
I'm getting the following CSP error:

image

Added `wss://*.intercom.io` host to `connect-src`.

Further, added:
- `form-action 'self'` -> To be sure in case we should use a form action which submits to a path/URL.

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -9,7 +9,7 @@
     '''
     Content-Security-Policy = '''
       default-src 'self' https://*.lunie.io https://*.netlify.app https://www.google-analytics.com https://monitoring.lunie.io:9000 https://js.intercomcdn.com;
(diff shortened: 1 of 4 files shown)


netlify.toml
```diff
@@ -20,6 +20,8 @@
       frame-src 'self' https://api-iam.intercom.io https://widget.intercom.io;
       frame-ancestors 'self';
       base-uri 'self';
+      form-action 'self';
+      navigate-to 'self';
```
Bitcoinera, line 24: I think we should remove this. Now users cannot navigate to anywhere. They cannot go to Coinbase using our referral, they cannot go to the marketing page, ... 😅  Too much security

Bitcoinera, line 24:
```suggestion

```

Bitcoinera:
Nice label btw 😉

PR #3922, opened 2020-04-27
WIP: Ana/moonpay integration

Bitcoinera:
Closes #ISSUE

**Description:**

Add Moonpay as our fiat on-ramp. Didn't do much yet actually. Saw their staging system is quite strange since you need to go through the whole verification process like normal and even add a regular card 🤷 So didn't get to the end of it yet

Thank you! 🚀

---

For contributor:

For reviewer:


Bitcoinera:
> Is this still relevant? If so, what is blocking it? Is there anything you can do to help move it forward?
> 
> This issue has been automatically marked as stale because it has not had recent activity. It will be closed if no further activity occurs.

Yes, still super relevant. I realized the env variable eas not being picked up 😅  But yeah, still asking for a real credit card even though 😆

src/components/common/TmBalance.vue
```diff
@@ -144,6 +144,14 @@
               <i class="material-icons">send</i></button
             ><span>Send</span>
           </div>
+          <div v-if="developmentMode" class="icon-button-container">
```
faboweb, line 147: change to experimental mode?

src/components/common/TmBalance.vue
```diff
@@ -250,6 +270,7 @@ export default {
       rewards: [],
       selectedFiatCurrency: "USD",
       preferredCurrency: "",
+      developmentMode: config.development,
```
faboweb, line 273: remove

src/components/common/TmBalance.vue
Bitcoinera: oh, ok. It is easier to test like this though. But ok

Bitcoinera: 👀 ?? What is the difference between experimental and developmentMode currently? In `session.js` I see:

```
    developmentMode: config.development, // can't be set in browser
    experimentalMode: config.development, // development mode, can be set from browser
```

Bitcoinera:
I just contacted the Moonpay team asking on behalf of the credit card for their staging version. Let's see what they answer.

Also, saw that our US users won't be able to purchase any of the coins from our current networks

jbibla:
design note: buttons should not be so close together:
https://www.sketch.com/s/8ffba04e-5055-44dd-984c-c60a74f5103b/a/mlOkjV

design question: should we have both buttons?

Screen Shot 2020-05-04 at 8 01 26 AM

src/components/common/TmBalance.vue
```diff
@@ -190,6 +201,18 @@
               ><span>Send</span>
             </div>
           </div>
+
+          <div
```
jbibla, line 205: note: the fact that we need to duplicate this indicates that there is a need for a refactor. https://github.com/luniehq/lunie/issues/3942#issuecomment-623424830

src/components/common/TmBalance.vue
```diff
@@ -243,6 +266,7 @@ export default {
       rewards: [],
       selectedFiatCurrency: "USD",
       preferredCurrency: "",
+      moonpayCoins: new Set(["ATOM", "LUNA", "SDT", "KAVA"]), // TODO: add warning for US custormers as they cannot purchase any of these coins
```
jbibla, line 269: why do we need to do `new Set()`?

src/components/common/TmBalance.vue
```diff
@@ -761,6 +796,10 @@ select option {
   color: var(--menu-bright);
 }
 
+.circle-buy-button {
+  background: #f67f70;
```
jbibla, line 800: this should be a variable — and the hover color should match

src/components/common/TmBalance.vue
```diff
@@ -730,6 +761,10 @@ select option {
   padding-left: 1rem;
 }
 
+.icon-button-container {
+  margin-right: 5px;
```
jbibla, line 765: use `rem` not pixels

Bitcoinera:
I added a minimal fiat currency selector. But we do need this (if we want our users to be able to buy in other fiat currencies apart from USD), I confirmed it:

src/components/common/TmBalance.vue
Bitcoinera: It is actually the same in the end of the day no?

image
(our variables.css)

I changed it anyways

src/components/common/TmBalance.vue
Bitcoinera: To know about if they are in the US or not we'd need to ask all our users for their location. They'd get a popup like this:



Or were we getting this info already for GA? 🤔

Bitcoinera: This is done now. Will ask @jrmoreau to confirm it is working

src/components/common/TmBalance.vue
jbibla: no. `5px` is `5px` ... `var(--px)` is `0.0625rem`.

jbibla:
so you think this is not available?

Screen Shot 2020-05-06 at 1 18 19 PM

i like the fiat selector but ... why is this separate from the existing currency dropdown?

Bitcoinera:
> so you think this screen would not be available in our moonpay integration?

Where is this coming from? 😮 I hadn't seen that before! 

Show me how to use the fire 🔥 

> i like the fiat selector but ... why is this separate from the existing currency dropdown?

Yeah, this is something I have also considered... It could be tied together and for the non-supported currencies on Moonpay we would open the modal. And only then... But yeah, users need to realize this (that they are coupled together). So, that is why I think this is a better UX

jbibla:
there's a menu in moonpay
Screen Shot 2020-05-06 at 2 05 07 PM

jbibla:
ana, i think we should remove the fiat selector for now. it is not needed for a first implementation. 

please also fix the unit tests and conflicts.

let's make sure we can merge this code without activating it yet? @faboweb ?

Bitcoinera:
> ana, i think we should remove the fiat selector for now. it is not needed for a first implementation.
> 
> please also fix the unit tests and conflicts.
> 
> let's make sure we can merge this code without activating it yet? @faboweb ?

Well, I can remove it but it will only allow people to buy crypto in USD. When ironically people in the USA cannot buy any of our tokens... 🤷 but as you prefer

jbibla:
this is not a priority right now. let's revisit when we knock off some more of our projects on priorities list.

Bitcoinera:
> this is not a priority right now. let's revisit when we knock off some more of our projects on priorities list.

OK, let's close it for now then

PR #3961, opened 2020-05-04
Michiel/add csp hash

michielmulders:
Seems like the intercom settings have changed as the CSP flags this as an insecure script.


faboweb:
A plug-in solution might be a better choice in the long term

PR #3962, opened 2020-05-04
Michiel/remove Firebase import warning 

michielmulders:
Simple PR to remove import warning for Firebase.


jbibla:
this is now only importing the module we need instead of all of firebase, right?

michielmulders:
@jbibla Yes!

PR #3970, opened 2020-05-04
limit staking

faboweb:
Optimizes queries. But takes the same time as the simple Mario version if you

### DIFF
--- a/src/signing/networkMessages/polkadot-testnet.js
+++ b/src/signing/networkMessages/polkadot-testnet.js
@@ -42,28 +42,68 @@ export async function StakeTx(senderAddress, { to, amount }, network) {
   return await getSignMessage(senderAddress, transactions)
 }
 
-export async function ClaimRewardsTx(senderAddress) {
+export async function ClaimRewardsTx(senderAddress, { from }) {
   let allClaimingTxs = []
   const api = await getAPI()
-  const stakerRewards = await api.derive.staking.stakerRewards(senderAddress)
-
-  if (stakerRewards.length === 0) {
-    allClaimingTxs = []
-  } else {
-    stakerRewards.forEach(reward => {
-      const validators = reward.nominating.map(
-        ({ validatorId, validatorIndex }) => [validatorId, validatorIndex]
-      )
-      allClaimingTxs.push(
-        api.tx.staking.payoutNominator(reward.era, validators)
-      )
-    })
+  console.time("rewards")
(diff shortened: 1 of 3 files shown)


src/signing/networkMessages/polkadot-testnet.js
```diff
@@ -42,28 +42,68 @@ export async function StakeTx(senderAddress, { to, amount }, network) {
...
+      allClaimingTxs.push(api.tx.staking.payoutStakers(validators[0], era))
     })
-  }
+  console.timeEnd("rewards")
   console.log(allClaimingTxs)
   if (allClaimingTxs.length === 0) {
     throw new Error("There are no claimable rewards")
   }
-  return await getSignMessage(senderAddress, allClaimingTxs)
+  return await getSignMessage(senderAddress, allClaimingTxs[0])
```
faboweb, line 106:
```suggestion
  return await getSignMessage(senderAddress, allClaimingTxs)
```

Bitcoinera:
> Optimizes queries. But takes the same time as the simple Mario version if you

This was cut out? What did you want to say here?

Always so cryptic 😎

src/signing/networkMessages/polkadot-testnet.js
Bitcoinera: ???????????????

faboweb:
Optimizes the query to be more performant. Reduces the amount of queries against the polkadot API. Maybe not needed. If we allow the user to query for `stakerRewards` why did we then store the rewards in the DB in the first place. Open for discussion.

src/signing/networkMessages/polkadot-testnet.js
faboweb: that was just to test if one single tx works

Bitcoinera:
> Optimizes the query to be more performant. Reduces the amount of queries against the polkadot API. Maybe not needed. If we allow the user to query for `stakerRewards` why did we then store the rewards in the DB in the first place. Open for discussion.

I 100% agree. DB is the way to go for Polkadot, since it itself is not so performant

jbibla:
> Always so cryptic

@faboweb so true

faboweb:
Next workshop: How Fabo can write so the team understands him. :D

Bitcoinera:
> Next workshop: How Fabo can write so the team understands him. :D

Yes please!

PR #3976, opened 2020-05-05
michiel/change client side flow registration push notification

michielmulders:
Change flow client-side using upsert logic from https://github.com/luniehq/lunie-api/pull/678 for push registration.

### DIFF
--- a/src/vuex/modules/pushNotifications.js
+++ b/src/vuex/modules/pushNotifications.js
@@ -28,12 +28,23 @@ const initializeFirebase = async () => {
 }
 
 const askPermissionAndRegister = async (activeNetworks, apollo) => {
-  // Only store for new registrations
   const isDeviceRegistered = localStorage.getItem(
     "registration-push-notifications"
-  ) // "true" if stored
+  ) // "allowed" / "blocked" if stored, null if not set
 
-  if (isDeviceRegistered !== "true") {
+  console.log("device", isDeviceRegistered)
+
+  if (isDeviceRegistered === "blocked") {
+    return
+  }
+
+  if (isDeviceRegistered === "allowed") {
+    const token = localStorage.getItem("registration-push-notifications-token")
+    registerDevice(token, activeNetworks, apollo, true) // non-blocking
+  }
+
+  if (!isDeviceRegistered) {
+    // === null
     try {
       messaging
         .requestPermission()
@@ -47,8 +58,9 @@ const askPermissionAndRegister = async (activeNetworks, apollo) => {
             try {
               await messaging.deleteToken(oldToken)
             } catch (error) {
-              console.log(
-                "bug FCM throws error while deleting token on first refresh"
+              console.error(
+                "bug FCM throws error while deleting token on first refresh",
+                error
               )
             }
           }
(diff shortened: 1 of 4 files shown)


src/vuex/modules/pushNotifications.js
```diff
@@ -66,25 +66,41 @@ const askPermissionAndRegister = async (activeNetworks, apollo) => {
     } catch (error) {
       console.log("Permission denied", error)
     }
+  } else {
+    // upsert
+    const token = localStorage.getItem("registration-push-notifications-token")
+    registerDevice(token, activeNetworks, apollo, true) // non-blocking
```
Bitcoinera, line 72: So the logic is that every time a registered device logs in, the token does in fact change, or why do we need to update these infos?

michielmulders: I've updated the flow a bit as this was not correct.

There are three options now:
1. device not registered === null
2. blocked push notifications === blocked
3. allowed push notifications === allowed

if null -> ask permission
if block -> skip flow
if allowed -> update device registration async when logging in (to detect new addresses so we can also create default topic subscriptions for those addresses/network).

src/vuex/modules/pushNotifications.js
```diff
@@ -57,22 +67,27 @@ const askPermissionAndRegister = async (activeNetworks, apollo) => {
...
           await registerDevice(token, activeNetworks, apollo)
         })
-        .catch(error =>
-          console.log(
-            "bug FCM throws error while deleting token on first refresh",
-            error
-          )
-        )
+        .catch(() => {
+          localStorage.setItem("registration-push-notifications", "blocked")
```
faboweb, line 71: is the only error that can happen, that the user blocked the notification?

Bitcoinera: I made the exact same comment on Slack! 🤣  amazing

Bitcoinera: neyo-1oC7B5v1VzaxJC9KeQ

faboweb: hahahah we already think the same! dangerous :D

Bitcoinera: xD

michielmulders: Sorry to interrupt this ET party 😄 

After some deep deep googling I found that you can check for the code:
1. error.code = blocked => set localstorage
2. other errors? console.error

Good catch to improve error catching, thanks guys! 👍

src/vuex/modules/pushNotifications.js
```diff
@@ -28,51 +28,67 @@ const initializeFirebase = async () => {
...
+
+        // Granted? Store device
+        const token = await messaging.getToken()
+        await registerDevice(token, activeNetworks, apollo)
+      })
+      .catch(error => {
+        if (error.code === "messaging/permission-blocked") {
+          localStorage.setItem("registration-push-notifications", "blocked")
+        } else {
+          console.error("Error occurred during permission request for push notifications", error);
```
faboweb, line 74: add Sentry here so we get notified about issues

PR #3978, opened 2020-05-05
Fabo/polkadot fees

faboweb:
Closes #ISSUE

**Description:**

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/src/ActionModal/components/ActionModal.vue
+++ b/src/ActionModal/components/ActionModal.vue
@@ -247,7 +247,9 @@
                 type="primary"
                 value="Next"
                 :disabled="
-                  disabled || (step === feeStep && $v.invoiceTotal.$invalid)
+                  disabled ||
+                    (step === feeStep && $v.invoiceTotal.$invalid) ||
+                    (step === feeStep && !estimatedFee)
                 "
                 @click.native="validateChangeStep"
               />
--- a/src/ActionModal/components/ActionModal.vue
+++ b/src/ActionModal/components/ActionModal.vue
@@ -249,7 +249,7 @@
                 :disabled="
                   disabled ||
                     (step === feeStep && $v.invoiceTotal.$invalid) ||
-                    (step === feeStep && !estimatedFee)
+                    (step === feeStep && !gasEstimateLoaded)
                 "
                 @click.native="validateChangeStep"
               />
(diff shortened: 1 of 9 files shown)


Bitcoinera:
It works! (just takes a while)



Amazing! Not at all a draft PR

Bitcoinera:
Well, actually we should change the next button for a loader or something like that... just did a quick fix

src/ActionModal/components/ActionModal.vue
```diff
@@ -247,7 +247,9 @@
                 type="primary"
                 value="Next"
                 :disabled="
-                  disabled || (step === feeStep && $v.invoiceTotal.$invalid)
+                  disabled ||
+                    (step === feeStep && $v.invoiceTotal.$invalid) ||
+                    (step === feeStep && !estimatedFee)
```
faboweb, line 252: doesn't work. if estimated fee is 0 this will always block.

faboweb:
```suggestion
                    (step === feeStep && !gasEstimateLoaded)
```

Bitcoinera: oh true!

faboweb:
getPolkadotFees is hard to test because it is actually querying the Polkadot API

Bitcoinera:
What do you think about this "Fetching fees..." message I just added? I think it is better so users understand what is going on:


Bitcoinera:
The loader somehow doesn't work very well with Polkadot. It gets stuck from time to time... Not the best

faboweb:
@jbibla any opinion?

jbibla:
why don't we hardcode a value like always?

faboweb:
I have no idea yet about the Polkadot fees and how dynamic they are and what good values are.

jbibla:
can we make the request for fees on the first step or do we have to wait for the second step?

faboweb:
> can we make the request for fees on the first step or do we have to wait for the second step?

easier to do as is. as we already know the properties of the request which you need for creating the transaction object to do the fees calculation. For Cosmos we could do the request earlier. But I decided to defer to improve speed of loading the actionModel initially.

faboweb:
Needs to be added in the extension as well

faboweb:
I confirmed that we need the transaction details for the fee calculation. We can connect to the API immediately though, which could speed up things.

Bitcoinera:
We could have it as a GraphQL query, is this what you have in mind? `polkadotFees(...transactionDetails)`

faboweb:
I didn't but we could do that as well.

PR #3997, opened 2020-05-07
Ana/3977 link success modal to activity

### LINKED ISSUE #3977: kusama integration feedback
> When I try to login, I need to agree that it’s a testnet address. Kusama shouldn’t be referred to as a testnet, see the following article https://medium.com/polkadot-network/kusama-polkadot-comparing-the-cousins-170e4fe6c280.

> “Waiting for confirmation from polkadot-testnet” should be changed. It’s not a testnet see above ;-).

i guess we'll just call it a mainnet of sorts. let's switch the chain-id to `kusama` or `polkadot-canary` instead of `polkadot-testnet`?

> On Kusama it only takes 7 days to unlock your tokens and not 21 (notification when I stake)

simple fix. let's update the language. ideally, we would pull this from somewhere instead of hardcoding — but for now let's just adjust the message assuming cosmos is 21 and polka/kusama is 7.

> When I click on stake or sending tokens it doesn’t show any transaction fee, but it removes the fee.

bug?

> Staking doesn’t seem to work. “No validators in your portfolio” 

bug again?

> KSM currently has no currency on android, even though it let’s you select Euro, Dollar, etc.

does coingecko have KSM support?

> I think a 1) transaction history or an 2) address book would be super useful.

1) let's change the message on the success screen in the action modal to link to the activity page instead of the block page.

2) one day @michielmulders 

> The recovery feature seems to be loading the address forever and if I just continue with the password and click on create it’s say unreachable.

bug again?

Bitcoinera:
Closes #3977
**Description:**

I think with this one we should have covered all their feedback remarks. Validators are showing for me when I stake and transactions are going through with fees being displayed.

We're good to go!

This is how it looks now:

Thank you! 🚀

---

For contributor:

For reviewer:


jbibla:
what happens if the transactions page isn't implemented yet?

jbibla:
nice! can you shorten the address `cosmos...1234` like we usually do?

Bitcoinera:
> what happens if the transactions page isn't implemented yet?

You just see the Activity page, just like with push notifications

Bitcoinera:
> nice! can you shorten the address `cosmos...1234` like we usually do?

This is what I got for now... I don't know why the `Address` component is not rendering at all 👀

faboweb:
unit tests are failing

jbibla:
> You just see the Activity page

yeah but we don't have this implemented for kusama yet... maybe if we don't have transaction history implemented it should just go back to profile page?

src/ActionModal/components/ActionModal.vue
```diff
@@ -191,12 +191,12 @@
             <div slot="subtitle">
               {{ notifyMessage.body }}
               <br />
-              <br />Block
+              <br />See your transaction
```
jbibla, line 194: sorry to be annoying - but i think this whole thing should be a link "See your transaction" and we don't need the word "here".

Bitcoinera: oh ok, makes sense. Let's change it

Bitcoinera:
> > You just see the Activity page
> 
> yeah but we don't have this implemented for kusama yet... maybe if we don't have transaction history implemented it should just go back to profile page?

Maybe link to Polkascan for now? 😅 or add a "Coming Soon"?

faboweb:
@jbibla can you clarify?

jbibla:
nevermind — if ana can change the link we should merge this!

Bitcoinera:
> nevermind — if ana can change the link we should merge this!

Done!

PR #4007, opened 2020-05-08
Ana/fix my mess polkadot fees in development mode

Bitcoinera:
Closes #ISSUE

**Description:**

I discovered that I disabled the Polkadot fees in a bad way 😅  Next button would be always enabled also in production even without the fees

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/src/ActionModal/components/ActionModal.vue
+++ b/src/ActionModal/components/ActionModal.vue
@@ -463,8 +463,10 @@ export default {
         this.gasEstimateLoaded = true
         return fee
       }
-
-      this.gasEstimateLoaded = true
+      // in development we don't need to worry about fees. Next button won't be disabled
+      if (this.session.developmentMode) {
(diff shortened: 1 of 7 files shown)


src/ActionModal/components/ActionModal.vue
```diff
@@ -463,8 +463,10 @@ export default {
         this.gasEstimateLoaded = true
         return fee
       }
-
-      this.gasEstimateLoaded = true
+      // in development we don't need to worry about fees. Next button won't be disabled
```
jbibla, line 466: if we don't need to worry about fees, why not just set it to true always?

Bitcoinera: because then users will have the "Next" button always enabled. But I see that you saw that

Bitcoinera:
Can we merge this? It is clearly just a timeout, you can check

PR #4010, opened 2020-05-08
Ana/4008 kusama send fees to extension

Bitcoinera:
Closes #ISSUE

**Description:**

Sending fees works, but tests are not xD (trying to fix)

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/src/ActionModal/components/ActionModal.vue
+++ b/src/ActionModal/components/ActionModal.vue
@@ -449,8 +449,9 @@ export default {
       }
 
       if (
-        (this.network.network_type === "polkadot" && this.step === feeStep) ||
-        (this.step === signStep && !this.session.developmentMode)
+        this.network.network_type === "polkadot" &&
+        (this.step === feeStep || this.step === signStep) &&
+        !this.session.developmentMode
       ) {
         const { type, ...message } = this.transactionData
         const fee = await this.transactionManager.getPolkadotFees({
--- a/src/ActionModal/components/ActionModal.vue
+++ b/src/ActionModal/components/ActionModal.vue
@@ -449,9 +449,8 @@ export default {
       }
 
       if (
-        this.network.network_type === "polkadot" &&
-        this.step === feeStep &&
-        !this.session.developmentMode
+        (this.network.network_type === "polkadot" && this.step === feeStep) ||
+        (this.step === signStep && !this.session.developmentMode)
       ) {
         const { type, ...message } = this.transactionData
         const fee = await this.transactionManager.getPolkadotFees({
@@ -470,7 +469,12 @@ export default {
   },
   computed: {
     ...mapState([`extension`, `session`]),
-    ...mapGetters([`connected`, `isExtensionAccount`, `networks`]),
+    ...mapGetters([
+      `connected`,
+      `isExtensionAccount`,
+      `networks`,
+      `currentNetwork`
+    ]),
     ...mapGetters({ networkId: `network` }),
     checkFeatureAvailable() {
       const action = `action_` + this.featureFlag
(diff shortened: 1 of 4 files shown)


src/ActionModal/components/ActionModal.vue
```diff
@@ -756,6 +760,24 @@ export default {
             }
           )
         }
+        if (this.network.network_type === "polkadot") {
+          transactionData = {
+            gasEstimate: this.estimatedFee,
```
faboweb, line 765: why not send the already calculated fees? and then in the extension switch between cosmos and polkadot?

Bitcoinera: well, that is another possibility, but it is basically the same thing no? But yeah, ok, let's do it that way. Maybe this is too hacky

Bitcoinera:
To test it lines 451 to 468 should look like this:

```
      if (
        this.network.network_type === "polkadot" &&
        (this.step === feeStep || this.step === signStep) //&&
        // !this.session.developmentMode
      ) {
        const { type, ...message } = this.transactionData
        const fee = await this.transactionManager.getPolkadotFees({
          messageType: type,
          message,
          senderAddress: this.session.address,
          network: this.network
        })
        this.gasEstimateLoaded = true
        return fee
      }

      // this.gasEstimateLoaded = true
      return 0
```

src/ActionModal/components/ActionModal.vue
faboweb: best way would get the fees directly in the extension as this data you send is only for displaying and not the actual value. But I guess it can be like that for Polkadot as it is just an estimate anyway.

src/ActionModal/components/ActionModal.vue
```diff
@@ -450,7 +450,7 @@ export default {
 
       if (
         this.network.network_type === "polkadot" &&
-        this.step === feeStep &&
+        (this.step === feeStep || this.step === signStep) &&
```
faboweb, line 453: this will trigger now twice and is an expensive process. can we do this just once?

Bitcoinera: I checked and the second time it is triggered it doesn't seem to do anything. Must take it from cache or something like that. But will change it anyway

Bitcoinera:
Let's merge! 🤩

src/ActionModal/components/ActionModal.vue
```diff
@@ -461,6 +462,7 @@ export default {
           network: this.network
         })
         this.gasEstimateLoaded = true
+        this.polkadotFee = fee
```
faboweb, line 465: why? `estimatedFee` is already the fee

src/ActionModal/components/ActionModal.vue
```diff
@@ -758,6 +765,14 @@ export default {
             }
           )
         }
+        if (this.network.network_type === "polkadot") {
+          transactionData = {
+            fee: {
+              amount: this.polkadotFee,
```
faboweb, line 771:
```suggestion
              amount: this.estimatedFee,
```

Bitcoinera: The problem is that in `signStep` this is undefined. This is actually the first thing I tried. And then I had to add `signStep` to the `v-if` because of that

src/ActionModal/components/ActionModal.vue
Bitcoinera: I answered that below

src/ActionModal/components/ActionModal.vue
faboweb: I see. the computed would trigger and return undefined.

PR #4018, opened 2020-05-12
Ana/add kava testnet

Bitcoinera:
Closes #ISSUE

**Description:**

Next time I think I will add the networkMessages file for both mainnet and testnet at the same time. Faster 😉 

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/src/signing/networkMessages/kava-testnet.js
+++ b/src/signing/networkMessages/kava-testnet.js
@@ -1,0 +1,1 @@
+export * from "./cosmos-hub-mainnet.js"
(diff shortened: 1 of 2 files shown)


src/signing/networkMessages/kava-testnet.js
```diff
@@ -0,0 +1 @@
+export * from "./cosmos-hub-mainnet.js"
```
faboweb, line 1: we should simplify and define a `network_message_constructor` property in the networks object. This way we can add the networks without needing to set these in most cases

faboweb: Also reduced client side code

Bitcoinera: oh yeah, good idea!

Issue #4017, opened 2020-05-11
Allow actions based on the account type in Polkadot


faboweb:
In Polkadot there are 2 account types: Stash and Controller. In Lunie we set both to be the same. But certain actions need certain account types. If not signing will fail.

TO DO:
1. PR
- [x] identify account type when logging into a Polkadot network
2. PR
- [ ] if Stash account and not the controller account for that account, disable delegation to validators (keep bonding enabled, disable validator selection)
- [ ] if Stash account and not the controller account for that account, show a warning "For staking to validators you need to sign in with your controller account ACCOUNT HERE. You can still stake tokens."
- [ ] if Controller account show a warning on the portfolio: "this is a controller account, if you want to see your validators, sign in with your stash account."
3. PR
- [ ] check batch message construction to not include the wrong tx

faboweb:
ping @jbibla for wording

Bitcoinera:
> if Stash account, disable delegation to validators (keep bonding enabled, disable validator selection)

How would this go? "Stake" button on the valitador page disabled right

Bitcoinera:
I'm struggling with the first one xD

faboweb:
> > if Stash account, disable delegation to validators (keep bonding enabled, disable validator selection)

> How would this go? "Stake" button on the valitador page disabled right

@jbibla any idea for the UI? I would on a stash account set the validator to undefined.

jbibla:
would the simplest thing be to identify stash accounts and block all staking related transactions? staking modal would pop up and say — "sorry this is stash account - please use a controller account you polka-ninja".

we could get more complex and try to cover all the edge cases but i'd start like this probably.

Bitcoinera:
I love polka-ninja! 😍

faboweb:
@jbibla ping re messaging for PR3
@Bitcoinera can you take care of PR3?

mariopino:
### **Possible cases (updated!)**

- **stash/controller**: same address is controller and stash, all actions allowed.

```
const address = `Ckus4fsdJHxVg69FMDPVbqre3cLUc5R6fWxkMR1HuyQC1Dk` // controller address = stash address
const bond = await api.query.staking.bonded(address)
if (bond === address) {
  console.log(`Address ${address} is stash/controller`)
}
```

- **controller**: address is a controller (!= stash), all actions allowed, currently we can not see the delegations, as they are indexed by stash and we can not derive the stash from controller yet (see NOTE2).

```
[...]

PR #4030, opened 2020-05-13
[WIP] Detect address role

mariopino:
Related  #4017 (will close the first proposed PR)

**Description:**

Detect address role that can be:

- stash (in polkadot type networks)
- controller (in polkadot type networks)
- undefined for the rest of the networks.

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/src/components/common/TmSessionSignIn.vue
+++ b/src/components/common/TmSessionSignIn.vue
@@ -150,7 +150,6 @@ export default {
         // Check address role, useful for Polkadot networks so we can limit actions based on it
         const addressRole = await this.checkAddressRole(this.signInAddress)
 
-        console.log(`address role is`, addressRole)
         await this.$store.dispatch(`setUserAddressRole`, {
           addressRole
         })
--- a/src/components/common/TmSessionSignIn.vue
+++ b/src/components/common/TmSessionSignIn.vue
@@ -76,6 +76,8 @@ const isPolkadotAddress = address => {
   const polkadotRegexp = /^(([0-9a-zA-Z]{47})|([0-9a-zA-Z]{48}))$/
   return polkadotRegexp.test(address)
 }
+import { getAPI } from "../../signing/networkMessages/polkadot-transactions"
+
 export default {
   name: `session-sign-in`,
   components: {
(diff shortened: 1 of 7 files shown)


src/vuex/modules/session.js
```diff
@@ -239,6 +240,10 @@ export default () =>
       setPreferredCurrency({ state, dispatch }, currency) {
         state.preferredCurrency = currency
         dispatch(`storeLocalPreferences`)
+      },
+      setUserAddressRole({ state, dispatch }, addressRole) {
+        state.addressRole = addressRole
+        dispatch(`storeLocalPreferences`)
```
Bitcoinera, line 246: better add this to `persistAddresses`? `storeLocalPreferences` is for users preferences not account specs right

mariopino: Good point! I think better is to remove it as it should not persist in any case, this needs to be checked in every sign in (you can change the controller account at any time, so we need to make sure).

src/components/common/TmSessionSignIn.vue
```diff
@@ -144,6 +146,13 @@ export default {
       })
       if (sessionCorrect) {
         this.selectNetworkByAddress(this.signInAddress)
+
+        // Set address role (stash | controller), useful for Polkadot networks so we can limit actions based on it
+        const addressRole = await this.checkAddressRole(this.signInAddress)
```
faboweb, line 151: move to store?

tests/unit/specs/store/session.spec.js
```diff
@@ -346,6 +353,29 @@ describe(`Module: Session`, () => {
...
+          commit,
+          dispatch,
+          rootState: {
+            connection: { network: "fabo-net" }
+          }
+        },
+        { sessionType, address, networkId: "kusama" }
+      )
+      expect(commit).toHaveBeenCalledWith(`setUserAddressRole`, {
+        addressRole: undefined
```
faboweb, line 376: undefined?

mariopino: nope nope nope!

PR #4034, opened 2020-05-13
Ana/make backtovalidators button smart

Bitcoinera:
@jbibla this one is for you!

It is not as perfect as I would have wished, since it seems there is no straightforward way to capture the last visited page. But I think just changing "Back to Validators" for "Back to previous page" is good enough, I checked responsiveness with the button and looks good.

Now the back button does go back to any page you were before within the app.

Thank you! 🚀

EDIT: OK, I acknowledge it is a bit hacky

---

For contributor:

For reviewer:

### DIFF
--- a/src/components/staking/PageValidator.vue
+++ b/src/components/staking/PageValidator.vue
@@ -12,14 +12,10 @@
       <div class="button-container">
         <button
           class="back-button"
-          @click="
-            $router.push(
-              `/${$router.history.current.params.networkId}/validators`
-            )
-          "
+          @click="$router.go(-1)"
         >
           <i class="material-icons notranslate arrow">arrow_back</i>
-          Back to Validators
+          Back to previous page
         </button>
         <button
           v-if="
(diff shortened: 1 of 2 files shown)


faboweb:
We wanted to scrap the button all together. @jbibla what is your thought?

Bitcoinera:
> We wanted to scrap the button all together. @jbibla what is your thought?

I will try today using Lunie on a Tablet, but I'm afraid that the three dots thing is not so UXish.

We should think on how to make Lunie more navigable imo

faboweb:
Tablet should probably have the same navigation as mobile in portrait and desktop in landscape mode

src/components/staking/PageValidator.vue
```diff
@@ -12,14 +12,10 @@
...
-          @click="
-            $router.push(
-              `/${$router.history.current.params.networkId}/validators`
-            )
-          "
+          @click="$router.go(-1)"
         >
           <i class="material-icons notranslate arrow">arrow_back</i>
-          Back to Validators
+          Back to previous page
```
jbibla, line 18:
```suggestion
          Back
```

jbibla:
this is a good improvement for now. one day i hope we don't need these buttons.

PR #4040, opened 2020-05-14
[WIP] Limit actions based on address role in Polkadot

### LINKED ISSUE #4017: Allow actions based on the account type in Polkadot

mariopino:
Partially closes #4017

**Description:**

Limit actions based on address role in Polkadot:

- [ ] If Stash account, disable delegation to validators (keep bonding enabled, disable validator selection)
- [x] If Stash account, show a warning "For staking to validators you need to sign in with your controller account. You can still send tokens."
- [ ] Check batch message construction to not include the wrong tx

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/src/ActionModal/components/ActionModal.vue
+++ b/src/ActionModal/components/ActionModal.vue
@@ -576,8 +576,8 @@ export default {
     },
     isRestrictedAddressRole() {
       return (
-        this.session.addressRole === "stash" ||
-        this.session.addressRole === "controller"
+        this.session.accountRole === "stash" ||
+        this.session.accountRole === "controller"
       )
     }
   },
--- a/src/ActionModal/components/ActionModal.vue
+++ b/src/ActionModal/components/ActionModal.vue
@@ -18,6 +18,11 @@
           <i class="material-icons notranslate">close</i>
         </div>
         <span class="action-modal-title">{{ title }}</span>
+        <span
+          >Currently you are logged in with a {{ session.addressRole }} account.
+          Only send action is allowed. To perform other actions, please create
+          and sign in with a Lunie account</span
+        >
         <Steps
           v-if="
             [defaultStep, feeStep, signStep].includes(step) &&
@@ -563,6 +568,12 @@ export default {
       // some API responses don't have gasPrices set
       if (!balance.gasPrice) balance.gasPrice = defaultBalance.gasPrice
       return balance
+    },
+    isRestrictedAddressRole() {
+      return (
+        this.session.addressRole === "stash" ||
+        this.session.addressRole === "controller"
+      )
     }
   },
   watch: {
--- a/src/ActionModal/components/ActionModal.vue
+++ b/src/ActionModal/components/ActionModal.vue
@@ -18,11 +18,16 @@
           <i class="material-icons notranslate">close</i>
         </div>
         <span class="action-modal-title">{{ title }}</span>
-        <span
-          >Currently you are logged in with a {{ session.addressRole }} account.
+        <TmFormMsg
+          v-if="isRestrictedAddressRole"
+          :msg="
+            `Currently you are logged in with a ${session.addressRole} account.
           Only send action is allowed. To perform other actions, please create
-          and sign in with a Lunie account</span
-        >
+          and sign in with a Lunie account`
+          "
+          type="custom"
+          class="tm-form-msg--desc"
+        />
         <Steps
           v-if="
             [defaultStep, feeStep, signStep].includes(step) &&
--- a/src/components/common/TmBalance.vue
+++ b/src/components/common/TmBalance.vue
@@ -12,11 +12,16 @@
     <div v-else>
       <div class="header-container">
         <h1>Your Portfolio</h1>
-        <span v-if="isRestrictedAddressRole"
-          >Currently you are logged in with a {{ session.addressRole }} account.
+        <TmFormMsg
+          v-if="isRestrictedAddressRole"
+          :msg="
+            `Currently you are logged in with a ${session.addressRole} account.
           If you want to stake and see your delegations, create and sing in with
-          a Lunie account.</span
-        >
+          a Lunie account.`
+          "
+          type="custom"
+          class="tm-form-msg--desc"
+        />
         <div class="buttons">
           <TmBtn
             class="send-button"
@@ -219,6 +224,7 @@ import { bigFigureOrShortDecimals } from "scripts/num"
 import { noBlanks } from "src/filters"
 import TmBtn from "common/TmBtn"
 import SendModal from "src/ActionModal/components/SendModal"
+import TmFormMsg from "common/TmFormMsg"
 import ModalWithdrawRewards from "src/ActionModal/components/ModalWithdrawRewards"
 import ModalTutorial from "common/ModalTutorial"
 import { mapGetters, mapState } from "vuex"
@@ -231,6 +237,7 @@ export default {
   components: {
     TmBtn,
     SendModal,
+    TmFormMsg,
     ModalWithdrawRewards,
     ModalTutorial
   },
--- a/src/components/common/TmBalance.vue
+++ b/src/components/common/TmBalance.vue
@@ -355,8 +355,8 @@ export default {
     },
     isRestrictedAddressRole() {
       return (
-        this.session.addressRole === "stash" ||
-        this.session.addressRole === "controller"
+        this.session.accountRole === "stash" ||
+        this.session.accountRole === "controller"
       )
     }
   },
--- a/src/components/common/TmBalance.vue
+++ b/src/components/common/TmBalance.vue
@@ -820,6 +820,11 @@ select option {
   .rewards {
     font-size: 12px;
   }
+
+  .tm-form-msg.tm-form-msg--desc {
+    padding-bottom: 1rem;
+    text-align: center;
+  }
 }
 
 @media screen and (min-width: 1254px) {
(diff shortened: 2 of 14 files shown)


mariopino:
I think we have 3 cases here:

- **stash/controller** address: same address is controller and stash, all actions allowed.
- **controller** address: address is a controller, all actions allowed, but we need to change some logic to work in this case, as some queries expect a stash address.
- **stash** address: address is a stash (no controller is set), so it can only send tokens. I Like the simplicity of @jbibla (no staking allowed at this moment, but we can handle in the future).

faboweb:
Don't show this as a maintenance bar. Show it in the action modal.

jbibla:
> For staking to validators you need to sign in with your controller account. You can still send tokens.

"To choose specific validators, you need to use your controller address."

Bitcoinera:
So, if I got it correctly, stash accounts can still stake, only that we need to add this controller account to it, right?

OK, I will see how to make this possible. And after doing that the stash account should be also completely functional?

faboweb:
> So, if I got it correctly, stash accounts can still stake, only that we need to add this controller account to it, right?

Did you read the documentation? https://github.com/luniehq/lunie/issues/4017

src/components/common/TmBalance.vue
```diff
@@ -340,6 +352,12 @@ export default {
     },
     isTestnet() {
       return this.networks.find(network => network.id === this.network).testnet
+    },
+    isRestrictedAddressRole() {
```
faboweb, line 356: why?

src/components/common/TmSessionExtension.vue
```diff
@@ -87,6 +90,31 @@ export default {
         }
       })
     }
+  },
+  apollo: {
+    addressRole: {
+      query: gql`
+        query polkadotSignIn($address: String!) {
```
faboweb, line 97: move to store to avoid douplicate code here and in TmSessionSignIn

faboweb: and TmSessionHardware

src/components/staking/PageValidator.vue
```diff
@@ -308,9 +310,15 @@ export default {
     }
   }),
   computed: {
-    ...mapState([`connection`]),
+    ...mapState([`connection`, `session`]),
     ...mapGetters([`network`, `stakingDenom`]),
-    ...mapGetters({ userAddress: `address` })
+    ...mapGetters({ userAddress: `address` }),
+    isRestrictedAddressRole() {
```
faboweb, line 316: why?

src/vuex/modules/session.js
```diff
@@ -105,6 +104,12 @@ export default () =>
           await commit(`setUserAddresses`, JSON.parse(addresses))
         }
       },
+      async checkForPersistedAddressRole({ commit }) {
```
faboweb, line 107: why?

faboweb: remove and always query?

src/components/wallet/PagePortfolio.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <TmPage :sign-in-required="true" :managed="true" :dark-background="true">
-    <template slot="managed-body">
+    <template v-if="!isRestrictedAddressRole" slot="managed-body">
```
faboweb, line 3: wrong. stash accounts can see the delegations

src/ActionModal/components/ActionModal.vue
```diff
@@ -18,6 +18,16 @@
           <i class="material-icons notranslate">close</i>
         </div>
         <span class="action-modal-title">{{ title }}</span>
+        <TmFormMsg
```
faboweb, line 21: put in the respective action modals: DelegationModal/UndelegationModal. Please read the acceptance criteria in #4017

src/ActionModal/components/ActionModal.vue
```diff
@@ -18,6 +18,16 @@
...
+          v-if="isRestrictedAddressRole"
+          :msg="
+            `Currently you are logged in with a ${session.addressRole} account.
```
faboweb, line 24: No! We allow bonding with a "stash" account and nominating with a "controller" account

faboweb:
This needs to be reworked

src/components/common/TmSessionExtension.vue
Bitcoinera: you mean I use apollo in store right

Bitcoinera: this was actually my first idea, but then I thought it is easier like this

Bitcoinera: I mean, why don't do this for everything else right? We have so many repeated code with `balances`, `rewards` and `delegations` queries for example. We could do the same for all of them. Query in the store. Why not? Maybe because they update much regularly?

Please, explain to me

src/components/staking/PageValidator.vue
Bitcoinera: xD

src/vuex/modules/session.js
Bitcoinera: state does not persist

Bitcoinera: you refresh and boom! restrictions are blown away

src/ActionModal/components/ActionModal.vue
Bitcoinera: I think it is easier like this. That looks like a rabbit hole to me

Bitcoinera: The idea is that people create their Lunie account like I did and pass funds to it. And then we can go on with our much more user-friendly approach of tying up bonding and nominating together.

Or what are you proposing? That is not a rabbit hole

faboweb: Please follow my proposed plan

src/vuex/modules/session.js
```diff
@@ -134,7 +142,7 @@ export default () =>
       },
       async signIn(
         { state, getters: { currentNetwork }, commit, dispatch },
-        { address, sessionType = `ledger`, networkId }
+        { address, sessionType = `ledger`, networkId, addressRole }
```
faboweb, line 145: remove and dispatch a "getPolkadotUserRole" from here

src/vuex/modules/session.js
faboweb: if you query for the role on any sign in it will always be available?

src/components/staking/PageValidator.vue
faboweb: following my instructions the user opens the action modal and has the option to do the action he can do

Bitcoinera: in PageValidator he can do nothing

src/vuex/modules/session.js
Bitcoinera: if we persist it in localStorage yes. Or I don't know if I'm following you

src/ActionModal/components/ActionModal.vue
Bitcoinera: xD

Bitcoinera: I love it when you are in discussing mood ^^

src/components/common/TmSessionExtension.vue
faboweb: yes we need to reduce duplicated code in general. and I am watching out that we don't make the mistakes of the past again.
in this case it doesn't make sense. you are passing the role anyway to the store. so you have 3 places doing the same things but passing it to the same function in the vuex store.

src/vuex/modules/session.js
mariopino: I think we should never persist the role, should be fetched in every sign-in as the user can i.e import the address one day in Lunie, change the controller in Polkadot JS another day, return to Lunie and boom!

src/components/staking/PageValidator.vue
faboweb: a controller can do staking
and he should be able to open the ActionModal in any case

src/vuex/modules/session.js
faboweb: don't store in local storage. run the query every time. we also don't know if the status of the account changes. one might add or change the controller account later.

Bitcoinera: Then run it every time in TmBalance. Still needs to be persisted though. What happens if the user just refreshes the page in PageValidator for instance? I hope I succeed pointing at the issue here

mariopino: Maybe we can query in TmSessionSignIn and store in session?

faboweb: > mean the function signIn in session.js is triggered when the user refresh the page, right?

yes

Bitcoinera:
Closed in favour of #4058

PR #4081, opened 2020-05-19
Ana/add default lunie theme

Bitcoinera:
Closes #ISSUE

**Description:**

Adds a default Lunie theme (lunie-dark) to stay safe and also because it is needed to integrate new networks.

Thank you! 🚀

---

For contributor:

For reviewer:


src/App.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <div id="app" :class="network">
+  <div id="app" :class="theme">
```
faboweb, line 2: @jbibla can we just do this?:

```suggestion
  <div id="app" class="lunie-dark" :class="network">
```

Bitcoinera: I just tried it and it only picks `lunie-dark`

Bitcoinera:
Wow, unit tests were failing from my prior PR with Dawn. How come they didn't fail there?

src/App.vue
```diff
@@ -38,10 +38,16 @@ export default {
   },
   data: () => ({
     isMobileApp: config.mobileApp,
+    networksThemes: ["cosmos", "terra", "emoney", "kava", "kusama"],
```
faboweb, line 41: I am a little worried, that this doesn't scale

Bitcoinera: How to do it then? CSS is quite dumb

Bitcoinera: OK, so what's the plan Stan

faboweb: I talked with @jbibla to experiment with default variables. We could just set default variables (lunie-dark) as a default and overwrite these variables with the network class

Bitcoinera: Sounds good. Unsure how to implement that without a similar logic than what we have in this PR (array with current networks slugs)

faboweb: how about setting .lunie-dark on #app?

Bitcoinera: just tried. It doesn't work

Bitcoinera:
OK, why not do this? I mean, to work with Akash I need this. I think it is also safer to have this logic.

I also think it is scalable. I can change the Array to a "networks" query in App.vue and then it is updated automatically.

How does that sound?

Bitcoinera:
Probably should have done the networks query thing from the beginning actually

Bitcoinera:
Another alternative is to add a new field in the networks Object, `hasTheme`, a Boolean

faboweb:
solution: add .lunie-dark to body

Bitcoinera:
> solution: add .lunie-dark to body

Yes, nice, that was it!

Only some minor CSS issues in lunie-dark... 👁️ 

image

src/components/common/AppMenu.vue
```diff
@@ -94,7 +94,7 @@
       </router-link>
 
       <router-link
-        v-if="session.experimentalMode"
+        v-if="session.developmentMode"
```
faboweb, line 97:
```suggestion
        v-if="session.experimentalMode"
```

PR #4016, opened 2020-05-11
Mario+Fabo+Ana/3996 polkadot unstake

faboweb:
Closes #3996

**Description:**

### WARNING! PLOT TWIST :-)

**Stash can:** 
- bond
- bondExtra

**Controller can:**
- nominate
- unbond (partially/all)

So actions should be:

- [x] Disable Unstake for stash accounts, we  should warn about it in some place.
- [x] No limits for unstake using a controller account

Thank you! 🚀

---

For contributor:

- [ ] Added changes entries. Run `yarn changelog` for a guided process.
- [ ] Reviewed `Files changed` in the github PR explorer
- [ ] Attach screenshots of the UI components on the PR description (if applicable)
- [ ] Scope of work approved for big PRs

For reviewer:

- [ ] Manually tested the changes on the UI

PR #4103, opened 2020-05-21
Ana/fix kusama staketx

Bitcoinera:
Closes #ISSUE

**Description:**

There was something left to do here. We were always replacing nominations adding them to empty array, since we weren't correctly fetching nominees (using controllerId instead of stashId)

Simple fix once you know how to extract stashId from controllerId

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/src/signing/networkMessages/kusama.js
+++ b/src/signing/networkMessages/kusama.js
@@ -43,7 +43,9 @@ export async function StakeTx(
   if (to.length > 0) {
     // only controller addresses can nominate (for not set controllers, we set the controller above)
     if (["controller", "stash/controller", "none"].includes(addressRole)) {
-      const response = await api.query.staking.nominators(senderAddress)
+      const stakingLedger = await api.query.staking.ledger(senderAddress)
+      const stashId = stakingLedger.toJSON().stash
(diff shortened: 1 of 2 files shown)


src/signing/networkMessages/kusama.js
```diff
@@ -43,7 +43,9 @@ export async function StakeTx(
   if (to.length > 0) {
     // only controller addresses can nominate (for not set controllers, we set the controller above)
     if (["controller", "stash/controller", "none"].includes(addressRole)) {
-      const response = await api.query.staking.nominators(senderAddress)
+      const stakingLedger = await api.query.staking.ledger(senderAddress)
```
faboweb, line 46: nice! but better to get this from the API?

faboweb: currently the controller will not show any validators in the first place

Bitcoinera: you merged #4016 with the exact same logic xD

faboweb: I retract. Let's do this for now and later get the validator list from the API

PR #4111, opened 2020-05-22
Ana/add alternative fee logic for emoney

### LINKED ISSUE #4050: Add fee denom selector on feesStep in ActionModal
Many users have requested this and I think it is also something our competitors allow: we should enable a fee selector so users can set their fee.

Some challenges:
- Kava and e-Money don't allow discounts: so going any lower than gasEstimate * gasPrice always results in an error and inability to broadcast.

- Some transactions will result in out of gas errors and users will only notice when they cannot find that transaction. Maybe we should capture this errors in the polling logic of our API and display some message in FE. But probably we'd need notifications working to do this properly

Bitcoinera:
Closes #4050

**Description:**

Tested and everything working great. Good idea @faboweb ! Makes a lot of sense

[This ](https://e-money.network/transactions/C32B25A53586FD00E0FA537B6F8A0EC9A2B2BA2BD182C4E6C4C187E5FB66FD9C)is the "mixed" transaction (other token paying the fees)

The ActionModal on feeStep handling a mixed transaction:

Thank you! 🚀

---

For contributor:

For reviewer:


Bitcoinera:
For Kava it doesn't make sense as they have only one token (for now)

Bitcoinera:
I'm thinking, makes actually sense to do the same for Terra. I tried sending a transaction with "SetMax" and it failed:

`	
TX failed: 23974A35D757000D7460989EE167A049048B217285BCDD644D7762FBEB86A71E Error: out of gas in location: WritePerByte; gasWanted: 43000, gasUsed: 44219`

faboweb:
> TX failed: 23974A35D757000D7460989EE167A049048B217285BCDD644D7762FBEB86A71E Error: out of gas in location: WritePerByte; gasWanted: 43000, gasUsed: 44219

This is out of gas not low fees. Need to increase gasEstimate for that TX

src/ActionModal/components/ActionModal.vue
```diff
@@ -85,8 +85,9 @@
           </TmFormGroup>
           <TableInvoice
             :amount="Number(subTotal)"
-            :estimated-fee="estimatedFee"
+            :estimated-fee="alternativeEstimatedFee || estimatedFee"
             :bond-denom="getDenom"
+            :fee-denom="alternativeFeeDenom || getDenom"
```
faboweb, line 90: instead make fee-denom optional?

src/ActionModal/components/ActionModal.vue
```diff
@@ -85,8 +85,9 @@
           </TmFormGroup>
           <TableInvoice
             :amount="Number(subTotal)"
-            :estimated-fee="estimatedFee"
+            :estimated-fee="alternativeEstimatedFee || estimatedFee"
```
faboweb, line 88: reduce to one variable? `fee`?

src/ActionModal/components/ActionModal.vue
```diff
@@ -830,6 +844,20 @@ export default {
     maxDecimals(value, decimals) {
       return Number(BigNumber(value).toFixed(decimals)) // TODO only use bignumber
     },
+    alternativeFeeSelector() {
+      const alternativeFeeBalance = this.balances.find(
+        ({ denom }) =>
+          denom !== this.getDenom && this.amount >= this.estimatedFee
```
faboweb, line 850: the fee is different in every denom I gues

src/ActionModal/components/ActionModal.vue
```diff
@@ -830,6 +844,20 @@ export default {
     maxDecimals(value, decimals) {
       return Number(BigNumber(value).toFixed(decimals)) // TODO only use bignumber
     },
+    alternativeFeeSelector() {
```
faboweb, line 847: make this a computed

src/ActionModal/components/ActionModal.vue
Bitcoinera: oh good point!

Bitcoinera:
> > TX failed: 23974A35D757000D7460989EE167A049048B217285BCDD644D7762FBEB86A71E Error: out of gas in location: WritePerByte; gasWanted: 43000, gasUsed: 44219
> 
> This is out of gas not low fees. Need to increase gasEstimate for that TX

true!

faboweb:
Total should display like this:

 Total  ...  eCHF
            ...  eNOK

faboweb:
This logic could also move to the api

Bitcoinera:
> Total should display like this:
> 
> Total ... eCHF
> ... eNOK

What do you mean?

src/ActionModal/components/ActionModal.vue
Bitcoinera: actually I just noticed this wasn't needed, as it was already optional

src/ActionModal/components/ActionModal.vue
Bitcoinera: yes, let's do that

src/ActionModal/components/ActionModal.vue
Bitcoinera: I cannot since I overwrite `this.gasPrice` and `this.gasEstimate` in there

Bitcoinera:
> This logic could also move to the api

Could be, but then we need to create the API PR first. It is not costly at all, so I think it is fine to have it in here to start with

faboweb:
> This logic could also move to the api

Sorry I will try to flag out of scope ideas better in the future, sorry

faboweb:
breaks e2e tests

faboweb:
User is somehow not signed in, in this PR in the E2E tests

Bitcoinera:
Let's merge 🤩 (in the second to last commit E2E tests did pass)

src/ActionModal/components/ActionModal.vue
faboweb: which is a bad sign

src/ActionModal/components/TableInvoice.vue
```diff
@@ -1,20 +1,25 @@
...
-          {{ bondDenom }}
+          {{ feeDenom || bondDenom }}
         </span>
       </li>
       <li class="total-row">
         <span>Total</span>
-        <span> {{ total | fullDecimals }} {{ bondDenom }} </span>
+        <div class="total-column">
+          <p>{{ total | fullDecimals }} {{ bondDenom }}</p>
+          <p v-if="feeDenom">
```
faboweb, line 19:
```suggestion
          <p v-if="feeDenom !== bondDenom">
```

src/ActionModal/components/TableInvoice.vue
```diff
@@ -49,7 +58,9 @@ export default {
       return this.amount
     },
     total() {
-      return this.estimatedFee + this.subTotal
+      // if there is a feeDenom, it means that subTotal and estimatedFee are different currencies and
+      // cannot be therefore added up together
+      return this.feeDenom ? this.subTotal : this.estimatedFee + this.subTotal
```
faboweb, line 63:
```suggestion
      return this.feeDenom !== this.bondDenom ? this.subTotal : this.estimatedFee + this.subTotal
```

faboweb:
Small adjustments to do, then let's merge. I think we should deprecate all the logic and put it into the API.

src/ActionModal/components/TableInvoice.vue
Bitcoinera: Shouldn't it rather be `feeDenom && feeDenom !== bondDenom`?

Bitcoinera:
> Small adjustments to do, then let's merge. I think we should deprecate all the logic and put it into the API.

I will open an issue once this is merged, don't worry. It is happening

src/ActionModal/components/TableInvoice.vue
faboweb: feeDenom defaults to `""`

faboweb: yes your right

faboweb:
In theory this code needs to be tested. But as we move this to the API I will ignore that.

Bitcoinera:
> In theory this code needs to be tested. But as we move this to the API I will ignore that.

I tested it all the way 😉 (like I nearly always do)

faboweb:
I meant unit tests

Bitcoinera:
> I meant unit tests

oh true. But yes, not needed since we are moving it right away 👍

PR #4134, opened 2020-05-25
Ana/improve maintenancebar with link and button

Bitcoinera:
Closes a 50% luniehq/lunie-api#734

Needs luniehq/lunie-api#820 but it doesn't break anything either

**Description:**

Looking good:

image

Still could make the button text and the pre-link message also customizable in the DB

I think I'm going to go for it

Thank you! 🚀

---

For contributor:

For reviewer:


src/components/common/Bar.vue
```diff
@@ -2,8 +2,14 @@
   <div v-if="showMessage" :class="`bar ${barType}`">
     <p>
       <slot />
+      <span v-if="link" class="link" @click="goToLink(link)">{{ link }}</span>
```
faboweb, line 5: replace with `router-link`?

src/components/common/MaintenanceBar.vue
```diff
@@ -25,6 +26,7 @@ export default {
         query Maintenance {
           maintenance {
             message
+            link
```
faboweb, line 29: will break if API isn't released first

Bitcoinera: ouch! Missed that one

src/components/common/Bar.vue
Bitcoinera: I don't think it makes that much difference since you still need to do the window.location thing and for that you need a method, which renders the native `to` from `router-link` useless

faboweb:
@jbibla would like your feedback on the two call to action

src/components/common/Bar.vue
faboweb: why do you need `window.location`?

jbibla:
1. the links should have the same font color as the rest of the message in the bar
2. we should not display the actual URL, it should be a link!

Screen Shot 2020-05-26 at 12 58 19 PM

3. let's not worry about buttons for now - it's unneeded complexity

Bitcoinera:
> 1. let's not worry about buttons for now - it's unneeded complexity 🤷‍♂️

Although that button might be the only element that could catch the eye of an absent-minded person like me xD

But it's OK. If nobody signs up for the newsletter, we can add it, go more aggressive

Bitcoinera:
It's looking like this rn:

image

src/components/common/Bar.vue
```diff
@@ -27,6 +39,12 @@ export default {
       this.showMessage = false
       this.$emit(`close`)
     },
+    goToLink(link) {
+      if (link) {
+        // make safe and independent from API
+        window.open(link, "_blank")
```
faboweb, line 45: what if the link it internal?

src/components/common/Bar.vue
```diff
@@ -2,8 +2,16 @@
   <div v-if="showMessage" :class="`bar ${barType}`">
     <p>
       <slot />
+      <span
+        >Come visit
+        <span v-if="link" class="link" @click="goToLink(link)">us!</span>
```
faboweb, line 7: you want to always show "Come visit us!"?

Bitcoinera: This is only in staging Bar ;)

src/components/common/Bar.vue
Bitcoinera: yes, true, I wrongly assumed

faboweb:
We need to add a "linkCaption" param to the API
We will only show a button on the right using the styles: `.button.small`
For now make the default caption of that button "Jump in!"

src/components/common/Bar.vue
Bitcoinera: ah no, you are right. Yesterday I must have been so sleepy 😅

Bitcoinera:
> We need to add a "linkCaption" param to the API
> We will only show a button on the right using the styles: `.button.small`
> For now make the default caption of that button "Learn more"

I think we could have both a "linkCaption" and a "buttonCaption". Wouldn't that be better?

Bitcoinera:
I'm thinking, linkCaption could be actually an Array, same than link. Like it is now for the CookieBar. Or is that overcomplicating?

faboweb:
I think an array is overcomplicating. We will revisit if we actually need that

Bitcoinera:
Good to go!

image

💃

src/components/common/Bar.vue
```diff
@@ -2,13 +2,10 @@
   <div v-if="showMessage" :class="`bar ${barType}`">
     <p>
       <slot />
-      <span v-if="link" class="link" @click="goToLink(link)">{{
-        linkCaption
-      }}</span>
     </p>
     <div class="right">
       <button v-if="link" class="button" @click="goToLink(link)">
```
faboweb, line 7:
```suggestion
      <button v-if="link" class="button small" @click="goToLink(link)">
```

faboweb:
unit test failing

src/components/common/Bar.vue
Bitcoinera: This doesn't change anything 🤔 What styles do you want me to add specifically?

Bitcoinera:
> unit test failing

There was a timeout on a Polkadot thing

faboweb:
missing unit tests for link handling

faboweb:
missing unit tests for link handling

Bitcoinera:
> missing unit tests for link handling

Added 💪

src/components/common/Bar.vue
faboweb: @jbibla that was your spec

src/components/common/Bar.vue
Bitcoinera: because if you do `<router-link to="https://lunie.io`>' and you click it, it will take you to `https://app.lunie.io/<currentNetworkId>/https://lunie.io`

Bitcoinera:
I have a solution for the circle thing, give me a brief inter-PR moment

Bitcoinera:
Let's merge! 💃

src/components/common/Bar.vue
jbibla: if it's always an http full url we can use an `a` tag..?

src/components/common/Bar.vue
jbibla: why doesn't it change anything?

jbibla:
did you change the font color?

Screen Shot 2020-05-29 at 12 06 28 PM

src/components/common/Bar.vue
jbibla: you're not using TmBtn....?

jbibla: you have to use TmBtn it looks like all our other buttons

jbibla:
why does the staging bar have a button now? this is not needed.

faboweb:
> why does the staging bar have a button now? this is not needed.

for testing ^^

Bitcoinera:
> * please make sure the link color is the same color as the font (not the link color from variables)
> * we should default to having no button and the staging bar doesn't need one
> * the button should use lunie button styles

All suggestions have been applied.

This is how it now looks like for the Bar type "info":

image

Also the links have now the same color than the font. And in staging I v-if the button so it doesn't show up

src/components/common/Bar.vue
```diff
@@ -4,9 +4,13 @@
       <slot />
     </p>
     <div class="right">
-      <button v-if="link" class="button" @click="goToLink(link)">
-        {{ linkCaption }}
-      </button>
+      <TmBtn
+        v-if="link && !session.developmentMode"
```
faboweb, line 8:
```suggestion
        v-if="link"
```

Bitcoinera: This is what @jbibla said.

"we should default to having no button and the staging bar doesn't need one"

Or maybe I misunderstood

faboweb: yeah, we can remove the link in the db again

faboweb:
button doesn't do anything

Bitcoinera:
> button doesn't do anything

Ah, the `.native` xD Good catch

PR #4136, opened 2020-05-25
Ana/fix actionmodal between networks

### LINKED ISSUE #4135: ActionModal Multi-issue
Some things seem to be very wrong with ActionModal and should be fixed asap:

- [ ] This one I am not so sure how to reproduce, will investigate:

Together with my e-Money picking feeDenom PR I get this in TableInvoice when the bug kicks in:

- [ ] ActionModal remains open and doesn't update when switching networks (related to the above):

Steps to reproduce: open any ActionModal in any network and then switch using the new slack bar on the left. You will see that the currency is present in all other networks (big bug)

- [ ] Last but not least. Styles that could be improved.

I personally think that ActionModal shouldn't overlap AppMenu when it is full-screen. What do you think?

Bitcoinera:
Closes #4135

**Description:**

All good. All issues fixed.

@jbibla , I think it is better if ActionModal does not overlap AppMenu. It doesn't look too nice

Thank you! 🚀

---

For contributor:

For reviewer:


src/ActionModal/components/SendModal.vue
```diff
@@ -286,7 +286,8 @@ export default {
   },
   methods: {
     open(denom = undefined) {
-      this.selectedToken = denom || this.selectedToken
+      this.selectedToken =
+        denom || this.balances.find(({ denom }) => denom).denom
```
faboweb, line 290: what should this do? `this.balances.find(({ denom }) => denom).denom` this part doesn't do anything

Bitcoinera: yes, it returns the first denom from balances, which makes total sense, because if the user hasn't selected any token yet, then he is going with the first balance (didn't touch the dropdown)

faboweb: ah I get it

PR #4154, opened 2020-05-27
Ana/812 move fee calculation to api

### LINKED ISSUE #4159: Make 'setMaxAmount' in action modals "smart" (discount fees)
Fees are now going to live exclusively in ActionModal.

We need to find a good way to make the `setMaxAmounts` functions in the modals take them into account

Bitcoinera:
Closes luniehq/lunie-api#812
Closes #4159 

**Description:**

Leaving this as WIP for now. Going to QA Kusama

Breaks extension but works with session `local`

Thank you! 🚀

---

For contributor:

For reviewer:


Bitcoinera:
Working well. Currently breaks extension and extension needs luniehq/lunie-browser-extension#264 to work fine with this.

Looking how to make it "friendly" so it doesn't break it 🤔 Or maybe not that important, we can release them together?

Another observation is: I think we need a plan for setMaxAmount. Wouldn't it be possible to have it in ActionModal? As we have now the fees in ActionModal this is what seems sensible

src/ActionModal/components/ActionModal.vue
```diff
@@ -698,11 +700,23 @@ export default {
...
+          )
           transactionData = await this.transactionManager.getCosmosTransactionData(
             {
               memo,
               gasEstimate: this.networkFees.gasEstimate,
-              fee: [this.networkFees.transactionFee],
+              // convert fee to chain values
+              fee: [
+                {
+                  amount: BigNumber(this.networkFees.transactionFee.amount)
```
faboweb, line 714: move the conversion into `getCosmosTransactionData`

src/ActionModal/components/ActionModal.vue
```diff
@@ -511,7 +511,9 @@ export default {
       const feeDenom = this.selectedDenom || this.network.stakingDenom
       let balance = this.balances.find(({ denom }) => denom === feeDenom)
       if (!balance) {
-        balance = defaultBalance
+        // HACK for now. If not balance we hard reload
```
faboweb, line 514: why not return an empty balance here? this gets reevaluated once this.balances updates, right?

Bitcoinera: I was debugging this and this is the only solution I could find for now... I don't know why it happens sometimes. You get the right balances, and then in the feeStep the old ones show up again... it is very strange

Bitcoinera: but with this hack it is safe

Bitcoinera:
E2E tests fail because the fee logic is not in productoin API yet

src/ActionModal/components/ActionModal.vue
faboweb: but it forces a refresh, it's a flag for me

faboweb: How can I reproduce this?

faboweb: This will also refresh the page if the user has no balance in that denom.

Bitcoinera: If it has no balance it cannot be selected in the dropdown in the first place (it doesn't appear)

Bitcoinera: To reproduce: you need to have your accounts signed in in at least two or three networks (I have sessions in all networks, so idk). Then you navigate to portfolio, make a transaction. Navigate to portfolio to another account. Try to make a transaction. You will probably be greeted with a "You don't have enough tokens to proceed" even though you have enough.

The reason is because in selectedBalance it is searching for a denom that only exists in another network

src/ActionModal/components/ActionModal.vue
Bitcoinera: Done

Bitcoinera:
Look at all that code gone! 🤩 !

And everything working! 💃

src/ActionModal/components/ActionModal.vue
faboweb: So something is not resetting correctly when closing/opening the actionmodal? Need to wait for some queries to finish?

Bitcoinera: 🤔 probably something like that, like reopening the modal doesn't refresh the values... 

But still, we don't need the balances query there. Getting it as a prop from `SendModal` and `ModalWithdrawRewards` covers it all

faboweb: passing the balances around creates unwanted dependencies between the parent and child components in my opinion

Bitcoinera: which already exists. We are already sending couple of required props from those "children" (see rewards)

Bitcoinera: but yeah, I get your point. You are concerned about going to the root of the thing: why those balances are not updating in ActionModal (but they do in SendModal)

faboweb: can we revert this change? as the network switching issue seems to come from somewhere else?

Bitcoinera: Ok, let me revert it and put balances back there as an apollo query. But we need to fix this asap, because for users it is annoying (they need to refresh to perform transactions and they will be confused by wrong validation)

Bitcoinera: Done

PR #4174, opened 2020-05-29
Fabo/fix balances not updating

### LINKED ISSUE #4163: Balances from other networks in ActionModal
Very often using the new networks navbar on the left, when we try to make a transaction we will see a wrong validation that doesn't allow us to continue:

The reason for this errors is because the balances in ActionModal don't get properly updated on network change (although they do in SendModal and the rest on the contrary), so `selectedBalance` is searching for a balance with a denom that actually belongs to another network, which returns `defaultBalance` with amount 0 🤷 

The user can only proceed by refreshing the page.

faboweb:
Closes #4163, #4172 

@Bitcoinera check this one please


src/ActionModal/components/ActionModal.vue
```diff
@@ -779,7 +778,7 @@ export default {
       /* istanbul ignore next */
       update(data) {
         this.balancesLoaded = true
-        return data.balances || []
+        return JSON.parse(JSON.stringify(data.balances)) || []
```
Bitcoinera, line 781: This works?? 👁️ I'm going to check

faboweb: Went on a hunch and worked for me. Assumed change is not triggered by the new data somehow.

Bitcoinera: Not working for me.

Could you reproduce the bug?

src/main.js
```diff
@@ -33,6 +34,7 @@ Vue.use(Vuelidate)
 Vue.use(VueClipboard)
 Vue.use(InfiniteScroll)
 Vue.use(AsyncComputed)
+Vue.use(DynamicReactiveRefs)
```
Bitcoinera, line 37: 👀

faboweb: makes the refs reactive so we can trigger logic on $refs.actionModal being available

Bitcoinera:
Seems fixed 👍 Awesome job!

src/main.js
Bitcoinera: thank you!

src/ActionModal/components/UndelegationModal.vue
```diff
@@ -411,7 +411,7 @@ export default {
       `,
       /* istanbul ignore next */
       skip() {
-        return !this.address
+        return !this.address || !this.$refs.actionModal.show
```
Bitcoinera, line 414:
```suggestion
        return !this.address ||  !this.$refs.actionModal || !this.$refs.actionModal.show
```

src/ActionModal/components/UndelegationModal.vue
```diff
@@ -474,7 +474,7 @@ export default {
       },
       /* istanbul ignore next */
       skip() {
-        return !this.address
+        return !this.address || !this.$refs.actionModal.show
```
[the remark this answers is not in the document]
Bitcoinera: Or is there a reason why they are not completely consistent?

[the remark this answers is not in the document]
faboweb: nice find. mussed have missed it

PR #4184, opened 2020-06-01
Ana/use optional chaining for refs.action modal.show

Bitcoinera:
Closes #ISSUE

**Description:**

A new Era has begun...

Fixes this in ValidatorPage:

image

Thank you! 🚀

---

For contributor:

For reviewer:


src/ActionModal/components/UndelegationModal.vue
```diff
@@ -411,7 +411,11 @@ export default {
       `,
       /* istanbul ignore next */
       skip() {
-        return !this.address ||  !this.$refs.actionModal || !this.$refs.actionModal.show
+        return (
+          !this.address ||
+          !this.$refs.actionModal ||
+          !this.$refs.actionModal?.show
```
faboweb, line 417:
```suggestion
          !this.$refs.actionModal?.show
```

the chaining does exactly reduce the need for that line

Bitcoinera: true xDDD

Bitcoinera:
I'd need to recheck, but `yarn install` seems now terribly slow? 🤷

babel.config.js
```diff
@@ -1,11 +1,14 @@
 module.exports = {
   presets: [
     [
-      "@vue/app",
+      "@vue/cli-plugin-babel/preset",
```
jbibla, line 4: what does this do?

jbibla:
@Bitcoinera how does chaining solve this problem?

faboweb:
> @Bitcoinera how does chaining solve this problem?

this doesn't change a problem. this is "refactoring" using an operator to reduce code. The operator is main ECMA script but not yet native in all browsers (already Chrome and others).

jbibla:
> Fixes this in ValidatorPage
?

faboweb:
> yep i see that.

> > Fixes this in ValidatorPage

https://github.com/luniehq/lunie/pull/4184/files#diff-35581415b274540f062489ab1c34e080R491

here is the fix. But it is already fixed on develop

Bitcoinera:
> 
> https://github.com/luniehq/lunie/pull/4184/files#diff-35581415b274540f062489ab1c34e080R491
> 
> here is the fix. But it is already fixed on develop

Should we leave this super mega advancement in Javascript for another time or...

go for it!!!! I can refactor something else. We have been having so many errors due to the same thing

faboweb:
I am very open to doing this! Do we need to switch the babel config? https://github.com/luniehq/lunie/pull/4184/files#diff-fc10a6682269c1fefc2abe05d1500b8aR4

babel.config.js
Bitcoinera: This is an array of babel plugins that are specifically for Vue. [Here ](https://babeljs.io/docs/en/presets)it explains what presets are (array of plugins so you don't need to configure them yourself)

We need this in order to have all the required plugins in our system to use ECMAScript 2020

Bitcoinera:
> I am very open to doing this! Do we need to switch the babel config? https://github.com/luniehq/lunie/pull/4184/files#diff-fc10a6682269c1fefc2abe05d1500b8aR4

It seems so. I found this specific way to do it for Vue [here](https://code.luasoftware.com/tutorials/vuejs/vue-cli-project-enable-optional-chaining/).  

Babel is what is compiling our Javascript so it makes sense we need to change the config so we are able to make our JS "smarter"

Bitcoinera:
Weird that the unit tests worked in the beginning and after merge they don't. Same than why suddenly  the levenary module was missing.

Probably dependabot changed something?

Bitcoinera:
Still another babel module? 🤦

Bitcoinera:
I think something is wrong... so weird. Why it was OK in the beginning and now there so many missing modules. What changed? 🤔
It is like it is not getting all the plugins it needs from @vue/cli-plugin-babel/preset

faboweb:
Let's skip this for now, as it seems to be more complicated than a simple change

Bitcoinera:
> Let's skip this for now, as it seems to be more complicated than a simple change

😓 so weird that it suddenly changed.

Was super excited to have optional chaining!

But hey, it will be much easier in the BE. We can add it there for now 🤩

PR #4200, opened 2020-06-04
Fabo/refactor event list

faboweb:
Refactors the event list with headers and infinity scroll into a component


src/components/common/EventList.vue
```diff
@@ -0,0 +1,117 @@
...
+  }),
+  computed: {
+    showingEvents() {
+      return this.events.slice(0, this.showing)
+    },
+    groupedEvents() {
+      return orderBy(
+        groupBy(this.categorizedEvents, "title"),
+        (group) => group[0].event.timestamp,
+        "desc"
```
michielmulders, line 58: Awesome!

src/components/wallet/PageTransactions.vue
```diff
@@ -126,19 +136,18 @@ const txFields = `
...
+    TransactionItem,
+    EventList,
     DataEmptyTx,
     TmDataMsg,
     TmPage,
   },
   data: () => ({
-    showing: 15,
     pageNumber: 0,
     validators: [],
```
michielmulders, line 147: Not sure, but shouldn't this be an object instead of array as we are assigning the validatorMap?

src/components/common/EventList.vue
```diff
@@ -0,0 +1,117 @@
...
+    title: "Today",
+    matcher: (event) => {
+      return moment(event.timestamp).isSame(moment(), "day")
+    },
+  },
+  {
+    title: "Yesterday",
+    matcher: (event) => {
+      return moment(event.timestamp).isSame(moment().subtract(1, "days"), "day")
+    },
```
Bitcoinera, line 30: Doesn't it make more sense categories like the ones we are already conceptually using throughout the whole app?

I mean, `Governance`, `Staking` and `Transfers`

Bitcoinera: Event is the same thing than notification?

faboweb: this is the same code as before in the transaction list. I just reduced this to be any event that would should in such a segmented list. In the case of the transaction list each transaction is an event.

Bitcoinera:
There seems to be some bug. Checking in Gaia with my address: `cosmos1rfg85mh5pmm5f0m05wmqg5qaknv3u7rfyqp36t`

faboweb:
This component is an event list not a transaction list. It wrapps around list items which can be anything. See PageTransactions for how this is solved for transactions.

faboweb:
The hash issue is in TransactionItem and has nothing to do with this PR.

src/components/common/EventList.vue
Bitcoinera: "In the case of the transaction list each transaction is an event."

? what do we gain here with the event thing? Ah, because you want to reuse the same component for notifications?

Bitcoinera:
> The hash issue is in TransactionItem and has nothing to do with this PR.

Yes, I saw... this must be a very old bug... The only thing that changes is that now we get the navigation duplicated error

faboweb:
What do you mean by navigation duplicated? and does it have to do with this PR?

Bitcoinera:
> What do you mean by navigation duplicated? and does it have to do with this PR?

This error is new (was silenced before, so good it came up):

Bitcoinera:
By the way, I don't know how to dismiss my earlier review of "Request changes". Guess I need to Approve?

src/components/common/EventList.vue
michielmulders: It's reused for notifications, same component

michielmulders:
> > What do you mean by navigation duplicated? and does it have to do with this PR?
> 
> This error is new (was silenced before, so good it came up):
> 
> 
> 
> Ah, it is probably the `v-bind` directive

Isn't this related to the bug (not related to this PR tho) that you see two times the same event, so the item has the same key? Think this is a different bug.

faboweb:
> This error is new (was silenced before, so good it came up):

could replicate. fixing

faboweb:
fixed

PR #4201, opened 2020-06-04
Fabo/api 874 notification pagination

faboweb:
Based on #4200 

Adds pagination to page notifications


src/components/notifications/PageNotifications.vue
```diff
@@ -75,8 +110,11 @@ export default {
         }
       },
       /* istanbul ignore next */
-      update(data) {
-        return data.notifications
+      update(result) {
+        this.dataLoaded = true
+        // assume that when the full page got loaded, that there is more
+        this.moreAvailable = result.notifications.length % 20 === 0
```
michielmulders, line 116: Can you wrap this with `( ... )` , makes it more clear that it's an evaluation (unless the linter removes this)

src/components/wallet/PageTransactions.vue
```diff
@@ -216,7 +214,7 @@ export default {
       },
       update(result) {
         this.dataLoaded = true
-        this.lastLoadedRecordsCount = result.transactionsV2.length
+        this.moreAvailable = result.transactionsV2.length > 0
```
michielmulders, line 217: same here, would rap it with `( ... )` for evaluation

src/components/notifications/PageNotifications.vue
faboweb: do you know the suggestion function in GitHub?
image

faboweb:
```suggestion
        this.moreAvailable = (result.notifications.length % 20 === 0)
```

src/components/wallet/PageTransactions.vue
faboweb:
```suggestion
        this.moreAvailable = (result.transactionsV2.length > 0)
```

Bitcoinera:
OK, now I completely got the idea of these PRs and love it!

But there seems to be a very strange bug here creating a loop inside "Today".

It is like Marmot Day, I cannot exit it. Checking again with `cosmos1rfg85mh5pmm5f0m05wmqg5qaknv3u7rfyqp36t`. Looking at the "Proposal status changed to rejected for Cyberd Takeoff" coming up over and over together with the rest, but this one as reference

Bitcoinera:
OK, I think I found the root for the "Marmot Day" bug: we are getting `events` as a prop, but how do you get more? Because right now it is only receiving 20 events. This is what creates the loop. We need a way to query for older events in EventList

src/components/common/EventList.vue
```diff
@@ -0,0 +1,117 @@
...
+</template>
+
+<script>
+import groupBy from "lodash.groupby"
+import orderBy from "lodash.orderby"
+import moment from "moment"
+
+const categories = [
+  {
+    title: "Today",
```
Bitcoinera, line 21:
```suggestion
    section: "Today",
```

Like this it cannot be mistaken with `event.title`

src/components/common/EventList.vue
```diff
@@ -0,0 +1,117 @@
...
+    matcher: (event) => {
+      return moment(event.timestamp).isSame(moment(), "day")
+    },
+  },
+  {
+    title: "Yesterday",
```
Bitcoinera, line 27:
```suggestion
    section: "Yesterday",
```

src/components/notifications/PageNotifications.vue
```diff
@@ -75,8 +110,11 @@ export default {
         }
       },
       /* istanbul ignore next */
-      update(data) {
-        return data.notifications
+      update(result) {
+        this.dataLoaded = true
+        // assume that when the full page got loaded, that there is more
+        this.moreAvailable = (result.notifications.length % 20 === 0)
+        return result.notifications
```
Bitcoinera, line 117: OK, I was debugging Marmot Day and I see that here it is always returning the same 20 transaction O.o

Debugging more

Bitcoinera: Probably an API thing. Maybe timestamp is not handled correctly

src/components/common/EventList.vue
michielmulders: In favor of this

src/components/notifications/PageNotifications.vue
michielmulders: I tried sending the latest timestamp of the 20 notifications to the API (in the playground) to get the next 20 notifications which was working fine.

faboweb:
hard to test right now

src/components/notifications/PageNotifications.vue
```diff
@@ -12,37 +12,45 @@
...
     >
-      <div class="content">
-        <img :src="notification.icon" />
-        <div>
-          <h3 class="title">{{ notification.title }}</h3>
-        </div>
-      </div>
-      <i class="material-icons notranslate">chevron_right</i>
-    </router-link>
+      <template scope="event">
```
Bitcoinera, line 21:
```suggestion
      <template slot-scope="event">
```

Fixes the "slot-scope" warning. Also needed in PageBlock and PageTransactions

Bitcoinera: Ok, I will push this better

Bitcoinera:
> hard to test right now

Yes.. I just realized. Now we only get notifications if we use staging API, to avoid duplicates. And also it seems the DB was clean yesterday, so not many notifications yet.

But I think after my fix adding the missing `timestamp` variable this should be working fine

Bitcoinera:
Missing changelog

src/components/notifications/PageNotifications.vue
```diff
@@ -68,13 +68,15 @@ export default {
...
+          .timestamp
+        const dateLastTimestamp = new Date(lastTimestamp)
         this.$apollo.queries.notifications.fetchMore({
           // New variables
           variables: {
-            addressObjects: this.session.allSessionAddresses,
             // get notifications that are older then the last one
-            timestamp: this.notifications[this.notifications.length - 1]
-              .timestamp,
+            timestamp: dateLastTimestamp.toISOString(),
```
faboweb, line 78: so the timestamp we send is not an ISOString?

faboweb: this feels like weird logic and should be addressed in the API @michielmulders

Bitcoinera: Yes, I can open an issue for this. It is weird. The timestamps in the DB are this format: 
2020-06-05T21:26:09.548319+00:00. Which looks very similar to ISO but are not exactly ISO (2020-06-06T15:40:23.925Z)

PR #4202, opened 2020-06-04
Fabo/api 874 subscribe to notifications

faboweb:
Based on https://github.com/luniehq/lunie/pull/4201

subscribe to new notifications


src/components/common/EventList.vue
```diff
@@ -0,0 +1,117 @@
...
+<script>
+import groupBy from "lodash.groupby"
+import orderBy from "lodash.orderby"
+import moment from "moment"
+
+const categories = [
+  {
+    title: "Today",
+    matcher: (event) => {
+      return moment(event.timestamp).isSame(moment(), "day")
```
michielmulders, line 23: Would be useful to add a small comment how this works (especially `isSame(moment(), "day")`)

faboweb:
```suggestion
      // tests if the timestamp has the same day as today
      return moment(event.timestamp).isSame(moment(), "day")
```

PR #4211, opened 2020-06-06
Fabo/app ci fixes

faboweb:


faboweb:
🚀 Deployed on https://5edc37ff3846e0d347a23411--lunieio.netlify.app

faboweb:
Working but failing e2e tests don't fail check

.github/workflows/app_test.yml
```diff
@@ -45,5 +45,4 @@ jobs:
       - name: Use Node.js
         uses: actions/setup-node@v1
       - name: Check if changlog was updated
-        run: npx simsala check --pending-path ./changes
-        working-directory: ${{env.WORKING_DIRECTORY}}
+        run: npx simsala check --pending-path ./app/changes
```
Bitcoinera, line 48: Ahh, so this is why it wasn't working? But checked in extension and WORKING_DIRECTORY was right

faboweb: yep, I don't know why it doesn't work with the working directory. But this syntax works.

Bitcoinera: I think it is related with the fact of app always searching for `node_modules` in the root? Like, the `app` folder still "remembers" it was root? 🤔

PR #4215, opened 2020-06-07
Fabo/extension ci fixes

faboweb:


extension/jest.config.js
```diff
@@ -3,29 +3,29 @@ module.exports = {
   verbose: true,
   testMatch: [`<rootDir>/test/unit/**/*spec.js`],
   transform: {
-    '^.+\\.(js|jsx)?$': `<rootDir>/node_modules/babel-jest`,
-    '.*\\.vue$': `<rootDir>/node_modules/vue-jest`
+    '.*\\.vue$': `vue-jest`,
+    '^.+\\.(js|jsx)?$': `babel-jest`
```
Bitcoinera, line 7: What is the idea here? Are we going then to have a global `node_modules`? Isn't it better to preserve modularity?

Using `yarn workspace` it is so easy. I would go for separate node_modules

faboweb: we are already using yarn workspaces

faboweb:
🚀 Deployed on https://5ede6dc4c9be27023e177062--lunieio.netlify.app

extension/jest.config.js
Bitcoinera: But your idea here is to have these two modules shared by all the subrepos right?

faboweb: I guess though

.github/workflows/extension_publish.yml
```diff
@@ -26,12 +26,12 @@ jobs:
...
-          APP_ID: ${{ secrets.EXTENSION_APP_ID }}
+      # - name: Deploy
+      #   run: sh ./scripts/deploy.sh ./dist-zip/lunie-browser-extension.zip
+      #   working-directory: ${{env.WORKING_DIRECTORY}}
+      #   env:
+      #     GAPI_CLIENT_ID: ${{ secrets.GAPI_CLIENT_ID }}
+      #     GAPI_CLIENT_SECRET: ${{ secrets.GAPI_CLIENT_SECRET }}
+      #     GAPI_REFRESH_TOKEN: ${{ secrets.GAPI_REFRESH_TOKEN }}
+      #     ACCESS_TOKEN: ${{ secrets.GAPI_ACCESS_TOKEN }}
+      #     APP_ID: ${{ secrets.EXTENSION_APP_ID }}
```
Bitcoinera, line 37: What is the GAPI_CLIENT? Google API?

faboweb: yep, for publishig the extension

Bitcoinera: Thanks

extension/jest.config.js
Bitcoinera: why make an exception here? Better have them always separate except when needed some meta-module, like in [this example](https://github.com/ferreiro/example-monorepo/blob/master/package.json) with concurrently, no?

To keep the same logic

faboweb: is there any problem right now with the logic?

extension/src/messageHandlers.js
```diff
@@ -5,7 +5,7 @@ const {
 } = require('@lunie/cosmos-keys')
 const {
   default: TransactionManager
-} = require('../lunie/src/signing/transaction-manager')
+} = require('app/src/signing/transaction-manager')
```
Bitcoinera, line 8: This means we don't have the submodule there right? 🤩 

Monorepo all the way! This is such a massive improvement! 💃

extension/jest.config.js
Bitcoinera: well, I would think so by looking at this. But maybe this is the same than with the `rootDir`, just stating `vue-jest`?

PR #4216, opened 2020-06-08
Ana/move dependabot config to root folder and add global lint command

Bitcoinera:
Move dependabot config to root folder, since it won't have any effect in the children directories


Bitcoinera:
Why is it triggering the checkChangelogs in API and extension? I think I did something wrong

Bitcoinera:
Why it says that there are no changes pending? I also do it locally and the same. Strange

Bitcoinera:
Extension testing /test is failing because node-modules are being installed in the root directory 👀

app/changes/ana_move-dependabot-config-to-root-folder
```diff
@@ -0,0 +1 @@
+[Code Improvements] [#4216](https://github.com/cosmos/lunie/pull/4216) Move dependabot config to root folder of monorepo @Bitcoinera
```
faboweb, line 1: why did you copy this in app as well?

app/src/components/notifications/PageNotifications.vue
```diff
@@ -121,7 +121,7 @@ export default {
       update(result) {
         this.dataLoaded = true
         // assume that when the full page got loaded, that there is more
-        this.moreAvailable = (result.notifications.length % 20 === 0)
+        this.moreAvailable = result.notifications.length % 20 === 0
```
faboweb, line 124: @michielmulders linting seems to kill the syntax

package.json
```diff
@@ -15,6 +15,7 @@
   "scripts": {
     "changelog:api": "cd api && yarn changelog",
     "changelog:app": "cd app && yarn changelog",
-    "changelog:extension": "cd extension && yarn changelog"
+    "changelog:extension": "cd extension && yarn changelog",
+    "lint:global": "cd api && yarn lint --fix && cd .. && cd app && yarn lint --fix && cd .. && cd extension && yarn lint:fix && cd .."
```
faboweb, line 19: do we really need this? will you want to lint all repos at once?

app/changes/ana_move-dependabot-config-to-root-folder
faboweb: I see.

> We should have some kind of global changelog 🤔 ? Idk, adding changelogs for all subrepos for now

yep, will think about an update to simsala

faboweb:
This feels wrong:
image

faboweb:
> Extension testing /test is failing because node-modules are being installed in the root directory 👀

extension CI is not finished yet

app/changes/ana_move-dependabot-config-to-root-folder
Bitcoinera: Great stuff! 🤩 Simsala will now be monorepo-compatible

package.json
Bitcoinera: yeah, why not. In global changes

app/changes/ana_move-dependabot-config-to-root-folder
Bitcoinera: egleniyormuyuzcom-cilBKvPHOrWHO7YCAs

package.json
Bitcoinera: How come that this is merged and I don't have it when I do `git pull`?

Bitcoinera: ah you revert it at some point right?

PR #4217, opened 2020-06-08
Ana/balances v2 return 0 balance

Bitcoinera:
Was already merged as luniehq/lunie-api#900, but open here manually since this is now our official repo

### DIFF
--- a/api/.eslintrc.json
+++ b/api/.eslintrc.json
@@ -14,5 +14,6 @@
       "ecmaVersion": 2018
   },
   "rules": {
+    "no-console": "off"
   }
 }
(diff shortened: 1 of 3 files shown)


Bitcoinera:
Here it is good, it just ran the API checks. Why not the same in #4216 ?

Bitcoinera:
Go away CircleCI!

api/.eslintrc.json
```diff
@@ -14,5 +14,6 @@
       "ecmaVersion": 2018
   },
   "rules": {
+    "no-console": "off"
```
faboweb, line 17: we probably should use one eslintrc for all repos

Bitcoinera: yeah, good point!

PR #4222, opened 2020-06-08
Fabo/fix currencies

faboweb:
The currency icon images had a wrong name


faboweb:
🚀 Deployed on https://5ede72c65363d201ba7fa3c5--lunieio.netlify.app

Bitcoinera:
But then the message is wrong no? It still says "This branch has not been deployed"

Where is this message coming from btw? Is this something you added or a new GH feature?

Bitcoinera:
I'd approve, but somehow the Publish on Netlify / e2e failed. But the currencies thing is fixed

PR #4223, opened 2020-06-08
Ana/enable akash testnet

Bitcoinera:
Enable Akash testnet

### DIFF
--- a/app/changes/ana_enable-akash
+++ b/app/changes/ana_enable-akash
@@ -1,0 +1,1 @@
+[Changed] [#4223](https://github.com/cosmos/lunie/pull/4223) Enables Akash testnet @Bitcoinera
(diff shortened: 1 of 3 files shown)


Bitcoinera:
Tested also with extension. Everything working great

Bitcoinera:
There is no faucet yet it seems (the one [here ](https://faucet.akash.vitwit.com/)is for an old testnet), so ping me if you need any AKT

faboweb:
🚀 Deployed on https://5ede96f1c623d7facea978e7--lunieio.netlify.app

app/changes/ana_enable-akash
```diff
@@ -0,0 +1 @@
+[Changed] [#4223](https://github.com/cosmos/lunie/pull/4223) Enables Akash testnet @Bitcoinera
```
faboweb, line 1: wrong folder for the changelog

Bitcoinera: oops, let me fix it

PR #4224, opened 2020-06-08
Ana/move dependabot config to root folder

Bitcoinera:
This went in once, so can go in again 🤩 

Will improve it so it is accepted, but dependabot in the right place and `yarn lint:globar` are good to have


faboweb:
🚀 Deployed on https://5ede9a46c65457f7b6933dd4--lunieio.netlify.app

package.json
```diff
@@ -16,6 +16,6 @@
     "changelog:api": "cd api && yarn changelog",
     "changelog:app": "cd app && yarn changelog",
     "changelog:extension": "cd extension && yarn changelog",
-    "lint:global": "cd api && yarn lint --fix && cd .. && cd app && yarn lint --fix && cd .. && cd extension && yarn lint:fix && cd .."
+    "lint:lunie": "cd api && yarn lint --fix && cd .. && cd app && yarn lint --fix && cd .."
```
faboweb, line 19: 🙈

faboweb: why this change? at least leave it global?

Bitcoinera: because usually we work with API and Lunie Core and it takes a good amount of time linting extension additionally

We can have both of them actually

Bitcoinera:
codecov went a bit crazy, it seems

PR #4226, opened 2020-06-08
Improve polkadot address validation, fix Error: Cannot use GraphQLSchema in frontend

mariopino:
- Improves polkadot address validation by using `checkAddress` fn of `@polkadot/util-crypto`. This fn allows you to check that a polkadot address/prefix tuple is valid. **NOTE**: we can use it to detect the specific network that is associated with that address.
- Fixes `Error: Cannot use GraphQLSchema "[object GraphQLSchema]" from another module or realm` error in frontend by using same graphql version (15.1.0) in both app and api


faboweb:
🚀 Deployed on https://5ede9b1719fecc6398cac717--lunieio.netlify.app

faboweb:
tests failing

mariopino:
> tests failing

Yap, already fixed!

app/tests/unit/specs/store/connection.spec.js
```diff
@@ -5,6 +5,11 @@ jest.mock(`src/../config.js`, () => ({
   network: `keine-ahnungnet`,
 }))
 
+jest.mock(`@polkadot/util-crypto`)
+import { checkAddress } from "@polkadot/util-crypto"
+
+checkAddress.mockReturnValue([true, 'The address is valid'])
```
faboweb, line 11: you're bypassing the actual implementation. why?

mariopino: True! not the best idea ... it was failing for some reason with a correct address.

mariopino: fixed!

app/src/vuex/modules/connection.js
```diff
@@ -79,6 +79,7 @@ export default function ({ apollo }) {
         // just by checking if address is valid for a known prefix
         const addressPrefix = 2 // Kusama
         const isValid = isValidPolkadotAddress(address, addressPrefix)
+        console.log(`isValid:`,isValid)
```
faboweb, line 82:
```suggestion
```

PR #4235, opened 2020-06-09
Ana/control network enable from db

Bitcoinera:
Right now just taking the 'enabled' field from DB and overwriting the one in store. 

Maybe I could store networks just on the first run. Makes much more sense


Bitcoinera:
So the idea of this PR is to give us the control to switch networks on/ off automatically, without having to release anything

Bitcoinera:
Mmm, actually instead of having `insertWithoutPrefix` and `readWithoutPrefix` I should some kind of boolean parameter for this.

OK, will do. Changing to WIP for now

api/lib/database/helpers.js
```diff
@@ -94,6 +94,45 @@ const insert = ({ hasura_url, hasura_admin_key }, upsert = false) => (
   return graphQLQuery({ hasura_url, hasura_admin_key })(query)
 }
 
+const insertWithoutPrefix = (
```
faboweb, line 97: what is this?

faboweb:
I remember we start with a proposal on how to do things. Proposal:
Store all the data from /data/networks.js in a table `networks` without coinLookup and network capabilities.  I would put coinLookup and networkCapabilities in different tables for cleanness using the id as a foreign key. The API polls the network table every minute (not every request to not cause more latency).

api/lib/database/helpers.js
Bitcoinera: Yes, this is repeating too much code. I will instead add a boolean as a parameter for insert, so we can use this function without the `schema _prefix` thing

Bitcoinera:
> I remember we start with a proposal on how to do things. Proposal:
> Store all the data from /data/networks.js in a table `networks` without coinLookup and network capabilities. I would put coinLookup and networkCapabilities in different tables for cleanness using the id as a foreign key. The API polls the network table every minute (not every request to not cause more latency).

Maybe another table for coinLookup is too much?

We could also do something else: store just capabilities, enabled and chaind_id & id in networks

faboweb:
We definitely want the connection properties in the DB as they shouldn't be in the repo.
I thought to put coinLookup in another table as you have multiple coins per network. Therefor the structure is different (multiple rows per network; PKS would be networkId, denom).

Bitcoinera:
> We definitely want the connection properties in the DB as they shouldn't be in the repo.
> I thought to put coinLookup in another table as you have multiple coins per network. Therefor the structure is different (multiple rows per network; PKS would be networkId, denom).

OK, let's do it.

Should I limit the scope of this PR to overwrite just enabled or go for it and overwrite everything?

It is quite simple actually, once you have the logic

api/lib/database/helpers.js
faboweb: I don't get the problem, can you explain?

faboweb:
Do it completely but this is not a high priority task. Please prio everything else before this

Bitcoinera:
Had some trouble with the foreign keys linking `coinLookups` and `networks` via `networkId`. I need to look into that

api/lib/database/helpers.js
```diff
@@ -59,15 +59,17 @@ function getColumns(objects) {
   return Object.keys(obj)
 }
 
-const insert = ({ hasura_url, hasura_admin_key }, upsert = false) => (
-  schema
-) => async (table, rows, height, chainId, returnKeys) => {
+const insert = (
+  { hasura_url, hasura_admin_key },
+  upsert = false,
+  usePrefix = true
```
faboweb, line 65: remove. Just don't send a schema and you get the root schema

api/lib/database/helpers.js
```diff
@@ -98,15 +100,12 @@ const insert = ({ hasura_url, hasura_admin_key }, upsert = false) => (
   return graphQLQuery({ hasura_url, hasura_admin_key })(query)
 }
 
-const read = ({ hasura_url, hasura_admin_key }) => (schema) => async (
-  table,
-  queryName,
-  keys,
-  filter
-) => {
+const read = ({ hasura_url, hasura_admin_key }, usePrefix = true) => (
```
faboweb, line 103: same

api/lib/database/methods.js
```diff
@@ -95,6 +95,93 @@ const getNotifications = ({ hasura_url, hasura_admin_key }) => (
   )
 }
 
+const getNetwork = ({ hasura_url, hasura_admin_key }) => () => async (id) => {
+  return await read(
```
faboweb, line 99: here you should return the first result already so you don't need to do it in the rest of the code

api/lib/stores/block-store.js
```diff
@@ -278,4 +325,56 @@ function enrichValidator(validatorInfo, validator) {
   }
 }
 
+function formatNetworkForDB(network) {
+  // removed enabled
+  return {
+    network: {
```
faboweb, line 331: why did you choose this nested structure? is the readable in the db?

api/lib/block-listeners/polkadot-node-subscription.js
```diff
@@ -223,6 +236,13 @@ class PolkadotNodeSubscription {
       Sentry.captureException(error)
     }
   }
+
+  async storeNetworkInDB() {
+    // only run at startup
+    await this.store.storeNetwork(this.store.network)
```
faboweb, line 242: why are we storing the network in the db? doesn't it need to be in there already at startup?

api/lib/stores/block-store.js
Bitcoinera: I chose it so I can be able to divide it in three different tables inside the `storeNetwork` function, using `payload[network]` for example

api/lib/database/helpers.js
Bitcoinera: oh, good to know!

api/lib/database/methods.js
Bitcoinera: what do you mean? You mean so I don't need to declare every and each field in the table?

api/lib/block-listeners/polkadot-node-subscription.js
Bitcoinera: It is a good idea. Imagine developing and adding a new network. The thing is that this part is automatic, storing the network in the DB. You only need to fill the network specs in data, and the rest will be automatically handled. And then, if we need to change something on the go, we do it simply using the DB

faboweb: that doesn't seem right. we would overwrite the persisted data. let's not manage the data twice. for development you could have a local networks.js file which adds development networks.

api/lib/database/helpers.js
Bitcoinera: There is an issue here. The DB instance is already initialized with `network.id` as its schema (network-container.js, line 22, `this.db = createDBInstance(network.id)`. So how to go around this? Create a new instance for this?

api/lib/stores/block-store.js
```diff
@@ -264,6 +264,53 @@ class BlockStore {
...
+    }
+  }
+
+  async updateNetworkInDB() {
+    try {
+      const storedNetwork = await this.db.getNetwork(this.network.id)
+      if (storedNetwork.length > 0) {
+        // getNetwork only returns one network, so it is safe to do storedNetwork[0]
+        // we update everything except the network id and powered
+        this.network.enabled = storedNetwork[0].enabled
```
faboweb, line 286: here you use `[0]` every time. I think `getNetwork` should return an object not an array to avoid code duplication

api/lib/stores/block-store.js
faboweb: I propose removing this function from the PR and using it once to fill the DB with the existing data we have so we don't need to type in every data point.

api/lib/stores/block-store.js
Bitcoinera: true true

api/lib/block-listeners/polkadot-node-subscription.js
Bitcoinera: OK, I see what you mean

api/lib/stores/block-store.js
Bitcoinera: ?? This got me kind of lost? How to use it once if we remove it? Ah, you mean moving it in that `networks.js` file you said below to use in development

Bitcoinera: right?

faboweb: no, removing it from the API code and only run it as a script once

api/lib/database/methods.js
faboweb: see this comment https://github.com/luniehq/lunie/pull/4235#discussion_r444338429

api/lib/stores/block-store.js
Bitcoinera: Wait, but scripts are also part of the API code :sweat_smile: OK, but I see what you mean

Bitcoinera: Not even have the call in the block-store, but have it as something completely independent. This is it, right

Bitcoinera:
Ready for final review.

I set the foreign key with "no action" in staging DB. Waiting on a final word on that to copy the table structure to production

api/data/network-coinlookups.js
```diff
@@ -9,11 +9,6 @@ const terraCoinLookup = [
     viewDenom: 'KRT',
     chainToViewConversionFactor: 1e-6
   },
-  {
```
faboweb, line 12: why did this one got deleted?

api/lib/stores/block-store.js
```diff
@@ -264,6 +266,68 @@ class BlockStore {
...
     )
   }
+
+  async updateNetworkInDB() {
+    try {
+      const storedNetwork = await database(config)('').getNetwork(
+        this.network.id
+      )
+      if (storedNetwork) {
+        // we update everything except the network id and powered
```
faboweb, line 276: why are we not updating those 2? why are we so explicit about all the values? can't we just do ```Object.assign(this.network, storedNetwork)```?

faboweb: powered should come from the DB, right?

api/lib/stores/block-store.js
```diff
@@ -264,6 +266,68 @@ class BlockStore {
       'operatorAddress'
     )
   }
+
+  async updateNetworkInDB() {
```
faboweb, line 270: why are we updating all the networks individually? can't we update all of them at once?

Bitcoinera: In store they are coming one at a time. With its validators and so on. I just update them as they come

api/data/network-coinlookups.js
Bitcoinera: It was a duplicate and destroying the index as a valid key-value reference

api/lib/stores/block-store.js
Bitcoinera: if we can update networkId, many things would fail. Static things such as the themes in FE. It is not a good idea to do it. Also the messageCreator. So it would be impossible to create transactions. This is based on file names and would get automatically broken

Bitcoinera: About powered, yes, we could do it. It is just a pain in the ass to edit it because it is in this format: `"{&quot;name&quot;:&quot;Forbole&quot;,&quot;providerAddress&quot;:&quot;kavavaloper14kn0kk33szpwus9nh8n87fjel8djx0y02c7me3&quot;,&quot;picture&quot;:&quot;https://lunie.fra1.digitaloceanspaces.com/validator-pictures/forbole.jpeg&quot;}"` and you need to be super careful with characters here. One little character extra and gone after editing would mean deserialization broken and error

Bitcoinera: We can seat and we can talk why this escape thing seems necessary. Perhaps there is a workaround, but I can confirm that directly sending stringified JSONs won't work (it is a quote `"` inside `"` problem, but then others quotes (` ') won't be parsed like in JS)

Bitcoinera: we could remove powered and id with `delete` and then we could do Object.assign. Or we could also have another table just for powered.

Bitcoinera: Waiting on instructions :v:

api/lib/stores/block-store.js
faboweb:
```suggestion
  async updateNetworkFromDB() {
```

api/lib/block-listeners/polkadot-node-subscription.js
```diff
@@ -143,6 +154,9 @@ class PolkadotNodeSubscription {
       const block = await this.polkadotAPI.getBlockByHeightV2(blockHeight)
       this.enqueueAndPublishBlockAdded(block)
 
+      // gives us the control to modify network parameters
+      this.store.updateNetworkInDB()
```
faboweb, line 158:
```suggestion
      this.store.updateNetworkFromDB()
```

api/lib/block-listeners/polkadot-node-subscription.js
```diff
@@ -131,6 +134,14 @@ class PolkadotNodeSubscription {
     }, POLLING_INTERVAL)
   }
 
+  async pollForUpdateNetworks() {
+    // gives us the control to modify network parameters
+    this.store.updateNetworkInDB()
```
faboweb, line 139:
```suggestion
    this.store.updateNetworkFromDB()
```

api/lib/block-listeners/cosmos-node-subscription.js
```diff
@@ -50,6 +53,14 @@ class CosmosNodeSubscription {
     }, PROPOSAL_POLLING_INTERVAL)
   }
 
+  async pollForUpdateNetworks() {
+    // gives us the control to modify network parameters
+    this.store.updateNetworkInDB()
```
faboweb, line 58:
```suggestion
    this.store.updateNetworkFromDB()
```

faboweb:
the initial set of networks should already come from the db

api/lib/database/methods.js
```diff
@@ -95,6 +95,131 @@ const getNotifications = ({ hasura_url, hasura_admin_key }) => (
   )
 }
 
+const getNetwork = ({ hasura_url, hasura_admin_key }) => () => async (id) => {
+  const storedNetwork = await read({
```
faboweb, line 99: you can run all 3 reads in parallel. but better use `query` and specify all 3 requests in 1 query.

faboweb:
Removed networks from code and query them on start. Fees should also go to db in the future

Bitcoinera:
> Removed networks from code and query them on start. Fees should also go to db in the future

But it is impossible to have dynamic fees like Terra's or Substrate in DB :thinking: Or what is your idea here?

faboweb:
Ah, I get it. Well dynamic fees should still be like they are. But all fixed fees should go into the db.

Bitcoinera:
> Ah, I get it. Well dynamic fees should still be like they are. But all fixed fees should go into the db.

Cool, I will open an issue about that

PR #4246, opened 2020-06-10
Validator list ux improvements

### LINKED ISSUE #3894: validator list UX improvements

- our validator list is an important page for our users to view, search, and assess the staking opportunities on a network
- there are many anonymous validators (no moniker, just address) and many without images which provide no information for users in a UI product. this reflects poorly on us and makes the list we show seem complex and unapproachable. 
- right now, we default to sort by rewards - but the top 20 or so all have the same rewards (0% commission). this means the above-the-fold view of our list includes a bunch of anonymous and undifferentiated validators.

AC:
- validators without profile pictures are not visible by default on Lunie
- on mobile you can sort validators by any visible property

- - [ ] API: let's remove them anonyomous validators from the "active" list and put make them visible with the "all" button.
~~- FE: we should make the "contact us" button more prominent for validators that want to add a name / update an image including adding it to the li-validator component (if missing an image or a name)~~
~~- we should sort the list by something that is special to Lunie — like popularity. this would give users a reason to check the list frequently to see which validators are trending and could incentivize validators to send people to Lunie (to improve their position in the list)~~
- - [ ] FE: on mobile, let users sort the list by more than 1 detail (popularity, voting power, yield)

**Additional context**

need to implement https://github.com/luniehq/lunie-api/issues/615

mariopino:
Closes #3894 

- [x] FE: on mobile, let users sort the list by more than 1 detail (popularity, voting power, yield). 

image


faboweb:
🚀 Deployed on https://5ee0b9175cf3ef867e3b77ff--lunieio.netlify.app

mariopino:
Design:

image

Bitcoinera:
> Design:

Popular is not in the design :thinking:  Is this how it is supposed to be? :eyes:

mariopino:
> > Design:
> 
> Popular is not in the design 🤔 Is this how it is supposed to be? 👀

I think we can keep the three-option toggle. This is what you mean?

faboweb:
> I think we can keep the three-option toggle. This is what you mean?

the three side by ide might be a problem

Bitcoinera:
This looks a bit weird, wouldn't it be better to left-align it?

image

app/src/components/staking/PageValidators.vue
```diff
@@ -138,12 +150,23 @@ export default {
...
+
+        return this.activeOnly
+          ? result.validators.filter(
+              ({ name, operatorAddress }) => name !== operatorAddress
+            )
+          : result.validators
       },
     },
     delegations: {
+      /* istanbul ignore next */
```
Bitcoinera, line 169:
```suggestion
      /* istanbul ignore next */
    delegations: {
```

app/tests/unit/specs/components/staking/TableValidators.spec.js
```diff
@@ -42,7 +42,10 @@ describe(`TableValidators`, () => {
     }
 
     wrapper = shallowMount(TableValidators, {
-      propsData: { validators },
+      propsData: {
+        validators,
+        showMobileSorting: true,
```
Bitcoinera, line 47: this is always set to true and therefore it is always visible, also on Desktop. Shouldn't it depend on the screen size?

faboweb:
Needs a merge of develop
Hide "Popular" radio button on mobile
Hide mobile filters on desktop

app/src/components/staking/PageValidators.vue
```diff
@@ -11,36 +11,48 @@
...
+              @click.native="defaultSelectorsController(`allValidators`)"
+            />
+            <TmBtn
+              value="Active"
+              class="btn-radio secondary"
+              :type="activeOnly ? `active` : `secondary`"
+              @click.native="defaultSelectorsController(`activeOnly`)"
+            />
+          </div>
+          <div class="show-mobile-sorting">
```
Bitcoinera, line 41: Is the idea that this is always visible or only on mobile?

Bitcoinera:
Little bug. Maybe the show-mobile-selector-options thing should only be visible on mobile?

image

Bitcoinera:
> Hide mobile filters on desktop

It would suffice hiding this: https://github.com/luniehq/lunie/pull/4246#discussion_r441694175

jbibla:
some thoughts about this (apologies in advance that these decisions weren't addressed before you started @mariopino @Bitcoinera)

- popular is a sorting function not a filter - as such it should not be part of the "active / all" button row
- charlie's design is good but it should be "active / all" not "active / inactive"
- from tablet size and up the dropdown _sorting_ functions should be in a dropdown on the right side close to the button (we can use the popover tooltip if that's helpful) - not floating full width

@mariopino did you try ionic components?

faboweb:
> popular is a sorting function not a filter - as such it should not be part of the "active / all" button row

that is not true. popular is a filtering function. it filters out all validators without identity

faboweb:
> @mariopino did you try ionic components?

little late? let's not do such a radical refactor on this PR

faboweb:
> charlie's design is good but it should be "active / all" not "active / inactive"

for another PR: maybe we should actually do "active / inactive" ? so the difference is clearer?

faboweb:
unit tests fail

app/src/components/staking/PageValidators.vue
mariopino: Nope, now it shows only in > 768px wide screens.

faboweb:
e2e fails

mariopino:
> e2e fails

mm weird resetting the search works in frontend. I triggered the job again and get a different error, I think e2e errors are related with the slow response of the api/db.

Bitcoinera:
@faboweb can you please merge? Two codecovs errors doesn't allow me

PR #4252, opened 2020-06-10
Ana/kava 3 upgrade

Bitcoinera:
All transactions failing with this error, even though hard-coded kava-3 chain_id again...

Fixed. It was just an endpoint issue


Bitcoinera:
OK, Kwun restarted the LCD and that was it it seems. Got kind of hung up at chain start. Now everything is just working fine again. Tested different transactions

Bitcoinera:
Let's merge!

PR #4266, opened 2020-06-11
Add polkadot testnet network (Westend)

### LINKED ISSUE #4248: Add Westend polkadot testnet network
Add westend network support to freely test polkadot.

**Denom:**

**Westie (WND)**
ss58 format: 42
Decimals: 12

**Public node:**

wss://westend-rpc.polkadot.io

**Faucet:**

The current testnet is called Westend and you can obtain its native tokens by posting !drip <WESTEND_ADDRESS> in the Riot chatroom #westend_faucet:matrix.org.

**Telemetry:**

https://telemetry.polkadot.io/#list/Westend

**Reference:**

https://polkadot.network/westend-introducing-a-new-testnet-for-polkadot-and-kusama/

mariopino:
Closes #4248 

Need to test send/stake/unstaking/all_the_things but fauce seems not operative, still waiting for my fresh WND tokens to test.

**Faucet**

The current testnet is called Westend and you can obtain its native tokens by posting !drip <WESTEND_ADDRESS> in the Riot chatroom #westend_faucet:matrix.org.


faboweb:
🚀 Deployed on https://5ee1e88f1845d392a3697505--lunieio.netlify.app

mariopino:
Tested Send transaction and it fails! :-/

```
2020-06-11 17:40:27        RPC-CORE: submitExtrinsic(extrinsic: Extrinsic): Hash:: 1010: Invalid Transaction: Inability to pay some fees (e.g. account balance too low)
```

I have enough balance (0.3 WND) and was trying to send 0,000001 sooo time to debug!

api/data/networks.js
```diff
@@ -222,6 +222,32 @@ module.exports = [
     slug: 'polkadot',
     lockUpPeriod: '28 days'
   },
+  {
+    id: 'westend',
```
faboweb, line 226: polkadot-testnet?

mariopino:
Found type not found errors:

```
Unknown types found, no types for Status
```

Upgrading polkadot-api to v1.19.0-beta.3 fixes them!

api/lib/controller/transaction/index.js
```diff
@@ -7,21 +7,54 @@ const networks = require('../../../data/networks')
 
 global.fetch = require('node-fetch')
 
-let api
-async function initPolkadotAPI() {
+let kusamaAPI, polkadotAPI, polkadotTestnetAPI
```
faboweb, line 10: 🤮 
there is no data structure that scales? like an object keyed by the networkId?

app/src/components/common/TmSessionSignIn.vue
```diff
@@ -72,11 +72,17 @@ import TmField from "common/TmField"
 import TmFormMsg from "common/TmFormMsg"
 import TmFormStruct from "common/TmFormStruct"
 import SessionFrame from "common/SessionFrame"
+import { checkAddress } from "@polkadot/util-crypto"
+
 const isPolkadotAddress = (address) => {
```
faboweb, line 77: is this still used?

app/src/signing/networkMessages/polkadot-testnet-transactions.js
```diff
@@ -0,0 +1,110 @@
+/*
```
faboweb, line 1: this file is duplicating a lot of code. can we refactor this?

app/src/signing/networkMessages/polkadot-testnet.js
```diff
@@ -0,0 +1,161 @@
+import { getSignMessage, getAPI } from "./polkadot-testnet-transactions"
```
faboweb, line 1: same here. can we refactor this?

app/src/signing/networkMessages/polkadot.js
```diff
@@ -0,0 +1,161 @@
+import { getSignMessage, getAPI } from "./polkadot-transactions"
```
faboweb, line 1: same

api/lib/controller/transaction/index.js
Bitcoinera: :sweat_smile:  the vomit emoji

Bitcoinera:
Still not working

app/src/components/common/TmSessionSignIn.vue
faboweb: ok let's leave in for now

app/src/gql/index.js
```diff
@@ -136,6 +136,7 @@ export const NetworksAll = gql`
         viewDenom
         chainToViewConversionFactor
       }
+      rpc_url
```
faboweb, line 139: hmm doesn't work as the rpc usually is whitelisted to the BE only. let's create a dictionary in the FE for those?

app/src/signing/networkMessages/polkadot-transactions.js
```diff
@@ -34,6 +44,7 @@ export async function getAPI(endpoint) {
 // }
 
 export async function multiMessage(transactions) {
+  const polkadotNetworks = await getPolkadotNetworks()
```
faboweb, line 47: this would be called every time. why not just get the normal networks list?

faboweb:
Consider that the extension will also need to provide the api url now

app/src/gql/index.js
Bitcoinera: Ah no sorry, I didn't update, it is working now

app/src/signing/networkMessages/polkadot-transactions.js
Bitcoinera: from where? There is no normal networks list, only in the store's state... But not sure how that would work out here

faboweb: we only need the rpc_urls right? let's go with just a dictionary for the rpc_urls

app/src/gql/index.js
faboweb: no I mean we need to use the public polkadot endpoints in the FE

app/src/signing/networkMessages/polkadot-transactions.js
Bitcoinera: but we need to update them no? What happens if they suddenly change? Then we need to remember updating them here too

app/src/gql/index.js
Bitcoinera: See my comment below: https://github.com/luniehq/lunie/pull/4266#discussion_r444560139

app/src/signing/networkMessages/polkadot-transactions.js
faboweb: True! We could pass public rpc urls independent from our internal rpc urls. But send both from the Api

Bitcoinera: You mean creating a query for rpcUrls specifically? Isn't is simpler to just stick to the current networks query?

faboweb: No I mean returning the public endpoints from the API for the networks query. We could define a property in the networks "public_rpc_endpoint" and only return that.

api/lib/schema.js
```diff
@@ -133,7 +133,7 @@ const typeDefs = gql`
     id: String
     title: String
     chain_id: String
-    rpc_url: String
+    public_rpc_url: String
```
Bitcoinera, line 136: This is wrong right. Why do you delete `rpc_url` here. Now you get the following error:

image

api/tests/network-configs.test.js
```diff
@@ -11,7 +11,7 @@ const schema = Joi.object({
   id: Joi.string().lowercase(),
   title: Joi.string(),
   chain_id: Joi.string(),
-  rpc_url: Joi.string().uri(),
+  public_rpc_url: Joi.string().uri(),
```
Bitcoinera: Same than above

extension/src/popup/gql/index.js
```diff
@@ -15,7 +15,7 @@ export const Networks = gql`
       chain_id
       testnet
       title
-      rpc_url
+      public_rpc_url
```
Bitcoinera, line 18: To be honest, I still don't quite understand the why for having `public_rpc_url`, but it's OK

app/src/signing/networkMessages/polkadot-transactions.js
Bitcoinera: But why not just return the rpc_urls? I think I'm missing something here. Will keep looking at the commits

api/lib/schema.js
faboweb: let's revert I changed it actually to return the public url

api/tests/network-configs.test.js
faboweb: also revert

faboweb: ah no, we want both here

extension/src/popup/gql/index.js
faboweb: the internal rpc url is blocked for users.

Bitcoinera: oh cool

api/tests/network-configs.test.js
Bitcoinera: OK, on it

api/lib/resolvers.js
```diff
@@ -340,16 +340,10 @@ const resolvers = {
...
-            // access the docker container from the outside via its port)
-            return {
-              ...network,
-              api_url: 'http://localhost:9071'
-            }
-          }
-          return network
+          return Object.assign({}, network, {
+            rpc_url: network.public_rpc_url,
+            public_rpc_url: undefined
```
Bitcoinera, line 345: Why this? This is what is making get `null` in the FE for `public_rpc_url`. What was your idea here?

faboweb: the idea is to only output `rpc_url` which is the public url. the frontend doesn't need to know about the private url

faboweb: need to adjust in polkadot-transactions line 32

Bitcoinera: ok I see

Bitcoinera: Shouldn't we do the same for the single network query?

PR #4274, opened 2020-06-12
Ana/add github templates and other gh docs

### LINKED ISSUE #4271: Add GitHub templates to monorepo
I just wanted to file a "Feature Request" but noticed that is not possible anymore as we don't have any `ISSUE_TEMPLATE` folder in our `.github` folder anymore... only CI files

Bitcoinera:
Closes #4271 

Adds the missing templates to our `.github` folder in the monorepo root directory, as well as some other docs like `CONTRIBUTING.md`, `pull_request_template.md` and so on...

I also added the `CODEOWNERS` file, which means @faboweb and @jbibla will be now automatically assigned as reviewers for each PR, but I can remove it if it is too much noise. What should I do?

Thank you 🚀 ! (this one was manually added)

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,8 @@
+# This is a comment.
+# Each line is a file pattern followed by one or more owners.
+
+# These owners will be the default owners for everything in
+# the repo. Unless a later match takes precedence,
+# @global-owner1 and @global-owner2 will be requested for
+# review when someone opens a pull request.
+*       @faboweb @jbibla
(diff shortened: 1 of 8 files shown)


Bitcoinera:
Here I guess I shouldn't add any changelog right? It is totally a root directory change, no inner repo changed

.github/CODEOWNERS
```diff
@@ -0,0 +1,8 @@
+# This is a comment.
+# Each line is a file pattern followed by one or more owners.
+
+# These owners will be the default owners for everything in
+# the repo. Unless a later match takes precedence,
+# @global-owner1 and @global-owner2 will be requested for
+# review when someone opens a pull request.
+*       @faboweb @jbibla
```
faboweb, line 8:
```suggestion
*       @faboweb @jbibla @Bitcoinera 
```

PR #4281, opened 2020-06-15
Ana/4280 optimize validator search

### LINKED ISSUE #4280: Validator search goes extremely slow
Testing in https://app.lunie.io validator search is going so slow now. It takes over 20 second or so to get the results... Could have to do with the popularity change? Maybe that is not optimized. Will check.

image

Bitcoinera:
Closes #4280

**Description:**

Optimizes validator search and validators handling in general in the block_listener. Now they are not queries every block but every 10 min.

Validator search goes much faster with this in

Thank you! 🚀

---

For contributor:

For reviewer:


api/lib/block-listeners/cosmos-node-subscription.js
```diff
@@ -51,6 +53,25 @@ class CosmosNodeSubscription {
     }, PROPOSAL_POLLING_INTERVAL)
   }
 
+  async updateValidators() {
```
faboweb, line 56: I don't really like the solution as it is only Cosmos. Can we have a solution that works for all networks?

faboweb: You can have the same logic of querying the database to get the popularity in the store. The store is used by every source the same way.

Bitcoinera: yes, sorry :facepalm:  At some point I forgot about Polkadot

Bitcoinera:
Maybe it would be good to trigger the app test too so we can see if E2E finally pass. Is making some change in the app the only way or can we do this at will too?

faboweb:
🚀 Deployed on https://5ee7abd61d0fd50266dd226e--lunieio.netlify.app

Bitcoinera:
Nooo, still failing with the same thing :cry:

Bitcoinera:
Wait, or does the E2E use another API instance?? I need to check that

Bitcoinera:
OK, let's merge this one then and bring some speed to validator search! :woman_dancing:

api/lib/block-listeners/cosmos-node-subscription.js
```diff
@@ -17,6 +17,7 @@ const EXPECTED_MAX_BLOCK_WINDOW = 120000
 // if we don't do this, we run into errors as the data is not yet available
 const COSMOS_DB_DELAY = 2000
 const PROPOSAL_POLLING_INTERVAL = 600000 // 10min
+const UPDATE_VALIDATORS_INTERVAL = 600000 // 10min
```
faboweb, line 20: so validators would only be visible after 10 minutes?

api/lib/block-listeners/cosmos-node-subscription.js
```diff
@@ -51,6 +53,25 @@ class CosmosNodeSubscription {
...
+  async updateValidators() {
+    const cosmosAPI = new this.CosmosApiClass(this.network, this.store)
+    let validators = await cosmosAPI.getAllValidators(this.height)
+    validators = await this.addPopularityToValidators(
+      validators,
+      this.network.id
+    )
+    const validatorMap = await this.getValidatorMap(validators)
+    this.updateDBValidatorProfiles(validators)
+    this.store.update({
```
faboweb, line 65: you could run

```
 validators = await this.addPopularityToValidators(
      validators,
      this.network.id
    )
```

in the store for every update. this way it scales across polkadot

you would leave the other update in place

api/lib/block-listeners/cosmos-node-subscription.js
```diff
@@ -149,14 +171,6 @@ class CosmosNodeSubscription {
         await cosmosAPI.newBlockHandler(block, this.store)
       }
 
-      const validators = await cosmosAPI.getAllValidators(block.height)
```
faboweb, line 152: this would stay here

api/lib/block-listeners/cosmos-node-subscription.js
Bitcoinera: no, because it runs it just on start. But they update every 10 minutes yes. I mean, like nearly everything I push I tested it and saw it works

api/lib/block-listeners/cosmos-node-subscription.js
Bitcoinera: why? Isn't it better to have it in a function on its own? Like we have proposals. There is no difference

api/lib/block-listeners/cosmos-node-subscription.js
Bitcoinera: Thank you for this observation! I just applied it and it goes even faster :star_struck: 

I also think the loop think is a great idea. Doesn't make sense to me updating validators every block.

Let's merge and enjoy :cocktail:

api/lib/block-listeners/cosmos-node-subscription.js
Bitcoinera: ah, I get it, because maybe in that block the validator got slashed or something like that no? ok :+1: putting it back

Bitcoinera: Ah, but the problem is that I don't want to store validators without popularity. So yeah, I will have to do as you said and add popularity inside the store

Bitcoinera: Moving it to the store made it terribly slow and created connection issues :shrug:   The fastest I have seen is with my last commit "applu one single DB connection". 

In the store the `update` function becomes async and everything slows down by x100

Bitcoinera: In the store what changes mainly is this, and this really kills performance it seems:

```
  async update({
    height,
    block = this.block,
    validators = this.validators,
    data = this.data // multi purpose block to be used for any chain specific data
  }) {
    if (Array.isArray(validators)) {
      // add popularity field
      validators = await this.addPopularityToValidators(
        validators,
        this.network.id
      )
      // convert to map
      validators = this.getValidatorMap(validators)
    } else {
      console.error(`Validators should be passed to store as an Array`)
    }
```

api/lib/block-listeners/cosmos-node-subscription.js
```diff
@@ -204,20 +204,20 @@ class CosmosNodeSubscription {
...
-        )
-        // we add the popularity field to the validator
-        return {
-          ...validator,
-          popularity: popularity || 0
-        }
-      })
-    )
+    // popularity is actually the number of views of a validator on their page
+    const validatorPopularity = await this.db.getValidatorsViews(networkId)
```
faboweb, line 208: make this a dictionary to speed up the search for each validator.

api/lib/block-listeners/cosmos-node-subscription.js
```diff
@@ -204,20 +204,20 @@ class CosmosNodeSubscription {
   }
 
   async addPopularityToValidators(validators, networkId) {
```
faboweb, line 206: can be put into store as it doesn't depend on any chain specific data

api/lib/block-listeners/cosmos-node-subscription.js
faboweb: Was this before or after you fixed the 200 requests to the db issue?

api/lib/block-listeners/cosmos-node-subscription.js
faboweb: > I also think the loop think is a great idea. Doesn't make sense to me updating db validators every block.

still we want the live validators even if we don't enrich them from db every block

api/lib/block-listeners/cosmos-node-subscription.js
Bitcoinera: after

Bitcoinera: I mean, I can push my tentative commit so you can see it with your own eyes, although I will probably need to later remove it

api/lib/block-listeners/cosmos-node-subscription.js
Bitcoinera: addressed above: perfomance killer for some reason

api/lib/block-listeners/cosmos-node-subscription.js
Bitcoinera: Yes, I understand your concern. I think what I'm going to do in the end is:

- keep loop in the listener only for updateDBValidators

- move the fixed `addPopularity`thing back again to the resolver

- everything else stays the same

End of the problem

faboweb: Can we move the loop for updating the db validators also into the store so it is available on all networks the same?

api/lib/block-listeners/cosmos-node-subscription.js
faboweb: I would like to cross check, if that is ok for you. Can you push?

api/lib/block-listeners/cosmos-node-subscription.js
Bitcoinera: Done

Bitcoinera:
Everything has been addressed and have also added the improvement of ordering validators alphabeticlally according to operatorAddress. Let's merge! :woman_dancing:

api/lib/block-listeners/cosmos-node-subscription.js
Bitcoinera: I will open another branch with it :wink:

Bitcoinera: Here it is #4294

api/lib/block-listeners/cosmos-node-subscription.js
Bitcoinera: Ouch I hadn't seen this one. OK, let's do it

Bitcoinera: It happens the same. Slows everything down... :shrug:  So strange Seems store is quite delicate

Bitcoinera: :thinking:  :sweat:  I think I never managed to fix this.. noooo

faboweb:
See #4294

PR #4288, opened 2020-06-16
Fabo/fix polkador reconnect

faboweb:
Problem was, that the reconnect would never resubscribe to blocks

### DIFF
--- a/api/lib/block-listeners/polkadot-node-subscription.js
+++ b/api/lib/block-listeners/polkadot-node-subscription.js
@@ -56,30 +56,35 @@ class PolkadotNodeSubscription {
     }
 
     // Subscribe to new block headers
-    const unsubscribe = await this.api.rpc.chain.subscribeNewHeads(async (blockHeader) => {
-      const blockHeight = blockHeader.number.toNumber()
-      if (this.height < blockHeight) {
-        this.height = blockHeight
-        this.newBlockHandler(blockHeight) // do not await as this can take some seconds
-      }
+    const unsubscribe = await this.api.rpc.chain.subscribeNewHeads(
+      async (blockHeader) => {
+        const blockHeight = blockHeader.number.toNumber()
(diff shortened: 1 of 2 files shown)


api/lib/block-listeners/polkadot-node-subscription.js
```diff
@@ -56,25 +56,35 @@ class PolkadotNodeSubscription {
...
+          DISCONNECTION_INTERVAL
+        ) {
+          console.log(
+            'Disconnecting Polkadot for network',
+            this.network.id,
+            'to avoid memory leaks'
+          )
+          // we keep the active connection alive so queries can finish
+          const oldApi = this.api
+          setTimeout(() => oldApi.disconnect(), 1000 * 60 * 5)
```
Bitcoinera, line 80: Then after 5 min it really disconnects right? But how do you know there are no other new queries waiting to be finished in the oldApi?

api/lib/block-listeners/polkadot-node-subscription.js
```diff
@@ -16,7 +16,7 @@ const { spawn } = require('child_process')
 
 const POLLING_INTERVAL = 1000
 // const NEW_BLOCK_DELAY = 2000
-// const DISCONNECTION_INTERVAL = 1000 * 60 * 60 * 6 // used to disconnect from API to free memory
+const DISCONNECTION_INTERVAL = 1000 * 60 * 60 * 6 // used to disconnect from API to free memory
```
Bitcoinera, line 19:
```suggestion
const DISCONNECTION_INTERVAL = 1000 * 60 * 60 * 6 // 6h. Used to disconnect from API to free memory
```

api/lib/block-listeners/polkadot-node-subscription.js
faboweb: new queries use `this.api` which is the new connection

PR #4301, opened 2020-06-17
Ana/4295 verify and store user accounts in db

### LINKED ISSUE #4295: Store user accounts in DB
Moving this from the archived lunie-api to be able to comment, reference, etc.

Resolve an OAuth user in the resolver.js and store it in the db. The frontend will add a header to send the uid of the user. The authentication remains completely with firebase.

In the resolver we will need to get the user from firebase. The FE will send an ID token in the header.
Verify Token
https://firebase.google.com/docs/auth/admin/verify-id-tokens#web
https://firebase.google.com/docs/auth/admin/custom-claims

DB
Create Schema (EMails remain in firebase)

user {
  uid
  email
  premium
  createdAt
  lastActive
}
On every verified token in resolver.js check if the user exists and either create the user or update lastActive.

Bitcoinera:
Closes #4295 

**Description:**

Working. Just needs the update part. I didn't finish it off because I wanted to confirm: `premium` would be a parameter coming in the query from the FE, right? This is assuming we will have some place in FE where people will be able to sign up for premium.

EDIT: Finally I added the premium thing as it seems quite sensible to me.

I have only one concern left: the `GOOGLE_APPLICATION_CREDENTIALS` variable seems only to be detected when doing `export GOOGLE_APPLICATION_CREDENTIALS`. So no matter what I do, in the `.env` file is always ignored. From a developer point of view, this is really annoying. All the time you need to remember this before running the API locally.

Any ideas here what I could be missing? All the other env variables are there but this one concretely is resisting

Thank you! 🚀

---

For contributor:

For reviewer:


api/lib/resolvers.js
```diff
@@ -294,6 +300,41 @@ const transactionMetadata = async (
   }
 }
 
+const storeUser = async (_, { idToken, premium }) => {
```
faboweb, line 303: let's move to a file like `authentication.js`

api/lib/resolvers.js
```diff
@@ -294,6 +300,41 @@ const transactionMetadata = async (
...
+    // get user creation date
+    const userRecord = await firebaseAdmin.auth().getUser(decodedToken.uid)
+    // check if user already exists in DB
+    const storedUser = await database(config)('').getUser(decodedToken.uid)
+    // check if user already has premium as a custom claim
+    if (!userRecord.customClaims) {
+      // set premium field
+      await firebaseAdmin
+        .auth()
+        .setCustomUserClaims(decodedToken.uid, { premium: premium || false })
```
faboweb, line 315: what is the reason to use firebase claims?

api/lib/resolvers.js
```diff
@@ -294,6 +300,41 @@ const transactionMetadata = async (
...
+    }
+    const user = {
+      uid: decodedToken.uid,
+      email: decodedToken.email,
+      premium: premium || false,
+      createdAt: userRecord.metadata.creationTime,
```
faboweb, line 321: as we have `createdAt` and `lastActive` in firebase, let's not store them in our db

api/lib/resolvers.js
```diff
@@ -294,6 +300,41 @@ const transactionMetadata = async (
...
+    // check if user already has premium as a custom claim
+    if (!userRecord.customClaims) {
+      // set premium field
+      await firebaseAdmin
+        .auth()
+        .setCustomUserClaims(decodedToken.uid, { premium: premium || false })
+    }
+    const user = {
+      uid: decodedToken.uid,
+      email: decodedToken.email,
```
faboweb, line 319: we don't want to store the email for privacy reasons (for now)

api/lib/subscription/slashing.js
```diff
@@ -26,18 +26,18 @@ class SlashingMonitor {
...
-        const missedBlocks = response.events['liveness.address'].map(
-          (address, index) => ({
-            validator: address,
-            missedBlocks: response.events['liveness.missed_blocks'][index]
-          })
-        )
-        console.log('Missed block', missedBlocks)
-      }
-    )
+    // this.client.subscribe(
```
faboweb, line 29: uncomment?

faboweb:
🚀 Deployed on https://5eea9d0711a0a985e47ca0c4--lunieio.netlify.app

api/lib/resolvers.js
Bitcoinera: we need it to store the premium field inside the user. You put the docs reference in the issue description

api/lib/resolvers.js
```diff
@@ -584,8 +573,10 @@ const resolvers = {
       if (!remoteFetch(dataSources, networkId).getAddressRole) return undefined
 
       return await remoteFetch(dataSources, networkId).getAddressRole(address)
-    },
-    storeUser: storeUser
+    }
+  },
+  Mutation: {
+    registerUser: registerUser
```
Bitcoinera, line 467: what is the advantage of setting this as a mutation?

api/lib/resolvers.js
```diff
@@ -300,28 +301,16 @@ const transactionMetadata = async (
...
-    const userRecord = await firebaseAdmin.auth().getUser(decodedToken.uid)
     // check if user already exists in DB
     const storedUser = await database(config)('').getUser(decodedToken.uid)
-    // check if user already has premium as a custom claim
-    if (!userRecord.customClaims) {
-      // set premium field
-      await firebaseAdmin
-        .auth()
-        .setCustomUserClaims(decodedToken.uid, { premium: premium || false })
-    }
```
Bitcoinera: we are going to need including this at some point. Why not do it now and just set it always to false?

api/lib/resolvers.js
Bitcoinera: it was in the issue description 🤷

api/lib/resolvers.js
Bitcoinera: https://github.com/luniehq/lunie-api/issues/833

api/lib/resolvers.js
```diff
@@ -18,9 +21,13 @@ const database = require('./database')
 const { getNotifications } = require('./notifications')
 const config = require('../config.js')
 const { logOverview } = require('./statistics')
-const BigNumber = require('bignumber.js')
 const networks = require('../data/networks')
 
+const firebaseServiceAccount = require('../firebaseCredentials.json')
+firebaseAdmin.initializeApp({
+  credential: firebaseAdmin.credential.cert(firebaseServiceAccount)
+})
```
Bitcoinera, line 29: Nice!

api/lib/resolvers.js
Bitcoinera: over a Query I mean

app/config.js
```diff
@@ -25,4 +25,6 @@ export default {
   referralLinks: {
     Coinbase: "https://coinbase-consumer.sjv.io/31vxX",
   },
+
+  firebaseConfig: JSON.parse(process.env.VUE_APP_FIREBASE_CONFIG || "{}"),
```
Bitcoinera, line 29: how would this look like in the `.env` file? I'm having trouble to test it

app/changes/ana_4295-verify-and-store-user-accounts-in-db
```diff
@@ -0,0 +1 @@
+[Added] Register users in DB @faboweb
```
Bitcoinera, line 1: usurper!

app/config.js
Bitcoinera: I'm getting all the time "unexpected end of JSON"

api/lib/resolvers.js
Bitcoinera: I mean, we need this so the firebase user object contains this piece of information, so it cna propagate and be shared both by BE and FE

Bitcoinera:
Tested and works. Would approve but would like this to get included: https://github.com/luniehq/lunie/pull/4301#discussion_r442156298

api/lib/resolvers.js
faboweb: just conceptional: a query delivers data, a mutation manipulates data

Bitcoinera:
Good to go!

Bitcoinera:
Let's merge!! 💃

api/lib/resolvers.js
Bitcoinera: Oh I see

Bitcoinera:
🤔 Don't know why E2E logs aren't complete

PR #4317, opened 2020-06-19
Fabo/credentials in secrets

faboweb:
Credentials are currently unavailable


faboweb:
We should put the credentials in a secret instead but can't succeed right now on that

PR #4334, opened 2020-06-22
Ana/4149 add authentication to apollo

### LINKED ISSUE #4149: Add authentication to apollo

We want to know if client side signed in users are actual users in the API.

We send the users uid as a header in apollo.
https://medium.com/risan/set-authorization-header-with-apollo-client-e934e6517ccf
Get the token of the user via `await firebase.auth().currentUser.getIdToken()`

**Additional context**

Bitcoinera:
Closes #4149 

**Description:**

This is about it, no?

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/api/lib/apollo.js
+++ b/api/lib/apollo.js
@@ -38,8 +38,8 @@ function startBlockTriggers() {
 
 async function checkIsValidIdToken(idToken) {
   try {
-    await firebaseAdmin.auth().verifyIdToken(idToken)
-    return true
+    const decodedToken = await firebaseAdmin.auth().verifyIdToken(idToken)
+    return decodedToken.uid
   } catch (error) {
     console.error(error)
     Sentry.captureException(error)
@@ -92,6 +92,7 @@ function createApolloServer(httpServer) {
           development: req.headers.development,
           authorization: {
             idToken: req.headers.authorization,
+            // valid will be the user uid in case the idToken is valid
             valid: validToken
           }
         }
--- a/api/lib/apollo.js
+++ b/api/lib/apollo.js
@@ -38,7 +38,7 @@ function startBlockTriggers() {
 
 async function checkIsValidIdToken(idToken) {
   try {
-    const decodedToken = await firebaseAdmin.auth().verifyIdToken(idToken)
+    await firebaseAdmin.auth().verifyIdToken(idToken)
     return true
   } catch (error) {
     console.error(error)
--- a/api/lib/apollo.js
+++ b/api/lib/apollo.js
@@ -1,5 +1,8 @@
 const { ApolloServer } = require('apollo-server-express')
 const responseCachePlugin = require('apollo-server-plugin-response-cache')
+const firebaseAdmin = require('firebase-admin')
+const Sentry = require('@sentry/node')
+
 const typeDefs = require('./schema')
 const resolvers = require('./resolvers')
 const Notifications = require('./notifications')
@@ -8,6 +11,11 @@ const { networkList } = require('./networks')
 const NetworkContainer = require('./network-container')
 
 const config = require('../config')
+
+const firebaseServiceAccount = require('../firebaseCredentials.json')
+firebaseAdmin.initializeApp({
+  credential: firebaseAdmin.credential.cert(firebaseServiceAccount)
+})
 
 const networks = networkList.map((network) => new NetworkContainer(network))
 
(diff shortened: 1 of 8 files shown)


faboweb:
🚀 Deployed on https://5ef0feed5ecc2a917549cc07--lunieio.netlify.app

app/src/vuex/modules/account.js
```diff
@@ -33,6 +33,7 @@ export default ({ apollo }) => {
         if (user) {
           commit(`userSignedIn`, true)
           commit(`setUserInformation`, user)
+          localStorage.setItem(`auth_token`, JSON.stringify(user.uid))
```
faboweb, line 36: missing remove on sign out

Bitcoinera: Oh true!

Bitcoinera: I should also refresh the token for ApolloLink right? Should I use this one (https://www.npmjs.com/package/apollo-link-token-refresh)?

Bitcoinera: Ah no, forget that. I saw we have the LinkObservable in `apollo.js`. But should we use then some kind of refreshToken logic in there like [here](https://stackoverflow.com/questions/50965347/how-to-execute-an-async-fetch-request-and-then-retry-last-failed-request/51321068#51321068)?

faboweb: let's improve later. can we check if the token is already outdated before sending the request?

Bitcoinera: This is what I'm about to test

Bitcoinera:
Found the solution here under "Reset store on logout": https://www.apollographql.com/docs/react/networking/authentication/

Tested it and it is working (looking at the headers, authorization token is not longer there)

Bitcoinera:
Still, there is some implementation still needed to get the "Unathorized" error on the GQL query (as seen [here](https://stackoverflow.com/questions/54568174/graphql-apollo-client-cant-see-authorization-header) in the OP).

But maybe this is beyond this PR?

app/src/vuex/modules/account.js
faboweb: Are u sure we need to send the UID? Don't we need to send a verification token?

Bitcoinera: The token can be anything. In this case, Firebase authentication system uses this JWT they call user "uid". We could also create our own, but it would be unnecessary complexity :shrug: 

Not sure though if I'm following you

faboweb: See here: https://firebase.google.com/docs/auth/admin/custom-claims#examples_and_use_cases
The UID is just a user ID anyone can copy. We need an oauth token which they call idToken

Bitcoinera: OK, I see. I will add this tomorrow :wink:

Bitcoinera: Thought it was actually the same but guess the idToken is a time-based randomly-generated one right? Like the 2FA

faboweb: 2FA has nothing to do with it. But yes ^^

Bitcoinera:
Just noticed I need to fix this now, cause by apollo.resetStore():

```
Error sending the query 'maintenance' Invariant Violation: Store reset while query was in flight (not completed in link chain)
```

Bitcoinera:
Ready to go! :woman_dancing:

api/lib/apollo.js
```diff
@@ -28,6 +36,17 @@ function startBlockTriggers() {
   networks.map((network) => network.initialize())
 }
 
+async function checkIsValidIdToken(idToken) {
+  try {
+    const decodedToken = await firebaseAdmin.auth().verifyIdToken(idToken)
+    return true
```
faboweb, line 42: can we return the UID of the user if the token is correct?

Bitcoinera: sure :ok_hand:

Bitcoinera:
Don't merge yet please. I'm still getting `In storeUser TypeError: Cannot read property 'auth' of undefined`...

Something is failing in the import of the `firebaseAdmin` from `apollo.js`

Bitcoinera:
Good to go!

api/lib/apollo.js
```diff
@@ -63,9 +77,19 @@ function createApolloServer(httpServer) {
             throw new Error(`Access Forbidden`)
           }
         }
+        const idToken = req.headers.authorization
+        let validToken = false
+        if (idToken) {
+          validToken = await checkIsValidIdToken(idToken)
```
faboweb, line 83:
```suggestion
        let validToken = undefined
        if (idToken) {
          validToken = await checkIsValidIdToken(idToken)
```

api/lib/apollo.js
```diff
@@ -63,9 +77,19 @@ function createApolloServer(httpServer) {
...
+        }
         return {
           fingerprint: req.headers.fingerprint,
-          development: req.headers.development
+          development: req.headers.development,
+          authorization: {
+            idToken: req.headers.authorization,
+            // valid will be the user uid in case the idToken is valid
+            valid: validToken
```
faboweb, line 91:
```suggestion
            uid: validToken
```

PR #4292, opened 2020-06-16
[Simsala] automatic release created for extension-v1.0.24

faboweb:
Please manually test before merging this to master

### Added

- [#4230](https://github.com/cosmos/lunie/issues/4230) Upgrade polkadot api to v1.18.1 @mariopino

PR #4346, opened 2020-06-23
Revert "[Simsala] automatic release created for extension-v1.0.24"

Bitcoinera:
Reverts luniehq/lunie#4292

I noticed that I actually need to merge to master to release extension


Bitcoinera:
What I don't get is why I was seeing all these changes being added from the `.changes` folder while comparing to develop... Develop should have these in no?

Bitcoinera:
Oops, ok, then it is all good actually :v:

PR #4357, opened 2020-06-24
Fabo/develop preview

faboweb:
Add develop preview and adds nice urls to previews

### DIFF
--- a/.github/workflows/app_deploy_develop.yml
+++ b/.github/workflows/app_deploy_develop.yml
@@ -29,7 +29,7 @@ jobs:
         env:
           SENTRY_DSN: ${{ secrets.SENTRY_DSN_APP }}
         working-directory: ${{env.WORKING_DIRECTORY}}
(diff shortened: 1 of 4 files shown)


faboweb:
🚀 Deployed on https://5ef382335ab7e8e550f7fefa--lunieio.netlify.app

.github/workflows/app_deploy_develop.yml
```diff
@@ -0,0 +1,45 @@
+name: Publish on Netlify
+on:
+  push:
+    branches:
+      - "develop"
```
Bitcoinera, line 5: Nice! But hey, how comes it was working before? I'm confused 👀

faboweb: I disabled the Netlify automatic deployment so we only build and deploy the things we want to reduce costs with Netlify

Bitcoinera: oh ok ok. Nice!

PR #4371, opened 2020-06-25
Ana/fix gql queries not coming from cache

### LINKED ISSUE #4367: GraphQL requests not comming from cache

When opening the ActionModal we see, that requests are done against the API which slow down opening the ActionModal.

The requests should be cached already and therefore should be instant. We need to investigate why this is not the case.
In the case of the ActionModal we should also investigate which requests are actually needed to be done there. The validators in Polkadot are 500 which is a huge payload. Maybe we can reduce them by defaulting to popular.

**Additional context**

Bitcoinera:
Closes #4367 

**Description:**

One line change 😛 

Thank you! 🚀

---

For contributor:

For reviewer:


Bitcoinera:
This is what it takes now:

faboweb:
🚀 Deployed on https://5ef5172d579e3a94f344e564--lunieio.netlify.app

app/src/gql/apollo.js
```diff
@@ -134,7 +134,7 @@ export const createApolloProvider = async () => {
     defaultOptions: {
       // apollo options applied to all queries in components
       $query: {
-        fetchPolicy: "cache-and-network",
+        fetchPolicy: "cache-first",
```
faboweb, line 137: cache-and-network: This fetch policy will have Apollo first trying to read data from your cache. If all the data needed to fulfill your query is in the cache then that data will be returned. However, regardless of whether or not the full data is in your cache this fetchPolicy will always execute query with the network interface unlike cache-first which will only execute your query if the query data is not in your cache. This fetch policy optimizes for users getting a quick response while also trying to keep cached data consistent with your server data at the cost of extra network requests.

Bitcoinera: This is the theory, but now try it

faboweb: Problem with cache first is, that it will not update. So why does cache-and-network not use existing queries?

Bitcoinera: Ok, will investigate

Bitcoinera: The problem seems to be the `loading` prop. We should remove it if we want the results to be coming from cache: https://github.com/apollographql/react-apollo/issues/2739

Bitcoinera: We can stop relying on loading and only look at `this.balances.length` and so on... I am going to give it a try

Bitcoinera: Nothing... 😞 I was also reading some SmartQueries concrete threads but cannot make it work (get results from cache) other than applying `cache-first`.

Look at this article: https://medium.com/@galen.corey/understanding-apollo-fetch-policies-705b5ad71980

I think if we explicitly have the `updates` (like we in fact do) in our SmartQueries we should be fine actually

Bitcoinera: I can open a Draft PR with the "no-apollo-loading" rewrite but I tested it and only `cache-fist` seems to be really working

Bitcoinera: OK, this is not a good solution. I tried undelegating and had to do a hard refresh in order to be able to see the undelegation in PagePortfolio... So not ideal

Bitcoinera:
OK, after reading [here ](https://www.apollographql.com/docs/react/caching/cache-interaction/) about the `cacheRedirects `and [here ](https://www.apollographql.com/docs/react/caching/cache-configuration/#custom-identifiers) about custom identifiers, I think I finally found the root of our problem with GQL queries: 
we should be using "id" fields for all of our GQL entities, same that we do with `Network` and `Proposal`.

Otherwise, the query response will fall in cache under the `ROOT_QUERY`, which is not good.

This can be visually confirmed with a breakpoint in the connection module and then entering `apollo.cache.data.data` (I set the breakpoint on the `const { data } = apollo.query` line).

Another option is to set custom identifiers like I was trying to do in the `add custom identifiers for queries`, but I think it is more elegant to use the `id` solution.

I noticed an improvement of performance by adding the identifiers and cacheRedirects (these last one I am still unsure if we need them or not)

Should I go then for the `ids` solution?

faboweb:
Yes, let'S add IDs for all the objects.

Bitcoinera:
Mmm, tried adding ids to rewards and delegations and was checking it in `PageValidator`, but performance seemed to decrease instead of increase :woman_shrugging: It was taking a while to load it. But now with ids in the validators query I can see a great improvement. Even in ActionModal for some reason

Bitcoinera:
Need to keep testing I think

api/lib/notifications.js
```diff
@@ -220,6 +220,7 @@ const startNotificationService = () => {
       const notificationResponse =
         response.data.insert_notifications.returning[0]
       const notification = {
+        id: getMessageTitle(notificationResponse),
```
faboweb, line 223: a title is not a good id. get the actual id of the notification from the db?

api/lib/reducers/cosmosV0-reducers.js
```diff
@@ -233,6 +234,7 @@ function validatorReducer(networkId, signedBlocksWindow, validator) {
 
 function blockReducer(networkId, block, transactions, data = {}) {
   return {
+    id: block.block_meta.header.height,
```
faboweb, line 237: that would conflict between different chains, right?
```suggestion
    id: `${networkId}_${block.block_meta.header.height}`,
```

api/lib/reducers/kavaV1-reducers.js
```diff
@@ -6,6 +6,7 @@ function setTransactionSuccess(transaction) {
 
 function blockReducer(networkId, block, transactions, data = {}) {
   return {
+    id: block.block.header.height,
```
faboweb, line 9: see above

api/lib/reducers/livepeerV0-reducers.js
```diff
@@ -3,6 +3,7 @@ const LPT_CONVERSION = `1000000000000000000`
```
faboweb, line 3: let's remove livepeer files

api/lib/reducers/polkadotV0-reducers.js
```diff
@@ -16,6 +16,7 @@ function blockReducer(
   data = {}
 ) {
   return {
+    id: blockHeight,
```
faboweb, line 19: see above

api/lib/reducers/cosmosV0-reducers.js
faboweb: or use block hash?

api/lib/reducers/cosmosV0-reducers.js
```diff
@@ -407,6 +414,7 @@ async function reduceFormattedRewards(
         ? await calculateFiatValue(lunieCoin, fiatCurrency)
         : undefined
       multiDenomRewardsArray.push({
+        id: lunieCoin.denom,
```
faboweb: do we need to specify an extra field `id` or can we specify in GQL what the ID field is?

api/lib/reducers/cosmosV0-reducers.js
Bitcoinera: :thinking: maybe better yes

api/lib/reducers/kavaV1-reducers.js
Bitcoinera: ok, let me change that

api/lib/reducers/cosmosV0-reducers.js
Bitcoinera: you mean when we do the query? Not sure... you think ID is a special field? Let's google :sweat_smile:

api/lib/notifications.js
Bitcoinera: changed

faboweb:
Looking good! Whats missing?

Bitcoinera:
> Looking good! Whats missing?

Missing time xD But today I'm killing this one off :+1:

Bitcoinera:
Mmm, getting kind of stuck with this error:

image

Have tried couple of things already but what I don't really get is why it is trying to overwrite a validator Object with a delegation one in cache...

Bitcoinera:
Working well. I found this other error just now and tried to fix it with the `map` method in `PageNotifications`. Unfortunately, hasn't worked. But it can be another PR, as it is unrelated to this one:

image

Bitcoinera:
Ready to go! :woman_dancing:

.vscode/launch.json
```diff
@@ -14,8 +14,8 @@
             "program": "${workspaceFolder}/api/index.js",
             "envFile": "${workspaceFolder}/api/.env",
             "env": {
-                "WHITELIST_ORIGIN": "http://localhost:4000",
-                // "RUN_ONLY_NETWORK": "cosmos-hub-testnet"
```
jbibla: should this be in the .vscode file?

Bitcoinera: OK, I just deleted this whole `env` block from the `launch.json`

faboweb:
e2e fails

Bitcoinera:
Mmm, where are the validators for the E2E to be found?

faboweb:
Still failing e2e tests

Bitcoinera:
> Still failing e2e tests

Yes, because of this:

image


We would need to set the E2E to run against the changed API from this PR. Is there no way to do that? :thinking: GitHub Instances, I don't know. We can be noisy with the GitHub people

faboweb:
Need to split the API from the PR and deploy independently

Bitcoinera:
> Need to split the API from the PR and deploy independently

:scream:

Bitcoinera:
Or we can E2E manually, which I already did, and see that it is effectively working, which it is :wink: And we save up dev time

Bitcoinera:
Or we can E2E manually, which I already did, and see that it is effectively working, which it is :wink: And we save up dev time

PR #4373, opened 2020-06-26
Ana/4369 connect to polkadotapi asap

### LINKED ISSUE #4369: Connect to Polkadot API asap

When the user wants to sign a tx we connect to the Polkadot API. This takes some time and causes the user to need to wait some seconds.

Connect to the Polkadot API in the background when the user is signed to a Polkadot network. The connection still takes time, but by the time the user is at the sign step, the connection should be established.

**Additional context**

Bitcoinera:
Closes #4369 

**Description:**

~~Working also signing with extension. ~~ --> this was just cache which made me think so. Fixing this now

It is really going faster now.

Question, is there a way to get the store state from a JS file outside the Vue app files like `polkadot.js`?

So we don't need to pass polkadotAPI as a parameter all the time, this is my idea

Thank you! 🚀

---

For contributor:

For reviewer:


Bitcoinera:
Sadly cannot test it on Ledger for obvious reasons 😅

faboweb:
🚀 Deployed on https://5ef5e80f7d5ef701cb997fab--lunieio.netlify.app

Bitcoinera:
But yeah, I am a bit surprised that it does work on extension with no further changes. On line 75 in `messageHandlers` it is calling `createAndSignLocally` like so:
```
 await transactionManager.createAndSignLocally(
          messageType,
          message,
          transactionData,
          senderAddress,
          network,
          'local',
          password,
        )
``` 
and it doesn't complain at all. When is this function triggered then?

Bitcoinera:
Mmm, it is going to be trickier with extension it seems. Cannot send the API, so I guess extension needs to connect itself:

faboweb:
> Mmm, it is going to be trickier with extension it seems. Cannot send the API, so I guess extension needs to connect itself:

obvio ^^

app/src/components/staking/PageValidators.vue
```diff
@@ -242,7 +242,6 @@ export default {
 .show-mobile-sorting {
   display: none;
   cursor: pointer;
-  display: none;
```
faboweb, line 245: ?

faboweb:
You can extract into a file `polkadotConnections.js` which you can import where needed. This file could also be reused in the extension.

Bitcoinera:
> > Mmm, it is going to be trickier with extension it seems. Cannot send the API, so I guess extension needs to connect itself:
> 
> obvio ^^

XD

app/src/components/staking/PageValidators.vue
Bitcoinera: Yeah, this is just a lint thing that I have fixed in all my open PRs now

faboweb:
This will reconnect every time. Please create an object in utils "Polkadot connections" which holds the API per network (dictionary). Then when getting the API, check if the API is already there and connect if not. Finally always await the API to be ready when returning.

extension/src/utils.js
```diff
@@ -1,9 +1,16 @@
...
+    const { WsProvider, ApiPromise } = await import('@polkadot/api')
+    const endpoint = polkadotNetwork.rpc_url
+    const polkadotAPI = new ApiPromise({
+      provider: new WsProvider(endpoint)
+    })
+    await polkadotAPI.isReady
+    // store it in the polkadotAPIs Object for a later use
+    polkadotAPIs[polkadotNetwork.id] = polkadotAPI
+  }
+  return polkadotAPIs[polkadotNetwork.id]
```
faboweb, line 15:
```suggestion
    // store it in the polkadotAPIs Object for a later use
    polkadotAPIs[polkadotNetwork.id] = polkadotAPI
  }
  await polkadotAPI.isReady
  return polkadotAPIs[polkadotNetwork.id]
```

app/src/vuex/modules/connection.js
```diff
@@ -150,8 +154,29 @@ export default function ({ apollo }) {
...
+    async connectToPolkadotAPI({ commit }, { networkId }) {
+      const { WsProvider, ApiPromise } = await import("@polkadot/api")
+
+      const polkadotNetworks = state.networks.filter(
+        (network) => network.network_type === `polkadot`
+      )
+
+      const polkadotNetwork = polkadotNetworks.find(
+        (network) => network.id === networkId
+      )
```
faboweb, line 172:
```suggestion
      const polkadotNetwork = state.networks.find(
        (network) => network.id === networkId
      )
```

app/src/vuex/modules/connection.js
```diff
@@ -150,8 +154,29 @@ export default function ({ apollo }) {
       }
       commit("setAddressType", network.address_creator)
       dispatch(`checkForPersistedSession`) // check for persisted session on that network
+      // if the network is a pokadot network
+      if (network.network_type === `polkadot`) {
+        actions.connectToPolkadotAPI({ commit }, { networkId: network.id })
```
faboweb, line 159:
```suggestion
        actions.connectToPolkadotAPI({ commit }, network
```

this way u save urself the extra network lookup

extension/src/utils.js
```diff
@@ -0,0 +1,16 @@
+// in the polkadotAPIs Object we will store all the different Polkadot APIs to sign in extension
```
faboweb, line 1: why not use this utils function also in the app?

app/src/vuex/modules/connection.js
Bitcoinera: yeah, much better

app/src/vuex/modules/connection.js
Bitcoinera: true!

extension/src/utils.js
Bitcoinera: true. It could be called from the connection module. Let me try that

Bitcoinera: OK, so what I did in the end is to have the `polkadotAPIs` object in state instead of the `polkadotAPI` one. Much better. Only it gets all erased with refresh, but logically happens the same for all objects.

Still, the performance improvement is quite significant. So let's merge!!

Bitcoinera:
All suggestions applied! Let's merge! :woman_dancing:

app/src/vuex/modules/connection.js
```diff
@@ -150,8 +154,26 @@ export default function ({ apollo }) {
...
+      if (!state.polkadotAPIs[polkadotNetwork.id]) {
+        const { WsProvider, ApiPromise } = await import("@polkadot/api")
+        const endpoint = polkadotNetwork.rpc_url
+        const polkadotAPI = new ApiPromise({
+          provider: new WsProvider(endpoint),
+        })
+        let polkadotAPIs = state.polkadotAPIs
+        // store it in the polkadotAPIs Object for a later use
+        polkadotAPIs[polkadotNetwork.id] = polkadotAPI
+        await polkadotAPIs[polkadotNetwork.id].isReady
```
faboweb, line 173:
```suggestion
```

we don't need to wait here. we want the api to be instantly available to i.e. the ActionModal. Then in the ActionModal we can wait for it to be ready.

extension/src/utils.js
faboweb: ur still duplicating code here. please use this utils function also in the app.

app/src/vuex/modules/connection.js
```diff
@@ -150,8 +154,26 @@ export default function ({ apollo }) {
       }
       commit("setAddressType", network.address_creator)
       dispatch(`checkForPersistedSession`) // check for persisted session on that network
+      // if the network is a pokadot network
+      if (network.network_type === `polkadot`) {
+        actions.connectToPolkadotAPI({ commit }, network)
```
faboweb, line 159: here call `getPolkadotAPI` so the API connects

app/src/ActionModal/components/ActionModal.vue
```diff
@@ -699,6 +699,9 @@ export default {
           network: this.network,
           signingType: this.selectedSignMethod,
           password: this.password,
+          polkadotAPI: this.$store.state.connection.polkadotAPIs[
+            this.network.id
+          ],
```
faboweb, line 704: Here you can also call `getPolkadotApi`

app/src/vuex/modules/connection.js
faboweb: you don't need to await it

app/src/ActionModal/components/ActionModal.vue
Bitcoinera: but it is already in store? Then where is the point?

Bitcoinera:
Can we do `import { getPolkadotAPI } from "../../../../extension/src/utils"` to import the same function than from extension? I think so, right?

We could also add a webpack config in the root folder so we get direct paths like `extension/..` or `app/..`

Also checked the `.isReady` thing and doesn't make any difference. You can test it by yourself. It is super fast now. You cannot tell the difference with a Tendermint account right now

app/src/vuex/modules/connection.js
```diff
@@ -1,6 +1,7 @@
 import config from "src/../config"
 import bech32 from "bech32"
 import { NetworksAll } from "../../gql"
+import { getPolkadotAPI } from "../../../../extension/src/utils"
```
faboweb, line 4: how about moving the script to `root/utils`?

app/src/vuex/modules/connection.js
```diff
@@ -156,24 +157,11 @@ export default function ({ apollo }) {
       dispatch(`checkForPersistedSession`) // check for persisted session on that network
       // if the network is a pokadot network
       if (network.network_type === `polkadot`) {
-        actions.connectToPolkadotAPI({ commit }, network)
+        const polkadotAPI = await getPolkadotAPI(network)
```
faboweb, line 160: as described earlier you don't want to wait here until the api is connected. In ActionModal in a bad case, the api object would not be set yet.

Bitcoinera: Done. At least what I think you meant

utils.js
```diff
@@ -0,0 +1,16 @@
...
+  if (!polkadotAPIs[polkadotNetwork.id]) {
+    const { WsProvider, ApiPromise } = await import('@polkadot/api')
+    const endpoint = polkadotNetwork.rpc_url
+    const polkadotAPI = new ApiPromise({
+      provider: new WsProvider(endpoint)
+    })
+    // store it in the polkadotAPIs Object for a later use
+    polkadotAPIs[polkadotNetwork.id] = polkadotAPI
+  }
+  await polkadotAPIs[polkadotNetwork.id].isReady
```
faboweb, line 14:
```suggestion
```

app/src/vuex/modules/connection.js
```diff
@@ -150,6 +155,11 @@ export default function ({ apollo }) {
       }
       commit("setAddressType", network.address_creator)
       dispatch(`checkForPersistedSession`) // check for persisted session on that network
+      // if the network is a pokadot network
+      if (network.network_type === `polkadot`) {
+        const polkadotAPI = getPolkadotAPI(network)
+        commit(`setPolkadotAPI`, polkadotAPI)
```
faboweb, line 161:
```suggestion
        // we initialize the API here so it is available when the user wants to use it
        getPolkadotAPI(network)
```

app/src/vuex/modules/connection.js
```diff
@@ -38,6 +40,9 @@ export default function ({ apollo }) {
     setNetworks(state, networks) {
       state.networks = networks
     },
+    setPolkadotAPI(state, polkadotAPI) {
+      state.polkadotAPI = polkadotAPI
+    },
```
faboweb, line 45:
```suggestion
```

app/src/vuex/modules/connection.js
```diff
@@ -23,6 +24,7 @@ export default function ({ apollo }) {
     externals: {
       config,
     },
+    polkadotAPI: {}, // in the polkadotAPI Object we will store all the different Polkadot APIs. It is erased on refresh
```
faboweb, line 27:
```suggestion
```

app/src/ActionModal/components/ActionModal.vue
```diff
@@ -699,6 +699,7 @@ export default {
           network: this.network,
           signingType: this.selectedSignMethod,
           password: this.password,
+          polkadotAPI: await this.$store.state.connection.polkadotAPI,
```
faboweb, line 702: use `getPolkadotApi` here

Bitcoinera: then what is the point of store? Wasn't the idea to connect there on network switch? Actually I think the best idea was to have the `polkadotAPIs` object in there. Then you can store all of them along a session. Let's do that better

Bitcoinera: My engineering 2 cents

Bitcoinera: Or you mean, delete it all from store and have the function just here? (still think the `polkadotAPIs` solution in store is better)

app/src/ActionModal/components/ActionModal.vue
faboweb: don't put it into the store. why maintain it twice?

app/src/ActionModal/components/ActionModal.vue
faboweb: why maintain the api object twice? if the api object is in the utils function, that is enough, isn't it? and this way you can reuse the script in FE,API,extension.

app/src/vuex/modules/connection.js
```diff
@@ -1,7 +1,7 @@
 import config from "src/../config"
 import bech32 from "bech32"
 import { NetworksAll } from "../../gql"
-import { getPolkadotAPI } from "../../../../utils"
+import { getPolkadotAPI } from "../../../../common/polkadotApiConnector"
```
Bitcoinera, line 4: :eyes:

app/src/ActionModal/components/ActionModal.vue
Bitcoinera: oh, saw this now. Yes, that is true. Taking advantage of the monorepo :thinking: 

I see you took over :+1:

app/src/vuex/modules/connection.js
Bitcoinera: oh I see

Bitcoinera:
We need to include `public_rpc_url` in the networks table too

faboweb:
This PR fails in the E2E tests and in general when broadcasting

PR #4376, opened 2020-06-26
Ana/fix many little Polkadot bugs

Bitcoinera:
Closes #ISSUE

**Description:**

Here I fix many little Polkadot/Kusama bugs that have been lying there for a while. Concretely, these three:

image

image

Well, this one is actually Polkadot-unrelated:

Thank you! 🚀

---

For contributor:

For reviewer:


faboweb:
🚀 Deployed on https://5ef64ae2e2adb201c69e57da--lunieio.netlify.app

api/lib/resolvers.js
```diff
@@ -397,7 +397,7 @@ const resolvers = {
       if (operatorAddress) {
         rewards = rewards.filter((reward) => {
           // multidenom rewards
-          if (reward.validator === undefined) {
+          if (reward.validator === undefined && Array.isArray(reward)) {
```
faboweb, line 400: reward can't be an object an array at the same time, can it?

Bitcoinera: :thinking: then we are doing something wrong. Our data types should be predictable.

Bitcoinera: Cannot reproduce this now...

faboweb: so what is reward? it is not an array, right?

Bitcoinera: this is what I'm trying to find out, why sometimes it fails / what it is. Maybe it is `null`?

Bitcoinera: ah no. I know what was happening. It was because of this error when you are switching networks and validators are not there yet. It was one of the switch-networks-navbar-related bugs. But now seems gone. I cannot reproduce it at least. Maybe because my cache is already "fixed" with the ids.

Actually I get to navigate now super fast between networks and validators pages. No errors.

So it is a cache problem. Will remove my "fix"

PR #4415, opened 2020-07-03
Fabo/store validator identity

faboweb:
Stores the identity in the DB to be able to let the script runner query and fill the profile pictures

### DIFF
--- a/api/lib/reducers/polkadotV0-reducers.js
+++ b/api/lib/reducers/polkadotV0-reducers.js
@@ -37,7 +37,7 @@ function validatorReducer(network, validator) {
       validator.identity.web && validator.identity.web !== ``
         ? validator.identity.web
         : ``,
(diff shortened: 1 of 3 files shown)


api/lib/reducers/polkadotV0-reducers.js
```diff
@@ -37,7 +37,7 @@ function validatorReducer(network, validator) {
       validator.identity.web && validator.identity.web !== ``
         ? validator.identity.web
         : ``,
-    identity: identityReducer(validator.accountId, validator.identity),
+    identity: validator.identity.twitter,
```
Bitcoinera, line 40: why twitter in particular? Is it the usually most complete one?

faboweb: We could use Twitter or Matrix. I don't know if Matrix actually has a profile picture.

PR #4438, opened 2020-07-06
Fabo/twitter images

faboweb:
Poll Twitter images for Polkadot validators


scriptRunner/src/twitterImages.js
```diff
@@ -0,0 +1,53 @@
+const fetch = require("node-fetch")
+const database = require("../../api/lib/database")
+const config = require("../../api/config")
+const db = database(config)
+
+module.exports.getTwitterImages = async () => {
+    const networks = await db("").getNetworks()
+    const cosmosNetworks = networks.filter(({network_type}) => network_type === "polkadot")
```
jbibla, line 8: shouldn't this be `substrateNetworks` or `polkadotNetworks`?

faboweb: There is a suggestion function :P

jbibla: i wasn't sure which one you'd prefer 😉

faboweb: sure... ;P

faboweb:
Needs to add the twitter API key to the deployment still

PR #4452, opened 2020-07-06
Fabo/slashing events

faboweb:
Closes #ISSUE

**Description:**

Thank you! 🚀

---

For contributor:

For reviewer:


faboweb:
On kava-mainnet we get liveness events

api/lib/notifications.js
```diff
@@ -105,6 +105,10 @@ function getMessageTitle(networks, notification) {
 
     case eventTypes.LUNIE_UPDATE:
       return data.title
+    case eventTypes.SLASH:
+      return `Validator ${data.operatorAddress} got slashed ${data.amount}.`
+    case eventTypes.LIVENESS:
+      return `Validator ${data.operatorAddress} was offline ${data.blocks}.`
```
jbibla, line 111: do we have any liveness events in the db?

api/lib/notifications.js
```diff
@@ -105,6 +105,10 @@ function getMessageTitle(networks, notification) {
 
     case eventTypes.LUNIE_UPDATE:
       return data.title
+    case eventTypes.SLASH:
+      return `Validator ${data.operatorAddress} got slashed ${data.amount}.`
```
jbibla, line 109: will this be a fiat amount or token amount?

api/lib/notifications.js
```diff
@@ -214,6 +222,10 @@ const startNotificationService = (networks) => {
     eventSubscription(async (event) => {
       if (event.properties.type === 'UnknownTx') return
 
+      // hack to not spam users on every liveness failure
+      // need to figure out how to handle this
+      if (event.eventType === eventTypes.LIVENESS) return
```
jbibla, line 227: so this happens a lot?

jbibla:
lots of lint warnings

jbibla:
sorry accidental closed the PR

api/lib/notifications.js
jbibla: ah on kava. why only kava?

api/lib/database/helpers.js
```diff
@@ -116,8 +117,16 @@ const read = ({ hasura_url, hasura_admin_key }) => (schema) => async (
...
-  return res.data[`${schema_prefix}${table}`]
+  try {
+    const res = await graphQLQuery({ hasura_url, hasura_admin_key })(query)
+    return res.data[`${schema_prefix}${table}`]
+  } catch (error) {
+    console.error("DB read failed for query", query, error)
+    Sentry.withScope(function (scope) {
+      scope.setExtra('query', query)
+      Sentry.captureException(error)
+    })
```
Bitcoinera, line 128: Nice. Adding try/catch here :+1:

api/lib/subscription/slashing.js
```diff
@@ -38,20 +53,58 @@ class SlashingMonitor {
...
+          const rows = await Promise.all(missedBlocks.map(async missedBlockEvent => {
+            const lastMissedBlockEvent = await getLastMissedBlock(this.networkId, missedBlockEvent.operatorAddress)
+            return aggregateMissedBlocks(lastMissedBlockEvent, missedBlockEvent)
+          }))
+          database(config)("").upsert("slashes", rows)
+          console.log('Wrote missed block events', rows.length)
+        } catch (error) {
+          console.error("Failed to write missed block events", error)
+          Sentry.captureException(error)
+        }
```
Bitcoinera, line 78: Probably this is an event that validators will be highly interested in (at least the rookie ones that are not measuring this themselves yet). Will we have in the future notifications specific for validators and others specific for "regular" users?

api/lib/subscription/slashing.js
```diff
@@ -38,20 +53,58 @@ class SlashingMonitor {
...
+      slashes(where:{networkId:{_eq:"${networkId}"}, operatorAddress:{_eq:"${operatorAddress}"}, reason:{_eq:"missed_blocks"}}, order_by:{id:desc}, limit: 1) {
+        height
+        id
+        reason
+      }
+    }
+  `)
+
+  return data.slashes[0]
+}
```
Bitcoinera, line 97: Another remark: I saw in DB that "slashes" is a different table. Shouldn't we have all events in a separate table of its own? Like, right now, where are we storing events in? In "notifications" and "slashes" and that is right?

Just trying to grasp why slashes get a different table here

faboweb: The idea was, that slashes are non query able. So I wanted a nondestructive way to store them. But I guess we could also just store them as notifications as the data part of the notifications also has all the data.

api/lib/subscription/slashing.js
faboweb: That is the idea. Validators would get an event on the first missed block and users once a validator actually gets slashed or similar.

Bitcoinera: Nice

api/lib/subscription/slashing.js
Bitcoinera: or at least we should store them in both tables to stay consistent i.e. collect all events in the notifications table

api/lib/notifications.js
faboweb: right now it is an onchain amount, need to adjust to show as coin value. should we show as fiat value? but which fiat value? so I guess we show the LunieCoin value

api/lib/notifications.js
faboweb: it could happen every block. if you check the slashes table you see some liveness failures >2000 blocks

api/lib/subscription/slashing.js
faboweb: I commented that we need to find a proper strategy for storing the liveness events. If not there can easily be a couple of thousand for one validator outage.

api/lib/notifications.js
faboweb: only tested on kava

api/lib/subscription/slashing.js
jbibla: just an idea - log the first liveness event, start a counter, log the count, log the last liveness event.

api/lib/notifications.js
jbibla: one day, users should be able to choose their preferred currency. for now though, if it's token amount it should include the denom and not the `$`.

api/lib/subscription/slashing.js
jbibla: > Will we have in the future notifications specific for validators and others specific for "regular" users?

probably, yes.

api/lib/notifications.js
faboweb: see update

api/lib/subscription/slashing.js
faboweb: similar: query last event of that type (operator address and liveness) compare with current and create a new notification if triggers apply

faboweb: should I put this in here or do another PR?

Bitcoinera: I'd put it in another one

Bitcoinera: Oh, could you also lint this file: `api/lib/database/helpers.js`

api/lib/notifications.js
```diff
@@ -105,6 +105,10 @@ function getMessageTitle(networks, notification) {
 
     case eventTypes.LUNIE_UPDATE:
       return data.title
+    case eventTypes.SLASH:
+      return `Validator ${data.operatorAddress} got slashed ${data.amount.amount} ${data.amount.denom}s.`
+    case eventTypes.LIVENESS:
+      return `Validator ${data.operatorAddress} was offline ${data.blocks}.`
```
jbibla, line 111:
```suggestion
      return `Validator ${data.operatorAddress} was offline for ${data.blocks} blocks.`
```

api/lib/notifications.js
jbibla: i don't see what changed with regard to fiat / token amounts

faboweb: was missing a push

api/lib/database/helpers.js
```diff
@@ -30,9 +31,22 @@ const graphQLQuery = ({ hasura_url, hasura_admin_key }) => async (query) => {
...
+    const clone = JSON.parse(JSON.stringify(value))
+    Object.keys(clone).forEach((key) => {
+      clone[key] = escapeValue(clone[key])
+    })
+    return JSON.stringify(clone)
+  } else {
+    return `"${escape(value)}"`
+  }
+}
+
```
Bitcoinera, line 46: We ended up doing this quite differently no? Strange that it doesn't show conflict. We should be careful with this, might overwrite

api/lib/database/helpers.js
```diff
@@ -30,9 +31,22 @@ const graphQLQuery = ({ hasura_url, hasura_admin_key }) => async (query) => {
...
+  if (typeof value === 'object') {
+    const clone = JSON.parse(JSON.stringify(value))
+    Object.keys(clone).forEach((key) => {
+      clone[key] = escapeValue(clone[key])
+    })
+    return JSON.stringify(clone)
+  } else {
+    return `"${escape(value)}"`
+  }
+}
```
Bitcoinera, line 45:
```suggestion
function escapeValue(value) {
  return value === undefined || value === null
    ? `""`
    : typeof value === 'boolean' || typeof value === 'number'
    ? value
    : typeof value === 'string'
    ? `"${escape(value)}"`
    : // we need to double stringify to double escape the quotations
      // if not, inserted in the query the object will have double quotes inside
      JSON.stringify(JSON.stringify(escapeObject(value)))
}
```

api/lib/subscription/slashing.js
```diff
@@ -38,20 +63,69 @@ class SlashingMonitor {
...
+  // get last liveness issue
+  const { data } = await database(config)('').query(`
+    query {
+      slashes(where:{networkId:{_eq:"${networkId}"}, operatorAddress:{_eq:"${operatorAddress}"}, reason:{_eq:"missed_blocks"}}, order_by:{id:desc}, limit: 1) {
+        height
+        id
+        reason
+      }
+    }
+  `)
```
Bitcoinera, line 112: Why not add a `liveness` query too? Or `missedBlocks`. Since liveness and slashes are two different things. Could be very useful for validators this query. Could look like this:

```
    query {
      missedBlocks(where:{networkId:{_eq:"${networkId}"}, operatorAddress:{_eq:"${operatorAddress}"}, order_by:{height:desc}) {
        height
        id # from Block
        total # total number from blocks missed. Could be a separate query by itself
      }

```

We can add this in a separate PR though

api/lib/subscription/slashing.js
```diff
@@ -38,20 +63,69 @@ class SlashingMonitor {
...
+          console.error('Failed to write missed block events', error)
+          Sentry.captureException(error)
+        }
+      }
+    )
+  }
+}
+
+async function getLastMissedBlock(networkId, operatorAddress) {
+  // get last liveness issue
```
Bitcoinera, line 103:
```suggestion
async function getLastSlash(networkId, operatorAddress) {
  // get last slash
```

I don't know if I'm missing something here but this is what this query looks like is doing

app/src/components/account/Paywall.vue
```diff
@@ -0,0 +1,342 @@
...
+    transform: translate3d(40%, 0, 0);
+    visibility: visible;
+  }
+
+  100% {
+    -webkit-transform: translate3d(-70%, 0, 0);
+    transform: translate3d(-70%, 0, 0);
+  }
+}
+</style>
```
Bitcoinera, line 342: Mmm, PR pollution

api/lib/subscription/slashing.js
```diff
@@ -38,20 +63,69 @@ class SlashingMonitor {
...
+                missedBlockEvent.operatorAddress
+              )
+              return aggregateMissedBlocks(
+                lastMissedBlockEvent,
+                missedBlockEvent
+              )
+            })
+          )
+          database(config)('').upsert('slashes', rows)
+          console.log('Wrote missed block events', rows.length)
```
Bitcoinera, line 92: Mmm, I see. Isn't this confusing? We are handling missing blocks as "slashes", but they are two separate things. Should have also different tables, as I already commented

api/lib/database/helpers.js
jbibla: 😆

faboweb: instead I merged develop

api/lib/subscription/slashing.js
faboweb: with this filter `reason:{_eq:"missed_blocks"}` I think the description is pretty decent

api/lib/subscription/slashing.js
faboweb: they are not. you also get slashed for liveness. need to find the numbers

api/lib/subscription/slashing.js
faboweb: I like it. Can you open an issue? We might need this for the validator profiles (uptime per chain and cross chain).

faboweb:
Specs on liveness slashes: https://github.com/cosmos/cosmos-sdk/blob/master/x/slashing/spec/04_begin_block.md#uptime-tracking
We could fire an event only on "validator goes offline for at least one block" and "validator got slashed and jailed for downtime".

api/lib/subscription/slashing.js
Bitcoinera: But it could be a slash due to double signing. I think we should avoid mixing up the concepts

Bitcoinera: oh, but that is why you are using this: `reason:{_eq:"missed_blocks"}`. OK, I see

api/lib/subscription/slashing.js
Bitcoinera: > We could fire an event only on "validator goes offline for at least one block" and "validator got slashed and jailed for downtime".

You mean as two separate events or you mean handling both as the same event?

faboweb: those would be 2 different events. one on the first block, one on the 5000 block or sth.

Bitcoinera:
Should we merge this? Just difficult to test, right?

jbibla:
> decrease coverage by 58.28%.

🤯

PR #4454, opened 2020-07-07
Ana/4365 cache api store in db

### LINKED ISSUE #4365: Cache store in db

On API startup the api takes a while to connect and therefore it takes a bit to deliver the first data.

Write the whole store into the db on an update. The db should have one table `storeCaches`. This table should have columns:
- networkId
- store // JSON serialized store

On API start fetch the cached store and populate the store with the cache to reduce waiting times on start.

This issue only happens when we deploy the API for a couple of seconds. So maybe we could just ignore the issue.

**Additional context**

Bitcoinera:
Closes #4365 

**Description:**

Reopening this. Also found some flaws in the new way we handle escaping values in helpers.js. Concretely, what concerns booleans.

EDIT: actually I had to revert the helper to my last commit @faboweb. It wasn't working with your last version.

Thank you! 🚀

---

For contributor:

For reviewer:


api/lib/database/helpers.js
```diff
@@ -32,16 +32,18 @@ const graphQLQuery = ({ hasura_url, hasura_admin_key }) => async (query) => {
...
-  if (typeof value === 'object') {
-    const clone = JSON.parse(JSON.stringify(value))
-    Object.keys(clone).forEach((key) => {
-      clone[key] = escapeValue(clone[key])
-    })
-    return JSON.stringify(clone)
-  } else if (typeof value === 'boolean') {
-    return escape(value)
-  } else {
+  try {
```
faboweb, line 35: let's write a test suite for this.

api/lib/database/helpers.js
```diff
@@ -31,26 +31,25 @@ const graphQLQuery = ({ hasura_url, hasura_admin_key }) => async (query) => {
...
-  } catch (error) {
-    return `"${escape(value)}"`
+    default: return value
   }
 }
 
 function gqlKeyValue([key, value]) {
   // escape all values but handle objects gracefully
-  return `${key}: ${escapeValue(value)}`
+  return `${key}: ${value ? escapeValue(value) : `""`}`
```
Bitcoinera, line 53: beware the booleans here. I was getting that error, `invalid boolean value, `""``

This was my initial PR description. Maybe I edited it

api/tests/source/database.test.js
```diff
@@ -0,0 +1,24 @@
+const {escapeValue, gqlKeyValue} = require("../../lib/database/helpers")
+
+describe("database helpers", () => {
+    describe("escape values", () => {
+        it("boolean", () => {
+            expect(escapeValue(true)).toBe(true)
+        })
```
Bitcoinera, line 7: can you try `false` here? I think it is not going to work

api/lib/stores/block-store.js
```diff
@@ -275,17 +275,8 @@ function enrichValidator(validatorInfo, validator) {
...
-  dbStore.block = store.block
-  dbStore.stakingDenom = store.stakingDenom
-  dbStore.annualProvision = store.annualProvision
-  dbStore.signedBlocksWindow = store.signedBlocksWindow
-  dbStore.validators = store.validators
-  dbStore.proposals = store.proposals
-  dbStore.newValidators = store.newValidators
   await database(config)('').storeStore({
-    store: JSON.stringify(dbStore),
+    store: JSON.parse(JSON.stringify(dbStore)), // clone to remove any functions
```
Bitcoinera: What functions do you mean?

api/tests/source/database.test.js
```diff
@@ -0,0 +1,24 @@
...
+        it("number", () => {
+            expect(escapeValue(42)).toBe(42)
+        })
+        it("object", () => {
+            expect(escapeValue({a: {b: "<script></script>", c: 42}})).toBe(`"{\"a\":{\"b\":\"&lt;script&gt;&lt;/script&gt;\",\"c\":42}}"`)
+        })
+        it("null", () => {
+            expect(gqlKeyValue(["x", undefined])).toBe(`x: ""`)
+        })
```
Bitcoinera, line 22: it would be better if "null" would be parsed as `null`, no?

api/lib/database/helpers.js
Bitcoinera: This happens when the Boolean is `false`

faboweb: uhhh good catch

api/tests/source/database.test.js
Bitcoinera: Ah no, ignore. It is working! :tada: bravo!

api/lib/database/helpers.js
Bitcoinera: ? but the test pass using `false`? :eyes:

Bitcoinera: I think it is ok, it goes to default then

api/lib/stores/block-store.js
faboweb: i.e. `addValidatorPictures`

api/tests/source/database.test.js
faboweb: is that possible?

Bitcoinera: should actually go to `default`? ah, but it is actually type string, doesn't convert it...

faboweb: not possible to set null

api/tests/source/database.test.js
faboweb: added a test for that

api/tests/source/database.test.js
Bitcoinera: but if we don't do JSON.stringify `null` should be `null` anyway, right

Issue #4435, opened 2020-07-06
remaining problems with user auth and menu


jbibla:
- [x] update menu copy (address sign in vs sign up)
- [x] clean up menu account list design (remove type and purple testnet icon?)
- [x] url should change when switching accounts @faboweb 
- [x] icon on session modal should be coloured and centered
- [ ] email validation is required for magic link email form @faboweb 
- [x] font sizing is wrong in new session modals
- [x] clean up urls (magic-link-sent-modal, sign-in-modal)

faboweb:
What do you mean by "clean up urls (magic-link-sent-modal, sign-in-modal)" ?

Can you take the CSS tasks? I would check "url should change when switching accounts" and can "email validation is required for magic link email form"

jbibla:
> What do you mean by "clean up urls (magic-link-sent-modal, sign-in-modal)" ?

i don't think the URLs should have "modal" in them and i want to check with the existing session modals to see if we can have some more consistency there. the combo of sign in, sign up, sign out, session, account, address is complex - wanted to see if things are consistent. 

> Can you take the CSS tasks

definitely. working on it already! 

> I would check "url should change when switching accounts" and can "email validation is required for magic link email form"

what do you mean by "check"?

faboweb:
> what do you mean by "check"?

I can check why this happening

faboweb:
> i don't think the URLs should have "modal" in them

doesn't feel like a strong problem to solve

Bitcoinera:
>  email validation is required for magic link email form

I don't understand the problem here?

Bitcoinera:
> url should change when switching accounts

This is actually working.

faboweb:
> >  email validation is required for magic link email form
> 
> I don't understand the problem here?

The input should check if it is an email.

Bitcoinera:
> The input should check if it is an email.

I believe Firebase already does it for us. Let me quickly check

jbibla:
url changing is working! nice! 

@Bitcoinera it shouldn't even let the user go to the next modal that says "sent" if it's not a valid email address.

Bitcoinera:
I tried again switching accounts, this time using extension accounts, and it is working well. No accounts mixings.

@faboweb can you check if can still reproduce this error? It would the last one to solve

faboweb:
just checked and looks good

Bitcoinera:
[...]

PR #4486, opened 2020-07-09
Fabo/fix session having wrong network

faboweb:
Closes https://github.com/luniehq/lunie/issues/4435

Session had a wrong network id s the current not the target network was used

### DIFF
--- a/app/tests/unit/specs/store/session.spec.js
+++ b/app/tests/unit/specs/store/session.spec.js
@@ -281,7 +281,7 @@ describe(`Module: Session`, () => {
           commit,
           dispatch,
         },
-        { sessionType: `explore`, address, networkId: "fabo-net" }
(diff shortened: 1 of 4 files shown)


faboweb:
🚀 Deployed on https://5f07849df4de1a77033c9756--lunieio.netlify.app

app/tests/unit/specs/store/session.spec.js
```diff
@@ -281,7 +281,7 @@ describe(`Module: Session`, () => {
           commit,
           dispatch,
         },
-        { sessionType: `explore`, address, networkId: "fabo-net" }
+        { sessionType: `explore`, address, networkId: "happy-net" }
```
Bitcoinera, line 284: :smile:

PR #4487, opened 2020-07-09
Fabo/email notifications

### LINKED ISSUE #4387: E-Mail Notifications
As a user, I want to be informed by E-Mail on critical notifications.

Topics:
- slashes (not there yet)
- PROPOSAL_CREATE
- PROPOSAL_UPDATE
- VALIDATOR_STATUS
- VALIDATOR_COMMISSION

E-Mail body should link back to the Notification Page noting that there is more to find there.
Add a note about unsubscribing by writing us an E-Mail.

PRs:
- Research: Find a mail provider (based on cost, do an estimate on amount of emails send)
- Api: Implement a way to send E-Mails to users based on any template
  - start with sending Hello World
  - continue with sending from a template
- Api: Send E-Mails to users on any of the critical topics
- Design: Design E-Mail template

faboweb:
Closes #4387

**Description:**

Thank you! 🚀

---

For contributor:

For reviewer:


api/config.js
```diff
@@ -22,5 +22,7 @@ module.exports = {
   scriptRunnerEndpoint:
     process.env.SCRIPT_RUNNER_ENDPOINT || 'http://localhost:9000',
   scriptRunnerAuthenticationToken:
-    process.env.AUTHENTICATION_TOKEN || secrets.read('AUTHENTICATION_TOKEN')
+    process.env.AUTHENTICATION_TOKEN || secrets.read('AUTHENTICATION_TOKEN'),
+  pepipostAPIKey:
```
Bitcoinera, line 26: Pepipost? :rofl: Sounds so funny!

Bitcoinera: Never heard of it O.o

api/lib/notificationController.js
```diff
@@ -0,0 +1,81 @@
+const database = require("./database");
```
Bitcoinera, line 1: I think we should make a folder called `notifications`. Otherwise the `lib` folder gets too messy

api/lib/notificationController.js
```diff
@@ -0,0 +1,81 @@
...
+            if (remove) {
+                delete this.registrations[topic][type][uid]
+                return
+            }
+            this.registrations[topic][type][uid] = true
+        })
+    }
+
+    getEmails(uids) {
+        return [`fabian@lunie.io`]
```
Bitcoinera, line 52: I also want to get the proto-emails

Bitcoinera, line 52:
```suggestion
        return [`fabian@lunie.io`, `ana@lunie.io`]
```

faboweb:
🚀 Deployed on https://5f0ca6767e906402086e6d2b--lunieio.netlify.app

api/lib/notificationController.js
faboweb: 🤸‍♂️

faboweb:
Gaia testnet is down that is why e2e tests are failing

faboweb:
you can only test this locally

Registrations can be found here
https://staging-db.lunie.io/console/data/schema/public/tables/notificationRegistrations/browse

Bitcoinera:
> Registrations can be found here

Ah, I see the problem with the firebase Id tokens now... because the expire we cannot use it to keep these registrations. Is that it?

Bitcoinera:
Mmm, probably we should handle this error better. I mean, it should be an error, it is a common case. Users can just not be logged in:

image

api/lib/notifications/notifications.js
```diff
@@ -157,7 +157,7 @@ function getPushLink(
 // Get relevant icon for notification
 // TODO: Upload icons to DO instead of passing relative links
 function getIcon({ eventType, data }) {
-  const notificationData = JSON.parse(data)
+  const notificationData = notification.properties || JSON.parse(data)
```
Bitcoinera, line 160: I get here "notification is not defined"

Bitcoinera: image

Bitcoinera, line 160:
```suggestion
  const notificationData= JSON.parse(data)
```

faboweb:
> Mmm, probably we should handle this error better. I mean, it should be an error, it is a common case. Users can just not be logged in:

hmm why are we deprecating the token? can't we just keep the token?

api/config.js
faboweb: I looked for the cheapest provider. It's not really high tech to send emails.

api/lib/notifications/notifications.js
faboweb: fixed

Bitcoinera:
> hmm why are we deprecating the token? 

No, this has nothing to do with the token. This happens when you enter the app without being logged in. Which is a very common case. I was trying to say that it shouldN'T be an error

faboweb:
The token is only evaluated in the API if it is present. So we need to remove invalid tokens.

api/README.md
```diff
@@ -45,7 +45,7 @@ When running you can access GraphQL Playground at http://localhost:4000/. You ca
 | `REDIS_URL` | Redis server URI in `<auth>@hostname:port` format |
 | `ENGINE_API_KEY` | Enable Apollo GraphQL metrics through https://engine.apollographql.com/  |
 | `SENTRY_DSN` | Sentry (https://sentry.io) data source name in format `https://<key>@sentry.io/<project>` |
-| `RUN_ONLY_NETWORK` | Run only the network with this ID (allows scaling and better development) |
+| `RUN_ONLY_NETWORKS` | Run only the networks with this ID comma separated (allows scaling and better development) |
```
Bitcoinera, line 48: uuu, yeeey! :tada:

api/config.js
```diff
@@ -7,6 +7,7 @@ module.exports = {
   env: env,
   port: process.env.PORT || '4000',
   queryPath: '/',
+  frontendURL: "https://app.lunie.io",
   transactionPath: '/transaction',
```
Bitcoinera, line 11: lint please

faboweb: linter doesn't complain 🤷‍♂️

Bitcoinera:
How does the registering work? I just logged in yesterday. So it happens automatically? 

Well, I saw you can easily unsubscribe... but how do you resubscribe if I change my mind?

Also find it kind of invasive to just start spamming users once the log in. Cannot we give them more control over their subscriptions?

api/config.js
Bitcoinera: :shrug: we never solved the different linter opinions then after all?

faboweb:
We can do a lot but this is simplest first step to provide users with critical emails. To be honest I think in no app you opt in to emails. They just start spamming you. So maybe we shouldn't be too careful here. You also don't get spammed. The amount of emails should be pretty low as the topics are only critical cases.

Bitcoinera:
> To be honest I think in no app you opt in to emails.

:thinking:  true

api/lib/notifications/notifications-types.js
```diff
@@ -79,8 +79,34 @@ const getDefaultSubscriptions = async (addresses, dataSources) => {
...
+    const delegations = await dataSources[
+      networkId
+    ].api.getDelegationsForDelegatorAddress(address)
+
+    delegations.forEach((delegation) => {
+      subscriptions.push(
+        `${delegation.validatorAddress}_${eventTypes.VALIDATOR_COMMISSION}_${networkId}`,
+        `${delegation.validatorAddress}_${eventTypes.VALIDATOR_STATUS}_${networkId}`,
+        `${delegation.validatorAddress}_${eventTypes.SLASH}_${networkId}`,
+        `${delegation.validatorAddress}_${eventTypes.LIVENESS}_${networkId}`
```
Bitcoinera, line 95: So this subscribes you by default to your own validators' updates right?

Bitcoinera: Nice

api/lib/notifications/notificationController.js
```diff
@@ -0,0 +1,139 @@
...
+    dataSources
+  ) {
+    let topics = []
+    if (notificationType === 'email') {
+      topics = (
+        await getDefaultEMailSubscriptions(addressObjects, dataSources)
+      ).map((topic) => ({ topic, type: 'email' }))
+    } else {
+      throw new UserInputError(
+        'Only Email notfication registration is supported on this endpoint'
```
Bitcoinera, line 74:
```suggestion
        'Only Email notification registration is supported on this endpoint'
```

api/lib/notifications/notificationController.js
```diff
@@ -0,0 +1,139 @@
...
+class NotificationController {
+  constructor(networks) {
+    this.db = database(config)
+    this.registrations = {
+      'cosmos1px678cw4m5j5ye52ydjapf2kcw5edn0p6u9cry_transactionSend_cosmos-hub-testnet': {
+        email: { abc: true },
+        push: {}
+      }
+    }
+    // TODO networks might not update here
```
Bitcoinera, line 19: This would be a very rare case no? I mean, networks need to change right after you pass them here. Highly unlikely?

api/lib/schema.js
```diff
@@ -409,6 +413,12 @@ const typeDefs = gql`
     networkId: String!
   }
 
+  input NotificationSetting {
+    topic: String!
+    type: String!
+    remove: Boolean
```
Bitcoinera, line 419: What is this remove thing?

Bitcoinera: ah, it is commented out in line 60 in notificationsController.js. Ah I see, for the future to delete notifications registrations

api/lib/source/cosmosV0-source.js
```diff
@@ -359,10 +359,13 @@ class CosmosV0API extends RESTDataSource {
     const balances = balancesResponse || []
     // the user might not have liquid staking tokens but have staking tokens delegated
     // if we don't add the staking denom, we would show a 0 total for the staking denom which is wrong
-    if (delegations.length > 0 && !balancesResponse.find(({denom}) => denom === this.network.stakingDenom)) {
+    if (
+      delegations.length > 0 &&
+      !balancesResponse.find(({ denom }) => denom === this.network.stakingDenom)
```
Bitcoinera, line 364:
```suggestion
      !balancesResponse.find(({ denom }) => this.reducers.denomLookup(denom) === this.network.stakingDenom)
```

api/lib/stores/block-store.js
```diff
@@ -258,6 +267,50 @@ class BlockStore {
...
+  checkProposalsUpdate(oldPropsals, newProposals) {
+    const oldProposalsDictionary = keyBy(oldPropsals, 'id')
+    const newProposalsDictionary = keyBy(newProposals, 'id')
+
+    // Safety check
+    // On the first run we don't have old proposals in the store
+    if (oldPropsals.length === 0) return
+
+    // case 1: New proposal
+    const newProposalIds = difference(
```
Bitcoinera, line 280: cool, I didn't know this difference function from Lodash

app/src/ActionModal/components/DelegationModal.vue
```diff
@@ -352,6 +352,9 @@ export default {
     },
     onSuccess(event) {
       this.$emit(`success`, event)
+
+      // update registered topics for emails as the validator set changed
+      this.$store.dispatch("updateEmailRegistrations")
```
Bitcoinera, line 357: smart

app/src/vuex/modules/session.js
```diff
@@ -172,6 +172,9 @@ export default ({ apollo }) => {
       const allSessionAddresses = await dispatch("getAllSessionAddresses")
       commit("setAllSessionAddresses", allSessionAddresses)
 
+      // update registered topics for emails
+      dispatch("updateEmailRegistrations")
```
Bitcoinera, line 176: We could check here if the user is signed in or not with rootState and thus prevent throwing the error of `Notifications need a signed in user` for no reason

Bitcoinera:
```
      rootState: {
        account: { userSignedIn },
      },
```
Althought I think we need to include a `setTimeOut` because it takes a bit for the Firebase observer to change this

api/lib/notifications/notifications-types.js
Bitcoinera: When does the Liveness notification get triggered? How many blocks does the validator need to miss?

app/src/vuex/modules/session.js
Bitcoinera: I had to add the Timeout here for example, otherwise it would redirect to the Paywall: https://github.com/luniehq/lunie/pull/4495/commits/80938ca978bd93943a2c889c89d6d838373f7f42

api/lib/notifications/notificationController.js
faboweb: The networks get passed only add API initialization. So whenever you update them in the db they don't get updated here? (maybe they are because we are actually manipulating the network object)

api/lib/notifications/notifications-types.js
faboweb: It doesn't get triggered yet. See #4504

app/src/vuex/modules/session.js
faboweb: https://github.com/luniehq/lunie/commit/80938ca978bd93943a2c889c89d6d838373f7f42#r40626643

Bitcoinera: Should we do this? Feels strange to get that error also if the user is simply not logged in

faboweb: o.O that is not the case, is it?

faboweb: ahh I get it

Bitcoinera:
Looking great!!

PR #4488, opened 2020-07-09
TmPage refactor

### LINKED ISSUE #4042: refactor TmPage

- padding is inconsistent across pages
- there is some unused code
- css may be duplicated / not in the correct component file

jbibla:
Closes #4042 

- pulled TmBalance out of TmPage and got rid of hide-header logic (which was really backwards!) 
- removed a bunch of unnecessary props and the "managed" concept from TmPage (mostly not being used anyways)
- quick clean up if proposal UI (just until we have the new designs)

---

For contributor:

For reviewer:


faboweb:
🚀 Deployed on https://5f079f83bf42baf62eb33260--lunieio.netlify.app

Bitcoinera:
> * pulled TmBalance out of TmPage and got rid of hide-header logic (which was really backwards!)
> * removed a bunch of unnecessary props and the "managed" concept from TmPage (mostly not being used anyways)
> * quick clean up if proposal UI (just until we have the new designs)

Awesome. This was really needed

app/src/components/common/FeatureNotAvailable.vue
```diff
@@ -2,11 +2,11 @@
   <TmDataMsg icon="timeline">
     <div slot="title">Feature not available</div>
     <div slot="subtitle">
-      This feature hasn't been implemented on this network yet.
+      We haven't implemented this feature for this network yet.
       <br />
       <br />
-      <a class="intercom-button" href="#"
-        >Reach out if you'd like us to enable this feature.</a
+      <a class="intercom-button" @click="$store.dispatch(`displayMessenger`)"
```
jbibla, line 8: i actually don't know if this works

app/src/components/common/TmPage.vue
```diff
@@ -1,101 +1,43 @@
...
-        <TmBalance v-if="session.signedIn" />
-      </slot>
-      <slot slot="header-buttons" name="header-buttons" />
-    </TmPageHeader>
-    <main
-      class="tm-page-main"
-      :class="{ 'dark-background': darkBackground && session.signedIn }"
-    >
-      <CardSignInRequired v-if="signInRequired && !session.signedIn" />
-      <template v-else-if="managed">
```
jbibla, line 20: there may be some issues related to removing this - would be helpful if you manually QA'd this by clicking around and seeing if there are any strange behaviours.

app/src/components/wallet/PagePortfolio.vue
```diff
@@ -1,23 +1,31 @@
 <template>
-  <TmPage :sign-in-required="true" :managed="true" :dark-background="true">
-    <template slot="managed-body">
+  <TmPage :sign-in-required="true" :dark-background="true">
```
jbibla, line 2: does it make sense to have `:sign-in-required="true"` and `<template slot="signInRequired">` ?

i think it does but there might be a smarter / simpler way.

app/src/components/common/TmPage.vue
```diff
@@ -109,19 +51,11 @@ export default {
     ...mapState([`session`]),
     ...mapGetters([`connected`]),
   },
-  watch: {
-    $route() {
-      this.scrollContainer.scrollTop = 0
```
jbibla, line 114: my understanding is that this isn't actually necessary / doing anything because of https://github.com/luniehq/lunie/pull/4488/files#diff-e2d8a102050232d3ad7188cabd6720b5R64

but i could be wrong

app/src/router.js
```diff
@@ -82,6 +88,6 @@ async function featureAvailable(store, networkSlug, feature) {
```
jbibla, line 88: this featureAvailable logic feels hacky to me. why not just display the message on the page? is the redirect really necessary?

app/tests/unit/specs/components/common/PageFeatureNotAvailable.spec.js
```diff
@@ -1,14 +1,30 @@
...
+const localVue = createLocalVue()
+
+localVue.use(Vuex)
 
 describe(`PageFeatureNotAvailable`, () => {
   let wrapper
+  let store
+
   beforeEach(() => {
+    store = new Vuex.Store({
```
jbibla, line 13: feels funny to me that a store is required to test this — but c'est la vie

app/src/components/common/FeatureNotAvailable.vue
Bitcoinera: mmm, I would rather stay safe. You can copy the code from somewhere else. Like this one:

```
    handleIntercom() {
      this.$store.dispatch(`displayMessenger`)
    },
```

You can find it in five different components in our app, so it is sure that this one works

app/src/components/common/TmPage.vue
Bitcoinera: that code looks cool. I didn't even know you could do that with Vue. Let me check

app/src/router.js
Bitcoinera: I think the redirect concretely is the least complex thing in that logic. You'd need to have nearly the same code to do what you are proposing. As long as we fetch availability from the network Object, that is. 

Maybe we could do this in another way, like having a super simple switch in DB that directly connects to the Vue router :shrug: I don't know. To me it looks fine the logic as it currently is

app/src/components/common/TmPage.vue
Bitcoinera: I saw that margins are suddenly disappearing in really thin screens like you can see here:

image

But I don't know if this is TmPage refactor related. Networks and the rest are doing fine. It is only Proposals it seems

Bitcoinera:
I feel tempted to merge this right away. It has been a while this TmPage was asking to be refactored

app/tests/unit/specs/components/common/PageFeatureNotAvailable.spec.js
faboweb: we don't need to init a store here, right? it can just be n object as in all the other tests.

app/src/components/common/FeatureNotAvailable.vue
faboweb: `displayMessenger`  only works on mobile, right?

app/src/components/common/TmPage.vue
```diff
@@ -1,101 +1,43 @@
...
-        <slot v-else name="managed-body" />
-      </template>
-      <slot />
-    </main>
+  <div
+    class="page"
+    :class="{ 'dark-background': darkBackground && session.signedIn }"
+  >
+    <CardSignInRequired v-if="signInRequired && !session.signedIn" />
+    <TmDataConnecting v-if="!connected" />
```
faboweb, line 7: are we showing this at any point?
```suggestion
```

app/src/components/governance/PageProposal.vue
```diff
@@ -1,6 +1,6 @@
 <template>
-  <TmPage data-title="Proposal" hide-header class="small">
-    <TmDataLoading v-if="$apollo.loading && !loaded" />
+  <TmPage data-title="Proposal" class="readable-width">
```
faboweb, line 2: use `loading` and `error` of TmPage?

app/src/components/network/PageBlock.vue
```diff
@@ -1,12 +1,11 @@
...
   <TmPage
     data-title="Block"
-    :managed="true"
     :loading="this.$apollo.queries.block.loading"
     :error="error"
-    hide-header
+    class="readable-width"
   >
-    <template slot="managed-body">
+    <template v-if="!this.$apollo.queries.block.loading">
```
faboweb, line 8: managed-body does this already plus it also hides on error, right?

app/src/components/notifications/PageNotifications.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <TmPage data-title="My alerts" hide-header>
+  <TmPage data-title="My alerts">
     <TmDataLoading v-if="$apollo.loading && notifications.length === 0" />
```
faboweb, line 3: use TmPage loading?

app/src/components/staking/PageValidator.vue
```diff
@@ -1,14 +1,11 @@
...
     :data-empty="!validator.operatorAddress"
     :loading="$apollo.queries.validator.loading"
-    :loaded="loaded"
-    :hide-header="true"
-    data-title="Validator"
-    class="small"
+    class="readable-width"
   >
-    <template v-if="validator.operatorAddress" slot="managed-body">
+    <template v-if="validator.operatorAddress">
```
faboweb, line 8: see above

app/src/components/staking/PageValidators.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <TmPage :managed="true" hide-header>
+  <TmPage>
     <div
```
faboweb, line 3: use TmPage loading

app/src/components/wallet/PagePortfolio.vue
faboweb: why the extra slot? `CardSignInRequired` should replace the managed content

app/src/components/wallet/PageTransactions.vue
```diff
@@ -1,17 +1,13 @@
...
+    :loading="$apollo.queries.transactions.loading"
+    :error="$apollo.queries.transactions.error"
     :sign-in-required="true"
-    :hide-header="true"
   >
-    <DataEmptyTx slot="no-data" />
-    <template slot="managed-body">
-      <div>
+    <template slot="signInRequired">
+      <DataEmptyTx v-if="!$apollo.queries.transactions.loading && dataEmpty" />
```
faboweb, line 9: add data empty to TmPag?

app/src/routes.js
```diff
@@ -45,7 +45,7 @@ export default (store) => {
         session: () => import(`./components/common/TmSessionSignIn`),
       },
       meta: {
-        feature: "Session",
+        feature: "session",
```
faboweb, line 48: ur aware that on PageFeatureNotAvailable now the message shows `session` instead of `Session`?

app/src/components/network/PageBlock.vue
Bitcoinera: What does `managed-body` actually do?

faboweb: `managed-body` is the content of the page that shows after everything is loaded and no errors appeared. It is defined as a slot in TmPage

Bitcoinera: Oh thank you. Now I found the code for it: `<slot v-else name="managed-body" />`

Yes, and it is already doing the loading check as you said (line 22)

app/src/components/common/FeatureNotAvailable.vue
jbibla: > You can find it in five different components in our app

smells funny. we shouldn't have to duplicate a method 5 times. i guess it can be moved to the store?

app/src/components/common/TmPage.vue
jbibla: which margins?

Bitcoinera: on the sides

app/src/components/common/TmPage.vue
jbibla: nope. nice find. the logic for connected doesn't even exist - it should all be removed. i'll do that as well.

https://github.com/luniehq/lunie/blob/develop/app/src/vuex/getters.js#L3
https://github.com/luniehq/lunie/blob/develop/app/src/vuex/modules/connection.js#L19

app/src/components/governance/PageProposal.vue
jbibla: yep i don't see why not!

app/src/components/network/PageBlock.vue
jbibla: true, thanks.

app/src/components/notifications/PageNotifications.vue
jbibla: yep. this page was not implemented correctly 😞

app/src/components/staking/PageValidators.vue
jbibla: ah so this is a bit tricky because we have the svg loader. should we / can we pass this in to TmPage?

app/src/components/wallet/PageTransactions.vue
jbibla: `DataEmptyTx` is just for transactions. i guess we should put a `DataEmpty` into `TmPage` and use props for the specific wording / language?

app/src/router.js
jbibla: it's not really about complexity - though i guess if this is the simplest solution that's an ok argument. why is a redirect the best way to solve this problem? it feels like a hacky way of keeping someone away from a page.

app/tests/unit/specs/components/common/PageFeatureNotAvailable.spec.js
jbibla: can you share an example of what you mean?

app/src/components/wallet/PagePortfolio.vue
jbibla: which slot is extra?

app/src/routes.js
jbibla: nope. we never actually used `meta` in the message.

https://github.com/luniehq/lunie/pull/4488/files/70cff90e8b23677f6a14523fd4baa0a1fb007ff5#diff-fa9c8bee29345ffb1ef448a7fd064882R9

app/src/components/common/TmPage.vue
jbibla: i think that's how it is in production already - not caused by this change

Bitcoinera:
Is this ready for second review?

app/src/routes.js
Bitcoinera: This is easily "checkable"

Bitcoinera: Yes, it doesn't appear:

image

Bitcoinera:
Bug:

app/tests/unit/specs/components/common/PageFeatureNotAvailable.spec.js
faboweb: check other tests for comparison. we never instantiate a store in a test

app/src/components/staking/PageValidators.vue
faboweb: yes. you can define the svg as the loader component

app/src/routes.js
faboweb: ok, then we removed it at some point

jbibla:
no, i'm still working on the updates from the comments.

@Bitcoinera - fix double balances should not go in this PR - please revert it and make a separate PR.

app/tests/unit/specs/components/common/PageFeatureNotAvailable.spec.js
jbibla: yeah we do in 3 places - that's where i copied it from

api/lib/source/cosmosV0-source.js
```diff
@@ -359,10 +359,13 @@ class CosmosV0API extends RESTDataSource {
     const balances = balancesResponse || []
     // the user might not have liquid staking tokens but have staking tokens delegated
     // if we don't add the staking denom, we would show a 0 total for the staking denom which is wrong
-    if (delegations.length > 0 && !balancesResponse.find(({denom}) => denom === this.network.stakingDenom)) {
```
jbibla, line 362: @Bitcoinera this is still here?

app/src/components/wallet/PagePortfolio.vue
jbibla: @faboweb ping

api/lib/source/cosmosV0-source.js
Bitcoinera: Hey. The code is the same, I reverted the fix, but if you merge with `develop` the fix should be there

Bitcoinera: ah, you need to do it anyway because of the conflict

api/lib/source/cosmosV0-source.js
```diff
@@ -361,10 +361,7 @@ class CosmosV0API extends RESTDataSource {
     // if we don't add the staking denom, we would show a 0 total for the staking denom which is wrong
     if (
       delegations.length > 0 &&
-      !balancesResponse.find(
-        ({ denom }) =>
-          this.reducers.denomLookup(denom) === this.network.stakingDenom
-      )
+      !balancesResponse.find(({ denom }) => denom === this.network.stakingDenom)
```
faboweb, line 364: ? this would fail right? the balance response has the balances not yet in Lunie format

Bitcoinera:
Did you resolve the empty thing discussion you were having? Can we merge this?

faboweb:
cardsignin is missing a padding
image

faboweb:
missing a padding on the top I think
image

jbibla:
unfortunately, this is already in production and not from this PR - it's because we didn't account for the hidden user menu styles

faboweb:
should we account for it?

jbibla:
yeah, but not in this PR CardSignIn doesn't use TmPage

Issue #4476, opened 2020-07-09
Mobile push notifications


faboweb:
As a mobile user, I want to be informed by Push on critical notifications.

Topics:
- slashes (not there yet)
- PROPOSAL_CREATE
- PROPOSAL_UPDATE
- VALIDATOR_STATUS
- VALIDATOR_COMMISSION

PRs:
- Api: Implement a way to push notifications to registered topics (recover PR)
- Api: Implement receiving push notifications in mobile app (recover code)
- FE: Register mobile apps to receive push notifications
- FE: Display push notifications and link to deep linked content

Bitcoinera:
Yeeey!!!

Bitcoinera:
We still need to test this

Bitcoinera:
We close it. I tested it and it is working. Also the link to the app. Only thing we need to change is that users shouldn't need to log in in the app as well. Webapp sessions should be enough to trigger the same notifications on mobile as for email

PR #4496, opened 2020-07-13
Fabo+Ana/4476 deep link for mobile sign in

### LINKED ISSUE #4293: User Authentication: Deep Links
This discussion is related to my PR #4242 and with the epic User Accounts #4237 

Everything was going just fine and happily with magic links until I was reminded about the app. Yes, how could users use the email magic link on the mobile app? I went on researching and realized that Dynamic links are a must to be able to have magic links work the expected way on mobile. This is easily noticeable because once you use the android field in the `[actionCodeSettings](https://firebase.google.com/docs/auth/web/passing-state-in-email-actions)` from Firebase, you automatically get the following error:

image

I just tried one of the magic links on mobile and, as I imagined, it directed me to the webapp. But the behavior we'd expect from the dynamic link on mobile would be to take the user to the app if they have it installed or even install it if they wouldn't have it (it is as easy as doing 

```
        android: {
          packageName: `org.lunie.lunie`, // no clue?
          installApp: true,
          minimunVersion: `minimum-version`
        },
```

So, to have the email sign in mobile app ready we need to set up the Firebase Dynamic Links, there is no question in that. Now, what @faboweb discovered is that in order to set them up, we'd need to have our app hosted on Firebase, which is too much of a tradeoff for what we want to achieve.

Possible solutions:

- We use another Firebase signin method. For starters, normal Email and Password. As a temporary solution, to later user the Email Magic links.

- We use another authentication service like OAuth or [Auth0](https://auth0.com/docs/connections/passwordless/guides/email-magic-link) using their magic link solution, or we could even [go native](https://medium.com/one-more-thing-studio/how-to-make-magic-links-with-node-1d164c036e29)! (although it has a lot of security concerns)

- We let mobile users login and use the webapp instead of the app for now? (Super bad solution)

Bitcoinera:
closes #4293 
Related to #4476 

**Description:**

The idea is to enable deep linking using capacitor to be able to send the proper link to Firebase depending if the user is on a mobile or PC/laptop device.

Stuck on the iOS part of [this guide](https://capacitorjs.com/docs/guides/deep-links)

Thank you! 🚀

---

For contributor:

For reviewer:


Bitcoinera:
Also not sure if the `**.keystore` file should be also in our codebase... I guess so? But where, in which folder. Is it OK to be public?

faboweb:
🚀 Deployed on https://5f0c85ae682c2f97bf6f3b18--lunieio.netlify.app

Bitcoinera:
OK, I put my name on that changelog, but we can change that later. I don't feel myself quite an expert on this field here xD

faboweb:
> Also not sure if the **.keystore file should be also in our codebase... I guess so

No, these files should only be added at buildtime in Bitrise

app/public/.well-known/assetlinks.json
```diff
@@ -0,0 +1,14 @@
+[
```
faboweb, line 1: I can't checkout this file ^^'

faboweb:
Missing the asset-links and the flow for sign in should be optimised with call to actions

Bitcoinera:
> Missing the asset-links and the flow for sign in should be optimised with call to actions

I shared with you the asset-links file or what do you mean?

Bitcoinera:
OK, so to test this I guess I need to use Android Studio right

faboweb:
The asset-link file you created wasn't signed with the keystone we use to sign the app so it will not work. This is also not yet tested on iOS. @jbibla will take care of that

Bitcoinera:
> The asset-link file you created wasn't signed with the keystone we use to sign the app so it will not work

Here you can then generate a working asset-link file using the right SHA-256: https://developers.google.com/digital-asset-links/tools/generator

jbibla:
waiting for validation: https://search.developer.apple.com/appsearch-validation-tool

Bitcoinera:
> waiting for validation: https://search.developer.apple.com/appsearch-validation-tool

cool :+1:

jbibla:
https://app.lunie.io/.well-known/apple-app-site-association

Bitcoinera:
> https://app.lunie.io/.well-known/apple-app-site-association

So good to go?

Bitcoinera:
OK, let's merge!! :woman_dancing:

Bitcoinera:
121 files changed? Seem there was some pollution between branches I think

app/.gitignore
```diff
@@ -6,6 +6,7 @@ node_modules
 /lunie-backend
 
 /tests/e2e/reports/
+/android/ANDROID-DEEP-LINK.keystore
```
faboweb, line 9:
```suggestion
```

app/src/vuex/modules/account.js
```diff
@@ -121,3 +142,43 @@ export default ({ apollo }) => {
     actions,
   }
 }
+
+export async function getLaunchUrl(router) {
```
faboweb, line 146:
```suggestion
// Deep Linking: The app can receive a deep link. This link needs to be forwarded to the web app. https://capacitorjs.com/docs/guides/deep-links
export async function getLaunchUrl(router) {
```

app/src/vuex/modules/account.js
```diff
@@ -121,3 +142,43 @@ export default ({ apollo }) => {
...
+
+  const queryObject = query
+    .split("&")
+    .map((keyValue) => keyValue.split("="))
+    .reduce((query, [key, value]) => {
+      query[key] = value
+      return query
+    }, {})
+
+  // if we receive a deeplink for firebase authentication we follow that link
```
faboweb, line 170:
```suggestion
  // if we receive a deeplink for firebase authentication we follow that link
  // the target will perform the authentication and then redirect back to lunie
```

PR #4514, opened 2020-07-14
Fabo/refactor listeners

faboweb:
Intent to simplify the block listeners as the code should work 90% the same everywhere as our source files already unify the interface to the data.


faboweb:
🚀 Deployed on https://5f0de1740b196365dd6a410d--lunieio.netlify.app

faboweb:
Would also mean we could remove the listener source class from the networks table

PR #4533, opened 2020-07-16
Ana/4518 fix expired idToken in API

### LINKED ISSUE #4518: Session Management

This is more a bug: Currently the user will have failing requests after using Lunie the next day as the idToken stored locally is invalid. We can't just create an idToken for every request as this would slow down the connection.

Implement: https://firebase.google.com/docs/auth/admin/manage-cookies

Store the lifetime of the idToken locally and refresh the token before it expires (on app start check and set a timeout to check)

**Additional context**

Bitcoinera:
Closes #4518

**Description:**

Fixes sending an expired idToken from `localStorage`to API. Concretely, the error coming from these lines in the `apollo.js` file:

```
        if (idToken) {
          uid = await checkIsValidIdToken(idToken)
        }
```

Thank you! 🚀

---

For contributor:

For reviewer:


faboweb:
🚀 Deployed on https://5f1096d4d6be160260d6ba50--lunieio.netlify.app

app/src/vuex/modules/account.js
```diff
@@ -42,6 +44,23 @@ export default ({ apollo }) => {
...
+          commit(`setUserInformation`, user)
+
+          // retrieving refreshed idToken
+          const idToken = await user.getIdToken(/* forceRefresh */ true)
+
+          // really important part, store idToken in localstorage so Apollo gets the newest
+          localStorage.setItem(`auth_token`, idToken)
+          // make sure new authorization token get added to header
+          apollo.cache.reset()
+        }
```
faboweb, line 61: is there a case where the user is null and we need to remove the auth token?

Bitcoinera: the problem is not if the user is `null`, but that apollo has already cached the old idToken when we do
```
    // Retrieve the authorization token from local storage.
    const token = localStorage.getItem("auth_token")
```
in `createApolloClient` in `apollo.js`

Bitcoinera:
Getting some errors here. Making a WIP for now:

image

app/src/vuex/modules/account.js
```diff
@@ -26,40 +26,46 @@ export default ({ apollo }) => {
     async listenToAuthChanges({ dispatch, commit }) {
       const Auth = (await getFirebase()).auth()
       // start listening for idToken changes too
-      dispatch(`listenToIdTokenChanges`)
-      await Auth.onAuthStateChanged(async (user) => {
-        if (user) {
-          commit(`userSignedIn`, true)
-          commit(`setUserInformation`, user)
+      await dispatch(`listenToIdTokenChanges`)
```
Bitcoinera, line 29: ah yeah, maybe that was it

Bitcoinera: No yes, it was most probably it

PR #4534, opened 2020-07-16
Ana/fix page notifications loading logic

Bitcoinera:
Closes #ISSUE

**Description:**

## Before:

Peek 2020-07-16 21-38

## After:

Peek 2020-07-16 21-42

I also think it is better to change the `moreAvailable` logic. If we want to filter some notifications in `notifications.js`, the current logic doesn't account for that.

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/app/src/components/notifications/PageNotifications.vue
+++ b/app/src/components/notifications/PageNotifications.vue
@@ -1,7 +1,9 @@
 <template>
   <TmPage
     data-title="My alerts"
-    :loading="$apollo.queries.notifications.loading"
+    :loading="
+      $apollo.queries.notifications.loading && notifications.length === 0
+    "
     :empty="notifications.length === 0"
     :empty-title="`You don't have any notifications yet`"
     :empty-subtitle="`Don't worry, they are on their way!`"
@@ -24,6 +26,14 @@
           </router-link>
         </template>
       </EventList>
+      <div
+        v-if="
+          $apollo.queries.notifications.loading && !dataLoaded && moreAvailable
+        "
+        class="spinner-container"
+      >
+        <img src="/img/spinner_blue@256.gif" class="spinner" />
+      </div>
     </template>
   </TmPage>
 </template>
@@ -76,6 +86,7 @@ export default {
           },
           // Transform the previous result with new data
           updateQuery: function (previousResult, { fetchMoreResult }) {
+            this.moreAvailable = fetchMoreResult.notifications.length === 0
             return {
               notifications: [
                 ...previousResult.notifications,
@@ -120,7 +131,6 @@ export default {
       update(result) {
         this.dataLoaded = true
         // assume that when the full page got loaded, that there is more
-        this.moreAvailable = result.notifications.length % 20 === 0
         return result.notifications
       },
       /* istanbul ignore next */
--- a/app/src/scripts/setNetwork.js
+++ b/app/src/scripts/setNetwork.js
@@ -74,7 +74,5 @@ export const setNetwork = async ({ to, next }, store) => {
 }
 
 function networkOrAny(network, networks) {
-  return network
-    ? network
-    : networks[0]
+  return network ? network : networks[0]
 }
--- a/app/src/vuex/modules/connection.js
+++ b/app/src/vuex/modules/connection.js
@@ -50,9 +50,10 @@ export default function ({ apollo }) {
       if (persistedNetwork && storedNetwork) {
         await dispatch(`setNetwork`, storedNetwork)
       } else {
-        const network = state.networks.find(
-          (network) => network.id === state.externals.config.network
-        ) || state.networks[0]
+        const network =
+          state.networks.find(
+            (network) => network.id === state.externals.config.network
+          ) || state.networks[0]
         // remove additional execution of checkForPersistedNetwork
         await dispatch(`setNetwork`, network)
       }
--- a/app/src/vuex/modules/notifications.js
+++ b/app/src/vuex/modules/notifications.js
@@ -1,29 +1,32 @@
 import gql from "graphql-tag"
 
-export default ({apollo}) => {
+export default ({ apollo }) => {
   const state = {}
 
   const mutations = {}
 
   const actions = {
-    async updateEmailRegistrations({rootState}) {
+    async updateEmailRegistrations({ rootState }) {
       // if the user is not signed in, we can't register topics for email notifications on the server
       if (!rootState.account.userSignedIn) return
-      
+
       const addressObjects = rootState.session.allSessionAddresses.map(
         ({ networkId, address }) => ({ networkId, address })
       )
       apollo.mutate({
         mutation: gql`
           mutation($addressObjects: [NotificationInput]!) {
-            notifications(addressObjects: $addressObjects, notificationType: "email")
+            notifications(
+              addressObjects: $addressObjects
+              notificationType: "email"
+            )
           }
         `,
         variables: {
-          addressObjects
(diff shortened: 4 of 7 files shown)


faboweb:
🚀 Deployed on https://5f10b279355f43025350c9d1--lunieio.netlify.app

jbibla:
we can merge this, but why not replace TmLoader instead of this?

Bitcoinera:
> we can merge this, but why not replace TmLoader instead of this?

Sure. I am just not sure of what you mean

app/src/components/notifications/PageNotifications.vue
```diff
@@ -1,7 +1,9 @@
 <template>
   <TmPage
     data-title="My alerts"
-    :loading="$apollo.queries.notifications.loading"
+    :loading="
+      $apollo.queries.notifications.loading && notifications.length === 0
```
faboweb, line 5:
```suggestion
      $apollo.queries.notifications.loading && !dataLoaded
```

Bitcoinera: oki

Bitcoinera:
Why so impatient? I am getting the same behavior than before. Jumping again.

Too impatient @jbibla :stuck_out_tongue:

Bitcoinera:
The reason is because when we do `loadMore` we assign `dataLoaded` to false. `dataLoaded`doesn't mean first load, that is why I had to do the `notifications.length === 0` thing

faboweb:
But it is 2 different ways of accounting for loading states. One is initial loading which should be handled by TmPage and one is updating of data or loading more data which should be in the components, shouldn't it?

Bitcoinera:
> we don't need to do this in this PR but we should do this at some point.

I mean, the logic in `TmPage` should be super easy. I'm just unsure how that spinning logo would look like as a replacement of `TmDataLoading`. I like the rectangle as a shape... would it be inside some kind of frame, the spinning logo?

Bitcoinera:
> But it is 2 different ways of accounting for loading states.

Yes, that is the logic that the TmPage would hangle. Two loading states. Amazing sum up

jbibla:
just the spinner would probably be fine

Bitcoinera:
> just the spinner would probably be fine

OK, then I can do this on this PR. Just I need to leave now since they are closing the CryptoInstitute. Give me a bit

Bitcoinera:
It is a bit more difficult than I though inserting the spinning loader from TmPage into the slot from the page... I would probably need a new named slot here and do like in PagePortfolio. Yes, I need some intermediary component to do the "fusion" I think

app/src/ActionModal/components/ActionModal.vue
```diff
@@ -43,9 +43,12 @@
       <template v-if="!checkFeatureAvailable">
         <FeatureNotAvailable :feature="title" />
       </template>
-      <TmDataLoading
+      <div
```
jbibla, line 46: i think it would have been easier to just delete the contents of TmDataLoading and replace with 🤷 

``

Bitcoinera: OK, yes, I will change it to that

Bitcoinera:
I have also fixed this (finally got rid of duplicates):

image


We can deprecate the `uniqBy` by fixing this in the API in another PR

app/src/components/notifications/PageNotifications.vue
```diff
@@ -1,7 +1,10 @@
 <template>
   <TmPage
     data-title="My alerts"
-    :loading="$apollo.queries.notifications.loading"
+    :loading="$apollo.queries.notifications.loading && !firstLoaded"
```
jbibla, line 4: not your fault ana, but i think this is still confusing logic. 

we have `loading` and `loading-more` props + `firstLoaded` `dataLoaded` `moreAvailable`

not necessarily in this PR but maybe we can simplify to:

`firstLoad`
`loadMore`

and use computed properties instead of `$apollo.queries.notifications.loading && !dataLoaded && moreAvailable` in the template?

app/src/components/notifications/PageNotifications.vue
```diff
@@ -216,14 +225,4 @@ img {
   font-weight: 400;
   overflow-wrap: anywhere; /** Important. Otherwise awful style bug */
 }
-
```
jbibla, line 219: great!

app/src/styles/app.css
```diff
@@ -130,6 +130,17 @@ input.tm-field {
   animation: fade 1.5s infinite;
 }
 
+.spinner-container {
```
jbibla, line 133: this should go in TmDataLoading not here.

jbibla:
do we know why there are so many lint changes?

app/src/components/notifications/PageNotifications.vue
Bitcoinera: true, we could use `firstLoaded` for loading and maybe `dataLoaded` for `loading-more`. Let me try that

Bitcoinera:
> do we know why there are so many lint changes?

:shrug:

Bitcoinera:
OK, all you just said fixed, Jordan!

app/src/components/common/TmPage.vue
```diff
@@ -22,14 +22,16 @@
       :subtitle="emptySubtitle"
     />
 
-    <slot v-else-if="!loading && !empty"></slot>
+    <template v-else-if="!loading && !empty">
+      <slot></slot>
+      <TmDataLoading v-if="!loading && loadingMore" class="loading-more" />
```
jbibla, line 27: we don't need this class "loading-more" the styles should go in `TmDataLoading`

app/src/components/common/TmPage.vue
```diff
@@ -153,6 +159,10 @@ h4 {
   line-height: 1rem;
 }
 
+.loading-more {
```
jbibla, line 162: move this to TmDataLoading

app/src/components/notifications/PageNotifications.vue
```diff
@@ -1,7 +1,8 @@
 <template>
   <TmPage
     data-title="My alerts"
-    :loading="$apollo.queries.notifications.loading"
+    :loading="!firstLoaded"
```
jbibla, line 4: that's not exactly what i meant - does this even work? it's not using `$apollo.queries.notifications.loading` anymore?

jbibla: just revert this - we will do a refactor later

app/src/components/common/TmPage.vue
Bitcoinera: we actually do. It needs to appear on the bottom when it is the `loadingMore`case

Bitcoinera: ah no sorry, the styles were simplified. OK, doing then

app/src/components/common/TmPage.vue
```diff
@@ -22,14 +22,16 @@
       :subtitle="emptySubtitle"
     />
 
-    <slot v-else-if="!loading && !empty"></slot>
+    <template v-else-if="!loading && !empty">
+      <slot></slot>
+      <TmDataLoading v-if="!loading && loadingMore" />
```
faboweb, line 27:
```suggestion
      <TmDataLoading v-if="loadingMore" />
```

Bitcoinera: then you get two different loaders at the beginning when opening the notifications page

Issue #4525, opened 2020-07-15
Cleaning up notifications


jbibla:
**now**
- [x] user menu ghost icon is not centred on windows
- [x] "Add an address" should go in user menu
- [x] `CardSignIn` should be modal instead of in page component
- [x] paywall variables and scroll bugs
- [x] message after sign in toggles states should be a modal that displays which addresses you'll receive notifications about -> link and CTA should go to notifications page
- [x] voting power changes message is broken (should show %)
- [x] remove new validator notification for now (and use network icon instead of lunie logo - for future use)
- [x] lunie theme on notifications page instead of network theme
- [x] hide app menu - add logo - deselect current network when on notifications page
- [x] notifications becomes home page / landing page if user is signed in with email
- [x] magic link copy needs to be updated
- [x] click the notifications icon from the paywall and the app gets stuck on a loading screen
- [x]  take user out from notifications back to Paywall when they sign out being on the notifications page

**next**
---
- [ ] connect user emails via intercom
- [ ] drip campaign 
- [ ] only display notifications that go back 30 days
- [ ] every user should have 1 notification from Lunie that says welcome

Bitcoinera:
I like this multi-issue! So much stuff to do! :sunglasses:

Bitcoinera:
> message after sign in toggles states should be a modal that displays which addresses you'll receive notifications about -> link and CTA should go to notifications page

You mean that the `email-authentication` page should be a modal with these characteristics? Sounds cool

Bitcoinera:
> voting power changes message is broken (should show %)

Seems correct? What do you think is broken?

![image](https://user-images.githubusercontent.com/40721795/87673496-f5ab4f00-c774-11ea-897e-6e38e709e335.png)

jbibla:
`increased from 0.214230 to 0.235088 (9.284% increase)`

should be

`increased from 0.21% to 0.24%`

faboweb:
Remember to do the issues as individual PRs

Bitcoinera:
> Remember to do the issues as individual PRs

But many of these things are super tiny... I would do all of the "now" ones simply in this PR

Bitcoinera:
Yes, I know what you are about to say: it's a one line change :laughing:

faboweb:
... I think this will be a lot smoother with multiple PRs

Bitcoinera:
> ... I think this will be a lot smoother with multiple PRs

But not 12 PRs?

[...]

PR #4546, opened 2020-07-20
Ana/4525 cleaning up notifications part 2

Bitcoinera:
Related to #4525
**Description:**

Fixes:

- [x] user menu ghost icon is not centred on windows
- [x] voting power changes message is broken (should show %)
- [x] take user out from notifications back to Paywall when they sign out being on the notifications page

Thank you! 🚀

---

For contributor:

For reviewer:


faboweb:
🚀 Deployed on https://5f159b5bf0c4e809f0a259f3--lunieio.netlify.app

jbibla:
let's get this merged! e2e is failing?

app/src/components/account/UserMenu.vue
```diff
@@ -358,6 +358,7 @@ h3 {
   font-size: 26px;
   transform: scale(0.75);
   top: 0;
+  padding-bottom: 0.25rem;
```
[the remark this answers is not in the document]
Bitcoinera: for the ghost icon to be centered

Bitcoinera:
> let's get this merged! e2e is failing?

this ValidatorList did not reset thing again. Rerunning

api/lib/notifications/notifications.js
```diff
@@ -69,9 +69,9 @@ function getMessageTitle(networks, notification) {
 
       return `Voting power decreased from ${Number(
         data.prevValidator.votingPower
-      ).toFixed(6)} to ${Number(data.nextValidator.votingPower).toFixed(
-        6
-      )} (${percentageDifference.toFixed(3)}% decrease) for ${
+      ).toFixed(2)}% to ${Number(data.nextValidator.votingPower).toFixed(
```
faboweb, line 72: refactor into some kind of helper function or at least the precision into a constant?

app/src/router.js
```diff
@@ -10,6 +10,10 @@ export const routeGuard = (store) => async (to, from, next) => {
   // Set any open modal to false
   store.state.session.currrentModalOpen = false
 
+  // notifications is landing page when user is signed in with email
+  if (to.fullPath.includes("portfolio") && store.state.account.userSignedIn) {
+    next(`/notifications`)
```
faboweb, line 15: don't need to return here?

app/tests/e2e/interface.spec.js
```diff
@@ -23,7 +23,7 @@ module.exports = {
 
         // checking validators did reset
         if (!isSameArray(validatorNames, updatedValidatorNames)) {
-          await sleep()
+          await browser.pause(2000)
```
faboweb, line 26: doesn't work. revert?

Bitcoinera: maybe it does. The last error is this:

```
   StatusCodeError: 503 - "<h1>This website is under heavy load</h1><p>We're sorry, too many people are accessing this website at the same time. We're working on this problem. Please try again later.</p>"
```

app/src/router.js
Bitcoinera: no :shrug:

api/lib/notifications/notifications.js
Bitcoinera: OK, but then it makes sense to also change this one (voting difference) to precision 2 no? To have it as a meaningful proper constant: `(${percentageDifference.toFixed(3)}% increase)`

faboweb: yes exactly

api/lib/notifications/notifications.js
```diff
@@ -69,9 +69,9 @@ function getMessageTitle(networks, notification) {
...
+        2
+      )}% (${percentageDifference.toFixed(3)}% decrease) for ${
```
faboweb, line 74:
```suggestion
      )}% (${percentageDifference.toFixed(2)}% decrease) for ${
```

app/src/router.js
```diff
@@ -10,6 +10,10 @@ export const routeGuard = (store) => async (to, from, next) => {
   // Set any open modal to false
   store.state.session.currrentModalOpen = false
 
+  // notifications is landing page when user is signed in with email
+  if (to.fullPath.includes("portfolio") && store.state.account.userSignedIn) {
```
faboweb, line 14: I guess this should actually be determined when entering the default route. This will open notifications always when the user goes to portfolio

api/lib/notifications/notifications.js
jbibla: should be done in another PR

app/src/components/notifications/PageNotifications.vue
```diff
@@ -47,8 +47,20 @@ export default {
...
   }),
   computed: {
-    ...mapState([`session`]),
+    ...mapState([`session`, `account`]),
     ...mapGetters([`networks`]),
+    userSignedIn() {
+      return this.account.userSignedIn
+    },
+  },
+  watch: {
```
jbibla, line 56: doesn't the router take care of this?

app/src/router.js
jbibla: vue router docs don't mention a need for returns

app/src/routes.js
```diff
@@ -11,7 +11,7 @@ export default (store) => {
   return [
     {
       path: `/`,
-      redirect: `/cosmos-hub/portfolio`,
+      redirect: `/notifications`,
```
jbibla, line 14: this is the simplest way to make notifications the home page. kind of radical - but also makes a lot of sense. 

if the user is not signed in they will be redirected to the paywall which is handled by the router guard. 

this will also really help with signups and exposure for notifications.

Bitcoinera: Mmm, I think this is too radical until we don't have really figured the final flow out. We can do this once we merge the "traditional" sign in with the new email-based one. But doing this right now is a bad idea (too confusing)

app/src/router.js
```diff
@@ -7,35 +7,33 @@ import { setNetwork } from "./scripts/setNetwork"
...
-      return
-    } else if (from.name === `paywall`) {
-      return
-    } else {
-      next(`/paywall`)
-    }
+
+  // redirect from notifications to paywall is user is not signed in
+  if (to.name === `notifications` && !store.state.account.userSignedIn) {
+    next(`/paywall`)
```
Bitcoinera, line 21: Well, it took me a while to fix this route. Maybe I should have left some comment about this, as it is not obvious. But if you are already in `/paywall` and you click on the notifications icon, then you are taken to notifications no matter what. That's why I had to add that `else if`:wink: It wasn't by whim

app/src/routes.js
jbibla: > merge the "traditional" sign in with the new email-based one.

what do you mean by this?

currently, we default to cosmos hub profile page. this is not really "not confusing" it's just how it's been for so long.

app/src/router.js
```diff
@@ -7,29 +7,35 @@ import { setNetwork } from "./scripts/setNetwork"
...
     const path = to.fullPath.substr(to.fullPath.indexOf("#") + 1)
     next(path)
-    return
   }
+
+  // redirect from notifications to paywall is user is not signed in
   if (to.name === `notifications`) {
     if (store.state.account.userSignedIn) {
       next()
+      // this is needed to prevent accessing notifications from paywall
```
jbibla, line 23: interesting! didn't realize this was the fix. but also, this feels hacky. do you know why double clicking would override the guard?

faboweb: did you debug what happens on the 2nd click? I think the bug with the 2nd click was on the other PR though.

app/src/routes.js
faboweb: We agreed, that if the user is not signed in, we forward him to the validators page not let him be stuck on notifications.

app/src/router.js
jbibla: also, i kind of like this being broken for now because otherwise there is no way to access notifications 😆

app/src/routes.js
jbibla: no no no we agreed that if the email modal closes and they are on a page that requires an address that we would send them to the validators page. we didn't discuss changing the landing page for non-email users.

anyways. we'll revert for now.

jbibla, line 14:
```suggestion
      redirect: `/cosmos-hub/portfolio`,
```

app/src/router.js
Bitcoinera: 2nd click? I mean, you are just taking to notifications for some reason, it ignores the `next(/paywall)`

Bitcoinera:
Let's merge!! :woman_dancing:

app/src/router.js
jbibla: > it ignores the next(/paywall)

yes. why?

jbibla: possibly: https://router.vuejs.org/guide/advanced/navigation-guards.html#global-before-guards

```
Make sure that the next function is called exactly once in any given pass through the navigation guard. It can appear more than once, but only if the logical paths have no overlap, otherwise the hook will never be resolved or produce errors. Here is an example of redirecting to user to /login if they are not authenticated:
```

faboweb: let'S fix the double click. try adding a return

> Make sure that the next function is called exactly once

that is what I meant before.

Bitcoinera: Okok. I can try that

Bitcoinera:
hhaha, loving those commits messages, but no luck it seems

jbibla:
the interface validator search test is failing again ...

Bitcoinera:
> the interface validator search test is failing again ...

Let's merge anyway, that one always fails

PR #4568, opened 2020-07-22
Ana/some polkadot fixes

Bitcoinera:
Closes #ISSUE

**Description:**

Fixes:

"currency selector doesn’t work" 
and
"blocks page doesn’t work - if you click it it just says loading forever"

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/api/lib/fiatvalues-api.js
+++ b/api/lib/fiatvalues-api.js
@@ -39,7 +39,8 @@ const denomToCoinGeckoIDDictionary = {
   KAVA: 'kava',
   BNB: 'binancecoin',
   // USDX: 'usdx', // not on coin gecko yet
-  KSM: 'kusama'
+  KSM: 'kusama',
+  DOT: 'polkadot'
 }
 
 const geckoIDToDenomDictionary = {
@@ -50,7 +51,8 @@ const geckoIDToDenomDictionary = {
   kava: 'KAVA',
   binancecoin: 'BNB',
   // usdx: 'USDX', // not on coin gecko yet
-  kusama: 'KSM'
+  kusama: 'KSM',
+  polkadot: 'DOT'
 }
 
 const fiatCurrenciesSymbolsDictionary = {
@@ -76,7 +78,7 @@ class fiatValueAPI {
     this.fiatValuesAPIDictionary = denomToCoinGeckoIDDictionary
     this.fiatValuesAPIReverseDictionary = geckoIDToDenomDictionary
     this.fiatCurrenciesSymbolsDictionary = fiatCurrenciesSymbolsDictionary
-    this.coins = ['ATOM', 'LUNA', 'KRT', 'SDT', 'KAVA', 'BNB', 'KSM'] // Lunie coins currently being traded in the open
+    this.coins = ['ATOM', 'LUNA', 'KRT', 'SDT', 'KAVA', 'BNB', 'KSM', 'DOT'] // Lunie coins currently being traded in the open
     this.priceFeed = {}
 
     this.pollNewPriceFeed()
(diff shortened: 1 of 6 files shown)


app/src/components/common/TmConnectedNetwork.vue
```diff
@@ -36,11 +36,19 @@
...
           v-if="block.height"
           v-tooltip.top="'Block Height'"
-          class="block-number"
-          :to="{
-            name: `block`,
-            params: { height: block.height, networkId: networkSlug },
-          }"
+          :class="
+            `block-number` && {
+              noClickable: currentNetwork.network_type === `polkadot`,
```
jbibla, line 41: why don't we use the feature flag instead of making it network specific?

app/src/components/common/TmConnectedNetwork.vue
```diff
@@ -173,6 +181,14 @@ export default {
   background: var(--app-nav);
 }
 
+.noClickable {
+  cursor: unset;
+}
+
+.noClickable:hover {
```
jbibla, line 188: no no - if it's not clickable nothing should happen on hover

faboweb:
🚀 Deployed on https://5f18af10cf096bbef039b80a--lunieio.netlify.app

app/src/components/common/TmConnectedNetwork.vue
Bitcoinera: yeah, that is why it is keeping the same color, overriding the previous hover. This is the idea

app/src/components/common/TmConnectedNetwork.vue
Bitcoinera: we can do that as well. But then it is clickable and will take you to the "Feature not available" page

Bitcoinera: whichever you prefer

faboweb:
```suggestion
              noClickable: !currentNetwork.feature_activity,
```

api/lib/fiatvalues-api.js
```diff
@@ -76,7 +78,7 @@ class fiatValueAPI {
     this.fiatValuesAPIDictionary = denomToCoinGeckoIDDictionary
     this.fiatValuesAPIReverseDictionary = geckoIDToDenomDictionary
     this.fiatCurrenciesSymbolsDictionary = fiatCurrenciesSymbolsDictionary
-    this.coins = ['ATOM', 'LUNA', 'KRT', 'SDT', 'KAVA', 'BNB', 'KSM'] // Lunie coins currently being traded in the open
+    this.coins = ['ATOM', 'LUNA', 'KRT', 'SDT', 'KAVA', 'BNB', 'KSM', 'DOT'] // Lunie coins currently being traded in the open
```
faboweb, line 81: can we handle this somehow without needing to hardcode?

Bitcoinera: Yes, I think the best would be to have this stored in DB too

faboweb: > without needing to hardcode

Bitcoinera: :thinking: any suggestions?

faboweb: try to get the coins and if they are not on coingecko ignore them

Bitcoinera: :shrug: maybe we just leave it like this (although I prefer the DB-centered solution)

faboweb: let's leave as is

Issue #4445, opened 2020-07-06
polkadot governance api endpoints


faboweb:
Part of #4456 

- [x] governance overview
  - totalStakedAssets
  - totalVoters (all staked accounts)
  - treasurySize (treasure/communty pool size)
  - recentProposals (make adjustable pagesize, maybe extra endpoint)
  - topVoters (council members in Polkad and highst voting power in Cosmos)
  - links (title, link, type, icon)

- [x] ~add full text search to proposals endpoint~
- [x] ~make title, description, summary overwritable from db~
- [x] add polkadot proposals
  - add text proposals like in Cosmos
  - add treasury proposals via a templte for the description and summary
  - add parameter change proposals via a template
- [ ] add to proposals detailed votes
  - deposits
  - deposits sum
  - percentage of needed deposits
  - votes (add filter to only show for your validators)
  - votes sum
  - voting threshold yes
  - voting threshold no
  - links (from db, have title, link, icon, type)
  - timeline [{title, time}] (maybe put in network config)

mariopino:
References:

- https://polkadot.network/a-walkthrough-of-polkadots-governance/

mariopino:
Referendums:

```
const startTime = Date.now()
const referendums = await api.derive.democracy.referendums()
console.log(JSON.stringify(referendums, null, 2))
const endTime = Date.now()
console.log(`execution time: ${(endTime - startTime) / 1000}s`)
```

```
[
  {
    "image": {
      "at": 2973417,
      "balance": 383333180,
      "proposal": {
        "callIndex": "0x1d02",
        "args": {
          "id": "0x2b3325",
          "when": 3200000,
          "maybe_periodic": [
            3600,
            336
          ],
          "priority": 254,
          "call": {
            "callIndex": "0x060a",
            "args": {
              "additional": 2
            }
          }
        }
      },
      "proposer": "H9eSvWe34vQDJAWckeTHWSqSChRat8bgKHG39GC1fjvEm7y"
    },
    "imageHash": "0x6bee0d2c7ac152a72ce4b617a26eba078b19a4b5f54fbcda45f87330e60472f8",
    "index": 71,
    "status": {
      "end": 3225600,
      "proposalHash": "0x6bee0d2c7ac152a72ce4b617a26eba078b19a4b5f54fbcda45f87330e60472f8",
      "threshold": "Supermajorityapproval",
      "delay": 115200,
      "tally": {
        "ayes": "0x000000000000000000f6f6272d994c00",
[...]

PR #4507, opened 2020-07-14
Polkadot governance api endpoints

mariopino:
Partially closes #4445 

**Description:**

Thank you! 🚀

---

For contributor:

- [ ] Added changes entries. Run `yarn changelog` for a guided process.
- [ ] Reviewed `Files changed` in the github PR explorer
- [ ] Attach screenshots of the UI components on the PR description (if applicable)
- [ ] Scope of work approved for big PRs

For reviewer:

- [ ] Manually tested the changes on the UI

PR #4607, opened 2020-07-28
Ana/4445 governance overview

Bitcoinera:
Related to #4445

**Description:**

Bringing in the governance overview endpoint.

Cosmos part pseudo-done, defaulting the topVoters (5) and the recentProposals (3).
Getting totalVoters as we know is problematic.

Next Substrate networks

Thank you! 🚀

---

For contributor:

For reviewer:


Bitcoinera:
What is this links field exactly about?

faboweb:
links are also for the network from the db.
we probably should have a table: networkInfo that holds a key, values we can query for

Bitcoinera:
Ready to go. Needs #4507 first in

api/lib/database/methods.js
```diff
@@ -248,6 +250,27 @@ const getNetwork = ({ hasura_url, hasura_admin_key }) => () => async (id) => {
   }
 }
 
+const getNetworkLinks = ({ hasura_url, hasura_admin_key }) => () => async (
```
faboweb, line 253: note: should have a been an extra PR

api/lib/database/methods.js
```diff
@@ -248,6 +250,27 @@ const getNetwork = ({ hasura_url, hasura_admin_key }) => () => async (id) => {
...
+    data: { networks }
+  } = await query({
+    hasura_url,
+    hasura_admin_key
+  })(`
+    query {
+      networks: networks(where: { 
+        id: {_eq: "${id}"}
+      }) {
+        links
```
faboweb, line 266: should we put this in another table: `networkGovernance` so we can normalize the table?

networkGovernance
- networkId
- link
- linkCaption

api/lib/source/cosmosV0-source.js
```diff
@@ -288,6 +290,65 @@ class CosmosV0API extends RESTDataSource {
     )
   }
 
+  async getRecentProposals() {
+    // while we don't have proposals in DB this is the only way
```
faboweb, line 294: yep

api/lib/source/cosmosV0-source.js
```diff
@@ -288,6 +290,65 @@ class CosmosV0API extends RESTDataSource {
...
+    const proposals = await this.getAllProposals()
+    return proposals.sort(
+      (a, b) => new Date(b.creationTime) - new Date(a.creationTime)
+    )
+  }
+
+  async getTopVoters() {
+    // for now defaulting to pick the 5 largest voting powers
+    const validatorsKeys = Object.keys(this.store.validators)
+    const validatorsArray = validatorsKeys.reduce(
```
faboweb, line 304: instead:
https://lodash.com/docs/4.17.15#sortBy (votingPower)
https://lodash.com/docs/4.17.15#take (5)

api/lib/source/cosmosV0-source.js
```diff
@@ -288,6 +290,65 @@ class CosmosV0API extends RESTDataSource {
...
+    const { bonded_tokens: totalBondedTokens } = await this.query(
+      '/staking/pool'
+    )
+    const communityPoolArray = await this.query('/distribution/community_pool')
+    const communityPool = communityPoolArray.find(
+      ({ denom }) => denom === this.network.coinLookup[0].chainDenom
+    ).amount
+    const links = await this.db.getNetworkLinks(this.network.id)
+    return {
+      totalStakedAssets: fixDecimalsAndRoundUp(
```
faboweb, line 333: this whole thing should be one function as we need to use it frequently

api/lib/source/polkadotV0-source.js
```diff
@@ -529,6 +532,151 @@ class polkadotAPI {
...
+  async getRecentProposals() {
+    // while we don't have proposals in DB this is the only way
+    const proposals = await this.getAllProposals()
+    // the problem in Substrate is that right now we don't have all creationTimes yet
+    return proposals.sort(
+      (a, b) => new Date(b.creationTime) - new Date(a.creationTime)
+    )
+  }
+
+  async getTopVoters() {
```
faboweb, line 618: should we also just return top n?

faboweb: @jbibla

api/lib/source/polkadotV0-source.js
```diff
@@ -529,6 +532,151 @@ class polkadotAPI {
...
+    // in Substrate we simply return council members
+    const api = await this.getAPI()
+    const councilMembers = await api.query.council.members()
+    const councilMembersWithStake = await Promise.all(
```
faboweb, line 622: how about instead mapping to our validators object and sorting by voting power?

api/lib/source/polkadotV0-source.js
```diff
@@ -529,6 +532,151 @@ class polkadotAPI {
...
+    return freeBalance.toString()
+  }
+
+  async getGovernanceOverview(network) {
+    const api = await this.getAPI()
+
+    const activeEra = parseInt(
+      JSON.parse(JSON.stringify(await api.query.staking.activeEra())).index
+    )
+    const erasTotalStake = await api.query.staking.erasTotalStake(activeEra)
```
faboweb, line 658: do all of the queries in parallel

api/lib/source/polkadotV0-source.js
```diff
@@ -529,6 +532,151 @@ class polkadotAPI {
...
+    const treasurySize = await this.getTreasurySize()
+    const links = await this.db.getNetworkLinks(this.network.id)
+    return {
+      totalStakedAssets: fixDecimalsAndRoundUp(
+        BigNumber(erasTotalStake).times(
+          network.coinLookup[0].chainToViewConversionFactor
+        ),
+        2
+      ),
+      totalVoters: await this.getTotalActiveAccounts(),
```
faboweb, line 668: do all of these requests in parallel

api/lib/database/methods.js
Bitcoinera: I think it is more organized per network? Otherwise it would look messy, don't you think, with multiple repetitions of networkId per each network link... what would the primary key be?
I think it is better to have all links in an array within the network. But we could also do `networkId` and `links` (array) in this new table, `networkGovernance`

api/lib/source/polkadotV0-source.js
Bitcoinera: this is something I was wondering too, actually, if we could make the slice limit a parameter here

api/lib/source/polkadotV0-source.js
Bitcoinera: I followed the specs in the issue. I think it depends on the voting type that council members vote are more important? Like in treasury proposals for example, while in referendums it doesn't make a difference. But will need to research more

api/lib/source/polkadotV0-source.js
Bitcoinera: true, in fact I realized this in the next one, the detailed votes PR

api/lib/source/polkadotV0-source.js
Bitcoinera: Also, it doesn't seem that in Polkadot validators necessarily have more voting power? I mean, each DOT holder votingPower depends on their stake. I think validators cannot vote on behalf of their stakers

api/lib/database/methods.js
faboweb: What is your opinion on the arrays being hard to read and edit in the db?

api/lib/source/polkadotV0-source.js
faboweb: I think you misunderstood me:
You are requerying the account data for each council member here. if all council members are validators, you can just lookup the staked balance from the validator and do not need to query again.

api/lib/source/polkadotV0-source.js
```diff
@@ -529,6 +532,151 @@ class polkadotAPI {
...
+    const councilMembersWithStake = await Promise.all(
+      councilMembers.map(async (member) => {
+        const account = await api.query.system.account(member)
+        const totalBalance = account.data.free
+        const freeBalance = BigNumber(totalBalance.toString()).minus(
+          account.data.miscFrozen.toString()
+        )
+        return {
+          address: member,
+          stakedBalance: totalBalance - freeBalance
```
faboweb, line 631: you are double negating here.
```suggestion
          stakedBalance: Number(account.data.miscFrozen.toString())
```

api/lib/source/polkadotV0-source.js
Bitcoinera: yeah, but not all of them are. Look Gav for example. It is safer to query for council members directly

api/lib/database/methods.js
Bitcoinera: IMO, it is worth to get the hang of it. Somethings it is just cleaner & better to store them like that, like for example the cryptos and the links in the Network Object. Plus you already have those as examples there

api/lib/source/polkadotV0-source.js
faboweb: ok, if that is the case then this way is safer

faboweb: I guess we still need to cache the data. at least put a caching for the query for 10 minutes or so

Bitcoinera: Done

Bitcoinera:
This one looks quite ready to me, but first we need to merge #4507 before this one

api/lib/source/polkadotV0-source.js
```diff
@@ -17,6 +18,7 @@ class polkadotAPI {
     this.store = store
     this.fiatValuesAPI = fiatValuesAPI
     this.db = db
+    this.topVoters = []
```
faboweb, line 21: this should not be here but in the store

faboweb: as the source is not persisted

Bitcoinera: oh ok. I thought it was once it is created? I can move it

api/lib/source/cosmosV0-source.js
```diff
@@ -1,14 +1,15 @@
 const { RESTDataSource, HTTPCache } = require('apollo-datasource-rest')
 const { InMemoryLRUCache } = require('apollo-server-caching')
 const BigNumber = require('bignumber.js')
-const { orderBy, keyBy, uniqBy } = require('lodash')
+const _ = require('lodash')
```
jbibla, line 4: does this mean we're importing all of lodash instead of just the functions we need?

api/lib/source/cosmosV0-source.js
```diff
@@ -288,6 +290,66 @@ class CosmosV0API extends RESTDataSource {
...
+          (validator) => {
+            return validator.votingPower
+          }
+        ])
+      ),
+      5
+    ).map(({ operatorAddress }) => operatorAddress)
+  }
+
+  async getGovernanceOverview() {
```
jbibla, line 315: should all these data points be in one function?

api/lib/source/polkadotV0-source.js
```diff
@@ -529,6 +534,166 @@ class polkadotAPI {
...
+          return this.reducers.councilProposalReducer(
+            this.network,
+            proposal,
+            councilMembers,
+            blockHeight
+          )
+        })
+      )
+
+    return orderBy(allProposals, 'id', 'desc')
```
jbibla, line 597: aren't the proposal IDs unique per type?

api/lib/source/polkadotV0-source.js
```diff
@@ -529,6 +534,166 @@ class polkadotAPI {
...
+  async getRecentProposals() {
+    // while we don't have proposals in DB this is the only way
+    const proposals = await this.getAllProposals()
+    // the problem in Substrate is that right now we don't have all creationTimes yet
+    return proposals.sort(
+      (a, b) => new Date(b.creationTime) - new Date(a.creationTime)
+    )
+  }
+
+  async pollForTopVoters() {
```
jbibla, line 620: why do we have to poll? can't we just query for council members when requested?

api/lib/source/polkadotV0-source.js
Bitcoinera: yes, they are. What conflict do you see?

api/lib/source/cosmosV0-source.js
Bitcoinera: I think it is fine like this, since they are just called as a bunch here

api/lib/source/cosmosV0-source.js
Bitcoinera: all of it since we are using 5 different lodash functions. I think it is cleaner like this. Or you think this sacrifices performance too much?

api/lib/source/polkadotV0-source.js
Bitcoinera: Fabo said it is better to cache them :shrug:

api/lib/source/cosmosV0-source.js
faboweb: on node it doesn't really matter that we import all of lodash.

api/lib/source/cosmosV0-source.js
```diff
@@ -288,6 +290,66 @@ class CosmosV0API extends RESTDataSource {
...
+        2,
+        this.network,
+        this.network.stakingDenom
+      ),
+      totalVoters: undefined,
+      treasurySize: fixDecimalsAndRoundUpBigNumbers(
+        communityPool,
+        2,
+        this.network,
+        this.network.stakingDenom
```
faboweb, line 345: why do you need to pass the stakingDenom when you already pass in the whole network?

api/lib/source/cosmosV0-source.js
faboweb: @jbibla can you clarify what you mean?

api/lib/database/methods.js
faboweb: Why is it cleaner and "better" to store them this way?

api/lib/source/cosmosV0-source.js
```diff
@@ -288,6 +290,66 @@ class CosmosV0API extends RESTDataSource {
     )
   }
 
+  async getRecentProposals() {
```
faboweb, line 293: Isn't this a duplicate of `getAllProposals`?

api/lib/source/polkadotV0-source.js
faboweb: I said cache them using a cache directive. Let's remove the polling to reduce code complexity.

api/lib/source/cosmosV0-source.js
jbibla: i mean, still seems funny to me - would be nice to reduce dependencies and only import the functions we need - the repo is over 1GB in size

api/lib/source/cosmosV0-source.js
jbibla: this one function is making 2 requests and calling two other functions - should one function make one request instead?

api/lib/source/polkadotV0-source.js
jbibla: if they are numbered by type than we can't order them by ID because it will not be chronological. 

```
4
4
4
3
3
2
2
2
1
1
1
1
```

Bitcoinera: it will be chronological but with mixed types. So we should also order by type, not only id, good observation

api/lib/source/polkadotV0-source.js
Bitcoinera: I don't know what you mean with cache directive. Just an object, like you did for `sessionCache` ?

api/lib/database/methods.js
Bitcoinera: because you get all you need in one place without needing multiple calls (more code to maintain + slower). You get your beautiful object,
```
[
{
name: "Edwards curve",
curve: "ed25519"
},
{
name: "Schnorrkel curve",
curve: "sr25519"
},
}
```

Otherwise you would need one column called curveName0 and another one called curveCurve0, and so on and so forth. Truly horrible, if you want my opinion

api/lib/source/cosmosV0-source.js
Bitcoinera: I guess we can just remove this function since it is not doing much

api/lib/source/cosmosV0-source.js
Bitcoinera: I don't see an issue here, if I am getting right what you mean. Why is a bad idea to have a function call many different ones? (in this case actually 5 or 6) I guess I am not getting straight what you mean

api/lib/source/cosmosV0-source.js
Bitcoinera: but lodash is already there Jordan, in the node_modules in full size

api/lib/source/polkadotV0-source.js
Bitcoinera: @jbibla any remark here?

api/lib/source/polkadotV0-source.js
faboweb: The id cant be the proposal Id but needs to be a combination of id and type as it would fuck with caching if multiple have the same id

api/lib/source/polkadotV0-source.js
faboweb: No, a graphql caching directive in schema.js

api/lib/source/polkadotV0-source.js
Bitcoinera: ouch true

faboweb:
🚀 Deployed on https://5f3282d1fc532983dca082eb--lunieio.netlify.app

api/lib/source/polkadotV0-source.js
Bitcoinera: Done

api/lib/source/polkadotV0-source.js
Bitcoinera: Done in #4507 and imported

api/lib/database/methods.js
faboweb: no, you can create a table `networkCrypto` then this table has columns:
- networkId
- curve
- name
- default

this way you can easily edit all of the rows.

api/lib/reducers/polkadotV0-reducers.js
```diff
@@ -470,6 +480,165 @@ function rewardReducer(network, validators, reward, reducers) {
...
+
+function councilTallyReducer(votes, councilMembers) {
+  const total = votes.ayes.length + votes.nays.length
+  return {
+    yes: votes.ayes.length,
+    no: votes.nays.length,
+    abstain: 0,
+    veto: 0,
+    total,
+    totalVotedPercentage: (total / councilMembers.length).toFixed(4) // the percent conversion is done in the FE. No need to multiply by 100
```
faboweb, line 623: each council member has the same voting power? who is voting on council members?

api/lib/resolvers.js
```diff
@@ -189,8 +204,18 @@ const resolvers = (networkList) => ({
       // a validator, and in that case fetch the current validator object from datasource
       // and attach it to proposal.
       //
-      if (proposal.proposer !== `unknown`) {
-        const proposerValAddress = encodeB32(
+      if (proposal.proposer && proposal.proposer !== `unknown`) {
+        let proposerValAddress = ''
+        if (
+          networkList.find(({ id }) => id === proposal.networkId) &&
```
faboweb, line 210: refactor this line so we don't duplicate the code

api/lib/schema.js
```diff
@@ -316,6 +323,21 @@ const typeDefs = gql`
...
+    link: String
+    type: String
+  }
+
+  type GovernanceOverview {
+    totalStakedAssets: Float
+    totalVoters: Int
+    treasurySize: Float
+    recentProposals: [Proposal]
+    topVoters: [String] @cacheControl(maxAge: 21600)
```
faboweb, line 337: this would only cache this property, but to answer the request, u still need to do all the calls, so topVoters would still be calculated every time. we should cache governance overview completely and remove "recentProposals". "recentProposals" should be a non cached request.

faboweb: Instead of using "recentProposals" I suggest we just use the already existing "proposals" query.

api/lib/source/cosmosV0-source.js
```diff
@@ -288,6 +299,60 @@ class CosmosV0API extends RESTDataSource {
...
+    return _.take(
+      _.reverse(
+        _.sortBy(this.store.validators, [
+          (validator) => {
+            return validator.votingPower
+          }
+        ])
+      ),
+      5
+    ).map(({ operatorAddress }) => operatorAddress)
```
faboweb, line 313: why return just the operatorAddress and not directly the whole validator object? this way you can directly display the validtor without needing to map in the FE

api/lib/source/cosmosV0-source.js
faboweb: let'S remove recent proposals from here (see above)

Bitcoinera: where is above exactly? Please link :pray:

api/lib/source/cosmosV0-source.js
Bitcoinera: because of what we already discussed. We need to agree on a new type, `topVoter`, that also deals with the council member case

Bitcoinera: some council members are not validators

Bitcoinera: I think this would be a different PR though

api/lib/source/cosmosV0-source.js
Bitcoinera: ah ok, found it: https://github.com/luniehq/lunie/pull/4607#discussion_r469638434

It was actually "below" :stuck_out_tongue:

api/lib/source/cosmosV0-source.js
Bitcoinera: OK, I'm going to investigate what data I can get from council members and create a `type topVoter`

api/lib/reducers/polkadotV0-reducers.js
Bitcoinera: oh, I see what you mean. True, it is not the real number

Bitcoinera: OK, done. Although I see another thing coming here:

```
Prime Members
The council, being an instantiation of Substrate's Collective pallet, implements what's called a prime member whose vote acts as the default for other members that fail to vote before the timeout.

The prime member is chosen based on a Borda count.

The purpose of having a prime member of the council is to ensure a quorum, even when several members abstain from a vote. Council members might be tempted to vote a "soft rejection" or a "soft approval" by not voting and letting the others vote. With the existence of a prime member, it forces councillors to be explicit in their votes or have their vote counted for whatever is voted on by the prime.
```

Bitcoinera: But this can be calculated once the voting period is closed :thinking: (the impact of the council prime member, I mean)

api/lib/source/cosmosV0-source.js
Bitcoinera: Created

api/lib/reducers/polkadotV0-reducers.js
Bitcoinera: This seems very advanced though. Let's do it afterwards. I just left a TODO comment

jbibla: yep let's leave it for now!

Bitcoinera:
Testing query:

```
query governanceOverview {
  governanceOverview(networkId: "polkadot") {
   totalStakedAssets
    totalVoters
    treasurySize
    topVoters {
      name
      address
      votingPower
      validator {
        name
      }
    }
    links {
      title
      link
      type
    }
  }
}
```

api/lib/reducers/polkadotV0-reducers.js
```diff
@@ -597,17 +589,13 @@ function tallyReducer(network, tally, totalIssuance) {
...
-  const yes = toViewDenom(network, tally.ayes, network.stakingDenom)
-  const no = toViewDenom(network, tally.nays, network.stakingDenom)
+  const total = toViewDenom(network, totalVoted.toString(10))
+  const yes = toViewDenom(network, tally.ayes)
+  const no = toViewDenom(network, tally.nays)
   const totalVotedPercentage = turnout
     .div(BigNumber(totalIssuance))
     .toNumber()
-    .toFixed(4) // the percent conversion is done in the FE. No need to multiply by 100
+    .toFixed(4) // the percent conversion is done in the FE. We just send the decimals here
```
jbibla, line 598: don't make the FE do anything if possible

Bitcoinera: OK. We already have this `percent` filter being used in PageProposal, but we can change this later :shrug:

PR #4611, opened 2020-07-28
Fabo/update bonding role

faboweb:
The bonding role would not update after the user bonded so consecutive staketx (the bonding part) would fail in polkadot as they need to send `bondMore` instead of `bond` extrinsics.

TODO: need to check if the role actually updates immediately or only after the era/session changes

### DIFF
--- a/app/src/ActionModal/components/DelegationModal.vue
+++ b/app/src/ActionModal/components/DelegationModal.vue
@@ -355,6 +355,7 @@ export default {
 
       // update registered topics for emails as the validator set changed
       this.$store.dispatch("updateEmailRegistrations")
+      this.$store.dispatch("checkAddressRole")
     },
   },
   validations() {
--- a/app/src/ActionModal/components/DelegationModal.vue
+++ b/app/src/ActionModal/components/DelegationModal.vue
@@ -355,7 +355,10 @@ export default {
 
       // update registered topics for emails as the validator set changed
       this.$store.dispatch("updateEmailRegistrations")
-      this.$store.dispatch("checkAddressRole")
+      // update the role of the user as it might change after bonding the first time
+      if (this.currentNetwork.network_type === "polkadot") {
+        this.$store.dispatch("checkAddressRole")
+      }
     },
   },
   validations() {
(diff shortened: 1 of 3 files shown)


faboweb:
🚀 Deployed on https://5f207e8cd905a7028d4d36af--lunieio.netlify.app

app/src/ActionModal/components/DelegationModal.vue
```diff
@@ -355,6 +355,10 @@ export default {
 
       // update registered topics for emails as the validator set changed
       this.$store.dispatch("updateEmailRegistrations")
+      // update the role of the user as it might change after bonding the first time
+      if (this.currentNetwork.network_type === "polkadot") {
+        this.$store.dispatch("checkAddressRole")
```
Bitcoinera, line 360:
```suggestion
     await this.$store.dispatch("checkAddressRole")
```

faboweb: why await?

app/src/ActionModal/components/DelegationModal.vue
```diff
@@ -355,6 +355,10 @@ export default {
 
       // update registered topics for emails as the validator set changed
       this.$store.dispatch("updateEmailRegistrations")
+      // update the role of the user as it might change after bonding the first time
+      if (this.currentNetwork.network_type === "polkadot") {
```
jbibla, line 359: anytime i see network specific stuff it feels wrong

faboweb: There are a bunch of network type logics. This endpoint actually works for Cosmos as well. Bit we safe a request. Happy to discuss further. But will leave this in for now.

app/src/ActionModal/components/DelegationModal.vue
Bitcoinera: ok, you want it non-blocking. Makes sense

Bitcoinera:
Mmm, E2E

Bitcoinera:
Ah, they fail because we are querying HDPaths and curves with API not released yet... We really need to release API asap

PR #4614, opened 2020-07-28
Fabo/handle batch failure

faboweb:
Batch transactions would fail partially but they would show as a success

### DIFF
--- a/api/lib/reducers/polkadotV0-reducers.js
+++ b/api/lib/reducers/polkadotV0-reducers.js
@@ -244,25 +244,20 @@ function parsePolkadotTransaction(
 }
 
 function getExtrinsicSuccess(extrinsicIndex, blockEvents, isBatch) {
-  const events = blockEvents
-  .filter(({ phase }) => {
+  const events = blockEvents.filter(({ phase }) => {
     return parseInt(phase.toHuman().ApplyExtrinsic) === extrinsicIndex // index is a string
   })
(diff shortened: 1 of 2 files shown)


api/lib/reducers/polkadotV0-reducers.js
```diff
@@ -243,15 +243,26 @@ function parsePolkadotTransaction(
...
+  .filter(({ phase }) => {
+    parseInt(phase.toHuman().ApplyExtrinsic) === extrinsicIndex // index is a string
+  })
+  // if tx is a batch, we need to check if all of the batched txs went through
+  if (isBatch) {
+    return !!events
+      .find(
+        ({ event }) =>
+          event.section === `utility` &&
+          event.method === `BatchCompleted`
```
faboweb, line 257: We should invert here and directly look for the errors so we can get the error codes

jbibla:
can you please run `lint --fix`

api/lib/reducers/polkadotV0-reducers.js
Bitcoinera: mmm, something seems to be failing here

Bitcoinera: image

Bitcoinera: Fixed

faboweb:
Great, thanks @Bitcoinera

PR #4626, opened 2020-07-30
Ana/fix navigation duplicated in recover flow

Bitcoinera:
Closes #ISSUE

**Description:**

One of those magic moments when the solution is actually deleting code :champagne: 

Thank you! 🚀

---

For contributor:

For reviewer:


jbibla:
so what happens if you click the network item now?

faboweb:
🚀 Deployed on https://5f230095a3eb0cf1b6208253--lunieio.netlify.app

faboweb:
the click is handled by networklist not  by networkitem

Bitcoinera:
> so what happens if you click the network item now?

in Networks it will work but in both create and recover flows will fail with a Navigation duplicated error.

The strange thing is that we didn't notice this happening yesterday :thinking:

Bitcoinera:
Let's still merge this one in develop? Or should we pull from master to develop instead? What is the best flow?

PR #4635, opened 2020-07-30
Fabo/fix extension account creation

faboweb:
Account creation would not work as the insecure warning would pop up and several other things would break


faboweb:
🚀 Deployed on https://5f234e8536927201f82dee2c--lunieio.netlify.app

PR #4660, opened 2020-08-04
WIP Ana/loop to get new networks from db

### LINKED ISSUE #4645: Add new network when set to enabled
I just realized while attempting to recover the akash-testnet network that, apparently, we don't have the logic there to add a new network from DB when setting it to enabled.

This should be fairly easy and we could include the logic for it in the same loop where we watch for other network changes, I think. It should be quite easy to do, in any case

Bitcoinera:
Closes #4645 

**Description:**

Looks good but sadly doesn't work. Nothing is that easy. There are no one line changes :shrug: 

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/api/index.js
+++ b/api/index.js
@@ -3,8 +3,12 @@ const http = require('http')
 const cors = require('cors')
 const { createApolloServer } = require('./lib/apollo')
 const { transaction } = require('./lib/routes')
-
+const database = require('./lib/database')
 const config = require('./config')
+const NetworkContainer = require('./lib/network-container')
+
+const db = database(config)('')
+const NEW_NETWORKS_POLLING_INTERVAL = 600000 // 10min
 
 if (config.SENTRY_DSN) {
   const Sentry = require('@sentry/node')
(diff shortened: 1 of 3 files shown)


Bitcoinera:
Currently it works adding networks that become `enabled` in DB, but not so when you change them to `disabled`

Any idea about this @faboweb ? Maybe I need to "kill" the app object and redefine it?

api/index.js
```diff
@@ -58,8 +58,9 @@ async function pollForNewNetworks(httpServer, app) {
     )
     console.log(`Transaction service ready at ${config.transactionPath}`)
   })
-  this.newNetworksPollingTimeout = setTimeout(async () => {
-    this.pollForNewNetworks()
+  newNetworksPollingTimeout = setTimeout(async () => {
+    httpServer.close({ port: config.port })
```
faboweb, line 62: what?!

faboweb:
I think the approach is wrong: I would query all networks and update all networks always. Then handle polling in the subscriptions based on the enabled state.

Bitcoinera:
> Then handle polling in the subscriptions based on the enabled state.

But the subscriptions are created only afterwards, per each filtered network. That is why you need to do this even before. Or am I missing something here?

api/index.js
Bitcoinera: I was getting an error here. You cannot do `httpServer.listen` without closing the port before, once opened

Bitcoinera:
Closing this as the logic to accomplish it would imply more serious changes, and with `enabled` and `experimental` we already have enough tools to play with

PR #4666, opened 2020-08-06
Fabo/sessions

faboweb:
Working on Desktop need to test on mobile
We store a session in the db and return it to the user. Then we validate only the session. Reduces requests and should allow multiple sign ins.
Still need to delete old sessions.

Edit: Mobile working

### DIFF
--- a/api/lib/accounts.js
+++ b/api/lib/accounts.js
@@ -19,7 +19,7 @@ const registerUser = async (idToken) => {
       lastActive: userRecord.metadata.lastSignInTime
     }
     database(config)('').upsert(`users`, user)
-    const session = await db.getNewSession(uid)
+    const session = await db.storeAndGetNewSession(uid)
 
     return {
       validUntil: session.valid_until,
@@ -41,6 +41,7 @@ async function validateSession(sessionToken) {
     session = sessionCache[sessionToken]
   } else {
     session = await db.getSession(sessionToken)
+    sessionCache[sessionToken] = session
   }
   if (
     !session ||
--- a/api/lib/accounts.js
+++ b/api/lib/accounts.js
@@ -1,7 +1,7 @@
 const firebaseAdmin = require('./notifications/firebase')
 const Sentry = require('@sentry/node')
 const database = require('./database')
-const config = require("../config")
+const config = require('../config')
 const { AuthenticationError } = require('apollo-server')
 const db = database(config)('')
 
@@ -42,9 +42,12 @@ async function validateSession(sessionToken) {
   } else {
     session = await db.getSession(sessionToken)
   }
-  if (!session || new Date(session.valid_until).getTime() - new Date().getTime() <= 0) {
+  if (
+    !session ||
+    new Date(session.valid_until).getTime() - new Date().getTime() <= 0
+  ) {
     delete sessionCache[sessionToken]
-    throw new AuthenticationError("Session is outdated")
+    throw new AuthenticationError('Session is outdated')
   }
 
   return session
--- a/api/lib/accounts.js
+++ b/api/lib/accounts.js
@@ -1,18 +1,16 @@
 const firebaseAdmin = require('./notifications/firebase')
 const Sentry = require('@sentry/node')
 const database = require('./database')
+const config = require("../config")
+const { AuthenticationError } = require('apollo-server')
+const db = database(config)('')
 
-const registerUser = async (uid) => {
+const registerUser = async (idToken) => {
   try {
+    // verify that the submitted tokens is valid
+    const { uid } = await firebaseAdmin.auth().verifyIdToken(idToken)
     // get user record, with custom claims and creation & activity dates
     const userRecord = await firebaseAdmin.auth().getUser(uid)
-    // check if user already exists in DB
-    const storedUser = await database(config)('').getUser(uid)
-    // check if user already has premium as a custom claim
-    if (!userRecord.customClaims) {
-      // set premium field
-      await firebaseAdmin.auth().setCustomUserClaims(uid, { premium: false })
-    }
     // we don't store user emails for now
     const user = {
       uid,
@@ -20,10 +18,12 @@ const registerUser = async (uid) => {
       createdAt: userRecord.metadata.creationTime,
       lastActive: userRecord.metadata.lastSignInTime
     }
-    if (!storedUser) {
-      database(config)('').storeUser(user)
-    } else {
-      database(config)('').upsert(`users`, user)
+    database(config)('').upsert(`users`, user)
+    const session = await db.getNewSession(uid)
+
+    return {
+      validUntil: session.valid_until,
+      sessionToken: session.session_token
     }
   } catch (error) {
     console.error(`In storeUser`, error)
@@ -32,6 +32,22 @@ const registerUser = async (uid) => {
       Sentry.captureException(error)
     })
   }
+}
+
+const sessionCache = {}
+async function validateSession(sessionToken) {
+  let session
+  if (sessionCache[sessionToken]) {
+    session = sessionCache[sessionToken]
+  } else {
+    session = await db.getSession(sessionToken)
+  }
+  if (!session || new Date(session.valid_until).getTime() - new Date().getTime() <= 0) {
+    delete sessionCache[sessionToken]
+    throw new AuthenticationError("Session is outdated")
+  }
+
+  return session
 }
 
 async function validateIdToken(idToken) {
@@ -50,5 +66,6 @@ async function validateIdToken(idToken) {
 
 module.exports = {
   registerUser,
-  validateIdToken
+  validateIdToken,
+  validateSession
 }
--- a/api/lib/apollo.js
+++ b/api/lib/apollo.js
@@ -6,7 +6,7 @@ const typeDefs = require('./schema')
 const resolvers = require('./resolvers')
 const Notifications = require('./notifications/notifications')
 const database = require('./database')
-const { validateIdToken } = require('./accounts')
+const { validateSession } = require('./accounts')
 
 const NetworkContainer = require('./network-container')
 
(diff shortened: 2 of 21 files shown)


faboweb:
🚀 Deployed on https://5f2c668b734c7000a752f510--lunieio.netlify.app

api/lib/database/index.js
```diff
@@ -61,6 +63,8 @@ function database({ hasura_url, hasura_admin_key }) {
       })(schema),
       storeUser: storeUser({ hasura_url, hasura_admin_key })(schema),
       getUser: getUser({ hasura_url, hasura_admin_key })(schema),
+      getSession: getSession({ hasura_url, hasura_admin_key })(schema),
```
jbibla, line 66: i don't think this is self explanatory enough. maybe `getCurrentSession` or `getExistingSession`

api/lib/database/methods.js
```diff
@@ -389,6 +389,44 @@ const getUser = ({ hasura_url, hasura_admin_key }) => (schema) => async (
   )
 }
 
+// write and return a new session
+const getNewSession = ({ hasura_url, hasura_admin_key }) => () => async (
```
Bitcoinera, line 393:
```suggestion
const storeNewSession = ({ hasura_url, hasura_admin_key }) => () => async (
```
It is actually storing the new session, isn't it?

Bitcoinera: oops I see

api/lib/database/index.js
```diff
@@ -61,6 +63,8 @@ function database({ hasura_url, hasura_admin_key }) {
...
+      getNewSession: getNewSession({ hasura_url, hasura_admin_key })(schema),
```
Bitcoinera, line 67:
```suggestion
      storeAndGetNewSession: storeAndGetNewSession({ hasura_url, hasura_admin_key })(schema),
```

api/lib/apollo.js
```diff
@@ -85,15 +85,16 @@ async function createApolloServer(httpServer) {
...
         }
-        const idToken = req.headers.authorization
-        let user = {}
-        if (idToken) {
-          user = await validateIdToken(idToken)
+
+        const sessionToken = req.headers.authorization
+        let session
+        if (sessionToken) {
+          session = await validateSession(sessionToken) // throws if session is outdated
```
Bitcoinera, line 92: Shouldn't then we add a try/catch block in this line in `index.js`?
```
const apolloServer = await createApolloServer(httpServer)
```

Bitcoinera: this could actually kill API altogether, throwing an error there?

api/lib/database/helpers.js
```diff
@@ -133,7 +133,11 @@ const insert = ({ hasura_url, hasura_admin_key }, upsert = false) => (
             }
     }
     `
-  return graphQLQuery({ hasura_url, hasura_admin_key })(query)
+  const response = await graphQLQuery({ hasura_url, hasura_admin_key })(query)
+  // return the inserted object
+  if (returnKeys.length > 0) {
+    return response.data[`insert_${schema_prefix}${table}`].returning
+  }
```
Bitcoinera, line 140: what is this? I had never seen that `returning` before?

api/lib/notifications/notifications.js
```diff
@@ -255,8 +255,11 @@ const startNotificationService = (networks) => {
 const getNotifications = (networks) => async (
   _,
   { timestamp = '', addressObjects },
-  { dataSources }
+  { dataSources, user: { uid } }
```
Bitcoinera, line 258: cool

app/src/gql/apollo.js
```diff
@@ -15,13 +15,34 @@ import { split } from "apollo-link"
...
 import { getGraphqlHost } from "scripts/url"
-import * as Sentry from "@sentry/browser"
 import config from "src/../config"
 import introspectionQueryResultData from "src/../fragmentTypes.json"
 import { getFingerprint } from "scripts/fingerprint"
 
 Vue.use(VueApollo)
 
+// to avoid circular dependencies we use this error handler, for errors happening in the router
+// those errors should trigger things in the store (that needs the router)
```
Bitcoinera, line 25: where is this error handler communicating with store? I don't see it

Bitcoinera: Another question: why we don't capture these router errors with Sentry anymore? They are not so critical?

app/src/vuex/modules/account.js
```diff
@@ -31,47 +34,25 @@ export default ({ apollo }) => {
...
-            commit(`userSignedIn`, false)
+            // user signed out
+            dispatch(`storeSession`, null)
             commit(`setUserInformation`, null)
           }
           resolve()
         })
       )
     },
-    async listenToIdTokenChanges({ commit }) {
```
Bitcoinera, line 55: This wasn't working in the end? You said you were still getting outdated idTokens right

app/src/vuex/modules/account.js
```diff
@@ -106,14 +109,14 @@ export default ({ apollo }) => {
...
-          bundleId: `1475911030.org.lunie.lunie`,
-        },
+        // android: {
+        //   packageName: `org.lunie.lunie`,
+        //   installApp: true,
+        //   minimumVersion: `1.0.221`, // the first version with deep linking enabled
+        // },
+        // iOS: {
+        //   bundleId: `1475911030.org.lunie.lunie`,
+        // },
```
Bitcoinera, line 119: did you test this against mobile? I guess you did

api/lib/apollo.js
faboweb: no, the error would be forwarded to the user

api/lib/database/index.js
faboweb: please take an example of @Bitcoinera and create a code change proposal for these things

api/lib/database/index.js
```diff
@@ -15,7 +15,9 @@ const {
   storeStore,
   getStore,
   storeNotificationRegistrations,
-  getNotificationRegistrations
+  getNotificationRegistrations,
+  getNewSession,
```
faboweb, line 19:
```suggestion
  storeAndGetNewSession,
```

api/lib/database/helpers.js
faboweb: this returns the inserted row. I used this to get the created UUID.

app/src/vuex/modules/account.js
faboweb: This is not needed as we only send the idToken once for the sign in and we get it in line 65

app/src/vuex/modules/account.js
faboweb, line 119:
```suggestion
         android: {
           packageName: `org.lunie.lunie`,
           installApp: true,
           minimumVersion: `1.0.221`, // the first version with deep linking enabled
         },
         iOS: {
           bundleId: `1475911030.org.lunie.lunie`,
         },
```

api/lib/database/index.js
jbibla: i am familiar with the option to do that - in this case i wasn't sure what you'd prefer.

api/lib/accounts.js
```diff
@@ -34,6 +34,25 @@ const registerUser = async (uid) => {
   }
 }
 
+const sessionCache = {}
```
Bitcoinera, line 37: will this be persisted?

faboweb: No, but it doesn't need to be as it just caches the db request. On a restart the first query for a session will just go to the db.

app/src/gql/apollo.js
```diff
@@ -15,13 +15,34 @@ import { split } from "apollo-link"
...
+class DeferredErrorHandler {
+  constructor() {
+    this.errors = []
+  }
+  error(error) {
+    if (this.errorCallback) {
+      this.errorCallback(error)
```
Bitcoinera, line 32: what does this do? This is the only part here that leaves me wondering

api/lib/accounts.js
```diff
@@ -34,6 +34,25 @@ const registerUser = async (uid) => {
...
+async function validateSession(sessionToken) {
+  let session
+  if (sessionCache[sessionToken]) {
+    session = sessionCache[sessionToken]
+  } else {
+    session = await db.getSession(sessionToken)
```
Bitcoinera, line 43: the Lunie sessions are long-lived right?

```
 SESSION {
  uid: 'R5KcpKptl0MaFWeiGBcDNzY2BSL2',
  valid_until: '2020-09-01T10:32:51.49+00:00'
}
```

app/src/gql/apollo.js
```diff
@@ -15,13 +15,34 @@ import { split } from "apollo-link"
...
 Vue.use(VueApollo)
 
+// to avoid circular dependencies we use this error handler, for errors happening in the router
+// those errors should trigger things in the store (that needs the router)
+class DeferredErrorHandler {
+  constructor() {
+    this.errors = []
+  }
+  error(error) {
+    if (this.errorCallback) {
```
Bitcoinera, line 31: this will be true pretty much all the time after the first error is triggered, won't it?

Bitcoinera: ah ok, actually looks good yes. Because this is the callback you are passing in initializeApp.js. Slick

```
  routerErrorHandler.onError((error) => {
    if (error.extensions && error.extensions.code === "UNAUTHENTICATED") {
      store.dispatch("signOutUser")
      return
    }
    // if sentry is enabled pass all error directly to sentry
    if (config.sentryDSN) {
      // pass errors to sentry
    } else {
      console.error(error)
    }
  })
```

app/src/gql/apollo.js
Bitcoinera: ok, already found out

api/lib/accounts.js
faboweb: 30 days. We can shorten that.

PR #4668, opened 2020-08-07
WIP Ana/4667 fix unstakes not shown correctly

Bitcoinera:
Closes #4667 

**Description:**

We didn't discuss about this but I actually did this "fix" in now time. How I would go about this: display the active and inactive delegations tables only when there are inactive present.

We have this boolean there in the Delegation Object to use it. I think this is the best solution. We need to account for both types in Substrate networks.

image

Open to any suggestions

Thank you! 🚀

---

For contributor:

- [ ] Added changes entries. Run `yarn changelog` for a guided process.
- [ ] Reviewed `Files changed` in the github PR explorer
- [ ] Attach screenshots of the UI components on the PR description (if applicable)
- [ ] Scope of work approved for big PRs

For reviewer:

- [ ] Manually tested the changes on the UI

PR #4671, opened 2020-08-10
Fabo/fix undelegating stake not in total

faboweb:
Undelegations were not counted towards the total balance


Bitcoinera:
Nice! I saw this, but I fixed this in #4668 

image

image

Or should we fix it here too (I simple used uniqWith to erase the duplicates)

jbibla:
why are there duplicates?

Bitcoinera:
> why are there duplicates?

Good question

Bitcoinera:
This was great the way it was. The duplicates will just need another PR

PR #4689, opened 2020-08-11
Fabo/returnlink desktop

faboweb:
The return link wasn't handled on desktop


faboweb:
🚀 Deployed on https://5f330de5096164dc6484e54b--lunieio.netlify.app

jbibla:
lint would be nice!

jbibla:
how can you test this if the link always goes back to app.lunie.io?

PR #4696, opened 2020-08-12
fix manifest path issue

jbibla:
- removed the extra router.push - doesn't seem necessary / used and is causing console errors


faboweb:
🚀 Deployed on https://5f3468a5b4bdbcc107963b1f--lunieio.netlify.app

faboweb:
this is needed for deep links

jbibla:
there are a few issues (i think). 

- the deeplinkhandler is handling all links (not only deeplinks) - is this on purpose?
- doesn't `window.location = link` handle the redirect? why do we need router.push also? 
- if we call router.push every time the deeplinkhandler is called it will cause errors because it's pushing to undefined links

faboweb:
> the deeplinkhandler is handling all links (not only deeplinks) - is this on purpose?

no, this is probably the issue here.

> doesn't window.location = link handle the redirect? why do we need router.push also?

this handles the redirect but not the deep linking. deep linking is when you click on a push notification and want to end up in a part of the app

jbibla:
> this handles the redirect but not the deep linking. deep linking is when you click on a push notification and want to end up in a part of the app

ah ok i see that now. but this `try` block is also being run every time...

PR #4697, opened 2020-08-12
Jordan/ghoost

jbibla:
the ghost emoji was not too complicated. emoji support is not great on the web and this will eliminate the recurring styling issues


faboweb:
🚀 Deployed on https://5f34740f62a87d00be253ab4--lunieio.netlify.app

PR #4706, opened 2020-08-13
Jordan/notificationzzz

jbibla:
- session frame click handler was too easy to set off by accident 
- the notifications page should always be 100% to avoid the title from moving
- paywall link was broken
- user menu addresses shouldn't display on notifications page
- add an intercom link on notifications page to encourage feedback

### DIFF
--- a/app/src/components/account/UserMenu.vue
+++ b/app/src/components/account/UserMenu.vue
@@ -1,6 +1,9 @@
 <template>
   <div class="user-menu">
-    <UserMenuAddress v-if="address" :address="address" />
+    <UserMenuAddress
+      v-if="address && $route.name !== `notifications`"
+      :address="address"
+    />
     <router-link
       v-tooltip="`Notifications`"
       :to="{ name: 'notifications' }"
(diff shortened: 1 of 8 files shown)


faboweb:
🚀 Deployed on https://5f359ce03c79c7541a144daf--lunieio.netlify.app

app/src/components/account/UserMenu.vue
```diff
@@ -1,6 +1,9 @@
 <template>
   <div class="user-menu">
-    <UserMenuAddress v-if="address" :address="address" />
+    <UserMenuAddress
+      v-if="address && $route.name !== `notifications`"
```
Bitcoinera, line 4: Why not? :thinking: I don't get it. That way you can see what addresses those notifications are related to, no? Currently it is actually the only way

jbibla: notifications is a network agnostic page - the address only shows the last used address. 

addresses relevant for notifications are in the user menu (always).

i agree, not having the addresses visible on the page stinks. we will solve it properly soon.

PR #4711, opened 2020-08-14
Fabo/fix notficiations not showing

faboweb:
The notifications were filtered too late. So we would return less than the desired 20. So the FE would think there are not more notifications to query.


jbibla:
api tests and lint are failing

PR #4725, opened 2020-08-17
Fabo/show boot error

faboweb:
Shows an error if there is an error in the bootprocess to avoid a spinner of death


faboweb:
🚀 Deployed on https://5f3ad1b8818927a5e70b257e--lunieio.netlify.app

app/src/initializeApp.js
```diff
@@ -37,7 +37,16 @@ function setOptions(urlParams, store) {
   }
 }
 
+export const bootError = (error) => {
+  document.querySelector("#app").innerText = `
+    Lunie failed to start. Please contact us at http://help.lunie.io/
```
jbibla, line 42: i'm not sure if the help center is the right place to send people.

faboweb: where should we send them? email?

jbibla: would be nice to have a link that opens intercom automagically - but i guess until then we don't have to send them anywere. yeah we can add a support email.

faboweb: Help Center has Intercom

faboweb:
@jbibla can you some basic styles in a free minute?

app/src/initializeApp.js
jbibla: it's not really obvious / in the corner - but it's ok - this is a rare screen 😄

app/src/initializeApp.js
```diff
@@ -37,7 +37,16 @@ function setOptions(urlParams, store) {
   }
 }
 
+export const bootError = (error) => {
+  document.querySelector("#app").innerText = `
```
jbibla, line 41: is this really the best way to do this?

app/src/initializeApp.js
jbibla, line 42:
```suggestion
    Uh oh. Lunie failed to start. Our team has been notified of the issue. Please try refreshing this page in a few minutes. 
```

what about something like this?

jbibla:
> @jbibla can you some basic styles in a free minute?

yeah - but it would be a lot simpler if it had a div or class. does it need to replace the whole app instead of using one of our existing error components?

app/src/initializeApp.js
faboweb: need to do basic JS here as the whole Vue app might have been crashed so we can't use virtual dom rendering.

app/src/initializeApp.js
faboweb: You don't want to give a way to contact us?

faboweb:
> yeah - but it would be a lot simpler if it had a div or class. does it need to replace the whole app instead of using one of our existing error components?

we can add a class to `#app` when calling the error function. or we add the error in another element. Overwriting the content of `#app` makes sure we don't display broken content and only show the error.

faboweb:
image

Bitcoinera:
Is this one ready to review?

PR #4752, opened 2020-08-20
Ana/4738 enable voting with time locking in polkadot

### LINKED ISSUE #4738: Enable Vote with Time Locking and Deposit in Polkadot/Substrate
Part of #4456

Bitcoinera:
Closes #4738 

**Description:**

https://polkascan.io/polkadot/transaction/0xeef94c98a9f959a14857ce223fc0682ef8662c96e7d1fd5d04f86e95b7a35d32

image

:star_struck: 
Thank you! 🚀

---

For contributor:

For reviewer:


faboweb:
🚀 Deployed on https://5f3ed8cadc14746ecc323f25--lunieio.netlify.app

Bitcoinera:
What does the multiplier card actually represent?

Bitcoinera:
How it is currently looking:

image

Bitcoinera:
We also need to enable the `Vote` query for Polkadot, in order to get the lastVote option. Although I keep on voting yes on the same proposal and don't get any error. Guess it only counts the first time though

Bitcoinera:
Just removed the Multiplier as we talked about in the meeting:

image

Bitcoinera:
I call this one ready!

jbibla:
looks great. conflict needs to be resolved. 

i'll QA after my meeting 😄

api/lib/schema.js
```diff
@@ -267,6 +267,8 @@ const typeDefs = gql`
     proposer: String
     initialDeposit: InputCoin
     voteOption: String
+    lockedBalance: Float
+    conviction: Int
```
faboweb, line 271: can we call this something understandable?

app/src/ActionModal/components/ModalVoteSubstrate.vue
```diff
@@ -0,0 +1,428 @@
...
+    feature-flag="vote"
+    @close="clear"
+    @txIncluded="onSuccess"
+  >
+    <div class="action-modal-group vote-options">
+      <h3>Select your vote</h3>
+      <div class="flex-row">
+        <TmBtn
+          id="vote-yes"
+          :class="{ active: vote === `Yes` }"
```
faboweb, line 22: use enums?

app/src/ActionModal/components/ModalVoteSubstrate.vue
```diff
@@ -0,0 +1,428 @@
...
+        Available to vote <span>{{ stakingDenomBalance.available }}</span>
+      </p>
+      <div class="locked-balance-container">
+        <TmFormGroup
+          :error="$v.lockedBalance.$error && $v.lockedBalance.$invalid"
+          class="action-modal-group"
+          field-id="lockedBalance"
+        >
+          <TmField
+            id="lockedBalance"
```
faboweb, line 52: why use an id here?

app/src/ActionModal/components/ModalVoteSubstrate.vue
```diff
@@ -0,0 +1,428 @@
...
+    </div>
+    <div class="totals flex-row">
+      <div class="card">
+        <span class="card-title">Total</span>
+        <span>{{ totalVotingPower }}</span>
+        <span>&nbsp;{{ currentNetwork.stakingDenom }}</span>
+      </div>
+    </div>
+    <div class="buttons">
+      <!-- <TmBtn value="Send your Vote" /> -->
```
faboweb, line 115:
```suggestion
```

app/src/ActionModal/components/ModalVoteSubstrate.vue
```diff
@@ -0,0 +1,428 @@
...
+  <!-- </SessionFrame> -->
+</template>
+<script>
+import { mapGetters } from "vuex"
+import { required, decimal } from "vuelidate/lib/validators"
+import { SMALLEST } from "src/scripts/num"
+import ActionModal from "./ActionModal"
+import TmFormGroup from "src/components/common/TmFormGroup"
+import TmField from "src/components/common/TmField"
+// import SessionFrame from "common/SessionFrame"
```
faboweb, line 132:
```suggestion
```

app/src/ActionModal/components/ModalVoteSubstrate.vue
```diff
@@ -0,0 +1,428 @@
...
+import TmBtn from "src/components/common/TmBtn"
+import TmFormMsg from "src/components/common/TmFormMsg"
+import { messageType } from "../../components/transactions/messageTypes"
+import gql from "graphql-tag"
+
+const isValid = (option) => option === `Yes` || option === `No`
```
faboweb, line 138: is this necessary?

app/src/ActionModal/components/ModalVoteSubstrate.vue
```diff
@@ -0,0 +1,428 @@
...
+
+export default {
+  name: `modal-vote-substrate`,
+  components: {
+    ActionModal,
+    // SessionFrame,
```
faboweb, line 144:
```suggestion
```

app/src/ActionModal/components/ModalVoteSubstrate.vue
```diff
@@ -0,0 +1,428 @@
...
+    TmFormGroup,
+    TmField,
+    TmFormMsg,
+    TmBtn,
+  },
+  props: {
+    proposalId: {
+      type: [Number, String],
```
faboweb, line 152: nope, please reduce to 1 type

app/src/ActionModal/components/ModalVoteSubstrate.vue
```diff
@@ -0,0 +1,428 @@
...
+  }),
+  computed: {
+    ...mapGetters([`address`, `currentNetwork`]),
+    transactionData() {
+      return {
+        type: messageType.VOTE,
+        proposalId: this.proposalId,
+        voteOption: this.vote,
+        lockedBalance: Number(this.lockedBalance) || 0,
+        conviction: this.selectedLockingOption
```
faboweb, line 187: how about `voteTokenTimeLock`?

app/src/ActionModal/components/ModalVoteSubstrate.vue
```diff
@@ -0,0 +1,428 @@
...
+          available: 0,
+        }
+      }
+    },
+  },
+  watch: {
+    lockedBalance: {
+      handler() {
+        this.totalVotingController()
+        const input = document.querySelector(".tm-field")
```
faboweb, line 214: massive hack. what are you trying to do here?

app/src/ActionModal/components/ModalVoteSubstrate.vue
```diff
@@ -0,0 +1,428 @@
...
+          return (this.lockingPeriod = "448")
+        case `6x`:
+          return (this.lockingPeriod = "896")
+        case `Set Max`:
+          return (this.lockingPeriod = "896")
+        default:
+          return (this.lockingPeriod = "0")
+      }
+    },
+    totalVotingController() {
```
faboweb, line 285: make this a computed and remove from the other calls

app/src/ActionModal/components/ModalVoteSubstrate.vue
```diff
@@ -0,0 +1,428 @@
...
+      },
+      /* istanbul ignore next */
+      skip() {
+        return !this.address
+      },
+      update(result) {
+        this.lockedBalance = result.balancesV2.find(
+          ({ denom }) => denom === this.currentNetwork.stakingDenom
+        ).available
+        this.totalVotingPower = this.lockedBalance
```
faboweb, line 318: yep, make computed

app/src/ActionModal/components/ModalVoteSubstrate.vue
```diff
@@ -0,0 +1,428 @@
...
+  display: block;
+  font-size: var(--sm);
+  color: var(--input-bc);
+}
+
+.line {
+  border: solid 1px var(--app-fg-hover);
+}
+
+.flex-row {
```
faboweb, line 349: @jbibla should we have such classes in the components or globally?

app/src/signing/networkMessages/polkadot.js
```diff
@@ -154,6 +154,26 @@ export async function ClaimRewardsTx(senderAddress, {}, network, api) {
...
 }
 
+// Vote
+export async function VoteTx(
+  senderAddress,
+  { proposalId, lockedBalance, voteOption, conviction },
+  network,
+  api
+) {
+  const referendumId = proposalId.split("-")[1]
```
faboweb, line 164: hack, can we get this without string manipulation?

Bitcoinera: not here, not in the FE. I think doing this is fine. I can add a comment in API though highlighting the importance of being consistent with this format

app/src/ActionModal/components/ModalVoteSubstrate.vue
Bitcoinera: this is to create the cool effect of the TmField/input adapting to the length of what you type. It is really cool

app/src/ActionModal/components/ModalVoteSubstrate.vue
Bitcoinera: this is copy pasted from the original ModalVote :shrug:

app/src/ActionModal/components/ModalVoteSubstrate.vue
Bitcoinera: OK, I remove this commented code

app/src/ActionModal/components/ModalVoteSubstrate.vue
Bitcoinera: This is also how it is done in ModalVote. I guess it is not so useful, this validation?

app/src/ActionModal/components/ModalVoteSubstrate.vue
Bitcoinera: removing

api/lib/schema.js
Bitcoinera: how would you call conviction? `timeLock`, right?

app/src/ActionModal/components/ModalVoteSubstrate.vue
faboweb: yep, I think it is not useful. the only useful check is, that there is any option selected

app/src/ActionModal/components/ModalVoteSubstrate.vue
faboweb: well, let's the improve the code

app/src/ActionModal/components/ModalVoteSubstrate.vue
faboweb: @jbibla is there any other way of doing this not using js dom manipulation?

app/src/signing/networkMessages/polkadot.js
faboweb: how about sending the actual id from the API?

Bitcoinera: yeah, that would work

Bitcoinera:
Just realized there is a but with the `0.1` option. Even converting to Float it throws an mysterious error coming from polkadot/api:

image

app/src/ActionModal/components/ModalVoteSubstrate.vue
Bitcoinera: at the end I need this to text-align the balance within the input

api/lib/schema.js
Bitcoinera: I renamed it to `timeLock`

app/src/signing/networkMessages/polkadot.js
Bitcoinera: Then we should add a new field in type Proposal too: `proposalType` , that would be an enum with `referemdum`, `democracy` and so on, so we can now in the FE which one to display as id.

But wait, now thinking about it, it is not a good solution, since we need different ids because of cache... otherwise we will get cache errors using the same id (same number)

app/src/ActionModal/components/ModalVoteSubstrate.vue
jbibla: oh interesting! 

yeah, i suppose an alternative would be to leave the input 100% width but center the text.

app/src/ActionModal/components/ModalVoteSubstrate.vue
jbibla: we don't use a lot of global styles. this is fine especially because it's scoped.

app/src/ActionModal/components/ModalVoteSubstrate.vue
```diff
@@ -0,0 +1,428 @@
...
+  height: 4rem;
+  width: 100%;
+}
+
+.vote-options button.active {
+  background: var(--highlight);
+  border-color: var(--highlight);
+}
+</style>
+<style>
```
jbibla, line 400: if this isn't scoped it will hack other components in unexpected ways 🤔

Bitcoinera: yeah, that is why I used id here. This CSS is only applied when the component is present right? Otherwise I can just use ids, actually, and scope it all

app/src/ActionModal/components/ModalVoteSubstrate.vue
Bitcoinera: yeah, I also thought of this option but the problem there is that how can users then know that there is an input there where they need to write something? :shrug:  leaving the border this is easier to see

app/src/ActionModal/components/ModalVoteSubstrate.vue
Bitcoinera: ah, but there is a problem. The classes are coming from TmField and so on. I need to do that in order to change those classes coming from other components. That is the issue here. How else can you do that? :shrug:  I think there is no other way

jbibla: do you have to use that component?

app/src/signing/networkMessages/polkadot.js
faboweb: you can't use the property `id`. it would be interpreted as a cachable id. you need to use sth like `proposalId` or `onChainId`

app/src/ActionModal/components/ModalVoteSubstrate.vue
jbibla: there will be a flashing vertical line like in all focussed inputs

app/src/signing/networkMessages/polkadot.js
Bitcoinera: Done

app/src/ActionModal/components/ModalVoteSubstrate.vue
Bitcoinera: Fixed in this component

app/src/ActionModal/components/ModalVoteSubstrate.vue
Bitcoinera: We can use a normal input :shrug:

app/src/ActionModal/components/ModalVoteSubstrate.vue
jbibla: done

api/lib/reducers/polkadotV0-reducers.js
```diff
@@ -560,6 +562,7 @@ function treasuryProposalReducer(
 ) {
   return {
     id: `treasury-`.concat(proposal.index || proposal.votes.index),
+    proposalId: proposal.index || proposal.votes.index,
```
faboweb, line 565: why are there 2 possibilities here?

app/src/components/governance/PageProposal.vue
```diff
@@ -170,7 +170,11 @@
       <ModalVoteSubstrate
         v-if="currentNetwork.network_type === `polkadot`"
         ref="modalVote"
-        :proposal-id="proposalId"
+        :proposal-id="
+          currentNetwork.network_type === `polkadot`
```
faboweb, line 174: why? this should be always the same. cosmos should also provide the proposalId, right?

app/src/components/governance/PageProposal.vue
```diff
@@ -170,7 +170,11 @@
...
+            ? proposal.proposalId
+            : proposalId
```
faboweb, line 176: if this is not substrate, we don't need this at all
```suggestion
            : `cosmos-networks-will-not-need-this-component`
```

faboweb, line 176:
```suggestion
        proposal.proposalId
```

faboweb: not needed as you already do  a v-if for the component

Bitcoinera: true

api/lib/reducers/polkadotV0-reducers.js
Bitcoinera: because treasury proposals can have two different formats: council or top. This has given me too some headaches

Bitcoinera:
This is ready! :woman_dancing: !!

Issue #4739, opened 2020-08-18
fix for purgatory balance in substrate chains


jbibla:
what to do about unbonding and unnominations?

- add unstake button on portfolio
- opens action modal with no validator selection option
- only for substrate chains
- add the total staked balance to validator section 
- rename "Your Validators" to "Your Stake"

---
not actual designs - just brainstorm ideas:

![Screen Shot 2020-08-18 at 10 57 21 AM](https://user-images.githubusercontent.com/6021933/90529983-55b15e80-e142-11ea-8ce1-9fa6516c9c4d.png)
![Screen Shot 2020-08-18 at 10 48 22 AM](https://user-images.githubusercontent.com/6021933/90529984-55b15e80-e142-11ea-9c27-9157e500e95c.png)

jbibla:
i believe this has been addressed

PR #4770, opened 2020-08-26
Ana/add bondedStake query to API

Bitcoinera:
Related to #4739 
**Description:**

The API PR to fix the problem of purgatory invisible stake in Polkadot/Substrate

Thank you! 🚀

---

For contributor:

For reviewer:


Bitcoinera:
image

Bitcoinera:
Query to test:

```
query proposals {
 bondedStake(networkId:"polkadot-testnet", address:"5EZjN7ZD6hLN9wLGju8Uzu4NjXtTFhCpS8wUAxVdvTGEQaQY") {
  total
  active
  unlocking
  stale
}
}
```

It still needs to wait for 1h in order for the era in Westend to pass.

Some considerations: in the beginning I thought of making a query only for the unnominated/stale bonded stake, but then I thought that having all this data already there, in stakingLedger, it wouln't make sense adding this as well.

The advantage is that we can also add the `totalBonded` row once we merge this one. In Cosmos networks we already have this info coming from balance (just need to do total - available). Or, if you prefer, we can also subtract `unlocking` there :shrug: for Cosmos

faboweb:
What do we need this for? We already have the staked tokens implied in the balances query.

Bitcoinera:
> What do we need this for? We already have the staked tokens implied in the balances query.

The original idea was getting the stale bonded balance. It is explained in the issue. We were discussing this in the meeting too: users can have bonded but unnominated balance/stake, Polkadot allows this. In Lunie, these balances won't just show.

But to get this stale bonded we need a new query. Would be bad practice to have it in FE

faboweb:
> In Lunie, these balances won't just show.

This is not true. We are displaying liquid stake. Ergo bonded stake is `total stake - liquid stake`. Even if display this explicit, we don't need another query for that.

Bitcoinera:
> This is not true. We are displaying liquid stake. Ergo bonded stake is `total stake - liquid stake`. Even if display this explicit, we don't need another query for that.

I think it give people more comfort to see that their tokens are somewhere. If they see total but then don't see any delegation, they can easily get scared. I know what I'm saying since I have been in this situation.

About calculating with `total stake - liquid`, the problem is that you need to search for staked with no nomination... Unless I am missing something here

faboweb:
> About calculating with total stake - liquid, the problem is that you need to search for staked with no nomination... Unless I am missing something here

you can search using `stakingDenom` right?

Bitcoinera:
> you can search using `stakingDenom` right?

What do you mean?

Bitcoinera:
Good bye, my sweet PR! :hand:

PR #4810, opened 2020-09-01
Fabo/check login url

faboweb:
Check redirect url to be a firebase url


faboweb:
🚀 Deployed on https://5f4ea74921da3708c82465b3--lunieio.netlify.app

PR #4817, opened 2020-09-02
Fabo/council voting

faboweb:
Removes council proposals and handles council voting for treasury proposals


Bitcoinera:
Mmm, I am thinking we can combine then the Council and Treasury proposals in one? So we need to get as much info as we can. Right now we are missing the tally, for instance... Need to combine them

Bitcoinera:
Was trying this but right then all the council / treasury proposals disappeared :shrug:

api/lib/source/polkadotV0-source.js
```diff
@@ -693,9 +693,14 @@ class polkadotAPI {
       )
     }
     if (type === `treasury`) {
-      description = `This is a Treasury Proposal whose description and title have not yet been edited on-chain. Only the proposer address (${
-        proposal.proposer || proposer
-      }) is able to change it.`
+      const { meta } = proposal.council[0]
```
faboweb, line 696: what do we need the meta for? the description should stay the same as the treasury proposal

Bitcoinera: Jordan and I just talked that there should be a description coming from the chain (for this we need `meta`) and the other default one that is custom from us in a separate field (`note`).

But this change will be done in another PR. Yes, we need `meta`

faboweb: we use the description of the treasury proposal, not from the council. so please revert the change

Bitcoinera: it is the same if you look at the object (OK, we need to wait until there are new Treasury proposals, but you can see here it is the same Object):

```
{
"council": [],
"id": 9,
"proposal": {
"proposer": "16XXocb76BpprdspsR7j5iWFHf4aTMnsVHxgGBXfixs1FFLb",
"value": 70000000000000,
"beneficiary": "16XXocb76BpprdspsR7j5iWFHf4aTMnsVHxgGBXfixs1FFLb",
"bond": 3500000000000
}
}
```

Bitcoinera: Like you said, they are actually the same treasury and council, and council it is just a way of voting, but the description is the same since they are both the same proposal

faboweb: the description is actually not the same

Bitcoinera: Are you sure about that? It should be. I will check myself

Bitcoinera: It is exactly the same description. You can check yourself by copy-pasting this on the Javascript section in polkadot.js

```
const councilProposals = await api.derive.council.proposals()
const treasuryProposals = await api.derive.treasury.proposals()

console.log(JSON.stringify(councilProposals))
console.log(JSON.stringify(treasuryProposals))

const councilDescriptionMeta = await api.registry.findMetaCall(councilProposals[0].proposal.callIndex)
const councilDescription = councilDescriptionMeta.meta.documentation.toString()

const treasuryDescriptionMeta = await api.registry.findMetaCall(treasuryProposals.proposals[0].council[0].proposal.callIndex)
const treasuryDescription = treasuryDescriptionMeta.meta.documentation.toString()

console.log('\n COUNCIL DESCRIPTION', councilDescription)
console.log('\n TREASURY DESCRIPTION', treasuryDescription)
```

Bitcoinera:
Need to fix this:

image

Bitcoinera:
Good! Everything working as expected :sunglasses: 

Let's merge! :woman_dancing:

faboweb:
Waiting on a treasury proposal

Bitcoinera:
> Waiting on a treasury proposal

There is one on Polkadot: https://polkadot.polkassembly.io/treasury/10

PR #4816, opened 2020-09-02
Ana/4812 get identity for proposer voter depositer

Bitcoinera:
Related to #4812 

**Description:**

- [x] depositers and voters and proposers should be identity objects (name, address, image if possible)

Thank you! 🚀

---

For contributor:

- [ ] Added changes entries. Run `yarn changelog` for a guided process.
- [ ] Reviewed `Files changed` in the github PR explorer
- [ ] Attach screenshots of the UI components on the PR description (if applicable)
- [ ] Scope of work approved for big PRs

For reviewer:

- [ ] Manually tested the changes on the UI

PR #4824, opened 2020-09-03
Ana/fix voter and depositer for proposal query

Bitcoinera:
Closes #ISSUE

**Description:**

All good! Very important PR that fixes things left behind in #4816 

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/api/lib/source/polkadotV0-source.js
+++ b/api/lib/source/polkadotV0-source.js
@@ -711,7 +711,9 @@ class polkadotAPI {
     }
   }
 
-  getDemocracyProposalDetailedVotes(proposal, links) {
+  async getDemocracyProposalDetailedVotes(proposal, links) {
+    const api = await this.getAPI()
+
     // in democracy proposals there is the first opening deposit made by the proposer
     // afterwards every account that seconds the proposal must deposit the same amount from the initial deposit
     const depositsSum = toViewDenom(
@@ -739,10 +741,15 @@ class polkadotAPI {
         ]
       }))
     )
-    const votes = proposal.seconds.map((secondAddress) => ({
-      voter: secondAddress,
-      option: `Yes`
-    }))
+    const votes = await Promise.all(
+      proposal.seconds.map(async (secondAddress) => {
+        const voterInfo = await api.derive.accounts.info(secondAddress)
+        return {
+          voter: this.reducers.networkAccountReducer(voterInfo),
+          option: `Yes`
+        }
+      })
+    )
     const votesSum = proposal.seconds.length
     return {
       deposits,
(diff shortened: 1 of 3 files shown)


Bitcoinera:
Fixes the detailedVotes query within proposal! :tada: @jbibla

jbibla (approved):

api/lib/source/polkadotV0-source.js
```diff
@@ -711,16 +711,19 @@ class polkadotAPI {
...
     // afterwards every account that seconds the proposal must deposit the same amount from the initial deposit
     const depositsSum = toViewDenom(
       this.network,
       BigNumber(proposal.balance).times(proposal.seconds.length).toNumber()
     )
+    const depositerInfo = await api.derive.accounts.info(proposal.proposer)
     const deposits = [
       {
-        depositer: proposal.proposer,
+        depositer: this.reducers.networkAccountReducer(depositerInfo),
```
jbibla, line 726: instead of this being a unique key, why not just `identity` or `networkAccount` for all?

api/lib/source/polkadotV0-source.js
```diff
@@ -729,20 +732,30 @@ class polkadotAPI {
...
-        depositer: second,
-        amount: [
-          {
-            amount: toViewDenom(this.network, proposal.balance),
-            denom: this.network.stakingDenom
+      Promise.all(
+        proposal.seconds.map(async (second) => {
+          const secondDepositerInfo = await api.derive.accounts.info(second)
+          return {
+            depositer: this.reducers.networkAccountReducer(secondDepositerInfo),
```
jbibla, line 739: instead of this being a unique key, why not just `identity` or `networkAccount` for all?

api/lib/source/polkadotV0-source.js
```diff
@@ -729,20 +732,30 @@ class polkadotAPI {
...
-        ]
-      }))
+        })
+      )
+    )
+    const votes = await Promise.all(
+      proposal.seconds.map(async (secondAddress) => {
+        const voterInfo = await api.derive.accounts.info(secondAddress)
+        return {
+          voter: this.reducers.networkAccountReducer(voterInfo),
```
jbibla, line 754: instead of this being a unique key, why not just `identity` or `networkAccount`?

Bitcoinera: I am not sure I understand what you mean? You mean a key like `voter` ?

jbibla: we have voter, depositer, and proposer but they are the same identity object. instead of changing the name of the key let's use the same key name. 

```
voter
depositer
proposer
```
```
networkAccount
networkAccount
networkAccount
```

Bitcoinera: I think that would bring a lot of confusion just like that. What about `enacter`? Not sure. I think it is better using those keys, more intuitive. The other is not so clearly understandable

Bitcoinera: I see that it makes sense standardizing, but how can you tell `account` or `identity` refers to the account voting, proposing or depositing?

faboweb: what else would it refer to?

PR #4827, opened 2020-09-04
redeploy scriptRunner when we redeploy API

Bitcoinera:
Closes #ISSUE

**Description:**

Apparently scriptRunner stopped working because it couldn't sync with the API after one of its last deployments.

This change in the workflow here is to make sure we redeploy scriptRunner also every time we deploy a new API.

We will check if it works the next time we merge something to develop for the `api` submodule (`scriptRunner_deploy_development`)

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/.github/workflows/scriptRunner_deploy_development.yml
+++ b/.github/workflows/scriptRunner_deploy_development.yml
@@ -6,6 +6,7 @@ on:
       - 'develop'
     paths:
       - 'scriptRunner/**'
+      - 'api/**'
 env:
   WORKING_DIRECTORY: ./scriptRunner
 jobs:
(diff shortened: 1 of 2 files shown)


.github/workflows/scriptRunner_deploy_development.yml
```diff
@@ -6,6 +6,7 @@ on:
       - 'develop'
     paths:
       - 'scriptRunner/**'
+      - 'api/**'
```
faboweb, line 9: we discussed this in the call. are you certain this works? or do now changes be present in scriptRunner AND API folder?

Bitcoinera: It will work:

"If at least one path matches a pattern in the paths filter, the workflow runs. "

Source: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions

PR #4861, opened 2020-09-09
Ana/beneficiary is also an Object (NetworkAccount)

Bitcoinera:
Closes #ISSUE

**Description:**

Tested voters, proposers and depositers. Now it is much faster.

Merge! :woman_dancing: 

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/api/lib/reducers/polkadotV0-reducers.js
+++ b/api/lib/reducers/polkadotV0-reducers.js
@@ -542,12 +542,12 @@ function democracyReferendumReducer(
   proposal,
   totalIssuance,
   blockHeight,
-  detailedVotes,
-  proposerInfo
+  detailedVotes
 ) {
   return {
     id: `referendum-`.concat(proposal.index),
     proposalId: proposal.index,
+    proposer: proposal.proposer,
     networkId: network.id,
     type: proposalTypeEnum.PARAMETER_CHANGE,
     title: `Proposal #${proposal.index}`,
@@ -558,7 +558,6 @@ function democracyReferendumReducer(
     statusEndTime: getStatusEndTime(blockHeight, proposal.status.end),
     tally: tallyReducer(network, proposal.status.tally, totalIssuance),
     deposit: toViewDenom(network, proposal.status.tally.turnout),
-    proposer: networkAccountReducer(proposerInfo),
     detailedVotes
   }
 }
--- a/api/lib/reducers/polkadotV0-reducers.js
+++ b/api/lib/reducers/polkadotV0-reducers.js
@@ -693,13 +693,12 @@ function topVoterReducer(
   validators,
   network
 ) {
-  const { identity, nickname } = accountInfo || {}
   const councilMemberInfo = electionInfo.members.find(
     (electionInfoMember) =>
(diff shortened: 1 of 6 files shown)


api/lib/source/polkadotV0-source.js
```diff
@@ -37,6 +37,13 @@ class polkadotAPI {
     return api
   }
 
+  async getNetworkAccountInfo(address, api) {
```
faboweb, line 40: nice

faboweb: the identities will get stale. but we shouldn't care about it right now as we release the API frequently

faboweb: how about returning the result after `networkAccountReducer`? that would reduce code

faboweb:
```suggestion
  async getNetworkAccountInfo(address, api, reducers) {
```

Bitcoinera: ah, I see what you mean. And calling `networkAccountReducer` only once inside `getNetworkAccountInfo`.

Good idea

faboweb: yep

faboweb: as a guide: things we store in "store" should already be in the Lunie data model so after reducers.

Bitcoinera: Done

api/lib/reducers/polkadotV0-reducers.js
```diff
@@ -693,6 +693,7 @@ function topVoterReducer(
   validators,
   network
 ) {
+  console.log('\n ACCOUNT INFO', accountInfo)
```
faboweb, line 696:
```suggestion
```

Bitcoinera:
why it suddenly says there is no changelog?

faboweb:
Can this be merged now?

Bitcoinera:
> Can this be merged now?

Yes

Bitcoinera:
There is changelog. I don't know why the check fails

api/lib/source/polkadotV0-source.js
```diff
@@ -1198,10 +1198,7 @@ class polkadotAPI {
       ),
       topVoters: await Promise.all(
         topVoters.map(async (topVoterAddress) => {
-          const accountInfo = await this.getNetworkAccountInfo(
-            topVoterAddress,
-            api
-          )
+          const accountInfo = await api.derive.accounts.info(topVoterAddress)
```
faboweb, line 1201: why this change?

Bitcoinera: because validators need to get more data from `identity` like web and so on, all fields that are not present in the `NetworkAccount` Object

faboweb: `topVoterReducer` doesn't use these infos. please double check

Bitcoinera: ah true

Issue #4812, opened 2020-09-01
Governance API questions and improvements


jbibla:
1. can depositers and voters be identity objects like proposer (name, address, image if possible)?
2. can proposer be an object instead of a string?
3. timeline and links don't need to be inside of detailedVotes
4. proposal descriptions should always include the content of the proposal. eg: `/polkadot/proposals/democracy-1` has been replaced with our "description" but the content of the parameter change is missing. let's always leave the actual proposal description in the description field and add our content as a `note` (https://kusama.polkassembly.io/referendum/81)
5. timeline should include all the steps of the governance process (including future steps - array of 3 or 4 objects) and one of the objects should be `current: true` to indicate where the proposal is in the timeline
6. can beneficiary be an object instead of a string?
7. what is the `validator` field for?
8. can we have actual IDs like polkassembly (https://kusama.polkassembly.io/referendum/81 appears as `democracy-1` for us)?
9. `percentageDepositsNeeded` is null - and why is it percentage as opposed to total number of tokens?
10. `/polkadot/proposals/democracy-1` has no date / time information at all anywhere ... is this expected?
11. `polkadot/proposals/referendum-2` all the votes (163) have the same `voter` - is `voter` supposed to be address? ideally this would be an identity object

Bitcoinera:
> can depositers and voters be identity objects like proposer (name, address, image if possible)?
> can proposer be an object instead of a string?

You mean like creating a new API Object for proposer? This doesn't exist. Proposer is always a String, like you said

jbibla:
> You mean like creating a new API Object for proposer?

yeah - if you look at the designs (😉 ) you would see what the requirements were intended to be ... the FE needs a name, address, and image always. for votes and deposits we need extra info (amount with denom and time):

<img width="369" alt="Screen Shot 2020-09-02 at 10 21 47 AM" src="https://user-images.githubusercontent.com/6021933/91995685-1cfab300-ed06-11ea-96ca-b6dccc3a1ef7.png">

<img width="1105" alt="Screen Shot 2020-09-02 at 10 22 53 AM" src="https://user-images.githubusercontent.com/6021933/91995821-43b8e980-ed06-11ea-961c-d17cb8a083c9.png">

Bitcoinera:
OK, I see. Getting the names and so on for proposers and voters is quite easy. The other thing with the deposits times I need to investigate a bit, but yeah, surely possible

Bitcoinera:
[...]

PR #4865, opened 2020-09-09
Ana/fix voter cache all voters the same

Bitcoinera:
Related to #4812  

**Description:**

Fixes all the voters being exactly the same in FE. Apollo needed ID to catch votes correctly. Probably need to do the same for deposits too actually

Fixes this:

image

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/api/lib/reducers/polkadotV0-reducers.js
+++ b/api/lib/reducers/polkadotV0-reducers.js
@@ -542,12 +542,12 @@ function democracyReferendumReducer(
   proposal,
   totalIssuance,
   blockHeight,
-  detailedVotes,
-  proposerInfo
+  detailedVotes
 ) {
   return {
     id: `referendum-`.concat(proposal.index),
     proposalId: proposal.index,
+    proposer: proposal.proposer,
     networkId: network.id,
     type: proposalTypeEnum.PARAMETER_CHANGE,
     title: `Proposal #${proposal.index}`,
@@ -558,7 +558,6 @@ function democracyReferendumReducer(
     statusEndTime: getStatusEndTime(blockHeight, proposal.status.end),
     tally: tallyReducer(network, proposal.status.tally, totalIssuance),
     deposit: toViewDenom(network, proposal.status.tally.turnout),
-    proposer: networkAccountReducer(proposerInfo),
     detailedVotes
   }
 }
--- a/api/lib/reducers/polkadotV0-reducers.js
+++ b/api/lib/reducers/polkadotV0-reducers.js
@@ -45,12 +45,12 @@ function validatorReducer(network, validator) {
     networkId: network.id,
     chainId: network.chain_id,
     operatorAddress: validator.accountId,
-    website:
-      validator.identity.web && validator.identity.web !== ``
-        ? validator.identity.web
-        : ``,
+    website: validator.identity.web ? validator.identity.web : ``,
     identity: validator.identity.twitter,
-    name: identityReducer(validator.accountId, validator.identity),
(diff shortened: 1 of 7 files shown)


Bitcoinera:
Comes from #4861 , which should be merged first and asap

api/lib/reducers/polkadotV0-reducers.js
```diff
@@ -45,12 +45,12 @@ function validatorReducer(network, validator) {
     networkId: network.id,
     chainId: network.chain_id,
     operatorAddress: validator.accountId,
-    website:
-      validator.identity.web && validator.identity.web !== ``
-        ? validator.identity.web
-        : ``,
+    website: validator.identity.web ? validator.identity.web : ``,
```
faboweb, line 48: do we need to check for `validator.identity`?

api/lib/source/polkadotV0-source.js
```diff
@@ -37,6 +37,16 @@ class polkadotAPI {
...
   }
 
+  async getNetworkAccountInfo(address, api) {
+    if (typeof address === `object`) address = address.toHuman()
+    if (this.store.identities[address]) return this.store.identities[address]
+    const accountInfo = await api.derive.accounts.info(address)
+    this.store.identities[address] = this.reducers.networkAccountReducer(
+      accountInfo
+    )
+    return this.reducers.networkAccountReducer(accountInfo)
```
faboweb, line 47:
```suggestion
    return this.store.identities[address]
```

api/lib/source/polkadotV0-source.js
```diff
@@ -858,26 +877,29 @@ class polkadotAPI {
...
+            voter,
             option: `Yes`
           }
         })
         .concat(
           proposal.allNay.map(async (nay) => {
+            const voter = await this.getNetworkAccountInfo(nay.accountId, api)
             return {
-              voter: this.reducers.networkAccountReducer(nay.accountId),
+              id: voter.address,
```
faboweb, line 901: shouldn't this be a combination of the proposal id and the voter? if not, the vote is not unique across several proposals. but that might not be a problem

api/lib/reducers/polkadotV0-reducers.js
Bitcoinera: no, we don't, since `api.derive.accounts.info()` always returns an Object, even when the fields are empty. I have checked

api/lib/source/polkadotV0-source.js
Bitcoinera: ah true :sweat_smile: slipped my mind there

api/lib/source/polkadotV0-source.js
Bitcoinera: I think it is not a problem since it will be first taken apart according to the proposal ID. Since it is an Object only living inside the proposal Object, Apollo won't mix them up.

My only concern here actually are deposits... maybe I can do it in this PR, add `id` to schema and then change the Vote ids to `"vote-"voter.address` and Deposit have `"deposit-"depositer.address`. Although maybe that is not either needed being separate Objects themselves :shrug:

Bitcoinera: But we need to include Deposit ids too, of that I'm sure. The only thing is that Jordan didn't include deposits in his new governance components PR, but still for the future is a good idea to get those in

faboweb: yeah, we need to add deposits, he just forgot I guess

PR #4874, opened 2020-09-10
Fabo/4873 claim polkadot performance api

faboweb:
Test via:

```
query {rewards(networkId:"polkadot" delegatorAddress:"13zLCfnA8kqShQrojhDQDT8EeRai17vVPW4W8442KKQ4HcjB" withHeight:true) {id amount height validator {operatorAddress}}}
```

### DIFF
--- a/api/lib/block-listeners/polkadot-node-subscription.js
+++ b/api/lib/block-listeners/polkadot-node-subscription.js
@@ -51,7 +51,7 @@ class PolkadotNodeSubscription {
     this.store.polkadotRPC = this.api
     this.store.polkadotRPCOpened = Date.now()
     await this.api.isReady
-    console.log(this.network.id +' API initialized')
+    console.log(this.network.id + ' API initialized')
   }
 
   async subscribeForNewBlock() {
--- a/api/lib/block-listeners/polkadot-node-subscription.js
+++ b/api/lib/block-listeners/polkadot-node-subscription.js
@@ -51,7 +51,7 @@ class PolkadotNodeSubscription {
     this.store.polkadotRPC = this.api
     this.store.polkadotRPCOpened = Date.now()
     await this.api.isReady
-    console.log('Polkadot initialized')
+    console.log(this.network.id +' API initialized')
   }
 
   async subscribeForNewBlock() {
--- a/api/lib/messageCreators/polkadot-transactions.js
+++ b/api/lib/messageCreators/polkadot-transactions.js
@@ -160,21 +160,34 @@ async function RestakeTx(senderAddress, api, { to, from, addressRole }) {
   return await getSignMessage(senderAddress, transactions, api)
 }
 
-async function ClaimRewardsTx(senderAddress, api, message, network, networkSource) {
+async function ClaimRewardsTx(
+  senderAddress,
+  api,
+  message,
+  network,
+  networkSource
+) {
   let allClaimingTxs = []
 
-  const rewards = await networkSource.getRewards(delegatorAddress, fiatCurrency, withHeight)
+  const rewards = await networkSource.getRewards(
+    delegatorAddress,
+    fiatCurrency,
+    withHeight
+  )
 
   if (rewards.length === 0) {
     allClaimingTxs = []
   } else {
     rewards
-    .sortBy((a,b) => a.height - b.height)
-    .forEach((reward) => {
-      allClaimingTxs.push(
-        api.tx.staking.payoutStakers(reward.validator.operatorAddress, reward.height)
-      )
-    })
+      .sortBy((a, b) => a.height - b.height)
+      .forEach((reward) => {
+        allClaimingTxs.push(
+          api.tx.staking.payoutStakers(
+            reward.validator.operatorAddress,
+            reward.height
+          )
+        )
+      })
   }
 
   if (allClaimingTxs.length === 0) {
--- a/api/lib/messageCreators/polkadot-transactions.js
+++ b/api/lib/messageCreators/polkadot-transactions.js
@@ -170,10 +170,7 @@ async function ClaimRewardsTx(senderAddress, api, { rewards }) {
       .sort((a, b) => a.height - b.height)
       .forEach((reward) => {
         allClaimingTxs.push(
-          api.tx.staking.payoutStakers(
-            reward.validator,
-            reward.height
-          )
+          api.tx.staking.payoutStakers(reward.validator, reward.height)
         )
       })
   }
--- a/api/lib/messageCreators/polkadot-transactions.js
+++ b/api/lib/messageCreators/polkadot-transactions.js
@@ -171,7 +171,7 @@ async function ClaimRewardsTx(senderAddress, api, { rewards }) {
       .forEach((reward) => {
         allClaimingTxs.push(
           api.tx.staking.payoutStakers(
-            reward.validator.operatorAddress,
+            reward.validator,
             reward.height
           )
         )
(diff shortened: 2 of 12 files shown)


api/lib/schema.js
```diff
@@ -327,6 +327,7 @@ const typeDefs = gql`
   type ClaimRewardsTx {
     amounts: [Coin]!
     from: [String]!
+    rewards: [Reward] # Polkadot only
```
Bitcoinera, line 330:
```suggestion
    rewards: [Reward] # Substrate only
```

Bitcoinera:
Can we please disable the codecov warnings? They make going through code changes really difficult

api/lib/block-listeners/polkadot-node-subscription.js
```diff
@@ -51,7 +51,7 @@ class PolkadotNodeSubscription {
     this.store.polkadotRPC = this.api
     this.store.polkadotRPCOpened = Date.now()
     await this.api.isReady
-    console.log('Polkadot initialized')
+    console.log(this.network.id + ' API initialized')
```
Bitcoinera, line 54: nice

api/lib/messageCreators/polkadot-transactions.js
```diff
@@ -160,30 +160,28 @@ async function RestakeTx(senderAddress, api, { to, from, addressRole }) {
   return await getSignMessage(senderAddress, transactions, api)
 }
 
-async function ClaimRewardsTx(senderAddress, api) {
+async function ClaimRewardsTx(
+  senderAddress,
+  api,
+  { rewards }
```
Bitcoinera, line 166: that's it. So simple actually. Great

api/lib/source/polkadotV0-source.js
```diff
@@ -363,7 +363,12 @@ class polkadotAPI {
     return filteredRewards
   }
 
-  async getRewards(delegatorAddress) {
+  async getRewards(delegatorAddress, fiatCurrency, withHeight) {
+    if (this.network.network_type !== 'polkadot' && withHeight) {
+      throw new Error(
+        'Rewards are only queryable per height in Polkadot networks'
```
Bitcoinera, line 369: Well, technically they are Substrate networks, but `network_type` is called `polkadot`. So I don't know. Maybe we can just leave it like that

api/lib/messageCreators/polkadot-transactions.js
```diff
@@ -160,30 +160,28 @@ async function RestakeTx(senderAddress, api, { to, from, addressRole }) {
...
-          )
-        }
+        )
       })
-    })
   }
+
   if (allClaimingTxs.length === 0) {
     throw new Error('There are no claimable rewards')
   }
```
Bitcoinera, line 187: When is this function from API used actually? I just tried it and was using the `polkadot.js` file in the signing folder in FE

Bitcoinera: ah, for network fees, I see it. It worked quite well actually

api/lib/reducers/polkadotV0-reducers.js
```diff
@@ -449,7 +449,15 @@ function rewardsReducer(network, validators, rewards, reducers) {
   return allRewards
 }
 
-function dbRewardsReducer(validatorsDictionary, dbRewards) {
+function dbRewardsReducer(validatorsDictionary, dbRewards, withHeight) {
+  if (withHeight) {
+    return dbRewards.map((reward) => ({
+      id: `${reward.validator}_${reward.denom}_${reward.height}`,
```
Bitcoinera, line 455: I don't get this so well, this `withHeight` logic. I mean, shouldn't it always be the case, that ID goes with height? Why not make it always like this?

Ah, maybe because Cosmos hasn't this logic of reward per height... Oh, I see. OK

Bitcoinera:
Mmm, for Polkadot seems to be still quite slow

Peek 2020-09-11 12-07

Bitcoinera:
OK, making it much faster by copying the API version

Bitcoinera:
It still fails... maybe claiming wasn't failing due a timeout after all?

image

api/lib/messageCreators/polkadot-transactions.js
faboweb: I opened an issue to remove this file and use the APP file to reduce code repetition #4879

app/src/ActionModal/components/ActionModal.vue
```diff
@@ -833,6 +833,10 @@ export default {
             denom,
           }))
         }
+        if (message.rewards) {
```
faboweb, line 836: we need this so we can calculate the fees

app/src/signing/networkMessages/polkadot.js
```diff
@@ -125,30 +125,24 @@ export async function RestakeTx(
   return await getSignMessage(senderAddress, transactions, api)
 }
 
-export async function ClaimRewardsTx(senderAddress, {}, network, api) {
+export async function ClaimRewardsTx(senderAddress, { rewards }, network, api) {
```
faboweb, line 128: I explicitly not did this in here to release the API first as the PR name suggests

Bitcoinera: :sweat_smile:

Bitcoinera: oh I see, didn't realize this API changes break the FE

app/src/ActionModal/components/ActionModal.vue
Bitcoinera: ah, you want to send it to API? It is quite complicated actually, you need `input Validator`, `input FiatValue`... all repeated code, and then the `__typename` thing was giving me errors. Maybe we don't need it

faboweb: ok, thanks for the analysis. will check it and find a solution.

app/src/signing/networkMessages/polkadot.js
faboweb: please revert the FE changes

Bitcoinera: give me a sec

Bitcoinera: Done

api/lib/schema.js
faboweb: Same comment: I would reduce the name "Substrate"

Bitcoinera: ok, fair enough

Bitcoinera:
I guess until the app PR is not in there is no performance improvement whatsoever

api/lib/messageCreators/polkadot-transactions.js
```diff
@@ -171,7 +171,7 @@ async function ClaimRewardsTx(senderAddress, api, { rewards }) {
       .forEach((reward) => {
         allClaimingTxs.push(
           api.tx.staking.payoutStakers(
-            reward.validator.operatorAddress,
+            reward.validator,
```
Bitcoinera: this looks like the whole validator Object, not only the address

api/lib/schema.js
```diff
@@ -257,6 +257,11 @@ const typeDefs = gql`
     denom: String
   }
 
+  input RewardInput {
+    validator: String! # just the address
```
Bitcoinera, line 261: Nice idea!

Bitcoinera:
The Codacy test fails due to this linting errors, but doing `yarn lint --fix` doesn't resolve them :shrug: Let's just leave it like it is

image

api/lib/messageCreators/polkadot-transactions.js
faboweb: the reward will only hold the address for the validator

faboweb:
this seems to work fine

Bitcoinera (approved):

PR #4876, opened 2020-09-10
Ana/add amount and time to vote

Bitcoinera:
Closes #ISSUE

**Description:**

WIP because I would like to investigate a bit if it is easy to get the cast date for votes & deposits

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/api/lib/schema.js
+++ b/api/lib/schema.js
@@ -72,6 +72,8 @@ const typeDefs = gql`
     id: String
     voter: NetworkAccount
     option: String
+    amount: Coin
+    time: String
   }
 
   type DetailedVotes {
--- a/api/lib/schema.js
+++ b/api/lib/schema.js
@@ -72,8 +72,8 @@ const typeDefs = gql`
     id: String
     voter: NetworkAccount
     option: String
-    amount: Coin
-    time: String
+    amount: Coin # Substrate only
+    time: String # Substrate only
   }
 
   type DetailedVotes {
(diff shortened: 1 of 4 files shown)


Bitcoinera:
Was investigating if it is also possible to get votes amounts for motions, but it seems this code only gives the votes to become council members:

```
const votes = await api.derive.council.votes()
console.log(votes.map(vote => {
  return JSON.stringify({
    address: vote[0],
    vote: JSON.stringify(vote[1])
  })
}))
```

Going to look at treasuries

Bitcoinera:
Ready to merge then

faboweb:
> This is what seems to be promoted by the W3F itself since they think the blockchain should always store minimal data

the blockchain yes, but the node could automatically index like in Cosmos...

api/lib/schema.js
```diff
@@ -72,6 +72,8 @@ const typeDefs = gql`
     id: String
     voter: NetworkAccount
     option: String
+    amount: Coin # Substrate only
+    time: String # Substrate only
```
faboweb, line 76: we don't get time in the end, am I wrong?

faboweb: ping

Bitcoinera: Yeah, I think this is already in anyway. Was part of #4887

Bitcoinera: yes, it is. closing this one now

PR #4910, opened 2020-09-15
Fabo/simple node subscription

faboweb:
Standardizes the node subscriptions as the logic is mostly duplicated

The old subscriptions are still in to be able to smoothly transition


Bitcoinera:
Great point! Yes, it was definitely totally duplicated, the whole node-subscription logic

api/lib/source/polkadotV0-source.js
```diff
@@ -37,6 +37,44 @@ class polkadotAPI {
...
+        body: JSON.stringify({
+          era,
+          networkId: this.network.id
+        })
+      }).catch((error) => {
+        console.error('Failed running Polkadot rewards script', error)
+        Sentry.captureException(error)
+      })
+    }
+  }
```
Bitcoinera, line 76: Ah, and you move this here. Interesting. Yes, good approach

Bitcoinera:
It's missing a `cosmos-node-subscription-inherit.js` file, no?

faboweb:
> cosmos-node-subscription-inherit.js

is not needed. cosmos networks so far would just use the base-node-subscription

faboweb:
alternatively we could just replace the old cosmos-node-subscription and polkadot-node-subscription files with inherit ones

Bitcoinera:
> is not needed. cosmos networks so far would just use the base-node-subscription

?? This is what I mean. You are not connecting `cosmos-node-subscription` to `base-node-subscription` in this PR :eyes: Let me check again

faboweb:
you could just set `block_listener_class_name` to `base-node-subscription` in the db for Cosmos. But I will change the existing ones to reduce effort to change things in the db

api/lib/block-listeners/polkadot-node-subscription.js
```diff
@@ -1,240 +1,29 @@
...
 
   // here we init the polkadot rpc once for all processes
   // the class gets stored in the store to be used by all instances
-  async initPolkadotRPC() {
-    this.api = new ApiPromise({
+  async initPolkadotRPC(network, store) {
+    const api = new ApiPromise({
       provider: new WsProvider(
-        this.network.rpc_url || this.network.public_rpc_url
+        network.rpc_url || thisnetwork.public_rpc_url
```
Bitcoinera, line 20: `        network.rpc_url || network.public_rpc_url`

PR #4911, opened 2020-09-15
Fabo/show all undelegations

faboweb:
Show undelegations as they are and not hide them when they are claimable
Show staked balance based on actual staking ledger and undelegations


faboweb:
🚀 Deployed on https://5f61504c2533688f21c42a59--lunieio.netlify.app

api/lib/reducers/polkadotV0-reducers.js
```diff
@@ -159,7 +160,7 @@ async function balanceV2Reducer(
     type: 'STAKE', // just a staking denom on Kusama for now
     available: availableLunieCoin.amount,
     total: totalLunieCoin.amount,
-    staked: stakedLunieCoin.amount,
+    staked: stakedLunieCoin.amount - undelegating,
```
Bitcoinera, line 163: oh woah, how could we miss that

app/src/components/staking/DelegationsOverview.vue
```diff
@@ -90,18 +88,8 @@ export default {
...
-      // substract the already unbonding balance in the case of Substrate networks.
-      if (this.undelegationsLoaded && this.undelegations.length > 0) {
-        stakedAmount = this.undelegations.reduce(
-          (stakedAmount, { amount }) => stakedAmount - Number(amount),
-          stakedAmount
-        )
-      }
       return {
-        total: stakedAmount.toFixed(3),
+        total: stakingDenomBalance ? stakingDenomBalance.staked.toFixed(3) : 0,
```
Bitcoinera, line 92: I'm a bit confused here. How does this fixes undelegations disappearing after reaching the claimable date?

api/lib/reducers/polkadotV0-reducers.js
Bitcoinera: mmm, actually you just need to use active, no? It is the same

Bitcoinera: For example:

```
"stash": "5GNtHmp1XmHJkGnM7swchrNGDcJ3r9yaQkumDZYK8tuEeZjG",
"total": 118362691275,
"active": 117362691275,
"unlocking": [
{
"value": 1000000000,
"era": 2717
}
],
```

Bitcoinera: We could also include unstaking in balanceV2 actually :shrug: It is not a bad idea

app/src/components/staking/DelegationsOverview.vue
faboweb: you could read the PR description: Show staked balance based on actual staking ledger and undelegations

api/lib/reducers/polkadotV0-reducers.js
faboweb: alright! let's use active. didn't see that. do we need to query for `unstaking` anywhere?

faboweb:
maybe we should have a different endpoint `stake` for all staking queries

PR #4933, opened 2020-09-17
Fabo/fix participant images

faboweb:
Show fallback when no image available

### DIFF
--- a/app/src/components/governance/ParticipantList.vue
+++ b/app/src/components/governance/ParticipantList.vue
@@ -10,7 +10,7 @@
         <div class="first-column">
           <span class="icon">
             <img
-              v-if="participant.validator"
+              v-if="participant.validator && participant.validator.picture"
               :src="participant.validator.picture"
             />
             <img v-else :src="currentNetwork.icon" />
@@ -76,6 +76,10 @@ export default {
       type: Array,
       required: true,
     },
+    showAmounts: {
+      type: Boolean,
+      default: false
+    }
   },
   data: () => ({
     showing: 5,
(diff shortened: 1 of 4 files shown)


jbibla (approved):

app/src/components/governance/ParticipantList.vue
```diff
@@ -76,6 +76,10 @@ export default {
       type: Array,
       required: true,
     },
+    showAmounts: {
+      type: Boolean,
+      default: false
+    }
```
Bitcoinera, line 82: I am just seeing in the `ParticipantList` code that the usage of `showAmounts` in quite counter-intuitive no?

Like it is exactly the opposite of what you would expect:

```
        <div v-if="!showAmounts && participant.option">
          <span class="option">{{ participant.option }}</span>
        </div>
        <div v-if="!showAmounts && participant.amount">
          <span class="amount">{{ participant.amount.amount }}</span>
          <span>{{ participant.amount.denom.concat(`s`) }}</span>
        </div>
```

We should definitely invert the values here

jbibla: yeah - shouldn't it be:

`<div v-if="showAmounts && participant.amount">`

Bitcoinera: OK, I am going to change it in this other PR #4939

faboweb: `showAmounts` was there before and I think it referred to the amount of total stake?

Bitcoinera: it refers whether or not the participants should display vote options & deposit amount. Yes, it was there and wrong from before

PR #4935, opened 2020-09-18
add changes / parameters to Cosmos & Treasury & Referendums

jbibla:
**Description:**
- this should include `changes` for parameter change proposals on cosmos and polkadot
- still need to adjust front end accordingly by adding `changes` to the description area 
- still need to remove `parameter` in favour of `changes`
- also removed some seemingly unused things from the reducers and auto lint made some adjustments

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/api/lib/reducers/cosmosV0-reducers.js
+++ b/api/lib/reducers/cosmosV0-reducers.js
@@ -271,7 +271,7 @@ function validatorReducer(networkId, signedBlocksWindow, validator) {
           ? validator.signing_info.missed_blocks_counter
           : 0
       ) /
-      Number(signedBlocksWindow),
+        Number(signedBlocksWindow),
     tokens: atoms(validator.tokens),
     commissionUpdateTime: validator.commission.update_time,
     commission: validator.commission.rate,
@@ -386,10 +386,10 @@ async function balanceReducer(coin, gasPrices, fiatValue) {
     fiatValue,
     gasPrice: gasPrices
       ? gasPriceReducer(
-        gasPrices.find(
-          (gasPrice) => denomLookup(gasPrice.denom) === coin.denom
-        )
-      ).price
+          gasPrices.find(
+            (gasPrice) => denomLookup(gasPrice.denom) === coin.denom
+          )
+        ).price
       : null
   }
 }
--- a/api/lib/reducers/cosmosV0-reducers.js
+++ b/api/lib/reducers/cosmosV0-reducers.js
@@ -182,7 +182,7 @@ function proposalReducer(
     type: proposal.proposal_content.type,
     title: proposal.proposal_content.value.title,
     description: proposal.proposal_content.value.description,
-    changes: JSON.stringify(proposal.proposal_content.value.changes),
+    changes: JSON.stringify(proposal.proposal_content.value.changes, null, 4),
     creationTime: proposal.submit_time,
     status: proposal.proposal_status,
     statusBeginTime: proposalBeginTime(proposal),
--- a/api/lib/reducers/cosmosV0-reducers.js
+++ b/api/lib/reducers/cosmosV0-reducers.js
@@ -173,7 +173,6 @@ function proposalReducer(
   proposer,
   totalBondedTokens,
   detailedVotes,
-  reducers,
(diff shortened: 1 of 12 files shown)


api/lib/schema.js
```diff
@@ -110,6 +110,7 @@ const typeDefs = gql`
     beneficiary: NetworkAccount
     summary: String
     parameter: String
+    changes: Changes
```
jbibla, line 113: maybe:
```suggestion
    changes: [Changes]
```

api/lib/reducers/cosmosV0-reducers.js
```diff
@@ -173,7 +173,6 @@ function proposalReducer(
   proposer,
   totalBondedTokens,
   detailedVotes,
-  reducers,
```
Bitcoinera, line 176: :eyes:

api/lib/schema.js
Bitcoinera:
```suggestion
    changes: [Change]
```
But not, actually just String it is the easiest. We just want to display a stringified JSON here. It is only a display thing after, we don't need another type to treat this data

Bitcoinera, line 113:
```suggestion
    changes: String
```
This is the right way

app/src/components/governance/ProposalHeader.vue
```diff
@@ -49,7 +49,7 @@
       <div class="proposer-and-summary-container">
         <div class="proposer">
           Proposed By:
-          {{ proposal.proposer.address | formatAddress }}
+          {{ proposal.address | formatAddress }}
```
Bitcoinera, line 52: ????

Bitcoinera: This looks quite wrong

app/src/gql/index.js
```diff
@@ -187,6 +187,7 @@ export const ProposalFragment = `
   }
   summary
   parameter
```
Bitcoinera, line 189:
```suggestion

```

faboweb:
🚀 Deployed on https://5f6497376a8e230f96d26831--lunieio.netlify.app

api/lib/reducers/polkadotV0-reducers.js
```diff
@@ -558,14 +556,14 @@ function democracyProposalReducer(
 function democracyReferendumReducer(
   network,
   proposal,
+  parameter,
```
Bitcoinera, line 559: ??? First you need to calculate it right?

I am not even sure the refenrendums have this field. I will check

api/lib/reducers/cosmosV0-reducers.js
jbibla: it was unused

api/lib/reducers/polkadotV0-reducers.js
jbibla: yeah i wasn't able to verify last night because my local API was having issues - just pushed and went to bed

app/src/components/governance/ProposalHeader.vue
jbibla: i think fabo changed this in one of his PRs yesterday - but we should confirm of course

api/lib/reducers/cosmosV0-reducers.js
Bitcoinera: :+1:

app/src/components/governance/ProposalHeader.vue
Bitcoinera: no, doesn't look like. Schema is the same than before

Bitcoinera:
This is how it is looking now for Kava:

image

api/lib/schema.js
faboweb: I think we should instead actually have Changes and map in the FE. this way we can actually map them nice line in a table or sth. get rid of parameter as well.

app/src/components/governance/ProposalHeader.vue
faboweb: sorry might have been a mistake yesterday

PR #4937, opened 2020-09-18
Ana/fix proposer in cosmos

Bitcoinera:
Closes #ISSUE

**Description:**

I was debugging this for a while. It seems the `this.query` in `cosmos-v0-source` doesn't trigger the `async query` function present in this file but another one.

Debugging this further after lunch 

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/api/lib/source/cosmosV2-source.js
+++ b/api/lib/source/cosmosV2-source.js
@@ -6,8 +6,8 @@ class CosmosV2API extends CosmosV0API {
     this.reducers = require('../reducers/cosmosV2-reducers')
   }
 
-  async query(url, resultSelector = 'result') {
-    const response = await this.getRetry(url)
+  async query(url, intent = 0, resultSelector = 'result') {
+    const response = await this.getRetry(url, intent)
     return response[resultSelector]
   }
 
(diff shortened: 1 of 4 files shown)


Bitcoinera:
All good

api/lib/source/cosmosV2-source.js
```diff
@@ -6,8 +6,8 @@ class CosmosV2API extends CosmosV0API {
     this.reducers = require('../reducers/cosmosV2-reducers')
   }
 
-  async query(url, resultSelector = 'result') {
-    const response = await this.getRetry(url)
+  async query(url, intent = 0, resultSelector = 'result') {
```
jbibla, line 9: in cosmosV0-source the argument is `noRetry` - can it be the same here?

Bitcoinera: yes, absolutely. You are right. Better stay consistent

Bitcoinera: although I see. I need to change and add the noRetry logic here too. Ah, thanks for catching this

faboweb (approved):

faboweb:
let's get rid of cosmosv0

PR #4949, opened 2020-09-18
Fabo/fix deposits

faboweb:
Displays the right amount of depositors
Displays the right amount of deposit

### DIFF
--- a/app/src/ActionModal/components/ModalDeposit.vue
+++ b/app/src/ActionModal/components/ModalDeposit.vue
@@ -99,10 +99,10 @@ export default {
       type: String,
       required: true,
     },
-    numberOfSeconds: {
-      type: Number,
-      default: 0,
-    },
+    deposits: {
+      type: Array,
+      default: () => []
+    }
   },
   data: () => ({
     amount: null,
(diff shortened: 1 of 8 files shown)


faboweb:
🚀 Deployed on https://5f652fa41e75b800bb748a64--lunieio.netlify.app

app/src/ActionModal/components/ModalDeposit.vue
```diff
@@ -175,7 +158,7 @@ export default {
   methods: {
     open() {
       if (this.currentNetwork.network_type === `polkadot`) {
-        this.amount = this.minimumDeposit
+        this.amount = this.deposits[0].amount[0].amount
```
Bitcoinera, line 161: well, this also makes sense, true. Actually it would be always the prior deposit, that is, the last one in the Array (`this.deposits[0].amount[this.deposits.length - 1].amount`)

Bitcoinera: Ah yeah, but logically must be all the same :sweat_smile:

app/src/ActionModal/components/ModalDeposit.vue
```diff
@@ -142,8 +126,7 @@ export default {
         amount: {
           amount: this.amount,
           denom: this.denom,
-        },
-        numberOfSeconds: this.numberOfSeconds,
```
Bitcoinera, line 146: we need this for the transaction. Now the `depositTx` is broken for Polkadot

faboweb: didn't know that

PR #4955, opened 2020-09-21
Ana/add message for polkadot treasuries

Bitcoinera:
Closes #ISSUE

**Description:**

Adds this message to Polkado Treasury proposals:

image

Thank you! 🚀

---

For contributor:

For reviewer:


faboweb:
🚀 Deployed on https://5f68ba726a02431af7b7250f--lunieio.netlify.app

jbibla:
this is not the right way to do this - if we want to include the message should alter the summary. the problem is that the summary is being displayed as a param change proposal - not a `type: treasury`

jbibla (changes requested):
not important right now and not the right implementation

Bitcoinera:
> this is not the right way to do this - if we want to include the message should alter the summary. the problem is that the summary is being displayed as a param change proposal - not a `type: treasury`

No, the problem is that we would need a network specific message for Polkadot networks, since we don't know how other Treasury proposals in other networks will look like. Council only exists in Polkadot networks

jbibla:
true. so in the API we can change the summary message based on whether or not it's a polkadot network.

Bitcoinera:
> true. so in the API we can change the summary message based on whether or not it's a polkadot network. but we shouldn't use a bar for this.

Done!

Bitcoinera:
Debugging this:

image

api/lib/reducers/polkadotV0-reducers.js
```diff
@@ -607,7 +607,9 @@ function treasuryProposalReducer(
     deposit: toViewDenom(network, Number(proposal.deposit)),
     proposer: proposal.proposer,
     beneficiary: proposal.beneficiary, // the account getting the tip
-    summary: getProposalSummary(proposalTypeEnum.TREASURY),
+    summary:
+      getProposalSummary(proposalTypeEnum.TREASURY) +
+      `\nTreasury proposals in ${network.title} can only be voted by the council members.`,
```
jbibla, line 612:
```suggestion
      `\nTreasury proposals on ${network.title} can only be voted on by council members.`,
```

jbibla (approved):

api/lib/reducers/polkadotV0-reducers.js
Bitcoinera: Native speaker powah!

PR #4957, opened 2020-09-21
Fabo/fix proposal network switch issue

faboweb:
Closes #ISSUE

**Description:**

Thank you! 🚀

---

For contributor:

- [ ] Added changes entries. Run `yarn changelog` for a guided process.
- [ ] Reviewed `Files changed` in the github PR explorer
- [ ] Attach screenshots of the UI components on the PR description (if applicable)
- [ ] Scope of work approved for big PRs

For reviewer:

- [ ] Manually tested the changes on the UI

PR #4959, opened 2020-09-21
Jordan/skip proposal queries

jbibla:
- only load proposals once to avoid console errors when switching networks
- reactivity doesn't really work for votes or deposits already so not sure if it needs to be addressed here
- don't query emoney gov params
- supporting links were not being displayed in the right hand column

---

For contributor:

For reviewer:

### DIFF
--- a/app/src/components/governance/PageProposal.vue
+++ b/app/src/components/governance/PageProposal.vue
@@ -204,6 +204,10 @@ export default {
         return data.proposal || {}
       },
       /* istanbul ignore next */
+      skip() {
+        return this.found
+      },
+      /* istanbul ignore next */
       variables() {
         return {
           id: this.proposalId,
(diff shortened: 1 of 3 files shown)


faboweb:
🚀 Deployed on https://5f691df257777e3e2cee72e6--lunieio.netlify.app

Bitcoinera:
This addresses the same than #4957, right?

app/src/components/governance/PageProposal.vue
```diff
@@ -204,6 +204,10 @@ export default {
         return data.proposal || {}
       },
       /* istanbul ignore next */
+      skip() {
+        return this.found
```
faboweb, line 208: but this is only set in the query after it is executed?!

app/src/components/governance/PageProposal.vue
```diff
@@ -227,7 +231,12 @@ export default {
       /* istanbul ignore next */
       skip() {
         // only Tendermint networks have this network-wide "governance parameters" logic
-        return !this.found || this.currentNetwork.network_type !== `cosmos`
+        return (
+          !this.found ||
+          this.currentNetwork.network_type !== `cosmos` ||
+          this.currentNetwork.id === `emoney-mainnet` ||
```
faboweb, line 237: why do parameters not work in emoney?

faboweb (approved):

app/src/components/governance/PageProposal.vue
jbibla: no governance

faboweb: why don't you use the general purpose feature flag for that?

faboweb: how does a user even get to this page for money?

Bitcoinera: I also think this part is not necessary

PR #4977, opened 2020-09-22
Ana/fix endpoint should start with ws

Bitcoinera:
Closes #ISSUE

**Description:**

Trying to fix this. But only hacky fix for now. I don't know why `polkadotNetwork.rpc_url` is `null` sometimes:

image

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/api/lib/resolvers.js
+++ b/api/lib/resolvers.js
@@ -280,7 +280,7 @@ const resolvers = (networkList, notificationController) => ({
         .map((network) => {
           // the server side nodes are mostly whitelisted or secret so we don't want to send them to the FE
           return Object.assign({}, network, {
-            rpc_url: network.public_rpc_url || network.rpc_url,
+            rpc_url: network.public_rpc_url,
             public_rpc_url: undefined
           })
         })
--- a/api/lib/resolvers.js
+++ b/api/lib/resolvers.js
@@ -280,7 +280,7 @@ const resolvers = (networkList, notificationController) => ({
         .map((network) => {
           // the server side nodes are mostly whitelisted or secret so we don't want to send them to the FE
           return Object.assign({}, network, {
-            rpc_url: network.public_rpc_url,
+            rpc_url: network.public_rpc_url || network.rpc_url,
             public_rpc_url: undefined
           })
         })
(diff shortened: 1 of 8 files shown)


faboweb:
🚀 Deployed on https://5f6a238950882b013db0958e--lunieio.netlify.app

faboweb:
let's close until you can reproduce

Bitcoinera:
I think this happens when changing branches. Going to check

Bitcoinera:
Exactly, this error happens every time branch is changed or API gets redeployed. Still investigating why

Bitcoinera:
Also fixes this minor style glitch:

Before:

image


Now:

image

api/lib/resolvers.js
```diff
@@ -280,7 +280,7 @@ const resolvers = (networkList, notificationController) => ({
         .map((network) => {
           // the server side nodes are mostly whitelisted or secret so we don't want to send them to the FE
           return Object.assign({}, network, {
-            rpc_url: network.public_rpc_url,
+            rpc_url: network.public_rpc_url || network.rpc_url,
```
faboweb, line 283:
```suggestion
            rpc_url: network.public_rpc_url
```

the comments states that we don't want to pass the private rpc out

Bitcoinera: there are some networks in staging-DB that lack of `rpc_url`, like cosmos for example. Guess we could use nylira there, but what about the others :shrug:

faboweb: is this a problem? you only had the issue with Polkadot, right?

Bitcoinera: yeah, true. Which left me wondering. OK, let me try this PR without that bit and see if it still works

Bitcoinera: Yeah, it works flawlessly. So the real fix was just adding `public-rpc-url` to the query body actually

Bitcoinera:
ping. Let's merge! (Changelog for `/app` not needed. Merely added a margin top for the Rewards message for Polkadot)

jbibla (approved):

PR #4989, opened 2020-09-24
Ana/add treasury proposals thresholds

### LINKED ISSUE #4988: Show thresholds for proposals

Currently proposals only show votes or deposits. We don't show required thresholds to succeed or pass.

I.e. in PolkadotJs
image

Display a bar that shows current votings/deposits and the thresholds. Negative votes could come from the right so we can show the threshold for those on the right. @jbibla ideas?

**Additional context**

Bitcoinera:
Closes #4988

**Description:**

All good

image

I can make the threshold span more prominent though

image

EDIT: Threshold is not displayed anymore

Thank you! 🚀

---

For contributor:

For reviewer:


faboweb:
🚀 Deployed on https://5f6c98f23f1f771a0115ba3c--lunieio.netlify.app

Bitcoinera:
WIP since there is a decision to be made on the display for Nays

app/src/components/governance/ProposalStatusBar.vue
```diff
@@ -26,13 +26,23 @@
...
+          <span>Threshold:</span>
+          <span>{{ proposal.detailedVotes.votingThresholdYes }} Votes</span>
+        </div>
       </div>
       <ProgressBar
         size="large"
-        :val="votePercentage * 100"
+        :val="
+          proposal.type === 'TREASURY'
+            ? proposal.detailedVotes.votingThresholdYes
```
faboweb, line 43: this shows the threshold not the current status am I wrong? Shouldn't it be: yesVotes / votingThresholdYes?

Bitcoinera: We could actually display both. But the latter is already visually displayed on the progress bar. But we could this display it as a percentage too :shrug: Not a bad idea

faboweb: let's have a call on this

Bitcoinera: Tuesday?

jbibla:
this should be implemented as a bar for all proposals. but also, this is not a priority right now.

Bitcoinera:
> this should be implemented as a bar for all proposals. but also, this is not a priority right now.

The others don't have voting threshold, or you mean Yes/No bars (two bars instead of one)?

It might not be a priority right now but it would nice to wrap governance up, with the cherry on the top, don't you think?

jbibla:
everything has a threshold (some measure of when the state will change). for treasury proposals it's a majority of council members for example.

Bitcoinera:
> everything has a threshold (some measure of when the state will change). for treasury proposals it's a majority of council members for example.

In Polkadot referendums and treasuries, yes. But in the case of Cosmos I thought there is merely a deadline, and when this is reached it is decided upon the votes percentage at that point

jbibla:
> in the case of Cosmos I thought there is merely a deadline, and when this is reached it is decided upon the votes percentage at that point

yes, true. but at the deadline the state change is determined by the voting outcomes which we can display as a threshold.

Bitcoinera:
> yes, true. but at the deadline the state change is determined by the voting outcomes which we can display as a threshold.

The threshold is always simple majority in Cosmos, right?

jbibla:
> The threshold is always simple majority in Cosmos, right?

i don't know - please confirm with the cosmos docs!

Bitcoinera:
'kay, will have a look

Bitcoinera:
> please confirm with the cosmos docs!

It seems it is this way: https://docs.cosmos.network/v0.40/modules/gov/01_concepts.html#threshold

The only two things to take into account is that the proposal won't be valid if more than 1/3rd votes NoWithVeto and also a minimum Quorum must be reached. But we don't need to display that for now. Threshold is this what stands on the docs

app/src/components/governance/ProposalStatusBar.vue
Bitcoinera: oops, we forgot about this, didn't we?

Bitcoinera:
Seems there are still some issues with Polkadot proposals times :eyes: 

image

Bitcoinera:
:thinking: 

image

 Guess we are missing something here

app/src/components/governance/ProposalStatusBar.vue
```diff
@@ -31,18 +31,14 @@
           <span>{{ votePercentage | percentInt }} of {{ stakingDenom }}</span>
           <span v-if="voteCount">({{ voteCount }} Votes)</span>
         </div>
-        <div v-if="true" class="vote-data">
+        <div v-if="proposal.type === governanceStatusEnum.TREASURY" class="vote-data">
```
jbibla, line 34: i don't understand why thresholds are only for treasury proposals...? every proposal type has a threshold.

Bitcoinera:
Alright, so I removed the simple display span for thresholds I added and this PR what is currently doing it simply fixing voting percentages in detailedVotes and in PropsalStatusBar

Bitcoinera:
Rerunning E2E

jbibla (approved):

jbibla:
thanks ana!

PR #5020, opened 2020-10-01
PROPOSAL Ana/add merge strategy in publish flow

Bitcoinera:
Closes #ISSUE

**Description:**

This strategy will always take the incoming change as the solution for the conflict. More information on merge strategies can be found in the Git official documentation site [here](https://www.git-scm.com/docs/merge-strategies)

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -26,7 +26,7 @@ jobs:
               git remote set-url origin https://${GIT_BOT_TOKEN}@github.com/luniehq/lunie.git &&
               git checkout master &&
               git pull &&
-              git merge --no-edit origin/develop &&
+              git merge --no-edit --strategy recursive --strategy-option theirs origin/develop &&
               git push;
             else
               echo "No release detected, so not publishing";


Bitcoinera:
This PR comes from [this failed attempt to release](https://github.com/luniehq/lunie/actions/runs/282439391).

Having hotfixes directly in masters can result in these conflicts that mess up our releasing process. It forces us to always stay attentive and see if they work and, if they don't, then manually merge to master like Jordan had to do in this case.

My proposed solution here is to change the merge strategy so changes coming from `develop` are always favored against changes done in master.

Alternatively, and I think also as better solution, there is the solution to always avoid to hotfix directly into master and always merge this "hotfixes" from develop, which the code owners can do, without required approvals

jbibla:
@Bitcoinera i'm going to close this PR. can you open an issue about our release flow and include the details above?

Bitcoinera:
I think we are going to need to do this if we are going to continue using Simsala. Otherwise Simsala is too much of a headache with the conflicts (we never see them directly since the Simsala branch has develop as its base).

But OK, opening an issue about this

PR #5063, opened 2020-10-07
Ana/(hopefully) fix cors errors

### LINKED ISSUE #5035: Sidecar failing
As you may notice we needed to revert the sidecar implementation for Polkadot in the API. The reason for this is a CORS error popping up and preventing a complete load of balances, block, proposals and so on. We released too fast since this error was present in the beginning, only not locally (but on https://staging--lunieio.netlify.app/ was also to be seen, when connecting to staging-api).

Here the console logs:

image

Kusama not loading as a result:

image

I opened a PR hacking this (#5024) a bit since I noticed subscriptions might be the problem here. But I am still not sure what is causing this. Some problem between netlify and the node we're connecting to for sidecar, but we are not getting the usual error for CSP. The error itself is quite strange, since we are always querying from those API URLs, but only for Polkadot networks with sidecar it will pop up.

Still investigating and hopefully we can solve this asap, as sidecar is a great performance improvement for Polkadot

Bitcoinera:
Closes #5035 

**Description:**

I think this should fix our CORS errors

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/api/changes/ana_fix-cors-errors
+++ b/api/changes/ana_fix-cors-errors
@@ -1,0 +1,1 @@
+[Fixed] [#5063](https://github.com/cosmos/lunie/pull/5063) Fix CORS errors when connecting to API (Sidecar, e-Money...) @Bitcoinera
--- a/api/index.js
+++ b/api/index.js
@@ -23,7 +23,11 @@ async function main() {
 
   const apolloServer = await createApolloServer(httpServer)
   app.use(
-    apolloServer.getMiddleware({ app, path: config.queryPath, cors: { origin: '*' } })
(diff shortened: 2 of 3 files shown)


jbibla:
why do you think we have CORS issues? 

you're just applying the default?
```
The default configuration is the equivalent of:

{
  "origin": "*",
  "methods": "GET,HEAD,PUT,PATCH,POST,DELETE",
  "preflightContinue": false,
  "optionsSuccessStatus": 204
}
```

Bitcoinera:
Mmm, is that is the default configuration then this one won't fix it, true

PR #5076, opened 2020-10-08
Ana/5053 duplicate notifications for multimessage

### LINKED ISSUE #5053: Too many emails notifications at once
Got 19 emails in a row saying I received 5,000 AKT.

Bitcoinera:
Closes #5053 

**Description:**

Need to try it a bit, but this would be the solution. Every notification should go to a maximum of 2 addresses (Send & Receive), since the notifications are per transaction message and not per transaction, and messages involve a maximum of 2 addresses.

The question is whether it makes sense to keep involvedAddresses or not.

I also noticed some error on FE in Akash due to this `E0D29BADA5CC678036279EE0771FD8A0A34840EF0ABA5884E417CD822D2E7F62`, but guess I should fix that in another PR

image

Thank you! 🚀

---

For contributor:

For reviewer:


api/lib/reducers/polkadotV0-reducers.js
```diff
@@ -225,6 +225,10 @@ function getMessageType(section, method) {
   }
 }
 
+function getTransactionMessageAddresses(agentAddress, transactionDetails) {
```
jbibla, line 228: if we have to use this function in multiple reducers, we should put it somewhere and import it as opposed to duplicating it.

api/lib/block-listeners/base-node-subscription.js
```diff
@@ -174,7 +174,8 @@ class BaseNodeSubscription {
       // transaction to each subscribed user.
       // TODO doesn't handle failing txs as it doesn't extract addresses from those txs (they are not tagged)
       block.transactions.forEach((tx) => {
-        tx.involvedAddresses.forEach((address) => {
```
jbibla, line 177: what is the difference between `involvedAddresses` vs. `transactionMessageAddresses`?

jbibla: where is this "published"?

https://github.com/luniehq/lunie/pull/5076/files#diff-0ad2d28223c295dd602e00c42eab16aeR187

Bitcoinera: in Redis. This is the subscription that `listenToNotifications` listens to, in line 31 in `notificationController.js`

Bitcoinera: > what is the difference between involvedAddresses vs. transactionMessageAddresses?

The difference is that involvedAddresses contains all the addresses involved per "transaction", whereas `transactionMessageAddresses` are the addresses involved per message

api/lib/reducers/polkadotV0-reducers.js
Bitcoinera: in this case I think it is fine, since all the Polkadot networks would take it from here. But we could also move it to common... although I think it is better here (like `coinReducer` and some others which are exactly the same)

Bitcoinera: across `cosmosV0-source.js` and `polkadotV0-source.js`, I mean

jbibla:
unit test is failing

Bitcoinera:
@faboweb , you want to share your expert take here? :pray:

api/lib/reducers/cosmosV2-reducers.js
```diff
@@ -266,6 +266,15 @@ function proposalReducer(
   }
 }
 
+function getTransactionMessageAddresses(agentAddress, transactionDetails) {
```
faboweb, line 269:
```suggestion
function getTransactionMessageAddresses(senderAddress, transactionDetails) {
```

api/lib/reducers/cosmosV2-reducers.js
```diff
@@ -337,7 +346,17 @@ function transactionReducerV2(network, transaction, reducers) {
             ? transaction.logs[index].log || transaction.logs[0] // failing txs show the first logs
             : transaction.logs[0].log || ''
           : JSON.parse(JSON.stringify(transaction.raw_log)).message,
-      involvedAddresses: uniq(reducers.extractInvolvedAddresses(transaction))
+      involvedAddresses: uniq(reducers.extractInvolvedAddresses(transaction)),
```
faboweb, line 349: instead this should be fixed to have the involved addresses for this message only. you can use the code you have here

api/lib/reducers/cosmosV3-reducers.js
```diff
@@ -161,7 +161,17 @@ function transactionReducerV2(network, transaction, reducers) {
             ? transaction.logs[index].log || transaction.logs[0] // failing txs show the first logs
             : transaction.logs[0].log || ''
           : JSON.parse(JSON.stringify(transaction.raw_log)).message,
-      involvedAddresses: uniq(reducers.extractInvolvedAddresses(transaction))
+      involvedAddresses: uniq(reducers.extractInvolvedAddresses(transaction)),
+      transactionMessageAddresses: reducers.getTransactionMessageAddresses(
```
faboweb, line 165: alarming that we have code duplication here. need to refactor

api/lib/reducers/polkadotV0-reducers.js
faboweb: I agree with Jordan, please refactor

api/lib/reducers/polkadotV0-reducers.js
```diff
@@ -263,6 +272,20 @@ function parsePolkadotTransaction(
       lunieTransactionType,
       signer,
       message
+    ),
+    transactionMessageAddresses: getTransactionMessageAddresses(
```
faboweb, line 276: remove and fix `involvedAddresses`

api/lib/block-listeners/base-node-subscription.js
```diff
@@ -174,7 +174,8 @@ class BaseNodeSubscription {
...
+        tx.transactionMessageAddresses.forEach((address) => {
+          if (!address) return // case for governance transactions where there is no recipient
```
faboweb, line 178: this case should be handled inside of `publishUserTransactionAddedV2`

api/lib/reducers/cosmosV2-reducers.js
Bitcoinera: ok

api/lib/block-listeners/base-node-subscription.js
Bitcoinera: I mean, if we fix involvedAddresses like you propose, we don't need to change anything here at all

Bitcoinera:
OK, found the real issue, it was this code block in extractInvolvedAddesses:

```
  const events = transaction.logs
    ? transaction.logs.reduce(
        (events, log) => (log.events ? events.concat(log.events) : events),
        []
      )
    : []
```

Logs were coming from the whole transaction

Bitcoinera:
ah ok, so the issue was in the `cosmosV3-reducers`. Now I see on files changed. Alright, then leaving the cosmosV2-reducers version. That one was actually right

api/lib/block-listeners/base-node-subscription.js
faboweb: `involvedAddresses` should never have an item `null`. something is wrong at a different place, if that happens

faboweb (approved):

PR #5104, opened 2020-10-14
Fabo/automerge changelog from develop to master

faboweb:
As a hotfix requires to edit the modules CHANGELOG.md on master there can be merging issues when editing the changelog in a release and merging it to master. Releases to master come from our new source of truth `develop`. We could just force push always to master. A slightly less destructive way to do this, is to merge but always assume that `develop` is the truth in case of a conflict. This will resolve all CHANGELOG.md conflicts in favor of `develop`. One bad edge case: We pushed a hotfix to master since we triggered a release and this hotfix conflicts with develop. Now develop will overwrite the hotfix.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -26,7 +26,7 @@ jobs:
               git remote set-url origin https://${GIT_BOT_TOKEN}@github.com/luniehq/lunie.git &&
               git checkout master &&
               git pull &&
-              git merge --no-edit origin/develop &&
+              git merge origin/develop -X theirs &&
               git push;
             else
               echo "No release detected, so not publishing";


jbibla:
can we specify that we only want to override the changelog file and not the whole branch?

jbibla:
> As a hotfix requires to edit the modules CHANGELOG.md on master

this is part of the problem IMO. we are always prioritizing the changelog and changelog management over the actual codebase when it should be the other way around.

Bitcoinera:
> we are always prioritizing the changelog and changelog management over the actual codebase 

:100: 

I think so too. Right now the way it all works it is just too cumbersome. Let's simplify and changelog should be just a graceful addition, not the base for the whole thing

Bitcoinera:
> One bad edge case: We pushed a hotfix to master since we triggered a release and this hotfix conflicts with develop. Now develop will overwrite the hotfix.

What about setting a GitHub Action that automatically creates a PR from master to develop when master is pushed? Then we just need to merge that hotfix to develop.

With that and this merge strategy here, we would be all setup

PR #5105, opened 2020-10-14
Fabo/fix proposal notifications not stored

faboweb:
Notifications for Proposals on Cosmos chains would not be stored

### DIFF
--- a/api/lib/notifications/notifications.js
+++ b/api/lib/notifications/notifications.js
@@ -212,7 +212,7 @@ const startNotificationService = (networks) => {
         topic,
         eventType: event.eventType,
         resourceType: event.resourceType,
-        resourceId: event.resourceId,
+        resourceId: String(event.resourceId),
         networkId: event.networkId,
         data: event.properties
       })
--- a/api/lib/notifications/notifications.js
+++ b/api/lib/notifications/notifications.js
@@ -210,11 +210,11 @@ const startNotificationService = (networks) => {
         ''
       ).storeNotification({
         topic,
-        eventType: event.eventType,
-        resourceType: event.resourceType,
+        eventType: String(event.eventType),
+        resourceType: String(event.resourceType),
         resourceId: String(event.resourceId),
-        networkId: event.networkId,
-        data: event.properties
+        networkId: String(event.networkId),
+        data: String(event.properties)
       })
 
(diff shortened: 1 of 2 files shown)


api/lib/notifications/notifications.js
```diff
@@ -212,7 +212,7 @@ const startNotificationService = (networks) => {
         topic,
         eventType: event.eventType,
         resourceType: event.resourceType,
-        resourceId: event.resourceId,
+        resourceId: String(event.resourceId),
```
jbibla, line 215: any reason not to `String()` everything here?

why does `resourceId` not being a string cause the whole row to be absent from the db?

Bitcoinera: Yes, this looks definitely weird. In addition, when would the resourceId (an address) not be a String?

faboweb: the resource ID is the proposal ID which is an Int in Cosmos. this is not matching the GQL schema of our db. we can stringify everything for safety. go ahead

Bitcoinera: shouldn't it be the proposer address? Ah no, alright, because it is a network related notification, I see.

OK, let's stringify everything then

api/lib/notifications/notifications.js
```diff
@@ -210,11 +210,11 @@ const startNotificationService = (networks) => {
...
         topic,
-        eventType: event.eventType,
-        resourceType: event.resourceType,
+        eventType: String(event.eventType),
+        resourceType: String(event.resourceType),
         resourceId: String(event.resourceId),
-        networkId: event.networkId,
-        data: event.properties
+        networkId: String(event.networkId),
+        data: String(event.properties)
```
faboweb, line 217: sure about this? properties is an object
```suggestion
        data: event.properties
```

Bitcoinera: I am quite sure:

image

faboweb: I am also quiet sure this is wrong. This logic would never be triggered:
https://github.com/luniehq/lunie/blob/develop/api/lib/database/helpers.js#L44

Bitcoinera: ah true! Good catch. Removing

PR #5116, opened 2020-10-15
Ana/select submodule to release

### LINKED ISSUE #5115: Select Submodule we want to release
Instead of opening three Simsala PRs every time we run `yarn release`, I propose that we add the submodule there as a parameter (i.e. `yarn release:app` or  `yarn release:api`) and then selectively open the Simsala PR concretely for that submodule.

This will reduce noise and help have cleaner & easier release flows

Bitcoinera:
Closes #5115
**Description:**

This PR makes it possible to select the wished submodule to release.

It could permit to automatically merge the Simsala PR into develop. But Fabo stressed the importance of QAing, so I guess we won't do that :shrug: at least for the time being

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,9 @@
     "serve": "cd app && yarn serve",
     "serve:win": "cd app && yarn serve:win",
     "lint:global": "cd api && yarn lint --fix && cd .. && cd app && yarn lint --fix && cd .. && cd extension && yarn lint:fix && cd ..",
-    "release": "git stash && git fetch --all && git checkout origin/develop -B develop && git pull && git checkout origin/release -B release && git pull && git merge origin/master && git merge origin/develop && git push && git stash pop"
+    "release:api": "git stash && git fetch --all && git checkout origin/develop -B develop && git pull && git checkout origin/release-api -B release-api && git pull && git merge origin/master && git merge origin/develop && git push && git stash pop",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -21,7 +21,9 @@
     "serve": "cd app && yarn serve",
     "serve:win": "cd app && yarn serve:win",
     "lint:global": "cd api && yarn lint --fix && cd .. && cd app && yarn lint --fix && cd .. && cd extension && yarn lint:fix && cd ..",
-    "release": "git stash && git fetch --all && git checkout origin/develop -B develop && git pull && git checkout origin/release -B release && git pull && git merge origin/master && git merge origin/develop && git push && git stash pop"
+    "release:api": "git stash && git fetch --all && git checkout origin/develop -B develop && git pull && git checkout origin/release-api -B release-api && git pull && git merge origin/master && git merge origin/develop && git push && git stash pop",
```
faboweb, line 24: soooo much code repetition

Bitcoinera: true true

Bitcoinera: but I am not sure how to avoid it... Need to investigate how to use that param in the script

Bitcoinera: There is a way to do this declaring a variable in package.json, but in Windows needs a different syntax:
https://stackoverflow.com/questions/43705195/how-can-i-use-variables-in-package-json

I think it is better to leave it like this. What do you think?

Bitcoinera: Great! That was it. Much better like this, thank you :pray: 

I leave the script in the root folder for now and we can create a scripts folder in the future, if we need another script.

Tried it and works

faboweb (approved):

PR #5056, opened 2020-10-06
Ana/move fixed network fee data to db

Bitcoinera:
Closes #4429 
Closes #4526 

**Description:**

Finally gas prices and gas estimates come from DB, giving us much more control and being able like this to fix fees-related issues on the fly

Thank you! 🚀

---

For contributor:

- [ ] Added changes entries. Run `yarn changelog` for a guided process.
- [ ] Reviewed `Files changed` in the github PR explorer
- [ ] Attach screenshots of the UI components on the PR description (if applicable)
- [ ] Scope of work approved for big PRs

For reviewer:

- [ ] Manually tested the changes on the UI

PR #5120, opened 2020-10-16
Ana/try to fix mem leak: maybe from network fees

Bitcoinera:
Closes #ISSUE

**Description:**

I think the slow mem leak we are experiencing could come from the gas estimates / prices coming from DB.

Looking at the code seems to be good, but I believe it might be related because of the times in Digital Ocean graphics and the time this PR, #5056 , was merged

image

image

Not 100% the same but this is the most likely one :shrug: 

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/api/data/network-fees.js
+++ b/api/data/network-fees.js
@@ -60,24 +60,28 @@ const pollForNewFees = async () => {
 }
 
 const getNetworkGasPrices = async (networkId) => {
-  const networkGasPrices = await db.getNetworkGasPrices()
-  networkGasPrices.forEach(networkGasPrice => {
-    if (!networkGasPricesDictionary[networkGasPrice.id]) networkGasPricesDictionary[networkGasPrice.id] = []
-    networkGasPricesDictionary[networkGasPrice.id].push({ denom: networkGasPrice.denom,
-      gasPrice: networkGasPrice.gasPrice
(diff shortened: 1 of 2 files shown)


Bitcoinera:
Well, the possible failure would be that this query about fees is being done too often and this is causing DB to fail or slow it down, and that is what is messing with API too. Because API is too dependent on DB. So maybe this fix is good, could be.

What do you think?

api/data/network-fees.js
```diff
@@ -60,24 +60,28 @@ const pollForNewFees = async () => {
...
+      networkGasPricesDictionary[networkGasPrice.id].push({ denom: networkGasPrice.denom,
+        gasPrice: networkGasPrice.gasPrice
+      })
     })
-  })
+  }
   // update emoneyGasPrices
   networkGasPricesDictionary['emoney-mainnet'] = emoneyGasPrices
-  return networkGasPricesDictionary[networkId] || []
+  return networkGasPricesDictionary[networkId]
```
jbibla, line 74: why this change?

Bitcoinera: because now it cannot be `undefined`

jbibla:
ok let's merge to staging and see if it helps

jbibla (approved):

PR #5142, opened 2020-10-20
Ana/5141 fix mem leak due to validator profiles

### LINKED ISSUE #5141: staging API mem leak due to validator profiles
Mem leak & disconnection problem due to a polling logic for validator profiles that I forgot was there

Bitcoinera:
Closes #5141

**Description:**

Still need to review a bit. Looks good, but still getting some queries failed for DB, even removing the faulty cosmos-hub-mainnet

But I think these errors with DB are due to some problem with staging DB. On production it works. I need to look

Thank you! 🚀

---

For contributor:

For reviewer:

### DIFF
--- a/api/lib/resolvers.js
+++ b/api/lib/resolvers.js
@@ -241,32 +241,15 @@ const resolvers = (networkList, notificationController) => ({
       const validators = Object.values(
         localStore(dataSources, validator.networkId).validators
       )
-      const network = networkList.find(({ id }) => id === validator.networkId)
-      const dataSourceClass = require(`./${network.source_class_name}`)
-      const fiatValuesAPI = new FiatValuesAPI()
-      const networkSchemaName = validator.networkId.replace(/-/g, '_')
-      const db = new database(config)(networkSchemaName)
-      const dataSource = new dataSourceClass(
-        network,
-        localStore(dataSources, validator.networkId),
-        fiatValuesAPI,
-        db
-      )
       if (validators) {
-        await getValidatorProfile(
+        const network = networkList.find(({ id }) => id === validator.networkId)
+        return await getValidatorProfile(
           validators,
           validator,
-          dataSource,
+          remoteFetch(dataSources, validator.networkId),
           network,
-          db
-        ).then((validatorWithProfile) => {
-          localStore(dataSources, validator.networkId)[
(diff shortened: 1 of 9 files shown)


api/lib/resolvers.js
```diff
@@ -236,11 +237,33 @@ const resolvers = (networkList, notificationController) => ({
...
-            validator.operatorAddress
-          ].profile
+    profile: async (validator, _, { dataSources }) => {
+      const validators = localStore(dataSources, validator.networkId).validators
+      const network = networkList.find(({ id }) => id === validator.networkId)
+      const dataSourceClass = require(`./${network.source_class_name}`)
+      const fiatValuesAPI = new FiatValuesAPI()
+      const networkSchemaName = validator.networkId.replace(/-/g, '_')
+      const db = new database(config)(networkSchemaName)
+      const dataSource = new dataSourceClass(
```
faboweb, line 247: why are you creating a new datasource here?

api/lib/resolvers.js
```diff
@@ -236,11 +237,33 @@ const resolvers = (networkList, notificationController) => ({
...
+        network,
+        localStore(dataSources, validator.networkId),
+        fiatValuesAPI,
+        db
+      )
+      if (validators && !validators[validator.operatorAddress].profile) {
```
faboweb, line 253: this will never update the validator profiles as it adds to the store once and then never runs again

Bitcoinera:
OK, it works really well! No errors in console and very fast :woman_dancing:  Let's merge :woman_dancing:

Bitcoinera:
Well, although I could change everything to just one validator actually. But it is very fast anyway now :shrug: 

The problem is that it won't update :thinking: Let me change that

api/lib/resolvers.js
```diff
@@ -236,12 +237,37 @@ const resolvers = (networkList, notificationController) => ({
...
+      )
+      if (validators) {
+        await getValidatorProfile(
+          validators,
+          validator,
+          dataSource,
+          network,
+          db
+        ).then((validatorWithProfile) => {
+          localStore(dataSources, validator.networkId)[
```
faboweb, line 263: why are you setting the profile here? you are never reading it

